PR #122, opened 2024-07-21
added translation tasks

Vasilije1990:
## Summary by CodeRabbit

- **New Features**
	- Introduced a language detection feature that identifies the language of input text.
	- Added a translation function that uses AWS Translate to convert text between specified languages.

- **Bug Fixes**
	- Enhanced error handling and logging for both language detection and translation processes.

- **Chores**
	- Added a new dependency for improved language code handling.
	- Expanded configuration options for AWS integration by adding optional AWS credential fields.


cognee/modules/data/processing/detect_chunk_language.py'.py
```diff
@@ -0,0 +1,46 @@
+import boto3
```
borisarzentar, line 1: There is `'` in filename, probably mistake.

cognee/modules/data/processing/detect_chunk_language.py'.py
```diff
@@ -0,0 +1,46 @@
...
+        if detected_lang_iso639_1 == 'hr':
+            return 'sr'
+        return detected_lang_iso639_1
+
+    except LangDetectException as e:
+        logging.error(f"Language detection error: {e}")
+    except Exception as e:
+        logging.error(f"Unexpected error: {e}")
+
+    return -1
```
borisarzentar, line 46: Why return `-1`?

cognee/modules/data/processing/translate_chunks.py
```diff
@@ -0,0 +1,48 @@
...
+from botocore.exceptions import BotoCoreError, ClientError
+from langdetect import detect, LangDetectException
+import iso639
+from dotenv import load_dotenv
+load_dotenv()
+
+import logging
+
+# Basic configuration of the logging system
+logging.basicConfig(level=logging.INFO, format='%(asctime)s - %(levelname)s - %(message)s')
```
borisarzentar, line 11: We shouldn't configure the global logging object. Let's create a local logger and configure that one.

cognee/modules/data/processing/detect_chunk_language.py'.py
borisarzentar: Do we want to detect it on the document level, or on the chunk level?

borisarzentar:
@Vasilije1990 Please check comments

cognee/modules/data/processing/detect_chunk_language.py'.py
Vasilije1990: How would tasks work on document level vs chunk from design side?

cognee/modules/data/processing/detect_chunk_language.py'.py
Vasilije1990: No good reason

PR #170, opened 2024-10-27
fix: Resolve chunking issue for notebook

dexters1:
Add cleaning of local data to resolve chunking issue with repeated notebook use

## Summary by CodeRabbit

- **Chores**
	- Updated the demo notebook for improved clarity and usability.

### DIFF
--- a/notebooks/cognee_demo.ipynb
+++ b/notebooks/cognee_demo.ipynb
@@ -577,6 +577,7 @@
                 "\n",
                 "import cognee\n",
                 "\n",
+                "await cognee.prune.prune_data()\n",
                 "await cognee.prune.prune_system(metadata=True)"
             ]
         },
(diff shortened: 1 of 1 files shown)


notebooks/cognee_demo.ipynb
```diff
@@ -667,6 +669,10 @@
                 "                    summarization_model = cognee_config.summarization_model,\n",
                 "                    collection_name = \"summaries\",\n",
                 "                ),\n",
+                "             Task(\n",
```
borisarzentar, line 672: Does this task work as expected? We didn't use it for a while.

dexters1: I don't know if it gives the expected output, but it doesn't cause problems anymore for the notebook and tests regarding UUIDs due to fixing the issue with cleaning the database. If the database would not be cleaned it would cause these issues though. The Task is needed for the insights search, otherwise it complains that it doesn't have a classification table.

Vasilije1990: Seems fine then, no obvious error, we can add tests for it later and figure out how to test each task

PR #183, opened 2024-11-08
Call os.getenv in BaseConfig

0xideas:
So that Graphistry Visualisation works

## Summary by CodeRabbit

- **New Features**
	- Enhanced configuration management by retrieving `graphistry_username` and `graphistry_password` from environment variables for improved security and flexibility.
  
- **Bug Fixes**
	- Resolved issues related to default values for configuration attributes, allowing for dynamic environment-based configurations.

### DIFF
--- a/cognee/base_config.py
+++ b/cognee/base_config.py
@@ -1,3 +1,4 @@
+import os
 from typing import Optional
 from functools import lru_cache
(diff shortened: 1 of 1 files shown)


dexters1:
Would be good to also add this to the .env.template file as an option then

PR #184, opened 2024-11-08
Remove automatic execution of test that require LLM inputs from merge…

0xideas:
… requests onto main.

To save money on OpenAI credits.

## Summary by CodeRabbit

- **New Features**
	- Workflows for Neo4j, Notebook, pgvector, Qdrant, and Weaviate can now be manually triggered, simplifying the initiation process.
  
- **Bug Fixes**
	- Removed automatic triggers for pull requests targeting the `main` branch across multiple workflows, ensuring more controlled execution.
  
- **Improvements**
	- Workflows now respond to pull requests labeled with `run-checks`, allowing for conditional execution based on specific criteria.

### DIFF
--- a/.github/workflows/test_neo4j.yml
+++ b/.github/workflows/test_neo4j.yml
@@ -2,6 +2,10 @@ name: test | neo4j
 
 on:
   workflow_dispatch:
+  pull_request:
+    branches:
+      - main
+    types: [labeled]
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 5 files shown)


0xideas:
Hey @Vasilije1990, the new workflow would be to add the label 'run-checks' to trigger the LLM-intensive tests before merging, and to allow merging only if this label is present. This isn't an ideal solution but the only alternative is an additional 'meta' GitHub action that manages the triggering of these tests. The solution implemented here is also the one suggested [here](https://stackoverflow.com/a/76346378)

PR #197, opened 2024-11-12
Cog 577 add unit test task

0xideas:
## Summary by CodeRabbit

- **New Features**
	- Introduced new pytest fixtures and test files for improved testing of various components, including `PdfDocument` and graph-related functionalities.
	- Added integration tests for chunking functionality and document classification.
	- Added a new pytest fixture to automate database setup for integration tests.

- **Bug Fixes**
	- Removed duplicate entry for "pdf" in document classification to ensure correct mapping.

- **Documentation**
	- Enhanced error messages in tests for better clarity during assertions.

- **Chores**
	- Updated testing workflows for Python 3.9, 3.10, and 3.11 to focus on unit and integration tests.
	- Removed unused test files and configuration settings to streamline the testing process.
	- Removed the `pytest.ini` configuration file.


0xideas:
Hey @borisarzentar, I moved the test for running the toy tasks out of the unit test folder and into an integration test folder, as they require a sqlite database. I tried monkeypatching the database access but that didn't work, so now I copy over a file that enables these pipelines. By relying on outside state, they are no longer unit tests, and so I moved the to `tests/integration`. This also introduces a difference between integration and end-to-end tests, which is what I'd call the`test_DATABASE` files. I hope this all works for you!

cognee/tests/integration/run_toy_tasks/conftest.py
```diff
@@ -0,0 +1,11 @@
+import os
+
+import pytest
+
+
+@pytest.fixture(autouse=True, scope="session")
+def copy_cognee_db_to_target_location():
```
borisarzentar, line 7: Why do we copy the database? Cognee will create a new one in each run, we just need to set system path like we already do in end-to-end tests.

cognee/tests/unit/interfaces/graph/conftest.py
```diff
@@ -0,0 +1,80 @@
+from datetime import datetime, timezone
```
borisarzentar, line 1: What `conftest` means?

cognee/tests/unit/processing/chunks/chunk_by_paragraph_test.py
```diff
@@ -0,0 +1,69 @@
...
+    "whole_text": """This is example text. It contains multiple sentences.
+    This is a second paragraph. First two paragraphs are whole.
+    Third paragraph is a bit longer and is finished with a dot.""",
+    "cut_text": """This is example text. It contains multiple sentences.
+    This is a second paragraph. First two paragraphs are whole.
+    Third paragraph is cut and is missing the dot at the end""",
+}
+
+
+def run_chunking_test(test_text, ground_truth):
```
borisarzentar, line 50: `ground_truth` in this case is a list of chunks, so let's name it `expected_chunks`.

cognee/tests/unit/interfaces/graph/util.py
```diff
@@ -0,0 +1,19 @@
+from datetime import datetime, timezone
+
+
+def run_test_against_ground_truth(
```
borisarzentar, line 4: I don't like this kind of abstractions because they usually expand over time in order to support more use cases. But let's keep it for now. If it starts growing we can kill it.

borisarzentar (changes requested):

cognee/tests/integration/run_toy_tasks/conftest.py
0xideas: This turned out to be necessary for the integration tests to run, I don't know the library well enough to know why exactly that is the case. But it looks for this file at this location when querying for a user, and errors out if it doesn't find it. 

I tried monkeypatching the database but that didn't work unfortunately.

cognee/tests/unit/interfaces/graph/conftest.py
0xideas: `conftest` is the pytest convention for setup and teardown functions for tests. Pytest automatically executes these before (and if required after) the tests themselves. They are the best practice for passing data to test functions, one I don't follow 100% in this PR, admittedly

borisarzentar (approved):

PR #228, opened 2024-11-15
COG-546 local script dependencies

lxobr:
A utility function, `get_local_script_dependencies`: 
- Extracts and resolves local dependencies of a Python script using `jedi` and `parso`. 
- Returns a sorted list of unique module paths 
- Optionally dependencies outside a specified repository path are filtered out

Will be used for creating a graph from a repo.

## Summary by CodeRabbit

- **New Features**
	- Introduced functionality for extracting and updating code entities from Python scripts.
	- Added a method to retrieve unique module paths that a script depends on.
	- Implemented a context manager for temporary system path modifications.
	- Included asynchronous handling for dependency extraction.
	- Demonstrated the new functionality with example paths.
	- Added a structured logging mechanism for improved tracking and debugging.

### DIFF
--- a/cognee/tasks/repo_processor/local_script_dependencies.py
+++ b/cognee/tasks/repo_processor/local_script_dependencies.py
@@ -1,13 +1,14 @@
-from contextlib import contextmanager
-from typing import List, Dict, Optional
+import argparse
 import asyncio
+import sys
+from contextlib import contextmanager
+from pathlib import Path
+from typing import List, Dict, Optional
+
 import aiofiles
 import jedi
 import parso
-import sys
-from pathlib import Path
 from parso.tree import BaseNode
-
 
 @contextmanager
 def add_sys_path(path):
@@ -91,10 +92,20 @@ async def get_local_script_dependencies(script_path: str, repo_path: Optional[st
         dependencies = await _extract_dependencies(script_path)
     return dependencies
 
+
 if __name__ == "__main__":
-    # Simple execution example, use absolute paths
-    script_path = ".../cognee/examples/python/simple_example.py"
-    repo_path = ".../cognee"
+    parser = argparse.ArgumentParser(description="Get local script dependencies.")
+
+    # Suggested path: .../cognee/examples/python/simple_example.py
+    parser.add_argument("script_path", type=str, help="Absolute path to the Python script file")
+
+    # Suggested path: .../cognee
+    parser.add_argument("repo_path", type=str, help="Absolute path to the repository root")
+
+    args = parser.parse_args()
+
+    script_path = args.script_path
+    repo_path = args.repo_path
 
     dependencies = asyncio.run(get_local_script_dependencies(script_path, repo_path))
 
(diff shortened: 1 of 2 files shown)


cognee/tasks/repo_processor/local_script_dependencies.py
```diff
@@ -0,0 +1,83 @@
+from typing import List, Dict, Optional
+import jedi
+import parso
+import sys
+from pathlib import Path
+from parso.tree import BaseNode
+
+
+def get_code_entities(node: parso.tree.NodeOrLeaf) -> List[Dict[str, any]]:
```
alekszievr, line 9: Doesn't this function need to be async?

cognee/tasks/repo_processor/local_script_dependencies.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    Recursively extract code entities using parso.
+    """
+    code_entity_list = []
+
+    if not hasattr(node, 'children'):
+        return code_entity_list
+
+    # Process nodes of type 'name', which correspond to code entities
```
alekszievr, line 18: unnecessary comment

hajdul88 (dismissed):

cognee/tasks/repo_processor/local_script_dependencies.py
lxobr: Turns out it does, thanks for the question! Updated it.

cognee/tasks/repo_processor/local_script_dependencies.py
```diff
@@ -0,0 +1,103 @@
...
+        with add_sys_path(repo_path_resolved):
+            dependencies = await _extract_dependencies(script_path)
+        dependencies = [path for path in dependencies if path.startswith(repo_path_resolved)]
+    else:
+        dependencies = await _extract_dependencies(script_path)
+    return dependencies
+
+if __name__ == "__main__":
+    # Simple execution example, use absolute paths
+    script_path = ".../cognee/examples/python/simple_example.py"
```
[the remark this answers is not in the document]
lxobr: Absolutely! Switched to `argparse` to handle it. Let me know if it works for you!

cognee/tasks/repo_processor/local_script_dependencies.py
```diff
@@ -0,0 +1,114 @@
...
+
+    module_paths = {
+        entity.get("module_path")
+        for entity in code_entities
+        if entity.get("module_path")
+    }
+
+    return sorted(str(path) for path in module_paths)
+
+async def get_local_script_dependencies(script_path: str, repo_path: Optional[str] = None) -> List[str]:
```
borisarzentar, line 82: Let's call the file also `get_local_script_dependencies.py`. Files should be named after functions or classes they contain.

cognee/tasks/repo_processor/local_script_dependencies.py
```diff
@@ -0,0 +1,114 @@
+import argparse
```
borisarzentar, line 1: Since this is used for code and I expect more tasks like this, we can put these kind of tasks inside `/cognee/tasks/code` directory.

cognee/tasks/repo_processor/local_script_dependencies.py
borisarzentar: Why `local` in the name?

PR #230, opened 2024-11-15
Cog 533 pydantic unit tests

0xideas:
## Release Notes

- **New Features**
  - Introduced a benchmarking script to evaluate the performance of graph generation based on recursive organizational structures.
  - Added new utility functions for creating recursive society structures and counting entities within them.
  - Implemented a new test suite to validate the functionality of the graph model generation.

- **Bug Fixes**
  - Enhanced type safety in handling graph edges and nodes.

- **Refactor**
  - Simplified logic in the `get_graph_from_model` and `get_model_instance_from_graph` functions for better readability and maintainability.
  - Updated test functions to improve structure and focus on specific relationships within the graph model.
  - Ensured models are rebuilt before being returned in the storage utility functions.

- **Documentation**
  - Improved test assertions and messages for better debugging and clarity.


0xideas:
Hey @borisarzentar,

I created two pydantic Classes, `Organization` and `SocietyPerson`, that contain each other. A nested pydantic structure of this kind is created by calling `create_organization_recursive` in the file `get_model_instance_from_graph_generative_test.py`.

Weirdly, the `get_model_instance_from_graph` call fails in Python 3.9 and 3.10, but succeeds in 3.11. The issue seems to be that the pydantic class `SocietyPerson` is undefined in the call to `model = create_model(model.__name__, **final_fields)` for reconstructing `Organization`. Python 3.11 seems to have solved this issue, although I'm unsure why/how. For now, I'm just skipping these tests for Python 3.9 and 3.10 and I have added a [bug ticket](https://linear.app/cognee/issue/COG-649/bug-recursive-pydantic-structure-cant-be-reconstructed-from-graph-in).

Thanks!
Leon

cognee/modules/graph/utils/get_model_instance_from_graph.py
```diff
@@ -1,29 +1,56 @@
...
 
         if edge_type == "list":
-            NewModel = copy_model(type(source_node), { edge_label: (list[type(target_node)], PydanticUndefined) })
-
-            node_map[edge[0]] = NewModel(**source_node.model_dump(), **{ edge_label: [target_node] })
+            NewModel = copy_model(
+                type(source_node),
+                {edge_label: (list[type(target_node)], PydanticUndefined)},
+            )
+            new_model_dict = merge_dicts(
```
borisarzentar, line 41: `merge_dicts` seems like an overkill here, we want to merge just one property here. It will iterate over properties of both objects for one property, right?

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -1,8 +1,16 @@
 from datetime import datetime, timezone
+
 from cognee.infrastructure.engine import DataPoint
 from cognee.modules.storage.utils import copy_model
 
-def get_graph_from_model(data_point: DataPoint, include_root = True, added_nodes = {}, added_edges = {}):
+
+def get_graph_from_model(data_point: DataPoint, added_nodes=None, added_edges=None):
+
+    if not added_nodes:
```
borisarzentar, line 9: Is there a difference between this and previous code in how the empty dict is created or function is called?

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -12,87 +20,94 @@ def get_graph_from_model(data_point: DataPoint, include_root = True, added_nodes
...
 
-    if include_root:
-        nodes.append(SimpleDataPointModel(**data_point_properties))
+    nodes.append(SimpleDataPointModel(**data_point_properties))
 
     return nodes, edges
 
 
+def add_nodes_and_edges(
+    data_point, field_name, field_value, nodes, edges, added_nodes, added_edges
```
borisarzentar, line 68: Do we actually need nodes and edges here? Could we hold that data in `added_nodes` and `added_edges`? Just thinking if we can make it simpler.

0xideas: There is also an implementation where we don't have separate `nodes` and `added_nodes`, and `edges` and `added_edges`, but I tried one and ran into difficulties. I expect the reason why they were separated in the first place is because it's not ideal to pass down a very 'fat' dictionary in recursive function calls, and the `added_` dicts are a lot thinner. My view is that either approach is fine, but currently we have this one and imo it's not worth the effort to change it at this point in time.

cognee/modules/graph/utils/get_model_instance_from_graph.py
0xideas: Yes, good point, I changed it to merge the key values directly.

0xideas (commented):
Thanks for the comments! I removed the merge_dicts call.

cognee/modules/graph/utils/get_graph_from_model.py
borisarzentar: Whaaat? This is the most bizarre thing I heard about any programming language. :D Ok, good to know. I'm not sure that other people know this, so please share it.

cognee/modules/graph/utils/get_graph_from_model.py
borisarzentar: Yeah, ok. Let's keep it this way then.

borisarzentar (approved):

PR #235, opened 2024-11-18
feat: add vector and graph dbs state to README file

borisarzentar:


borisarzentar:
I am setting LanceDB and NetworkX to be default in template env file because that is "no setup" configuration.

0xideas (approved):
Looks good to me!

PR #268, opened 2024-12-09
COG-810 Implement a top-down dependency graph builder tool

lxobr:
- Generates a dictionary representing an (approximate) call graph for a given repository path.
- Dictionary structure:
  - Keys: Paths to Python files in the repository.
  - Values: Lists of subdictionaries, each representing an importable entity.

## Summary by CodeRabbit

- **New Features**
	- Introduced functionality for parsing Python source code to extract details about top-level objects, including functions, classes, and variable definitions.
	- Added methods for identifying entity usages within a project and aggregating results from multiple files.
	- Enhanced parsing capabilities with new functions for object extraction and usage tracking.
- **Bug Fixes**
	- Implemented error handling to improve robustness during parsing and extraction processes.


alekszievr:
@lxobr can I test this in the code graph pipeline? Does it replace get_repo_file_dependencies directly?

lxobr:
@alekszievr The answer to both questions is no. The output of this function has a different structure than `get_repo_file_dependencies`. It is meant to be tested on its own using a local repo path, and will be integrated into the pipeline in a separate PR.

cognee/tasks/repo_processor/top_down_repo_parse.py
```diff
@@ -0,0 +1,171 @@
...
+    EXCLUDE_DIRS = {'venv', '.git', '__pycache__', 'build'}
+
+    python_files = [
+        os.path.join(directory, file)
+        for directory, _, filenames in os.walk(repo_path)
+        if not any(excluded in directory for excluded in EXCLUDE_DIRS)
+        for file in filenames
+        if file.endswith(".py")
+    ]
+
```
alekszievr, line 164: Empty files should be filtered somewhere, otherwise they will cause an error at the embedding step. This would be a good place to filter them.

cognee/tasks/repo_processor/top_down_repo_parse.py
```diff
@@ -0,0 +1,171 @@
...
+    logger.debug(f"Inferred entity: {inferred[0].name}, type: {inferred[0].type}")
+
+    try:
+        references = script.get_references(line=line, column=column, scope="project", include_builtins=False)
+    except Exception as e:
+        logger.error(f"Error retrieving references for entity at {module_path}:{line},{column}: {e}")
+        references = []
+
+    # for ref in tqdm(references, desc="Processing references"):
```
alekszievr, line 120: delete this comment line

cognee/tasks/repo_processor/top_down_repo_parse.py
lxobr: Nice suggestion!

borisarzentar:
@lxobr What is your plan with this one? Do we want to benchmark it first or it is good to be merged like this?

lxobr:
@borisarzentar I would like to merge it as is. There are follow up tasks that will improve on this and integrate it into the library.

PR #289, opened 2024-12-09
feat: outsources chunking parameters to extract chunk from documents …

hajdul88:
…task

## Summary by CodeRabbit

- **New Features**
	- Introduced a new `ChunkerConfig` class for flexible chunking strategy selection.
	- Updated document processing methods to support dynamic chunker selection.

- **Bug Fixes**
	- Enhanced error handling for unsupported chunker names.

- **Documentation**
	- Updated method signatures across various document classes to include a new `chunker` parameter.

- **Chores**
	- Removed direct dependencies on `TextChunker`, streamlining chunking implementation.


cognee/modules/data/processing/document_types/TextDocument.py
```diff
@@ -15,6 +15,8 @@ def get_text():
 
                     yield text
 
-        chunker = TextChunker(self, chunk_size = chunk_size, get_text = get_text)
+        chunker_func = ChunkerConfig.get_chunker(chunker)
```
borisarzentar, line 18: Why we don't pass the `chunker` function itself to `read` method?
That way we don't have to have this config and chunker can be customized.

dexters1:
What about the UnstructuredDocument class? Do these changes need to be implemented there as well?

hajdul88:
> What about the UnstructuredDocument class? Do these changes need to be implemented there as well?

If we outsource chunkers like this, then yes, but we discuss this with Boris.

cognee/modules/data/processing/document_types/TextDocument.py
borisarzentar: Do we want that map? It is an additional step to do in order to use a new chunker. What I see is that we need some chunker interface, so that all custom chunkers we add are compatible with the core.

hajdul88: Lets discuss this in person

borisarzentar (approved):

PR #293, opened 2024-12-10
COG 870 Remove duplicate edges from the code graph

lxobr:
- Refactored the `summarize_code` task into a generator that does not call the `add_points` function.
- `summarize_code` now:
  - Yields input data points.  
  - Creates and yields summary data points.  
  
  This ensures all data points reach the `add_datapoints` function.
- Moved the `add_datapoints` task to the end of the pipeline.
- Extracted `run_code_graph_pipeline` into a standalone function.
- Added a script example for running the code graph pipeline with the new function.

## Summary by CodeRabbit

- **New Features**
	- Introduced a new Python script for executing a code graph pipeline.
	- Added a new function to handle the code graph processing pipeline.

- **Improvements**
	- Updated method signatures for clarity and consistency.
	- Enhanced error handling and streamlined function calls for better performance.

- **Bug Fixes**
	- Adjusted formatting and logic within existing functions to improve functionality. 

- **Deprecation Notices**
	- Marked existing functions as deprecated in favor of new implementations. 

- **Documentation**
	- Updated comments and structure for better readability and understanding.


evals/eval_swe_bench.py
```diff
@@ -41,7 +41,7 @@ def check_install_package(package_name):
             return False
 
 
-async def generate_patch_with_cognee(instance, llm_client, search_type=SearchType.CHUNKS):
+async def run_code_graph_pipeline(repo_path):
```
alekszievr, line 44: this should be in cognee/api/v1/cognify/code_graph_pipeline.py

alekszievr: and ../../cognee/notebooks/cognee_code_graph_demo.ipynb should also call this function

examples/python/code_graph_example.py
```diff
@@ -0,0 +1,13 @@
...
+import asyncio
+from evals.eval_swe_bench import run_code_graph_pipeline
+
+async def main(repo_path):
+    async for result in await run_code_graph_pipeline(repo_path):
+        print(result)
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("repo_path", type=str, help="Path to the repository")
```
alekszievr, line 11: I prefer arguments starting with "--", but it doesn't really matter while there is only one argument.

evals/eval_swe_bench.py
lxobr: Great points! I moved the function to `code_graph_pipeline.py` for now. I propose creating a separate issue to clean up that file and update the notebook, though. If you are fine with this please let me know and I will create the issue and resolve this conversation.

cognee/tasks/summarization/summarize_code.py
```diff
@@ -1,39 +1,42 @@
...
 from cognee.shared.CodeGraphEntities import CodeFile
-from cognee.tasks.storage import add_data_points
-
 from .models import CodeSummary
 
 
 async def summarize_code(
-    code_files: list[DataPoint],
-    summarization_model: Type[BaseModel],
+        code_graph_nodes: list[DataPoint],
```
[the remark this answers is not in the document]
lxobr: Good catch, thanks!

cognee/tasks/summarization/summarize_code.py
```diff
@@ -1,39 +1,42 @@
...
+    }
+
+    for node in code_graph_nodes:
+        if not isinstance(node, DataPoint):
+            continue
+        yield node
+
+        if not isinstance(node, CodeFile):
+            continue
+        yield CodeSummary(
```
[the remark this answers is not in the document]
lxobr: You are absolutely right, updated it.

examples/python/code_graph_example.py
lxobr: Absolutely, it's neater. Improved it

evals/eval_swe_bench.py
alekszievr: code_graph_pipeline.py is quite confusing now that it contains both implementations, I wouldn't leave it like this :(
These are the options I see, from least effort to most effort:
- Leave it as is, but add big bold docstrings to the old functions saying "deprecated, use run_code_graph_pipeline instead"
- Remove the old functions, leave the notebook as is, but test the new implementation in test_code_generation.py, otherwise it will break. I would choose this option.
- Remove old functions, update both the notebook and the tests

lxobr: Thanks, @alekszievr! I will go for the least effort option for now, and then we can all sync about it for the next cycle. 
I created [a new issue](https://linear.app/cognee/issue/COG-906/update-cognifycode-graph-pipelinepy-and-related-code) to address this. In the meanwhile, I added some comments and updated the docstring as you suggested.

alekszievr (approved):

borisarzentar (approved):

PR #372, opened 2024-12-14
Update get_cognify_router.py with graph endpoint

Vasilije1990:
## Summary by CodeRabbit

- **New Features**
	- Introduced a new optional parameter, `graph_model`, to the cognify function, enhancing its functionality.

- **Workflow Improvements**
	- Conditional execution added to multiple test workflows, ensuring they run only if the `run-checks` label is present on the pull request.

### DIFF
--- a/cognee/api/v1/cognify/routers/get_cognify_router.py
+++ b/cognee/api/v1/cognify/routers/get_cognify_router.py
@@ -1,5 +1,5 @@
 from fastapi import APIRouter
-from typing import List
+from typing import List, Optional
 from pydantic import BaseModel
 from cognee.modules.users.models import User
 from fastapi.responses import JSONResponse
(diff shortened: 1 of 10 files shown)


cognee/api/v1/cognify/routers/get_cognify_router.py
```diff
@@ -6,22 +6,24 @@
...
 class CognifyPayloadDTO(BaseModel):
     datasets: List[str]
+    graph_model: BaseModel
 
 def get_cognify_router() -> APIRouter:
     router = APIRouter()
 
     @router.post("/", response_model=None)
-    async def cognify(payload: CognifyPayloadDTO, user: User = Depends(get_authenticated_user)):
+    async def cognify(payload: CognifyPayloadDTO, user: User = Depends(get_authenticated_user), graph_model:BaseModel=None):
```
dexters1, line 18: If graph_model is part of the CognifyPayloadDTO I don't think we need it as an input parameter in the endpoint.

cognee/api/v1/cognify/routers/get_cognify_router.py
```diff
@@ -6,22 +6,24 @@
 from cognee.modules.users.methods import get_authenticated_user
 from fastapi import Depends
 
+
 class CognifyPayloadDTO(BaseModel):
     datasets: List[str]
+    graph_model: BaseModel
```
dexters1, line 12: If we want to have graph_model be optional we should add the type in an Optional wrapper

Vasilije1990: fixed

dexters1 (approved):

PR #382, opened 2024-12-18
Cog 650 replace pylint

dexters1:
- add check for ruff formatting
- add check for ruff linting
- add git hook for ruff ( run locally: `pre-commit install` to add git hook for ruff. Make sure to run poetry install beforehand to install pre-commit )

- Run `ruff check` to lint code in local
- Run `ruff format` to format code in local

## Summary by CodeRabbit

- **New Features**
	- Introduced new GitHub Actions workflows for code formatting and linting using Ruff.
	- Added pre-commit hooks for enhancing code quality and consistency.
  
- **Dependencies**
	- Added `pre-commit` as a new dependency for the project.
  
- **Configuration**
	- Updated linting rules and configurations for Ruff in the project settings, including new ignore rules.


.github/workflows/ruff_format.yaml
```diff
@@ -0,0 +1,11 @@
+name: ruff format
```
borisarzentar, line 1: I don't think we want to format the code with ruff in pull request. Linting should happen and notify developer if errors are there. Developers should fix those themselves.

dexters1: This is not formatting code in the PR itself, just checking if it is formatted properly.

I've added a git hook which should be installed to verify this as well, would be good to keep the code formatted uniformly in the future.

dexters1: if a developer disagrees with the proposed formatting he can add format suppression as described here:

https://docs.astral.sh/ruff/formatter/#format-suppression

PR #401, opened 2024-12-30
feat: adding ruff, visualization, and fixing ollama

Vasilije1990:
## Summary by CodeRabbit

- **New Features**
	- Enhanced graph visualization capabilities with new functions for styling, rendering, and logo embedding
	- Added advanced graph manipulation utilities for serialization and layout generation
	- Introduced network visualization with customizable node colors and centrality measures


pyproject.toml
```diff
@@ -18,24 +18,24 @@ classifiers = [
 ]
 
 [tool.poetry.dependencies]
-python = ">=3.9.0,<3.12"
-openai = "1.52.0"
+python = ">=3.10.0,<3.12"
```
dexters1, line 21: Did we intentionally raise minimum python version to 3.10? Should we enable the usage of python 3.12 if compatible?

Vasilije1990: Yes, bokeh is not supported for 3.9

Vasilije1990: 3.12 can work, need to find which lib is causing segmentation fault now

dexters1: It doesn't cause a segmentation fault for me when running locally on 3.12 with poetry version 1.8.5

Vasilije1990: Neither does it for me. But actions seem to fail

dexters1 (approved):

PR #415, opened 2025-01-07
fix: Fixes duplicated edges in cognify by limiting the recursion dept…

hajdul88:
…h in add datapoints

## Summary by CodeRabbit

- **New Features**
  - Added an optional `only_root` parameter to control data point processing behavior across multiple functions
  - Enhanced graph model generation with more flexible node and edge handling

- **Refactor**
  - Updated method signatures in data processing pipeline to support more granular control of data point inclusion

### DIFF
--- a/cognee/api/v1/cognify/cognify_v2.py
+++ b/cognee/api/v1/cognify/cognify_v2.py
@@ -94,7 +94,7 @@ async def run_cognify_pipeline(dataset: Dataset, user: User, graph_model: BaseMo
                 summarization_model = cognee_config.summarization_model,
                 task_config = { "batch_size": 10 }
             ),
-            Task(add_data_points, task_config = { "batch_size": 10 }),
+            Task(add_data_points, only_root = True, task_config = { "batch_size": 10 }),
         ]
 
         pipeline = run_tasks(tasks, data_documents, "cognify_pipeline")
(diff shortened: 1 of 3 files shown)


Vasilije1990 (commented):
@hajdul88 adding a boolean seems a bit hacky. Should it not work on any depth and the problem be solved generally instead of limiting recursion?

hajdul88:
> @hajdul88 adding a boolean seems a bit hacky. Should it not work on any depth and the problem be solved generally instead of limiting recursion?

Let's discuss this tomorrow. The main problem is that when we create the chunk descriptions, the whole nested structure (which was already added to the graph db) is added to the TextSummary. Since we use the same add_datapoint to add summaries, it extracts the subgraphs from the DocumentChunks and adds them again. The add datapoints is therefore used twice on the same structure.

There are multiple ways to solve this:

1. We add everything at the end, but then we have to have everything as a nested object passed through the whole pipeline (some edges are not added to the document chunks). I think it's not a good solution and also increases the mass we have to drag til the end of the pipeline. Also some edges only exist in a previous step but are not the part of the nested structure.
2. We introduce two ways to use the get graph from model method. First, when we don't limit the recursion, we add the whole nested structure, and second, when we limit the recursion and add only one layer. In this way we can use it to add the entire structure or just one layer.
3. We implement a new get graph from model that only adds one layer, but then its gonna be exactly the same. I think we need the functionality of adding only the outer layer from the nested objects.

Vasilije1990 (approved):

PR #418, opened 2025-01-08
Feat/cog 946 abstract eval dataset

alekszievr:
QA eval dataset as argument, with hotpot and 2wikimultihop as options. Json schema validation for datasets.

## Summary by CodeRabbit

- **New Features**
	- Enhanced evaluation script to support multiple QA datasets
	- Added dynamic dataset selection through command-line arguments
	- Implemented dataset download and loading mechanisms

- **Improvements**
	- Generalized evaluation function to work with different question-answering datasets
	- Improved dataset handling with metadata and validation


evals/qa_dataset_utils.py
```diff
@@ -0,0 +1,74 @@
+from cognee.root_dir import get_absolute_path
+import json
+import wget
```
lxobr, line 3: `wget` is not in `pyproject.toml`. Also, it's a package that has not been maintained since 2015.  Maybe we can use the built in `requests.get` instead of `wget.download`?

lxobr (commented):
Nice work!

lxobr (approved):

PR #422, opened 2025-01-09
fix: Fix ubuntu 24.04 segmentation fault

dexters1:
Current testing seems to suggest issues with running Cognee on Ubuntu 24.04 when using SQLite. When running Cognee on Ubuntu 24.04 with PostgreSQL there are no segmentation fault issues.

## Summary by CodeRabbit

- **Infrastructure**
  - Updated GitHub Actions workflows to use `ubuntu-latest` across multiple workflow files
  - Added a new `trigger_deployment` job in the continuous deployment workflow

- **Code Optimization**
  - Updated SQLAlchemy query loading strategy in user retrieval method from `joinedload` to `selectinload`


dexters1:
For some reason unknown to me using the joinedload for getting the default user would cause a segmentation fault in Ubuntu 24.04, problem is solved by using the selectinload option for joining data

Vasilije1990 (approved):

PR #442, opened 2025-01-15
COG-989 feat: make tasks a configurable argument in the cognify function

lxobr:
## Summary by CodeRabbit

- **New Features**
  - Enhanced `cognify` function with optional custom task list support.
  - Added ability to specify and validate custom tasks during data processing.
  - Introduced default task generation mechanism when no tasks are provided.

- **Improvements**
  - Updated function signatures to provide more flexibility in task management.
  - Improved error handling and telemetry for task creation.

### DIFF
--- a/cognee/api/v1/cognify/cognify_v2.py
+++ b/cognee/api/v1/cognify/cognify_v2.py
@@ -160,6 +160,7 @@ async def get_default_tasks(
                 summarization_model=cognee_config.summarization_model,
                 task_config={"batch_size": 10},
             ),
+            Task(add_data_points, only_root=True, task_config={"batch_size": 10}),
         ]
     except Exception as error:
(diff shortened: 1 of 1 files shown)


cognee/api/v1/cognify/cognify_v2.py
```diff
@@ -55,18 +56,19 @@ async def cognify(
 
     awaitables = []
 
+    if tasks is None:
```
hajdul88, line 59: I wouldn't necessarily do this;, it seems we are just outsourcing the pipeline into a function. I think it was good and clean as it was before and this can also cause misunderstanding and an added layer of complexity for the users and to the examples we have. But It's just my opinion, maybe I am not right.

hajdul88: After getting the context, now I understand and agree with Lazar.

borisarzentar (approved):

PR #480, opened 2025-01-28
Cog 1055 node extraction task

lxobr:
## Description

- Created a node extraction tasks that performs multiple rounds of extractions
- Created a corresponding prompt
- Included a simplified version of the cognifypipeline to test integration

## DCO Affirmation


cognee/modules/data/extraction/knowledge_graph/extract_content_graph.py
```diff
@@ -13,3 +14,36 @@ async def extract_content_graph(content: str, response_model: Type[BaseModel]):
...
+    existing_names = set()  # Track existing entity names in lowercase
+
+    for round_num in range(n_rounds):
+        context = {
+            "previous_entities": all_entities,
+            "round_number": round_num + 1,
+            "total_rounds": n_rounds,
+        }
+
+        system_prompt = render_prompt("extract_graph_nodes_prompt.txt", context)
```
alekszievr, line 38: The context is not being rendered into the prompt because the prompt is missing the fields where the context should be added, e.g. `{{ all_entities }}`. 

See cognee/infrastructure/llm/prompts/context_for_question.txt for reference.

alekszievr: I would render both the content and the context into user prompt ("text_input") rather than system prompt.

alekszievr:
This is out of this PR's diff range, but in extract_content_graph.py in extract_content_graph()
instead of 
`system_prompt = render_prompt("generate_graph_prompt.txt", {})`
we should use
`system_prompt = read_query_prompt("generate_graph_prompt.txt")`

cognee/tasks/graph/extract_nodes_from_data.py
```diff
@@ -0,0 +1,27 @@
...
+    """Extracts and integrates entities from document chunks using multi-round entity extraction."""
+    chunk_entities = await asyncio.gather(
+        *[extract_content_nodes(chunk.text, n_rounds) for chunk in data_chunks]
+    )
+
+    # Update chunks with their entities and collect all entities for storage
+    all_entities = []
+    for chunk, entities in zip(data_chunks, chunk_entities):
+        chunk.contains = entities
+        all_entities.extend(entities)
```
borisarzentar, line 25: Why do we have this `all_entities` array?

lxobr: Good catch. It's a leftover from another thing I tried. Will remove it.

cognee/modules/data/extraction/knowledge_graph/extract_content_graph.py
lxobr: Thanks, this was very useful! I improved it.

alekszievr:
> This is out of this PR's diff range, but in extract_content_graph.py in extract_content_graph() instead of `system_prompt = render_prompt("generate_graph_prompt.txt", {})` we should use `system_prompt = read_query_prompt("generate_graph_prompt.txt")`

what about this, @lxobr ?

lxobr:
@alekszievr, makes sense to me. As it is part of the core functionality, I would prefer for @borisarzentar to decide whether to do it here or elsewhere.

alekszievr (approved):

PR #481, opened 2025-01-28
Feat: add number of tokens and descriptive graph metrics to metric table [COG-1132]

alekszievr:
In this PR, num_nodes and num_edges are calculated only as a placeholder for all the descriptive metrics which are coming in another PR.

## DCO Affirmation

## Summary by CodeRabbit

- **New Features**
	- Added a new task to store descriptive metrics for data points.
	- Introduced functions to fetch token counts and calculate graph metrics.

- **Improvements**
	- Enhanced data model flexibility by allowing nullable metric fields.
	- Updated database schema to support more dynamic metric storage.
	- Expanded data tracking capabilities with new metric calculation methods.


alekszievr:
image

cognee/tasks/storage/descriptive_metrics.py
```diff
@@ -0,0 +1,46 @@
...
+    """
+    Fetches and sums token counts from the database.
+
+    Returns:
+        int: The total number of tokens across all documents.
+    """
+
+    async with db_engine.get_async_session() as session:
+        document_data_points = await session.execute(select(Data))
+        token_count_sum = sum(document.token_count for document in document_data_points.scalars())
```
borisarzentar, line 21: We can do sum in `sql` as well and return just that number instead of all Data objects.

cognee/tasks/storage/descriptive_metrics.py
```diff
@@ -0,0 +1,46 @@
...
+async def calculate_graph_metrics(graph_data):
+    nodes, edges = graph_data
+    graph_metrics = {
+        "num_nodes": len(nodes),
+        "num_edges": len(edges),
+    }
+    return graph_metrics
+
+
+async def store_descriptive_metrics(data_points: list[DataPoint]):
```
borisarzentar, line 35: Should we get metrics only for the documents that were processed?

cognee/modules/data/models/MetricData.py
```diff
@@ -1,6 +1,6 @@
 from datetime import datetime, timezone
```
borisarzentar, line 1: Name of the file is `MetricData` and name of the class inside is `GraphMetricData`, not aligned.
Please either rename the class or the file. And since we store metrics, let's make it plural, like `GraphMetrics`.

cognee/tasks/storage/descriptive_metrics.py
```diff
@@ -0,0 +1,46 @@
...
+    graph_engine = await get_graph_engine()
+    graph_data = await graph_engine.get_graph_data()
+
+    token_count_sum = await fetch_token_count(db_engine)
+    graph_metrics = await calculate_graph_metrics(graph_data)
+
+    table_name = "graph_metric_table"
+    metrics_dict = {"id": uuid.uuid4(), "num_tokens": token_count_sum} | graph_metrics
+
+    await db_engine.insert_data(table_name, metrics_dict)
```
borisarzentar, line 46: It's always good to return `data_points` at the end of the task, in case someone puts another task after this.

borisarzentar (changes requested):

cognee/tasks/storage/descriptive_metrics.py
alekszievr: done!

cognee/tasks/storage/descriptive_metrics.py
alekszievr: No, we will get graph metrics as well (e.g. number of nodes and edges, mean degree, etc.) 
Maybe I don't understand the question here

cognee/modules/data/models/MetricData.py
alekszievr: done

cognee/tasks/storage/descriptive_metrics.py
alekszievr: fixed

lxobr (approved):

PR #504, opened 2025-02-06
Comment out the postgres configuration from docker-compose.yml

soekja:
## Description
Comment out the postgres configuration from docker-compose.yml.  I think leaving postgres details commented is better than removing them entirely. This keeps simple optionality visible for newcomers.

## DCO Affirmation

## Summary by CodeRabbit

- Chores  
  - Improved service configuration by standardizing resource allocation settings.  
  - Temporarily deactivated a non-essential database service along with its related storage setup, while keeping reactivation instructions for future needs.


dexters1:
I wouldn't comment out postgres from docker compose since we use it all the time, you can run docker compose with specifying only the containers you need if you don't need postgres.

soekja:
I understand you guys are using postgres, but I was approaching this from the perspective of the experience of new community members who may not have docker experience or technical background. In that frame of mind, default should enable smoothest onboarding in the SQLite-NetworkX-LanceDB scenario, hence leaving *postgress details visible but commented.

Vasilije1990 (approved):

PR #519, opened 2025-02-10
feat: eval with modal cognify25q1 [COG-1184]

lxobr:
## Description
- Implemented minimal functionality to run QA evaluations in parallel using Modal.
- Refactored parts of the existing evaluation framework to expose functionality for parallel execution.

## DCO Affirmation


cognee/tasks/graph/extract_graph_from_data_chunks.py
```diff
@@ -0,0 +1,50 @@
...
+    data_chunks: List[DocumentChunk], n_rounds: int = 1
+) -> List[DocumentChunk]:
+    """Extract and update graph data from document chunks in multiple steps."""
+    chunk_nodes = await asyncio.gather(
+        *[extract_nodes(chunk.text, n_rounds) for chunk in data_chunks]
+    )
+
+    chunk_results = await asyncio.gather(
+        *[
+            extract_content_nodes_and_relationship_names(chunk.text, nodes, n_rounds)
```
hajdul88, line 25: Why do we separate relationship name extractrion from  edge triplet extraction?

cognee/tasks/graph/extract_graph_from_data_chunks.py
```diff
@@ -0,0 +1,50 @@
...
+from cognee.tasks.storage import add_data_points
+from cognee.tasks.graph.extract.extract_nodes import extract_nodes
+from cognee.tasks.graph.extract.extract_relationship_names import (
+    extract_content_nodes_and_relationship_names,
+)
+from cognee.tasks.graph.extract.extract_edge_triplets import extract_edge_triplets
+
+
+async def extract_graph_from_data_chunks(
+    data_chunks: List[DocumentChunk], n_rounds: int = 1
```
hajdul88, line 16: Here n_rounds is 1 but then the initial value in the extract tasks is 2. What's the golden rule for this parameter based on your experience?

evals/eval_on_hotpot.py
```diff
@@ -117,19 +148,21 @@ async def eval_on_QA_dataset(
...
         promptfoo_results = await eval_metrics["promptfoo_metrics"].measure(
             instances, context_provider, contexts_filename
         )
     else:
         promptfoo_results = {}
 
-    answers_filename = out_path / Path(
-        f"answers_{dataset_name_or_filename.split('.')[0]}_{context_provider_name}.json"
-    )
+    # answers_filename = out_path / Path(
```
hajdul88, line 162: This remained here

hajdul88 (commented):
In general, I didn't run the code I have just some comments in general:

- I think I understand what it does, but I can't decide whether if its better than cognify so it would be nice to see quality and running time comparisons. 

- I imagine this as a parallel direction next to cognify so I wouldn't replace it even if its working better for now, because this pipeline limits the future flexibility of the graph creation. The way we build the graph here is a few shot learning based version focusing on prompting the extraction in different rounds. I imagine this as an additional direction that we explore parallel to the one we have. And yes its possible that this is gonna be the best, I would just want to keep cognify as a baseline to go to different directions from there for now.

Additional:
This PR also changes a lot in the old eval, do we want to keep that? For me it seems its the merge of two topics, eval and the new 25q1 pipeline.

lxobr:
@hajdul88 - Great comments! I agree with most. Indeed, this is just an experimental exploration at this point, which is why we are merging to a separate branch. Once we see the effect on the metrics we can see what we do with all of this work.

This is really exactly few-shot-learning based (we are not giving it any, nor few, reference examples), but definitely multiple round extraction.

The question about changes in the old eval approach is fair. I think it should be part of the wider conversation about what the status of the old eval code is. I think it is ripe for reorganization and/or removal.

cognee/infrastructure/llm/prompts/answer_hotpot_using_cognee_search.txt
```diff
@@ -1,2 +1,9 @@
...
\ No newline at end of file
+You are a highly optimized question-answering system that extracts only the most essential information from the provided context.
+
+Rules:
+	•	Minimize words – Use as few as possible
+	•	Yes/no questions → Answer with only “yes” or “no”
+	•	What/Who/Where questions → Answer with a single word or phrase (no full sentences).
+	•	When questions → Give only the relevant date, time, or period.
+	•	How/Why questions → Answer with the shortest possible phrase.
+	•   No punctuation, just the answer, preferably in dry, concise lowercase
```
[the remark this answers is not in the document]
lxobr: Noted, will fix it.

evals/qa_context_provider_utils.py
```diff
@@ -127,6 +138,18 @@ async def get_context_with_brute_force_triplet_search(instance: dict) -> str:
...
 
+async def get_cognee_25q1_context(instance):
+    return await get_context_with_cognee(instance, task_getter=get_25q1_tasks)
+
+
+async def get_simple_rag_25q1_context(instance):
+    return await get_context_with_simple_rag(instance, task_getter=get_25q1_tasks)
+
+
+async def get_brute_force_25q1_context(instance):
```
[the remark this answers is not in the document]
lxobr: In order to test the graph improvements and calculate metrics for both new and old approach in one go, I need to call both "old" and "new" cognify with different pipelines and different metrics. I can call with `partial`, but it seems to me some wrapping needs to happen somewhere. Am I missing something?

evals/run_hotpotqa_evals_on_instance_ids.py
```diff
@@ -0,0 +1,76 @@
...
+        result_metrics = await calculate_eval_metrics(instances, answers, eval_metrics)
+        results[rag_option] = result_metrics | {"answers": answers}
+
+    return {"instance_ids": paramset["hotpotqa_instance_ids"], "results": results}
+
+
+async def main():
+    paramset = {
+        "dataset": ["hotpotqa"],
+        "rag_option": ["cognee_25q1", "brute_force_25q1"],
```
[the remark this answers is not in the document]
lxobr: Absolutely! Good point, will iterate on this.

cognee/tasks/graph/extract_graph_from_data_chunks.py
lxobr: As this is all highly experimental, I did... well.. experiment... with different values. I will default to the general spirit of my comments: we will see what makes sense after some more testing.

cognee/tasks/graph/extract_graph_from_data_chunks.py
lxobr: Analogous to how it could make sense to separate entity type extraction from entity extraction, I decided to explore separating relationship name extraction (what type of edges do we allow) from the extraction of triplets themselves. Seemed promising, actually, will iterate on this depending on the results of metrics.

evals/qa_context_provider_utils.py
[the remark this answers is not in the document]
hajdul88: I thought it could be separated because 25q1 only contains changes in the graph creation but not in retrieval. It is not important, and it's the old eval anyhow.

cognee/tasks/graph/extract_graph_from_data_chunks.py
hajdul88: We discussed this and agree, good idea.

PR #528, opened 2025-02-12
feat: Update parameters in search API route to match search function parameters order

soekja:
## Description

**Updated handling of SearchType through the chain:**
Router receives JSON with  searchType  Enum, example: "searchType": "CHUNKS"
FastAPI converts to SearchType enum via SearchPayloadDTO
search_v2.py expects SearchType enum
search.py takes SearchType enum and extracts value
log_query.py takes string value
Query model stores string in database

**get_search_router.py**

Matched the exact field name from JSON payload searchType instead of search_type in the SearchPayloadDTO class.
Changed cognee_search() params to use payload.query and payload.searchType

**search.py**
Changed query_type to SearchType
log_query to accept query_type.value parameter instead of str(query_type)

## DCO Affirmation

## Summary by CodeRabbit

- **Refactor**
  - Updated the search functionality to improve consistency and reliability.
  - Enhanced validation by switching to stricter search type checks, ensuring only valid search types are processed.
  - Maintained robust error handling for uninterrupted search operations.


cognee/api/v1/search/routers/get_search_router.py
```diff
@@ -10,7 +10,7 @@
 
 
 class SearchPayloadDTO(InDTO):
-    search_type: SearchType
+    searchType: SearchType
```
borisarzentar, line 13: `searchType` sent from the client is automatically converted to `search_type`, you can see that there is `alias_generator=to_camel` in `InDTO`, so this is not needed.

soekja: Thank you for pointing this out, I will make the necessary adjustments.

soekja: Refactor completed on latest commit.

Vasilije1990 (approved):

PR #530, opened 2025-02-13
Cognee gui

dexters1:
## Description
Add a simple GUI to add documents to Cognee and use GRAPH_COMPLETION search to get answers

## DCO Affirmation


pyproject.toml
```diff
@@ -84,6 +84,8 @@ tree-sitter = {version = "^0.24.0", optional = true}
 tree-sitter-python = {version = "^0.23.6", optional = true}
 plotly = {version = "^6.0.0", optional = true}
 gdown = {version = "^5.2.0", optional = true}
+pyside6 = "^6.8.2.1"
```
borisarzentar, line 87: Should we make it optional? For `gui` extra?

cognee-gui.py
```diff
@@ -0,0 +1,106 @@
...
+        # Disable the entire window
+        self.setEnabled(False)
+        try:
+            await cognee.add(self.selected_file)
+            await cognee.cognify()
+        except Exception as e:
+            QMessageBox.critical(self, "Error", f"File processing failed: {str(e)}")
+        # Once finished, re-enable the window
+        self.setEnabled(True)
+
```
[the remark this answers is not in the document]
borisarzentar: @dexters1 This is a good suggestion from coderabbit, to have some kind of feedback that stuff is happening in the background. Where does it get `loading_spinner` here?

borisarzentar (dismissed):

cognee-gui.py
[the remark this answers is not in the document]
dexters1: I can try to add some loading bar, for now I've disabled the whole GUI while it's processing as a sign it's working ( and to stop users from breaking cognee )

pyproject.toml
dexters1: Good question, I'll check the size of the dependency first if it's small I'd maybe like to keep it in the base so people can easily test the GUI ( or at least add a custom error to tell people how to install it if it's not installed )

dexters1: PySide6 is huge, I've just checked, so it has to be optional

borisarzentar (approved):

borisarzentar (approved):

PR #543, opened 2025-02-17
feat: naive correctness [COG-1348]

lxobr:
## DCO Affirmation


evals/eval_framework/evaluation/evaluation_executor.py
```diff
@@ -1,19 +1,31 @@
...
+        if isinstance(evaluator_engine, EvaluatorAdapter):
             self.eval_adapter = evaluator_engine.adapter_class()
-        else:
+            return
+
+        if not isinstance(evaluator_engine, str):
             self.eval_adapter = evaluator_engine
+            return
+
+        if evaluator_engine == "Naive":
```
borisarzentar, line 17: Let's always pass the evaluator class itself, not strings. That way we remove this abstraction and we don't have to have dependencies on those classes.

PR #546, opened 2025-02-17
fix: entrypoint.sh to not fail on first docker up, improved handling of migrations, signals and errors.

soekja:
## Description

In it's current form, the entrypoint.sh script will run but fail with exit code 3 on the first docker compose up. Technically, running docker compose up a second time will not throw the same error and the application works fine. The new changes will improve the  first time user experience and improve on some other aspects.

Summary of Changes:
1- entrypoint.sh to not fail with exit code 3 on first docker up.
2- Improved error and signal handling with set -e.
3- Improved database migration, verification and error handling. Avoids schema version mismatch and ensures db schema is always in sync with application code.
4- Added exec before Gunicorn commands to ensure proper signal handling.

## DCO Affirmation

## Summary by CodeRabbit

- **Chores**
	- Improved error handling for smoother database migrations and startup.
	- Updated process management to ensure reliable application launch.
	- Optimized worker configuration and introduced a startup delay to guarantee database readiness.


entrypoint.sh
```diff
@@ -1,33 +1,34 @@
...
-#     echo "Migrations completed successfully."
-# else
-#     echo "Migration failed, exiting."
-#     exit 1
-# fi
-
+# Run Alembic migrations with proper error handling
+echo "Running database migrations..."
+MIGRATION_OUTPUT=$(poetry run alembic upgrade head 2>&1) || {
+    if [[ $MIGRATION_OUTPUT == *"UserAlreadyExists"* ]] || [[ $MIGRATION_OUTPUT == *"User default_user@example.com already exists"* ]]; then
```
borisarzentar, line 10: Can you put a comment describing the reasoning behind this check please? So it doesn't confuse us in the future.

entrypoint.sh
```diff
@@ -1,33 +1,34 @@
...
 
-    python -m debugpy --wait-for-client --listen 0.0.0.0:5678 -m gunicorn -w 3 -k uvicorn.workers.UvicornWorker -t 30000 --bind=0.0.0.0:8000 --log-level debug --reload cognee.api.client:app
-  else
-    gunicorn -w 3 -k uvicorn.workers.UvicornWorker -t 30000 --bind=0.0.0.0:8000 --log-level debug --reload cognee.api.client:app
-  fi
+# Modified Gunicorn startup with error handling
+if [ "$ENVIRONMENT" = "dev" ]; then
+    if [ "$DEBUG" = true ]; then
+        echo "Waiting for the debugger to attach..."
+        exec python -m debugpy --wait-for-client --listen 0.0.0.0:5678 -m gunicorn -w 1 -k uvicorn.workers.UvicornWorker -t 30000 --bind=0.0.0.0:8000 --log-level debug --reload cognee.api.client:app
```
borisarzentar, line 28: Why do we decrease the number of workers to 1?

entrypoint.sh
soekja: Yes I will make the updates to have the comments.

entrypoint.sh
soekja: You're right it should be set to 3 workers like the previous script.

soekja:
Updated workers count and comments for UserAlreadyExists check as requested @borisarzentar

Vasilije1990 (approved):

PR #552, opened 2025-02-18
feat: entity completion skeleton [COG-1318]

lxobr:
## Description

- Modular implementation of entity completion search
- Added base classes that define entity extractors and context providers
- Created dummy implementations that return test data
- Set up adapters that let us switch between different entity extractors and context providers using strings
- Added configuration class to control which implementations to use
- Entity completion: query → find entities → get context → interact with LLM → return answer
## DCO Affirmation

## Summary by CodeRabbit

- **New Features**
	- Enhanced the query completion experience with integrated language model response generation, improved validation, and robust error handling.
	- Introduced sample modules for context retrieval and entity extraction that simulate key processing steps.
	- Established foundational abstractions to support flexible context and entity handling strategies.


cognee/tasks/entity_completion/context_providers/base_context_provider.py
```diff
@@ -0,0 +1,13 @@
...
+from typing import List
+
+from cognee.modules.engine.models import Entity
+
+
+class BaseContextProvider(ABC):
+    """Base class for context retrieval strategies."""
+
+    @abstractmethod
+    async def get_context(self, entities: List[Entity], query: str) -> str:
```
borisarzentar, line 11: Wouldn't use `Entity` here, as it is not something that will exist every time. `DataPoint` maybe?

cognee/tasks/entity_completion/entity_completion.py
```diff
@@ -0,0 +1,119 @@
...
+        try:
+            return ContextProviderAdapter(getter)
+        except ValueError:
+            raise ValueError(f"Unsupported context provider: {getter}")
+    return getter
+
+
+async def entity_completion(
+    query: str,
+    extractor_type: Union[str, EntityExtractorAdapter] = None,
```
borisarzentar, line 72: Let's pass only the `extractor_type` class itself, not string. We don't want that complexity where we need to store adapters in order to be able to get them by string. Passing adapter class itself is flexible enough for all use cases. Same for `context_provider_type`.

cognee/tasks/entity_completion/entity_completion_config.py
```diff
@@ -0,0 +1,12 @@
+from pydantic_settings import BaseSettings, SettingsConfigDict
+
+
+class EntityCompletionConfig(BaseSettings):
```
borisarzentar, line 4: If this config class doesn't have to exist after the changes, delete it please. We are moving away from global configs.

hajdul88 (dismissed):

cognee/tasks/entity_completion/entity_completion.py
lxobr: Good point, changed it.

cognee/tasks/entity_completion/entity_completion_config.py
lxobr: I removed the config, it is more streamlined now.

cognee/tasks/entity_completion/entity_completion.py
```diff
@@ -0,0 +1,103 @@
...
+        ContextProviderAdapter,
+    )
+
+    logging.basicConfig(level=logging.INFO)
+
+    async def run_entity_completion():
+        # Uses config defaults
+        result = await entity_completion(
+            "Tell me about Einstein",
+            EntityExtractorAdapter.DUMMY.adapter_class(),
```
borisarzentar, line 98: I don't think we need those adapters. We can just pass the `EntityExtractor` and `ContextProvider` as we want when we use `entity_completion`.

cognee/tasks/entity_completion/context_providers/base_context_provider.py
```diff
@@ -0,0 +1,13 @@
+from abc import ABC, abstractmethod
+from typing import List
+
+from cognee.infrastructure.engine import DataPoint
+
+
+class BaseContextProvider(ABC):
```
borisarzentar, line 7: Let's move this stuff to infrastructure. You can create a new directory there for context and the one for retrievers as well.

borisarzentar: And file name should follow the class name, so `BaseContextProvider.py`.

cognee/tasks/entity_completion/entity_completion.py
lxobr: Adapters removed!

cognee/tasks/entity_completion/context_providers/base_context_provider.py
lxobr: I renamed the files. I also moved BaseContextProvider and BaseEntityExtractor to their respective new folders in infrastructure.

cognee/tasks/entity_completion/entity_completion.py
```diff
@@ -0,0 +1,103 @@
...
+            text_input=user_prompt,
+            system_prompt=system_prompt,
+            response_model=str,
+        )
+    except Exception as e:
+        logger.error(f"LLM response generation failed: {str(e)}")
+        raise
+
+
+async def entity_completion(
```
borisarzentar, line 46: Why in tasks? How should we use this function?

lxobr: The function implements [the third point from the issue description](https://linear.app/cognee/issue/COG-1318/develop-a-skeleton-for-new-search-type-with-entity-extraction). If `tasks` isn’t the best place, I’d appreciate your thoughts on where it should go instead.

borisarzentar: Later we can move it to modules.

borisarzentar (approved):

PR #556, opened 2025-02-18
feat: Implement optional neo4j metrics and improve tests [cog-1262]

alekszievr:
## DCO Affirmation

## Summary by CodeRabbit

- **New Features**
	- Enhanced graph analytics now offer detailed metrics—including shortest path lengths, diameter, and clustering coefficients—to provide deeper insights.
	- Added new functions for creating connected test graphs and validating metrics against predefined ground truth values.
	- Introduced a new JSON file containing metrics for connected and disconnected graph structures.

- **Improvements**
	- Updated how graphs are projected to consistently use undirected representations, ensuring more accurate and reliable metric calculations.
	- Streamlined metric consistency checks across different graph processing methods for robust, reliable results.
	- Simplified testing logic by consolidating metric assertions into a single function call.

- **Chores**
	- Removed unnecessary secret variables from the workflow configuration, potentially affecting access to certain resources.
	- Updated secret management to include the new `OPENAI_API_KEY`.


cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
```diff
@@ -564,13 +564,17 @@ async def project_entire_graph(self, graph_name="myGraph"):
             raise ValueError("No node labels or relationship types found in the database.")
```
lxobr, line 564: We could first check whether there are nodes and only then calculate relationship_types and check whether they exist. Optionally: we might extract two functions: one that gets the node labels string and the other that gets the relationship types string and do the checks within them.

cognee/tests/tasks/descriptive_metrics/metric_consistency_test.py
```diff
@@ -1,13 +1,27 @@
...
 import asyncio
 
 
-async def check_graph_metrics_consistency_across_adapters():
-    neo4j_metrics = await get_neo4j_metrics(include_optional=False)
-    networkx_metrics = await get_networkx_metrics(include_optional=False)
-    assert networkx_metrics == neo4j_metrics
+async def check_graph_metrics_consistency_across_adapters(include_optional=False):
+    connected_example = include_optional
+    neo4j_metrics = await get_metrics(
```
lxobr, line 8: It seems that the function get_metrics is always used with the same values of include_optional and connected_example, so it may be worthwile to change it and how it's called.

cognee/tests/tasks/descriptive_metrics/metrics_test_utils.py
```diff
@@ -23,3 +25,69 @@ async def create_disconnected_test_graph():
...
+    return graph_metrics
+
+
+async def assert_metrics(provider, include_optional=True):
+    connected_example = include_optional
+    metrics = await get_metrics(
+        provider=provider, connected_example=connected_example, include_optional=include_optional
+    )
+    if connected_example:
+        ground_truth_metrics = {
```
lxobr, line 61: Maybe these values should not be hardcoded here, but elsewhere, separately? And possibly in their own .json file. 
Also, since we already have a very useful docs page about the metrics, maybe we could reference it somewhere in a comment within code?

cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
```diff
@@ -636,19 +640,31 @@ async def _count_self_loops():
             result = await self.query(query)
             return result[0]["self_loop_count"] if result else 0
 
-        async def _get_diameter():
-            logging.warning("Diameter calculation is not implemented for neo4j.")
-            return -1
+        async def _get_shortest_path_lengths():
```
lxobr, line 643: There are quite a few inner functions here, we may want to consider extracting them in a separate module.
Also, a lot of the functions drop and project the graph over and over again, which seems unnecessary and probably could be done only once within the function.

PR #576, opened 2025-02-24
Feat/cog 1331 modal run eval

lxobr:
## Description
- Split metrics dashboard into two modules: calculator (statistics) and generator (visualization)
- Added aggregate metrics as a new phase in evaluation pipeline
- Created modal example to run multiple evaluations in parallel and collect results into a single combined output
## DCO Affirmation


evals/eval_framework/modal_run_eval.py
```diff
@@ -0,0 +1,113 @@
...
+)
+
+
+@app.function(image=image, concurrency_limit=2, timeout=1800, retries=1)
+async def modal_run_eval(eval_params=None):
+    """Runs evaluation pipeline and returns combined metrics results."""
+    if eval_params is None:
+        eval_params = EvalConfig().to_dict()
+
+    print(f"Running evaluation with params: {eval_params}")
```
alekszievr, line 57: Shouldn't we use logger instead of print lines?

evals/eval_framework/modal_run_eval.py
```diff
@@ -0,0 +1,113 @@
...
+            building_corpus_from_scratch=True,
+            answering_questions=True,
+            evaluating_answers=True,
+            calculate_metrics=True,
+            dashboard=False,
+        ),
+    ]
+
+    # Run evaluations in parallel with different configurations
+    modal_tasks = [modal_run_eval.remote.aio(config.to_dict()) for config in configs]
```
alekszievr, line 99: These jobs are running correctly, the output files are also generated, but the actual meaningful logs are getting lost between scary error messages. I think this is fine for now, but if there's a way to handle these errors by logging something shorter, that would be less confusing for the user.
image

evals/eval_framework/modal_run_eval.py
soobrosa: print must die IMHO

lxobr: Absolutely agree. Fixed it.

evals/eval_framework/modal_run_eval.py
lxobr: Good point, thanks! I don't know how to fix this quickly, though, so I will create a follow up issue.

alekszievr (approved):

PR #593, opened 2025-02-28
feat: user authorization [COG-1189]

dexters1:
## Description
Added user authorization through JWT header, reworked user and relevant RBAC models to accompany future User Permission system.

## DCO Affirmation

## Summary by CodeRabbit

- **New Features**
  - Introduced an automated workflow to validate server startup.
  - Added secure JWT token generation for improved session handling.
  - Enabled a new structure for permission management with role and tenant-based controls, including endpoints for creating roles, tenants, and assigning permissions.
  - Added methods for assigning default permissions to roles, tenants, and users.
  - Introduced new classes for managing default permissions for roles, tenants, and users.

- **Refactor**
  - Streamlined authentication and user management flows with enhanced error handling.

- **Tests**
  - Upgraded integration tests with improved database initialization and data pruning for a more stable environment.


cognee/modules/users/authentication/get_auth_backend.py
```diff
@@ -1,11 +1,20 @@
...
     JWTStrategy,
 )
+from datetime import datetime, timedelta
+from typing import Optional
+
+
+class CustomJWTStrategy(JWTStrategy):
+    async def write_token(self, user: str, lifetime_seconds: Optional[int] = None) -> str:
+        data = {"user_id": str(user.id), "tenant": user.tenant, "roles": user.roles}
+        return generate_jwt(data, self.encode_key, self.lifetime_seconds, algorithm=self.algorithm)
```
borisarzentar, line 17: Just to be sure that I understand it, customers will need to store jwt token on their own for each user they manage and send it with each request?

cognee/modules/users/methods/create_user.py
```diff
@@ -19,17 +26,51 @@ async def create_user(
         async with relational_engine.get_async_session() as session:
             async with get_user_db_context(session) as user_db:
                 async with get_user_manager_context(user_db) as user_manager:
+                    # Check if the tenant already exists
```
borisarzentar, line 29: Can we have this new method split from the old one? So we can turn it on/off how we want.
Maybe we won't allow role to be created if it doesn't exist for some use cases.

cognee/modules/users/authentication/get_auth_backend.py
dexters1: They won't need to store it, they will provide user information to a function (there is a different function for this) which will generate the token for them which then they will give our endpoint

cognee/modules/users/methods/create_user.py
dexters1: Yeah, I used this creation process more for testing the new model than as a permanent solution

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -1,29 +1,117 @@
...
+            raise RoleNotFoundError
+
+        permission_entity = (
+            (await db.execute(select(Permission).where(Permission.name == permission)))
+            .scalars()
+            .first()
+        )
+
+        if not permission_entity:
+            stmt = insert(Permission).values(name=permission)
```
borisarzentar, line 47: `stmt` -> `create_permission_statement`

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -1,29 +1,117 @@
...
 
 def get_permissions_router() -> APIRouter:
     permissions_router = APIRouter()
 
-    @permissions_router.post("/groups/{group_id}/permissions")
-    async def give_permission_to_group(
-        group_id: str, permission: str, db: Session = Depends(get_user_db)
+    @permissions_router.post("/roles/{role_id}/permissions")
+    async def give_default_permission_to_role(
+        role_id: str, permission: str, db: Session = Depends(get_async_session)
```
borisarzentar, line 33: Is `role_id` `UUID` instead of `str`?
Also, `permission_name`, to be explicit what it is.

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -1,29 +1,117 @@
...
     ):
-        group = (
-            (await db.session.execute(select(Group).where(Group.id == group_id))).scalars().first()
+        role = (await db.execute(select(Role).where(Role.id == role_id))).scalars().first()
+
+        if not role:
+            raise RoleNotFoundError
```
borisarzentar, line 38: Why do we raise exception when there is no role, but we don't when there is no permission?

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -1,29 +1,117 @@
...
+
+        try:
+            # add default permission to role
+            await db.execute(
+                insert(RoleDefaultPermission).values(
+                    role_id=role.id, permission_id=permission_entity.id
+                )
+            )
+        except IntegrityError:
+            raise EntityAlreadyExistsError(message="Role permission already exists.")
```
borisarzentar, line 63: Should we create an error specific to this?

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -1,29 +1,117 @@
...
 
 
 def get_permissions_router() -> APIRouter:
     permissions_router = APIRouter()
 
-    @permissions_router.post("/groups/{group_id}/permissions")
-    async def give_permission_to_group(
-        group_id: str, permission: str, db: Session = Depends(get_user_db)
+    @permissions_router.post("/roles/{role_id}/permissions")
+    async def give_default_permission_to_role(
```
borisarzentar, line 32: Why `default`?

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -77,4 +163,65 @@ async def add_user_to_group(user_id: str, group_id: str, db: Session = Depends(g
...
+        tenant_id: str,
+        permissions: Optional[list[str]] = None,
+        db: Session = Depends(get_async_session),
+    ):
+        try:
+            # Add association directly to the association table
+            role = Role(name=role_name, tenant_id=tenant_id)
+            db.add(role)
+        except IntegrityError:
+            raise EntityAlreadyExistsError(message="Role already exists for tenant.")
```
borisarzentar, line 178: Custom error?

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -77,4 +163,65 @@ async def add_user_to_group(user_id: str, group_id: str, db: Session = Depends(g
...
+        tenant_name: str,
+        permissions: Optional[list[str]] = None,
+        db: Session = Depends(get_async_session),
+    ):
+        try:
+            # Add association directly to the association table
+            tenant = Tenant(name=tenant_name)
+            db.add(tenant)
+        except IntegrityError:
+            raise EntityAlreadyExistsError(message="Tenant already exists.")
```
borisarzentar, line 210: Custom error?

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -77,4 +163,65 @@ async def add_user_to_group(user_id: str, group_id: str, db: Session = Depends(g
...
+
+        await db.commit()
+        await db.refresh(tenant)
+
+        tenant = (
+            (await db.execute(select(Tenant).where(Tenant.name == tenant_name))).scalars().first()
+        )
+
+        if permissions:
```
borisarzentar, line 219: Do you think we should mix these two? I would have it separated into two functions. It gives you flexibility and you can still achieve the same result.

cognee/modules/users/methods/get_authenticated_user.py
```diff
@@ -1,5 +1,31 @@
...
+        scheme, token = authorization.split()
+        if scheme.lower() != "bearer":
+            raise HTTPException(status_code=401, detail="Invalid authentication scheme")
+
+        payload = jwt.decode(
+            token, os.getenv("FASTAPI_USERS_JWT_SECRET", "super_secret"), algorithms=["HS256"]
+        )
+
+        # SimpleNamespace lets us access dictionary elements like attributes
+        ret_val = SimpleNamespace(
```
borisarzentar, line 23: `ret_val` -> `user` or `auth_data`

cognee/modules/users/methods/get_default_user.py
```diff
@@ -21,4 +23,7 @@ async def get_default_user():
         if user is None:
             return await create_default_user()
 
-        return user
+        # We return a SimpleNamespace to have the same user type as our SaaS
+        # SimpleNamespace is just a dictionary which can be accessed through attributes
+        ret_val = SimpleNamespace(id=user.id, tenant_id=user.tenant_id, roles=[])
```
borisarzentar, line 28: `ret_val` -> `auth_data` or `user`

cognee/modules/users/models/RoleDefaultPermission.py
```diff
@@ -0,0 +1,22 @@
+from datetime import datetime, timezone
+from sqlalchemy import Column, ForeignKey, DateTime, UUID
+from cognee.infrastructure.databases.relational import Base
+
+
+class RoleDefaultPermission(Base):
```
borisarzentar, line 6: `permission` -> `permissions`

cognee/modules/users/models/UserDefaultPermission.py
```diff
@@ -0,0 +1,18 @@
+from datetime import datetime, timezone
+from sqlalchemy import Column, ForeignKey, DateTime, UUID
+from cognee.infrastructure.databases.relational import Base
+
+
+class UserDefaultPermission(Base):
```
borisarzentar, line 6: What are these default tables for?

dexters1: I want to allow tenants to be able to define which permissions will be default when they upload documents. So when a document is uploaded in a tenant without specifying permissions these default permissions will be applied, if any permission is specified when adding a document these default permissions won't be applied.

cognee/api/v1/permissions/routers/get_permissions_router.py
dexters1: I don't know, depends on our frontend. They are separate already, as if no permission is given it won't assign permissions, and we can assign this permission later. With this I wanted to be able to define permissions as well at the beginning, but I can split them up it will work either way

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -1,29 +1,117 @@
...
-    @permissions_router.post("/groups/{group_id}/permissions")
-    async def give_permission_to_group(
-        group_id: str, permission: str, db: Session = Depends(get_user_db)
+    @permissions_router.post("/roles/{role_id}/permissions")
+    async def give_default_permission_to_role(
+        role_id: str, permission: str, db: Session = Depends(get_async_session)
     ):
-        group = (
-            (await db.session.execute(select(Group).where(Group.id == group_id))).scalars().first()
+        role = (await db.execute(select(Role).where(Role.id == role_id))).scalars().first()
```
borisarzentar, line 35: I think this logic belongs to modules, we should not expose infrastructure in api router.

cognee/get_token.py
```diff
@@ -0,0 +1,21 @@
+import jwt
+import os
+import datetime
+
+SECRET_KEY = os.getenv("FASTAPI_USERS_JWT_SECRET", "super_secret")
```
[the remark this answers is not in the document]
borisarzentar: @dexters1 This is a good suggestion from coderabbit.

cognee/modules/users/authentication/get_auth_backend.py
```diff
@@ -1,11 +1,29 @@
...
     JWTStrategy,
 )
+from datetime import datetime, timedelta
+from typing import Optional
+
+from cognee.modules.users.methods import get_user
+
+
+class CustomJWTStrategy(JWTStrategy):
+    async def write_token(self, user: str, lifetime_seconds: Optional[int] = None) -> str:
```
borisarzentar, line 17: `user: str` -> `user: User`

cognee/get_token.py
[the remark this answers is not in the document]
dexters1: At the moment it's doing the same as cognee, trying to get "FASTAPI_USERS_JWT_SECRET"

If it can't get it it's defaulting to "super_secret"

Do we want to make this change for Cognee as well? It will make running the backend server locally a bit more complicated for users

cognee/api/v1/permissions/routers/get_permissions_router.py
dexters1: As we'll automatically create a permission specified if it doesn't exist. 

I'm not sure in general what kind of permissions can exist for Cognee outside of:

Create, Delete, Read, Write and Update

Might be worth considering if we should have a complete list of permissions for now instead of allowing users to define their own permissions

cognee/api/v1/permissions/routers/get_permissions_router.py
dexters1: I think that would be overkill at the moment, as we have 5-6 different situations here where the problem is straightforward and the http status return code is the same. Should be enough to define a message what entity already exists instead of defining exceptions for: RoleAlreadyExists, TenantAlreadyExists, UserRoleAlreadyExists, UserPermissionAlreadyExists and etc.

The database entity already exists and it's this specific one in question

cognee/get_token.py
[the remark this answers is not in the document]
borisarzentar: Don't know what is better. Do we want to make it easier for users, but less secure if they don't change the default? Or more secure by always requesting the secret to be set? This secret is important only in the case where customers control users and permissions?

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -1,80 +1,68 @@
...
 
 def get_permissions_router() -> APIRouter:
     permissions_router = APIRouter()
 
-    @permissions_router.post("/groups/{group_id}/permissions")
-    async def give_permission_to_group(
-        group_id: str, permission: str, db: Session = Depends(get_user_db)
+    @permissions_router.post("/roles/{role_id}/permissions")
+    async def give_default_permission_to_role(
+        role_id: UUID, permission_name: str, db: Session = Depends(get_async_session)
```
borisarzentar, line 15: We don't need db here anymore, right?

cognee/modules/users/authentication/get_auth_backend.py
borisarzentar: Then we expect users to use python and install cognee, which is usually not necessary for running the API. Should we expose some API method for this?

cognee/api/v1/permissions/routers/get_permissions_router.py
borisarzentar: @dexters1 Why not just `give_permission_to_role`?

cognee/modules/users/authentication/get_auth_backend.py
dexters1: We have an API method for this with login, it stays the same as before for our SaaS. But this function is when someone wants to integrate Cognee as a third party app into their app to allow them to handle their own users and just forward user information to Cognee.

cognee/get_token.py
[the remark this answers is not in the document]
dexters1: I'd leave it as is for now, I don't think this is a big problem

cognee/api/v1/permissions/routers/get_permissions_router.py
dexters1: I was hoping to make a differentiation between giving a permission to a role for a specific document, and giving default permissions for roles when no specific permission is set for a document.

So there would be less confusion

borisarzentar (approved):

PR #597, opened 2025-03-04
bug: added fix and added my pat

Vasilije1990:
## DCO Affirmation

## Summary by CodeRabbit

- **Chores**
  - Introduced an automated process that routinely updates project dependencies. This enhancement minimizes manual maintenance and helps ensure optimal system stability and security for users.


.github/workflows/autoupdate.yml
```diff
@@ -0,0 +1,60 @@
+name: Auto Update Dependencies
```
borisarzentar, line 1: What is the difference between this and existing one?

Vasilije1990: I added actual PAT from the secrets vs hardcoded one that was hallucinated with chatgpt

borisarzentar: Should we delete the other one then?

PR #599, opened 2025-03-04
Allowing parallel edges in graph projection when using graph completion search

hajdul88:
## Description
Allows parallell edges in graph projection when using graph completion search

## DCO Affirmation


hajdul88:
I tested it with Dreamify on benchmarks, and the search behaviour didn't change.

dexters1 (approved):

PR #603, opened 2025-03-04
feat: Anthropic model support for Google Vertex AI

dougnukem:
## Description

implemented support for Anthropic via Google Vertex AI by:

Updated the dependency in pyproject.toml to include the Vertex AI integration:

```
anthropic = {version = "^0.26.1", extras = ["vertex"]}
```
Created a new AnthropicVertexAdapter class in cognee/infrastructure/llm/anthropic/vertex_adapter.py that uses the AnthropicVertex client instead of the standard Anthropic client.

Updated the __init__.py file to expose both adapter classes.

Added a new provider option in get_llm_client.py:
```
class LLMProvider(Enum):
    # Existing providers...
    ANTHROPIC_VERTEX = "anthropic_vertex"
    # Other providers...
```
Added the handler for the new provider in get_llm_client.py with support for Google Cloud specific configuration.

## Authentication
The AnthropicVertex client uses Google Cloud's Application Default Credentials (ADC) for authentication, which is different from the API key-based authentication used by the standard Anthropic client.

ADC looks for credentials in the following order:

GOOGLE_APPLICATION_CREDENTIALS environment variable pointing to a service account key file
Default service account when running in Google Cloud environments (Compute Engine, GKE, Cloud Run, etc.)
User credentials from gcloud CLI (after running gcloud auth application-default login)

## Configuration
To use the Anthropic Vertex AI integration:

Set the environment variable LLM_PROVIDER to anthropic_vertex
Set the LLM_MODEL to the appropriate Vertex AI model name (e.g., "claude-3-5-sonnet-v2@20241022")
Optionally set GCP_PROJECT_ID and GCP_LOCATION environment variables if you need to specify a different project or region
Ensure Google Cloud authentication is properly set up using one of the ADC methods
The implementation supports additional configuration parameters:

project_id: Google Cloud project ID (optional, defaults to the project from ADC)
location: Google Cloud region (optional, defaults to "us-central1")
These can be set via environment variables GCP_PROJECT_ID and GCP_LOCATION which will be picked up by the Pydantic settings.

## DCO Affirmation


cognee/infrastructure/llm/anthropic/vertex_adapter.py
```diff
@@ -0,0 +1,86 @@
...
+                {
+                    "role": "user",
+                    "content": f"""Use the given format to extract information
+                from the following input: {text_input}. {system_prompt}""",
+                }
+            ],
+            response_model=response_model,
+        )
+
+    def show_prompt(self, text_input: str, system_prompt: str) -> str:
```
borisarzentar, line 70: I think that you don't have to implement this, it is already implemented in `LLMInterface` class.

cognee/infrastructure/llm/anthropic/vertex_adapter.py
```diff
@@ -0,0 +1,86 @@
+from typing import Type, Optional
+from pydantic import BaseModel
+import instructor
+from anthropic import AnthropicVertex
```
borisarzentar, line 4: Please import this where it is needed. Otherwise the code will break if this optional dependency is not installed.

pyproject.toml
```diff
@@ -61,7 +61,7 @@ litellm = "1.57.2"
 groq = {version = "0.8.0", optional = true}
 langfuse = "^2.32.0"
 pydantic-settings = "^2.2.1"
-anthropic = "^0.26.1"
+anthropic = {version = "^0.26.1", extras = ["vertex"]}
```
borisarzentar, line 64: Extra should be called `vertex`. I guess that if there are more anthropic providers we want to install only one and not all of them if we want to use anthropic.

cognee/infrastructure/llm/anthropic/vertex_adapter.py
Vasilije1990: @dougnukem let us know if you need help and clarifications on this one

pyproject.toml
Vasilije1990: @dougnukem could you please check comments? we would love to merge this one

borisarzentar:
Closing due to inactivity.

PR #644, opened 2025-03-14
feat: Relational db to graph db [COG-1468]

dexters1:
## Description
Add ability to migrate relational database to graph database

## DCO Affirmation


cognee/modules/visualization/cognee_network_visualization.py
```diff
@@ -182,8 +184,8 @@ async def cognee_network_visualization(graph_data, destination_file_path: str =
     </html>
     """
 
-    html_content = html_template.replace("{nodes}", json.dumps(nodes_list))
-    html_content = html_content.replace("{links}", json.dumps(links_list))
+    html_content = html_template.replace("{nodes}", json.dumps(nodes_list, default=str))
+    html_content = html_content.replace("{links}", json.dumps(links_list, default=str))
```
dexters1, line 188: There were issues for some Neo4j data types when encoding, if we don't care if everything is a str regrading information in the graph visualization it should be more robust to handle everything as a str instead of adding all possible data type conversions to our JsonEncoder

cognee/infrastructure/databases/graph/networkx/adapter.py
```diff
@@ -315,11 +317,13 @@ async def load_graph_from_file(self, file_path: str = None):
...
 
-                        if isinstance(edge["updated_at"], int):  # Handle timestamp in milliseconds
+                        if isinstance(
+                            edge.get("updated_at"), int
+                        ):  # Handle timestamp in milliseconds
                             edge["updated_at"] = datetime.fromtimestamp(
                                 edge["updated_at"] / 1000, tz=timezone.utc
                             )
-                        elif isinstance(edge["updated_at"], str):
+                        elif isinstance(edge.get("updated_at"), str):
```
dexters1, line 326: There were issues with a missing "update_at" key, this was a temporary solution to get it to work. Do we need a permanent one?

borisarzentar: Missing as not set on the DataPoint when saved?

dexters1: This was before the refactor that is using DataPoints

cognee/infrastructure/databases/vector/embeddings/LiteLLMEmbeddingEngine.py
```diff
@@ -64,17 +64,28 @@ async def exponential_backoff(attempt):
...
             else:
-                response = await litellm.aembedding(
-                    model=self.model,
-                    input=text,
-                    api_key=self.api_key,
-                    api_base=self.endpoint,
-                    api_version=self.api_version,
-                )
+                embedd = []
+                batch_size = 100
```
dexters1, line 68: This whole part is a work in progress at the moment, there are issues when trying to embed all the migrated nodes at once. We need a better solution for this than what I've done here

cognee/infrastructure/databases/relational/sqlalchemy/SqlAlchemyAdapter.py
```diff
@@ -345,3 +343,94 @@ async def delete_database(self):
...
+                        schema[table_name]["foreign_keys"].append(
+                            {
+                                "column": fk[3],  # Column in the current table
+                                "ref_table": fk[2],  # Referenced table
+                                "ref_column": fk[4],  # Referenced column
+                            }
+                        )
+            else:
+                schema_list = await self.get_schema_list()
+                for schema_name in schema_list:
```
dexters1, line 387: Relational database migration has not been tested yet with multiple schemas in PostgreSQL

cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
```diff
@@ -199,13 +200,14 @@ async def add_edge(
         serialized_properties = self.serialize_properties(edge_properties)
 
         query = dedent(
-            """MATCH (from_node {id: $from_node}),
-            (to_node {id: $to_node})
-            MERGE (from_node)-[r]->(to_node)
-            ON CREATE SET r += $properties, r.updated_at = timestamp(), r.type = $relationship_name
+            f"""\
```
dexters1, line 203: Old code did not work for me with regards to relationship_name

cognee/infrastructure/databases/graph/kuzu/adapter.py
```diff
@@ -923,3 +925,6 @@ async def load_graph_from_file(self, file_path: str) -> None:
         except Exception as e:
             logger.error(f"Failed to import graph from file: {e}")
             raise
+
+    async def migrate_relational_database(self, schema):
+        await migrate_relational_database(self, schema=schema)
```
borisarzentar, line 930: Isn't this one calling itself?

cognee/modules/engine/models/TableType.py
```diff
@@ -0,0 +1,9 @@
+from cognee.infrastructure.engine import DataPoint
+
+
+class TableType(DataPoint):
+    name: str
+    description: str
+    ontology_valid: bool = False
```
borisarzentar, line 7: This property should probably exist on DataPoint, since it is used in the core of the system.

cognee/tasks/storage/index_data_points.py
```diff
@@ -38,7 +38,11 @@ async def index_data_points(data_points: list[DataPoint]):
         index_name = index_name_and_field[:first_occurence]
         field_name = index_name_and_field[first_occurence + 1 :]
         try:
-            await vector_engine.index_data_points(index_name, field_name, indexable_points)
+            # In case the ammount if indexable points is too large we need to send them in batches
+            batch_size = 1000
```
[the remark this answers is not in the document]
borisarzentar: That's a good question. I think let's leave it here for now, and we will consider moving it in the future. We need to find some solution so we don't have to do this in every adapter.

cognee/infrastructure/databases/graph/migrate_relational_database.py
```diff
@@ -0,0 +1,164 @@
...
+from cognee.tasks.storage.index_graph_edges import index_graph_edges
+
+from uuid import uuid4
+
+from cognee.modules.engine.models import TableRow, TableType
+
+logger = logging.getLogger(__name__)
+
+
+async def migrate_relational_database(graph_db, schema):
```
borisarzentar, line 19: I think that you can move this to tasks, looks like a task.

cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
```diff
@@ -650,3 +652,6 @@ async def get_graph_metrics(self, include_optional=False):
             }
 
         return mandatory_metrics | optional_metrics
+
+    async def migrate_relational_database(self, schema):
```
borisarzentar, line 656: Do we need it on the adapter itself? If we have a task to which we pass schema and graph engine then we don't need it on the adapter.

cognee/modules/visualization/cognee_network_visualization.py
borisarzentar: For the visualization I think it is fine, the type of data is not so important there.

cognee/infrastructure/databases/graph/kuzu/adapter.py
dexters1: No, as it's not referencing self, it's calling the imported function

cognee/modules/engine/models/TableType.py
dexters1: Which property? ontology_valid? I've just reused the properties specified in Entity and EntityType. Should I move it to DataPoint then and remove it from Entity, EntityType, TableRow and TableType?

cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
dexters1: It can be a task now as there is only one implementation for all graphs ( compared to before where each graph had it's own ) 

My only concern with moving this to a Task is that this migration might not be best suited for pipelines, the intended workflow is to get the schema from the relational database and the users should confirm and edit the schema manually in case they don't want to migrate some information, afterwards the migration function should be called. If this kind of workflow would be used do we want to have this function as a Task still?

borisarzentar: The Task accepts schema, so the previous task can provide it to this one, or we can send it to migration task when creating the pipeline. We can have user intervention in between, but for now it will just extract everything.
I looked at the puppygraph, and they can define schema beforehand and what should be extracted, so we can add something like that later.

cognee/modules/engine/models/TableType.py
borisarzentar: Yes, `ontology_valid`. Please move it then to DataPoint.

borisarzentar (approved):

PR #646, opened 2025-03-16
feat: Text to Cypher Retriever

dm1tryG:
## Description

New Text to Cypher Retriever

## DCO Affirmation

## Summary by CodeRabbit

- **New Features**
	- Introduced a new "Cypher prompting" search option in the interface that expands and enhances search capabilities for graph-based queries.

### DIFF
--- a/cognee/modules/retrieval/cypher_query_retriever.py
+++ b/cognee/modules/retrieval/cypher_query_retriever.py
@@ -5,6 +5,7 @@ from cognee.infrastructure.databases.graph import get_graph_engine
 from cognee.infrastructure.llm.get_llm_client import get_llm_client
 
 import logging
+
 logger = logging.getLogger(__name__)
 
 
@@ -56,10 +57,9 @@ class CypherQueryRetriever(BaseRetriever):
     {previous_attempt_context}
                 """,
                 system_prompt=system_prompt,
-                response_model=str
+                response_model=str,
             )
             try:
-
                 context = await graph_engine.query(cypher_query)
                 if context:
                     return context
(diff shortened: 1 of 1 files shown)


dm1tryG:
@hajdul88 Hey, I see that you added https://github.com/topoteretes/cognee/pull/648

I added same logic here 🤔 

Text to cypher as well. How can we merge it? Can you check text 2 cypher logic?

hajdul88:
> @hajdul88 Hey, I see that you added #648
> 
> I added same logic here 🤔
> 
> Text to cypher as well. How can we merge it? Can you check text 2 cypher logic?

I will talk to @borisarzentar about this. I think it's slightly different. In my PR, it's about exposing the cypher search parameter of the DB adapter, so the query is ultimately just a cypher query, not a natural language query. 

Text to cypher means the user is forming a natural language query and you have to create a corresponding cypher right?

borisarzentar:
> > @hajdul88 Hey, I see that you added #648
> > I added same logic here 🤔
> > Text to cypher as well. How can we merge it? Can you check text 2 cypher logic?
> 
> I will talk to @borisarzentar about this. I think it's slightly different. In my PR, it's about exposing the cypher search parameter of the DB adapter, so the query is ultimately just a cypher query, not a natural language query.
> 
> Text to cypher means the user is forming a natural language query and you have to create a corresponding cypher right?

@hajdul88 @dm1tryG Yes, this one is different from the one we implemented last week. I would name it differently, so we know that it is a natural language search. Cypher is the implementation detail, right? Users don't care what happens behind the scene, they just send the natural language query and get results.

dm1tryG:
@borisarzentar @hajdul88 I will rename the retriever, maybe good idea to inherit from new cypher retriever from https://github.com/topoteretes/cognee/pull/648 will check

PR #648, opened 2025-03-17
feat: adds cypher search to retrievers module

hajdul88:
## Description
Exposes the query method of the adapter in the search interface for Kuzu and Neo4j (cypher compatible adapters)

## DCO Affirmation

## Summary by CodeRabbit

- **New Features**
  - Introduced a new cypher-based search option that expands the app's search functionality.
  - Enabled asynchronous processing for advanced query execution.
  - Enhanced error messaging for unsupported search types and query execution issues.
  - Added a new enumeration value for `CYPHER` to support the new search type.


cognee/modules/retrieval/cypher_search_retriever.py
```diff
@@ -0,0 +1,46 @@
...
+
+            result = await graph_engine.query(query)
+        except SearchTypeNotSupported:
+            raise
+        except Exception as e:
+            logger.error("Failed to execture cypher search retrieval: %s", str(e))
+            raise CypherSearchError() from e
+        return result
+
+    async def get_completion(self, query: str, context: Optional[Any] = None) -> Any:
```
hajdul88, line 42: Since its just a search get_completion returns only the context similarly to INSIGHTS. @borisarzentar please check if you were thinking something like this

cognee/modules/retrieval/cypher_search_retriever.py
```diff
@@ -0,0 +1,46 @@
...
+        try:
+            graph_engine = await get_graph_engine()
+
+            if isinstance(graph_engine, NetworkXAdapter):
+                raise SearchTypeNotSupported(
+                    "CYPHER search type not supported for NetworkXAdapter."
+                )
+
+            result = await graph_engine.query(query)
+        except SearchTypeNotSupported:
```
borisarzentar, line 35: We don't need to catch it if we want just to raise it again.

cognee/modules/retrieval/cypher_search_retriever.py
borisarzentar: Yeah, that's fine.

cognee/modules/retrieval/exceptions/exceptions.py
```diff
@@ -0,0 +1,22 @@
...
+    ):
+        super().__init__(message, name, status_code)
+
+
+class CypherSearchError(CogneeApiError):
+    def __init__(
+        self,
+        message: str = "An error occurred during the execution of the Cypher query.",
+        name: str = "CypherSearchError",
+        status_code: int = status.HTTP_500_INTERNAL_SERVER_ERROR,
```
borisarzentar, line 20: Shouldn't be 500. 500 means something crashed, but we handled it here.

cognee/modules/retrieval/cypher_search_retriever.py
hajdul88: Fixed

cognee/modules/retrieval/exceptions/exceptions.py
hajdul88: Fixed

borisarzentar (approved):

PR #658, opened 2025-03-20
fix: human readable logs

soobrosa:
## Description
Introducing scructlog.

## DCO Affirmation


cognee-mcp/src/server.py
```diff
@@ -3,6 +3,8 @@
 import os
 import cognee
 import logging
+import structlog
```
borisarzentar, line 6: I imagined that there will be a way to integrate structlog into logging, so we don't have to do this replacement everywhere. Usually other logging libs do it.

cognee/infrastructure/llm/gemini/adapter.py
```diff
@@ -10,7 +12,8 @@
 from cognee.infrastructure.llm.prompts import read_query_prompt
 from cognee.base_config import get_base_config
 
-logger = logging.getLogger(__name__)
+setup_logging(logging.DEBUG)
+logger = structlog.get_logger(__name__)
```
borisarzentar, line 16: After looking at this, makes sense to return `logger` from the `setup_logging` function. That way we will always have one way where we do this setup and return logger. So we can return structlog, or something else.

cognee-mcp/src/server.py
Vasilije1990: @coderabbitai try handling this

Vasilije1990:
@coderabbitai review

cognee-mcp/src/server.py
soobrosa: @borisarzentar @Vasilije1990 thanks for the feedback, I refactored to one import and one init.

cognee/api/v1/add/routers/get_add_router.py
```diff
@@ -4,12 +4,12 @@
 from typing import List
 import aiohttp
 import subprocess
-import logging
+from cognee.shared.logging_utils import get_logger, INFO
```
borisarzentar, line 7: Where do we use `INFO`?

soobrosa: Good catch, `INFO` is default, so removing the import where it was not explicitly specified.

borisarzentar (dismissed):

PR #670, opened 2025-03-26
feat: Add initial cognee pipeline simplification [COG-1705]

dexters1:
## Description
Simplify Cognee pipeline usage for users

## DCO Affirmation


soobrosa:
@dexters1 how would you tl;dr the changes?

borisarzentar:
Let's park this for now, until we have the new pipeline implementation. Couple of days to a week probably.

dexters1:
> @dexters1 how would you tl;dr the changes?

Handle all complexities of getting user, datasets, creating db and tables and testing LLM connection in the new Cognee Pipeline so users can just create tasks and call our cognee pipeline which will handle all of this overhead without them having to do it themselves if they want to create and use custom tasks

dexters1:
NOTE: At this moment dreamify will not work together with Cognee with these changes, it needs to be updated as well to accommodate to the workflow

dexters1:
> NOTE: At this moment dreamify will not work together with Cognee with these changes, it needs to be updated as well to accommodate to the workflow

Seems to be working fine actually upon testing

borisarzentar (approved):

hajdul88 (approved):

PR #672, opened 2025-03-27
feat: Add logging to file [COG-1715]

dexters1:
## Description
Add logging to logs file

## DCO Affirmation


cognee/shared/logging_utils.py
```diff
@@ -31,6 +122,36 @@ def get_logger(name=None, level=INFO):
...
+
+        # Sort log files by modification time (newest first)
+        log_files.sort(key=lambda x: x.stat().st_mtime, reverse=True)
+
+        # Remove old files that exceed the maximum
+        if len(log_files) > max_files:
+            for old_file in log_files[max_files:]:
+                try:
+                    old_file.unlink()
+                    print(f"Deleted old log file: {old_file}")
```
borisarzentar, line 145: It would be good to use logger here as well, so we can turn off these logs.

PR #686, opened 2025-03-31
Feature: add graph creation prompt

lxobr:
## Description

- Added new graph creation prompts
- Exposed graph creation prompts in .cognify via get_default tasks
- Exposed graph creation prompts in eval framework
## DCO Affirmation


cognee/api/v1/cognify/cognify.py
```diff
@@ -111,6 +111,7 @@ async def get_default_tasks(  # TODO: Find out a better way to do this (Boris's
     chunker=TextChunker,
     chunk_size: int = None,
     ontology_file_path: Optional[str] = None,
+    graph_prompt_path: Optional[str] = None,
```
borisarzentar, line 114: We need a better way to handle this kind of stuff. Seems that settings this on the task level would make sense, but adding this parameter on the pipeline introduces too many parameters.

lxobr: Thanks for the comment, Boris. I agree, this is not ideal. I would be happy with moving most of the parameters into of the BaseSetting config or similar, but I understand that's against our current guidelines. That said, we do need this parameter exposed for various evaluation and research related reasons.
Do you have any suggestion how you want to handle it in this PR?

borisarzentar: I'm thinking if we should put this in the llm config. It will be against our current guidelines, but maybe better to solve it that way for now and migrate later to something else we come up with.

lxobr: Moved it there. Please let me know what you think.

PR #687, opened 2025-03-31
Fix logging [COG-1747]

dexters1:
## Description
Add fixes for logger, specifically:
- Add handler when users don't have microseconds
- Resolve issue with multiprocess logging

## DCO Affirmation

### DIFF
--- a/cognee/shared/logging_utils.py
+++ b/cognee/shared/logging_utils.py
@@ -69,7 +69,7 @@ class PlainFileHandler(logging.FileHandler):
                 logger_name = record.msg.get("logger", record.name)
 
                 # Format timestamp
-                timestamp = datetime.now().strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+                timestamp = get_timestamp()
 
                 # Create the log entry
                 log_entry = f"{timestamp} [{record.levelname.ljust(8)}] {message}{context_str} [{logger_name}]\n"
@@ -337,3 +337,17 @@ def get_log_file_location():
     for handler in root_logger.handlers:
         if isinstance(handler, logging.FileHandler):
             return handler.baseFilename
+
+
+def get_timestamp():
+    # NOTE: Some users have complained that Cognee crashes when trying to get microsecond value
+    #       Added handler to not use microseconds if users can't access it
+    logger = structlog.get_logger()
+    try:
+        return datetime.now().strftime("%Y-%m-%dT%H:%M:%S.%fZ")
+    except Exception as e:
+        logger.debug(f"Exception caught: {e}")
+        logger.debug(
+            "Could not use microseconds for the logging timestamp, defaulting to use hours minutes and seconds only"
+        )
+        return datetime.now().strftime("%Y-%m-%dT%H:%M:%S")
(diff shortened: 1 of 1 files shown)


cognee/shared/logging_utils.py
```diff
@@ -226,7 +226,7 @@ def exception_handler(logger, method_name, event_dict):
             structlog.stdlib.add_logger_name,
             structlog.stdlib.add_log_level,
             structlog.stdlib.PositionalArgumentsFormatter(),
-            structlog.processors.TimeStamper(fmt="iso"),
```
dexters1, line 229: Note: ISO is no longer used to the timezone/format for the logs defaults to the users timezone, is this fine?

cognee/shared/logging_utils.py
```diff
@@ -328,3 +337,23 @@ def get_log_file_location():
...
+def get_timestamp_format():
+    # NOTE: Some users have complained that Cognee crashes when trying to get microsecond value
+    #       Added handler to not use microseconds if users can't access it
+    logger = structlog.get_logger()
+    try:
+        # We call datetime.now() here to test if microseconds are supported.
+        # If they are not supported a ValueError will be raised
+        datetime.now().strftime("%Y-%m-%dT%H:%M:%S.%f")
+        return "%Y-%m-%dT%H:%M:%S.%f"
+    except Exception as e:
```
dexters1, line 351: I had no way to reproduce specific issue users were having so we will need to ask them later to confirm this resolution works for them. This handler was written based on the error message they had provided.

cognee/shared/logging_utils.py
soobrosa: I wouldn't open a religious conversation on this, so sure!

cognee/shared/logging_utils.py
soobrosa: Let's see.

cognee/shared/logging_utils.py
```diff
@@ -288,9 +288,18 @@ def emit(self, record):
...
     stream_handler.setLevel(log_level)
 
+    # Check if we already have a log file path from the environment
+    # NOTE: environment variable must be used here as it allows us to
+    # log to a single file with a name based on a timestamp in a multiprocess setting.
+    # Without it, we would have a separate log file for every process.
+    log_file_path = os.environ.get("LOG_FILE_NAME")
+    if not log_file_path:
+        # Create a new log file name with the cognee start time
+        start_time = datetime.now().strftime("%Y-%m-%d_%H-%M-%S")
```
soobrosa, line 298: Shouldn't we go all in with `get_timestamp_format()` then, and no `strftime("%Y-%m-%d_%H-%M-%S")`? It's not a biggie, just asking.

soobrosa (approved):
@dexters1 great fix for the `multiprocessing` issue, thank you!

borisarzentar (approved):

cognee/shared/logging_utils.py
dexters1: I thought about it, but this formatting is for the file and is different compared to the actual timestamp.

PR #690, opened 2025-04-01
feat: pipeline tasks needs mapping

borisarzentar:
## DCO Affirmation


cognee/api/v1/cognify/cognify.py
```diff
@@ -121,24 +123,33 @@ async def get_default_tasks(  # TODO: Find out a better way to do this (Boris's
...
+        ),
         Task(
             summarize_text,
             summarization_model=cognee_config.summarization_model,
-            task_config={"batch_size": 10},
+            task_config=TaskConfig(needs=[extract_chunks_from_documents]),
+        ),
+        Task(
+            add_data_points,
+            task_config=TaskConfig(needs=[merge_needs(summarize_text, extract_graph_from_data)]),
```
dexters1, line 151: Is it possible to have merge_needs to be done automatically in TaskConfig if more than one task is needed?

It's not user friendly I think most would expect for this to work with just having a list of tasks

cognee/modules/pipelines/operations/run_tasks.py
```diff
@@ -271,19 +107,22 @@ async def run_tasks_with_telemetry(tasks: list[Task], data, pipeline_name: str):
 
 async def run_tasks(
     tasks: list[Task],
-    dataset_id: UUID = uuid4(),
+    dataset_id: UUID = None,
     data: Any = None,
     pipeline_name: str = "unknown_pipeline",
+    context: dict = None,
 ):
+    dataset_id = dataset_id or uuid4()
```
dexters1, line 115: Currently dataset_id is defined in the following way during the add pipeline:

# Dataset id should be generated based on dataset_name and owner_id so multiple users can use the same dataset_name
id=uuid5(NAMESPACE_OID, f"{dataset_name}{str(owner_id)}")

We might need to reconsider how the dataset_id is determined and have it be uniform

cognee/modules/pipelines/operations/needs.py
```diff
@@ -0,0 +1,60 @@
...
+
+def get_need_task_results(results, task: Task):
+    input_results = []
+
+    for task_dependency in task.task_config.needs:
+        if isinstance(task_dependency, MergeNeeds):
+            task_results = []
+            max_result_length = 0
+
+            for task_need in task_dependency.needs:
```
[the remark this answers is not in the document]
borisarzentar: It is not a lookup, just a for loop. :)

cognee/api/v1/cognify/code_graph_pipeline.py
```diff
@@ -45,43 +46,64 @@ async def run_code_graph_pipeline(repo_path, include_docs=False):
...
-        yield run_status
+    async for run_info in run_tasks(tasks, dataset_id, repo_path, "cognify_code_pipeline"):
+        yield run_info
 
 
 if __name__ == "__main__":
 
     async def main():
-        async for run_status in run_code_graph_pipeline("REPO_PATH"):
+        async for run_status in run_code_graph_pipeline("/Users/borisarzentar/Projects/graphrag"):
```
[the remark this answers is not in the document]
borisarzentar: Good catch!

cognee/modules/pipelines/operations/run_tasks.py
borisarzentar: We don't send `dataset_id` in rare cases, so then we create one. `dataset_id` should be always present, but for some demos and examples we omit it.

cognee/api/v1/cognify/cognify.py
borisarzentar: It is optional. If you don't provide it, inputs will just go directly in the task, in the order they are in the list. But in some cases you don't want inputs to be passed that way, so this `merge_needs` is for that. You use it if you want to change the data from needs before they enter the task.

cognee/modules/pipelines/operations/needs.py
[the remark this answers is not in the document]
hajdul88: lol I completely overlooked that. I just saw in :D fair sorry

PR #695, opened 2025-04-01
test: relational db migration

hande-k:
## DCO Affirmation

### DIFF
--- a/.github/workflows/test_ollama.yml
+++ b/.github/workflows/test_ollama.yml
@@ -110,7 +110,7 @@ jobs:
           LLM_MODEL: "phi4"
           EMBEDDING_PROVIDER: "ollama"
           EMBEDDING_MODEL: "avr/sfr-embedding-mistral:latest"
-          EMBEDDING_ENDPOINT: "http://localhost:11434/v1/"
+          EMBEDDING_ENDPOINT: "http://localhost:11434/api/embeddings"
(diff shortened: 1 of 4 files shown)


cognee/tests/integration/test_relational_db_migration.py
```diff
@@ -0,0 +1,141 @@
...
+        ("Employee:4", "Employee:2"),
+        ("Employee:6", "Employee:1"),
+        ("Employee:8", "Employee:6"),
+        ("Employee:7", "Employee:6"),
+        ("Employee:3", "Employee:2"),
+    }
+    for e in expected_edges:
+        assert e in found_edges, f"Edge {e} not found in the actual '{relationship_label}' edges!"
+
+    print(f"All checks passed for {graph_db_provider} provider with '{relationship_label}' edges!")
```
dexters1, line 141: Might be also good to add a test that verifies the number of all nodes and edges in the migration, as it should always be the same number

cognee/tests/integration/test_relational_db_migration.py
```diff
@@ -0,0 +1,206 @@
+import json
```
dexters1, line 1: We should move this from the integration tests. It should be in the directory above with test_kuzu.py, test_neo4j.py and etc.

dexters1 (approved):

PR #726, opened 2025-04-13
test: make search unit tests deterministic

borisarzentar:
## DCO Affirmation


cognee/modules/retrieval/chunks_retriever.py
```diff
@@ -2,15 +2,32 @@
...
+
     async def get_context(self, query: str) -> Any:
         """Retrieves document chunks context based on the query."""
         vector_engine = get_vector_engine()
-        found_chunks = await vector_engine.search("DocumentChunk_text", query, limit=5)
+
+        try:
+            found_chunks = await vector_engine.search("DocumentChunk_text", query, limit=self.top_k)
+
+            if len(found_chunks) == 0:
```
hajdul88, line 26: I think empty context should be allowed in all searches, but I may be wrong. It's a feasible scenario and not an error imho.

borisarzentar: Ok, makes sense also, so if there is no context we return empty array.
We only throw an error if the system databases are not set up. I will change it to be like that.

borisarzentar: Pushed changes that allow this now.

PR #730, opened 2025-04-14
feature: tighten run_tasks_base

lxobr:
## Description

- Extracted run_tasks_base function into a new file run_tasks_base.py.
- Extracted four executors that execute core logic based on the task type.
- Extracted a task handler/wrapper that safely executes the core logic with logging and telemetry.
- Fixed the inconsistency with the batches of size 1.

## DCO Affirmation


cognee/modules/pipelines/operations/run_tasks_base.py
```diff
@@ -0,0 +1,186 @@
...
+async def handle_task(
+    running_task: Task,
+    args: list,
+    leftover_tasks: list[Task],
+    next_task_batch_size: int,
+    user: User,
+):
+    """Handle common task workflow with logging, telemetry, and error handling around the core execution logic."""
+    # Get task information using the helper functions
+    task_type = get_task_type(running_task)
```
borisarzentar, line 126: I see that we often use this "pattern" when we handle multiple types of data, but there is a real pattern for that.
https://en.wikipedia.org/wiki/Polymorphism_(computer_science)#Subtyping

We can solve this by moving the execution type methods to `Task` and just call `Task.run` like I did it here -> https://github.com/topoteretes/cognee/pull/690/files#diff-54f3065eb9ce36810ac0c36306830f3dffadee4efa80a9b9a5140f72f8b5feafR42

It is cleaner solution and we don't have to juggle with strings here and there.

lxobr: Thanks for the comment. It was outside the original scope, but worth improving.

lxobr:
I also fixed a little bug with the TaskGetters.

cognee/modules/pipelines/tasks/task.py
```diff
@@ -0,0 +1,97 @@
...
+        """Execute coroutine task and yield the result."""
+        task_result = await self.run(*args)
+        yield task_result
+
+    async def execute_function(self, args):
+        """Execute function task and yield the result."""
+        task_result = self.run(*args)
+        yield task_result
+
+    async def execute(self, args, next_batch_size=None):
```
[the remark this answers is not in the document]
lxobr: Ah yes, the classic case of: do we run it or do we execute it?
As you know .run() was the original method, it just calls the executable with stored params.
The new .execute() method came from refactoring logic that used to live around where we called .run(). So of course it adds batching and async handling in one place. Crucially, it depends on the info outside the task itself, which is the _next_batch_size argument. Which is why I named it .execute() since it doesn't merely run the task, but  does some managing and connects it to the next task.
I tried to follow our guidelines about naming functions by what they actually do. On a technical note, I am a bit concerned about renaming .execute() into .run(), since that would mean reviewing every .run() call and updating them from sync to async generators where needed.
I hope this clears the confusion. I would recommend keeping the name as is. Please let me know how you would like to proceed.

borisarzentar (approved):

soobrosa:
Great job @lxobr !

PR #731, opened 2025-04-14
feature: Adds S3 functionality

hajdul88:
## Description
Adds S3 support

## DCO Affirmation


borisarzentar:
@hajdul88  Just to make sure that we are aligned on the expectations here.
We want couple of things with s3:
- To be able to attach cognee on s3 bucket and read the data that is in that bucket. This is similar to what we have for the local filesystem, just works with s3. We should create a dataset per bucket and attach data to that dataset.
- To be able to store data/files into that s3 bucket. When user adds a new file, we want to store that file in the s3 bucket for that dataset.
- To be able to store task and pipeline artifacts into s3 bucket, and read them later. This one doesn't require any additional change except in task handling. Each pipeline run should create a run artifacts directory and put stuff there.

Which ones are we solving now?

Additional note: user/developer doesn't need to know where the data is saved, they should just use the system as it is.

hajdul88:
> @hajdul88 Just to make sure that we are aligned on the expectations here. We want couple of things with s3:
> 
> * To be able to attach cognee on s3 bucket and read the data that is in that bucket. This is similar to what we have for the local filesystem, just works with s3. We should create a dataset per bucket and attach data to that dataset.
> * To be able to store data/files into that s3 bucket. When user adds a new file, we want to store that file in the s3 bucket for that dataset.
> * To be able to store task and pipeline artifacts into s3 bucket, and read them later. This one doesn't require any additional change except in task handling. Each pipeline run should create a run artifacts directory and put stuff there.
> 
> Which ones are we solving now?
> 
> Additional note: user/developer doesn't need to know where the data is saved, they should just use the system as it is.

Discussed

soobrosa (approved):
`pyproject.toml` still might need a bit of TLC on the `boto3`, `botocore`, `s3fs` front but LGTM

PR #733, opened 2025-04-14
fix: cognee ui with new visualization

borisarzentar:
## DCO Affirmation


dexters1:
If it's ready to be used would be good to also uncomment it from docker-compose

dexters1:
The issue with docker-compose and mounting the local cognee .venv to the docker is still present and it causes my python environment to break completely, would be good to fix this as well before we release the new UI

dexters1:
image

Defining the vector_db from the wizard doesn't work (no option in dropdown, no option to write manually the vector_db)

Can't proceed at all with the gui at the moment so can't test it further

Issue #729, opened 2025-04-14
[COG-1883] Each word processed via Ollama + Hugging Face tokenizer triggers HF request in chunking


Vasilije1990:
<sub>From [SyncLinear.com](https://synclinear.com) | [COG-1883](https://linear.app/cognee/issue/COG-1883/each-word-processed-via-ollama-hugging-face-tokenizer-triggers-hf)</sub>

RyabykinIlya:
I’ve identified an issue causing excessive requests to the Hugging Face server, drastically slowing down data processing

steps to reproduce:
.env
```
LOG_LEVEL="DEBUG"

# LLM Configuration
LLM_API_KEY="ollama"
LLM_MODEL="gemma3:4b"
LLM_PROVIDER="ollama"
LLM_ENDPOINT="http://host.docker.internal:11435/v1"
LLM_API_VERSION=""
LLM_MAX_TOKENS="128000"

GRAPHISTRY_USERNAME=
GRAPHISTRY_PASSWORD=

SENTRY_REPORTING_URL=

# Embedding Configuration
EMBEDDING_PROVIDER="ollama"
EMBEDDING_MODEL="avr/sfr-embedding-mistral:latest"
EMBEDDING_ENDPOINT="http://host.docker.internal:11435/api/embeddings"
EMBEDDING_API_VERSION=""
EMBEDDING_DIMENSIONS=4096
EMBEDDING_MAX_TOKENS=8191

HUGGINGFACE_TOKENIZER="Salesforce/SFR-Embedding-Mistral"
```

invoke api
```
access_token=$(curl --location 'http://127.0.0.1:8000/api/v1/auth/login' \
  --form 'username="default_user@example.com"' \
  --form 'password="default_password"' | sed -n 's/.*"access_token":"\([^"]*\)".*/\1/p')

curl --location 'http://127.0.0.1:8000/api/v1/add' \
  --header "Authorization: Bearer $access_token" \
  --form 'data=@"/home/ilya/home/cognee/alice_in_wonderland.txt"' \
  --form 'datasetId="test-dataset"'

curl --location 'http://127.0.0.1:8000/api/v1/cognify' \
--header 'Content-Type: application/json' \
--header "Authorization: Bearer $access_token" \
--data '{
 "datasets": ["test-dataset"]
}'
```

results in logs:
```
2025-04-14T16:34:11.664822 [info     ] Async generator task started: `extract_chunks_from_documents` [run_tasks(tasks: [Task], data)]
[...]

PR #734, opened 2025-04-14
Initialize embedding engine once

RyabykinIlya:
COG-1881 fix

## Description

fixed bug: each chunk processed via Hugging Face tokenizer triggers HF request
#729 

## DCO Affirmation


RyabykinIlya:
image
image
checkboxes are disabled @Vasilije1990

cognee/tasks/chunks/chunk_by_sentence.py
```diff
@@ -3,6 +3,16 @@
 from .chunk_by_word import chunk_by_word
 from cognee.infrastructure.databases.vector.embeddings import get_embedding_engine
 
+# Initialize embedding engine once
+_embedding_engine = get_embedding_engine()
```
borisarzentar, line 7: If you change the embedding engine after you start the app, this will still keep the old one and call the wrong tokenizer.

dexters1: I've actually got blocked by HuggingFace due to this: "429 Client Error: Too Many Requests for url" as it calls HuggingFace each time the embedding is used instead of once, so we should consider moving the embedding engine to be a singleton based on the configuration ( like with the databases )

borisarzentar: Ok, then let's do that instead.

RyabykinIlya: @borisarzentar thanks for pointing that out! I hadn’t considered the issue with the tokenizer persisting after switching the engine. That’s a really helpful insight.

RyabykinIlya: @dexters1, your idea about making the embedding engine a singleton sounds like a solid fix for the rate-limiting issue. Do you think you’d be able to implement this? Or is it something you’re already planning to tackle?

dexters1: @RyabykinIlya Thanks for pointing out the issue to us and opening the PR! We've already handled using the vector and graph databases as a singleton based on configuration so I assume I can easily modify the embedding engine to work in the same way. I'll see to to make a PR for it today. I'll do this work from your commit so you'd be also part of the contributors :)

PR #742, opened 2025-04-15
Embedding string fix [COG-1900]

dexters1:
## Description
Allow embedding of big strings to support full row embedding in SQL databases

## DCO Affirmation


cognee/infrastructure/databases/vector/embeddings/LiteLLMEmbeddingEngine.py
```diff
@@ -77,20 +78,34 @@ async def exponential_backoff(attempt):
...
+
+                # Recursively embed the split parts in parallel
+                (left_vec,), (right_vec,) = await asyncio.gather(
+                    self.embed_text([left_part]),
+                    self.embed_text([right_part]),
+                )
 
-                return all_embeddings
+                # POOL the two embeddings into one
+                pooled = (np.array(left_vec) + np.array(right_vec)) / 2
```
borisarzentar, line 107: @dexters1 Did you discuss with @lxobr if merging embedding like this works fine?

dexters1: Yes, it works good enough for now. It should only be used if data can't be chunked ( like a table row when migrated from a relational DB )

Vasilije1990 (approved):

PR #785, opened 2025-04-25
Mcp sse support

dexters1:
## Description
Add both sse and stdio support for Cognee MCP

## DCO Affirmation


dexters1:
NOTE: did not yet test how the docker implementation works with the new MCP. Only stdio and sse ran locally with Cursor

Vasilije1990 (approved):

Vasilije1990:
@dexters1 check dlt function naming + strings, it is important to have good names so LLM can pick up the right tool

cognee-mcp/src/server.py
```diff
@@ -1,253 +1,141 @@
...
+
+@mcp.tool()
+async def cognify(text: str, graph_model_file: str = None, graph_model_name: str = None) -> list:
+    async def cognify_task(
+        text: str, graph_model_file: str = None, graph_model_name: str = None
+    ) -> str:
+        """Build knowledge graph from the input text"""
+        # NOTE: MCP uses stdout to communicate, we must redirect all output
+        #       going to stdout ( like the print function ) to stderr.
+        #       As cognify is an async background job the output had to be redirected again.
```
borisarzentar, line 31: I don't understand the last sentence here.

dexters1: It doesn't apply anymore, it used to make sense before we switch to FastMCP will remove it

PR #793, opened 2025-04-28
feat: add MCP check status tool [COG-1784]

dexters1:
## Description
Added tools to check current cognify and codify status

## DCO Affirmation


dexters1:
To release this new MCP version first cognify v0.1.40 will need to be released and the uv.lock will need to be rebuilt with cognee v0.1.40

cognee-mcp/src/server.py
```diff
@@ -138,6 +138,28 @@ async def prune():
         return [types.TextContent(type="text", text="Pruned")]
 
 
+@mcp.tool()
+async def cognify_status():
+    """Get status of cognify pipeline"""
+    with redirect_stdout(sys.stderr):
+        user = await get_default_user()
+        status = await cognee_datasets.get_status(
+            [await cognee_datasets.get_unique_dataset_id("main_dataset", user)]
```
dexters1, line 147: Would also be better to not have the dataset_name hardcoded for cognify and codify and import the names from Cognee

cognee/modules/data/methods/create_dataset.py
```diff
@@ -4,8 +4,13 @@
 from sqlalchemy.orm import joinedload
 from cognee.modules.data.models import Dataset
 
+from cognee.api.v1.datasets.datasets import datasets as cognee_datasets
```
borisarzentar, line 7: We are importing api from module, which should not happen.
Should we move `get_unique_dataset_id` to a module instead?

cognee-mcp/src/server.py
borisarzentar: What default names?

cognee/modules/data/methods/create_dataset.py
dexters1: Probably, but should we move the whole datasets class as well? And use it as a general dataset adapter

cognee-mcp/src/server.py
dexters1: main_dataset and code_dataset (or a similar name)

cognee/modules/data/methods/create_dataset.py
dexters1: get_unique_datset_id moved to module instead

borisarzentar (approved):

PR #796, opened 2025-04-29
Update dependency versions [WARNING - high risk PR ]

dexters1:
## Description
Updates all dependencies to their latest compatible versions with each other. This is needed to reduce cognee install time through pip as we have previously supported too many versions and version combinations for pip to handle properly.

This PR will likely cause issues due to all these version changes (also to third party integrations)

## DCO Affirmation


dexters1:
The more I think about this the less I want to do it. No reason to limit previous versions of software that work fine

PR #820, opened 2025-05-13
fix: cognee docker image

borisarzentar:
## DCO Affirmation


soobrosa:
@borisarzentar neat caching, might still make sense to leave a bit of a documentation on which extra is what -- I mean it could also be in `pyproject.toml`

Dockerfile
```diff
@@ -1,60 +1,60 @@
-FROM python:3.11-slim
-
-# Define Poetry extras to install
-ARG POETRY_EXTRAS="\
```
dexters1, line 4: Might be good to define the Dockerfile used extras at the beginning of the Dockerfile like before, it will make it easier for updating later on

borisarzentar: We don't pass all extras at once, like we did with poetry. Here we have to pass one by one, so I would have to iterate  somehow through extras and add each separately. I don't want to spend time on that now, we can change it later.

Vasilije1990 (dismissed):

hajdul88 (approved):

pyproject.toml
```diff
@@ -127,6 +126,7 @@ dev = [
     "mkdocs-minify-plugin>=0.8.0,<0.9",
     "mkdocstrings[python]>=0.26.2,<0.27",
 ]
+debug = ["debugpy==1.8.9"]
```
dexters1, line 129: Maybe add this under dev dependencies?

cognee/modules/pipelines/operations/pipeline.py
```diff
@@ -88,31 +101,48 @@ async def run_pipeline(
...
+
+    dataset_id = dataset.id
 
     if not data:
         data: list[Data] = await get_dataset_data(dataset_id=dataset_id)
 
     # async with update_status_lock: TODO: Add UI lock to prevent multiple backend requests
     if isinstance(dataset, Dataset):
-        task_status = await get_pipeline_status([dataset_id])
+        task_status = await get_pipeline_status([dataset_id], pipeline_name)
```
dexters1, line 133: I think this is a breaking change fot cognee-mcp checking of pipeline statuses. As there both add and cognify are under one pipeline

cognee-mcp/src/server.py
```diff
@@ -144,7 +144,9 @@ async def cognify_status():
     """Get status of cognify pipeline"""
     with redirect_stdout(sys.stderr):
         user = await get_default_user()
-        status = await get_pipeline_status([await get_unique_dataset_id("main_dataset", user)])
+        status = await get_pipeline_status(
+            [await get_unique_dataset_id("main_dataset", user)], "cognify_pipeline"
```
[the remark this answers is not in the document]
dexters1: What would happen now if the user would check the status while the add pipeline is running?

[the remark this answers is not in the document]
borisarzentar: In mcp `add` and `cognify` are merged into one, so I think that checking the `cognify_pipeline` status should be enough. Add happens usually very quickly, so I'm not sure if that can happen.

cognee/modules/pipelines/operations/pipeline.py
borisarzentar: They are not technically one pipeline, they are just called from one function. But two pipelines will run. What do you think will break?

pyproject.toml
borisarzentar: It was there, but sometimes I need just `debugpy` and nothing else. And in case of Docker container, we have a debug option there so we install now only this extra instead of the whole dev.

cognee-mcp/src/server.py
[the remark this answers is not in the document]
dexters1: If the users check right away I it will most likely happen, which is pretty realistic for them to do. Would be great to treat the return status as one pipeline then as well by checking state of both and handling the return based on combined states (or something like this)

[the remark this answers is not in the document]
Vasilije1990: Extension of scope or unclear scope, let's create a ticket for this one and merge with existing functionality

cognee/modules/pipelines/operations/pipeline.py
dexters1: I didn't initially see cognee-mcp was updated as well, you can disregard this comment

Vasilije1990 (approved):

PR #824, opened 2025-05-15
chore: Clean up core cognee repo

Vasilije1990:
## DCO Affirmation


deployment/Dockerfile_modal
```diff
@@ -21,12 +21,12 @@ WORKDIR /app
...
+COPY ../pyproject.toml poetry.lock /app/
 
 
 RUN pip install poetry
 
 RUN poetry install --all-extras --no-root --without dev
 
-COPY cognee/ /app/cognee
-COPY README.md /app/README.md
+COPY ../cognee /app/cognee
```
borisarzentar, line 31: This won't work, because docker can't access files and directories out of the current running context. You can leave it as it was, we just need to specify a file when calling `docker build -f deployment/Dockefile_modal .` and it has to be called from the cognee root dir. The same I did already in parallelization PR.

deployment/docker-compose.yml
```diff
@@ -4,8 +4,8 @@ services:
     networks:
       - cognee-network
     build:
-      context: .
-      dockerfile: Dockerfile
+      context: ..
+      dockerfile: ../Dockerfile
```
borisarzentar, line 8: Should we move `Dockerfile` to also be inside `deployment`?

borisarzentar (commented):
Alembic configuration is expected to be in the root dir, we probably need to specify directory when calling alembic now. Probably some param in `entrypoint.sh`.

deployment/docker-compose.yml
Vasilije1990: Makes sense

cognee-mcp/entrypoint.sh
```diff
@@ -13,7 +13,7 @@ echo "Environment: $ENVIRONMENT"
 # inconsistencies and should cause the startup to fail. This check allows for
 # smooth redeployments and container restarts while maintaining data integrity.
 echo "Running database migrations..."
-MIGRATION_OUTPUT=$(uv run alembic upgrade head 2>&1) || {
+MIGRATION_OUTPUT=$(uv run alembic -c alembic.ini upgrade head 2>&1) || {
```
dexters1, line 16: Does this account for the moved alembic.ini location?

deployment/Dockerfile
```diff
@@ -34,8 +34,8 @@ RUN --mount=type=cache,target=/root/.cache/uv \
     uv sync --extra debug --extra api --extra postgres --extra weaviate --extra qdrant --extra neo4j --extra kuzu --extra llama-index --extra gemini --extra ollama --extra mistral --extra groq --extra anthropic --frozen --no-install-project --no-dev --no-editable
```
dexters1, line 34: `Docker build .` for this Dockerfile does not work anymore, it can't access the cognee folder

dexters1:
Also the helm Dockerfile won't work anymore, do we use Helm anywhere?

deployment/Dockerfile_modal
borisarzentar: We will put it in the docs and readme. It is not unusual, quite common in monorepos where you have multiple Dockerfiles.
Users will probably use docker-compose that will point to correct Dockerfiles.

cognee/tests/unit/eval_framework/answer_generation_test.py
```diff
@@ -1,35 +1,35 @@
...
-    assert answers[0]["question"] == qa_pairs[0]["question"], (
-        "AnswerGeneratorExecutor is passing the question incorrectly"
-    )
-    assert answers[0]["golden_answer"] == qa_pairs[0]["answer"], (
-        "AnswerGeneratorExecutor is passing the golden answer incorrectly"
-    )
-    assert answers[0]["answer"] == "Mocked answer", (
-        "AnswerGeneratorExecutor is passing the generated answer incorrectly"
-    )
+# import pytest
```
borisarzentar, line 1: You can delete this file, since the test is not used, and tests mock.

cognee/tests/unit/eval_framework/benchmark_adapters_test.py
```diff
@@ -1,80 +1,80 @@
...
+#
+# ADAPTER_CLASSES = [
+#     HotpotQAAdapter,
+#     MusiqueQAAdapter,
+#     DummyAdapter,
+#     TwoWikiMultihopAdapter,
+# ]
+#
+#
+# @pytest.mark.parametrize("AdapterClass", ADAPTER_CLASSES)
```
borisarzentar, line 23: This one as well.

deployment/Dockerfile
borisarzentar: It has to be called from the cognee repo root, so it has cognee folder in context. `docker build -f deployment/Dockerfile .` Note `.` at the end, giving it root directory as context.

cognee-mcp/entrypoint.sh
borisarzentar: Nope, it should be changed.

borisarzentar (commented):
Should we remove those commented tests?
They also test mocks, so not so useful.

PR #833, opened 2025-05-18
feat: Add developer rules

Vasilije1990:
## DCO Affirmation

### DIFF
--- a/cognee-mcp/src/server.py
+++ b/cognee-mcp/src/server.py
@@ -24,9 +24,46 @@ log_file = get_log_file_location()
 
 
 @mcp.tool()
-async def cognify(text: str, graph_model_file: str = None, graph_model_name: str = None) -> list:
+async def cognify(data: str, graph_model_file: str = None, graph_model_name: str = None) -> list:
+    """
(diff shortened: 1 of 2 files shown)


cognee-mcp/src/server.py
```diff
@@ -23,6 +23,89 @@
...
+        ".sourcegraph/memory.md",
+        "AGENT.md",
+        "AGENTS.md"
+    ]
+
+    async def cognify_task(file_path: str) -> None:
+        with redirect_stdout(sys.stderr):
+            logger.info(f"Starting cognify for: {file_path}")
+            try:
+                await cognee.add(file_path, nodeset="developer_rules")
```
dexters1, line 80: Do we also want to add nodeset support for the regular cognify call in MCP? If we do it would be better to reuse the cognify_task by either calling cognify directly in cognee_add_developer_rules or by having cognify_task be a reusable function. The more we can avoid duplicate code the better and easier it will be to maintain it

Vasilije1990: In the first iteration, we should keep them separate. I agree we should refactor it later

dexters1 (dismissed):

PR #843, opened 2025-05-20
feat: add relational db migration and dlt example

hande-k:
## DCO Affirmation


pyproject.toml
```diff
@@ -53,7 +53,7 @@ dependencies = [
     "fastapi==0.115.7",
     "python-multipart==0.0.20",
     "fastapi-users[sqlalchemy]==14.0.1",
-    "dlt[sqlalchemy]>=1.9.0,<2",
+    "dlt[sqlalchemy]>=1.9.0",
```
dexters1, line 56: if poetry.lock has changes most likely the uv lock has changed as well. Please run uv lock and upload that file as well. Btw, do we need a higher dlt version to run this? Will it work with 1.9.0 as well?

hande-k: thanks! dlt[sqlalchemy]>=1.9.0,<2 works so i took this back

dexters1 (approved):

PR #846, opened 2025-05-20
fix: use default threading in Fastembed

lxobr:
## Description

- set the parallel option to None in Fastembed's embedding function

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/databases/vector/embeddings/FastembedEmbeddingEngine.py
+++ b/cognee/infrastructure/databases/vector/embeddings/FastembedEmbeddingEngine.py
@@ -44,7 +44,7 @@ class FastembedEmbeddingEngine(EmbeddingEngine):
                 embeddings = self.embedding_model.embed(
                     text,
                     batch_size=len(text),
-                    parallel=0,
+                    parallel=None,
                 )
 
                 return list(embeddings)


soobrosa (approved):
LGTM

cognee/infrastructure/databases/vector/embeddings/FastembedEmbeddingEngine.py
```diff
@@ -44,7 +44,7 @@ async def embed_text(self, text: List[str]) -> List[List[float]]:
                 embeddings = self.embedding_model.embed(
                     text,
                     batch_size=len(text),
-                    parallel=0,
+                    parallel=None,
```
dexters1, line 47: We were using fast embed due to it's parallelization mostly, is there a way to forward the EmbeddingEngine to these processes so they won't have to create requests that cause us the issue?

lxobr: Let's talk about it! I will merge this for now so we have something that works, and then we can create a follow up issue.

Vasilije1990: This fixes the issue. Please open a ticket to enable parallelization to work with fast embed

PR #851, opened 2025-05-21
feat: websockets for pipeline update streaming

borisarzentar:
## DCO Affirmation


hajdul88:
LGTM, lets wait for @dexters1 but I think its good, also tested it locally, messages are sent from API

dexters1:
> LGTM, lets wait for @dexters1 but I think its good, also tested it locally, messages are sent from API

I can't get it to work locally for me

cognee/eval_framework/corpus_builder/corpus_builder_executor.py
```diff
@@ -61,4 +61,7 @@ async def run_cognee(self, chunk_size=1024, chunker=TextChunker) -> None:
         await cognee.add(self.raw_corpus)
 
         tasks = await self.task_getter(chunk_size=chunk_size, chunker=chunker)
-        await cognee_pipeline(tasks=tasks)
+        pipeline_run = cognee_pipeline(tasks=tasks)
```
dexters1, line 64: This should be await cognee_pipeline I think

cognee/api/v1/cognify/cognify.py
```diff
@@ -31,13 +35,50 @@ async def cognify(
...
+
+    async for run_info in cognee_pipeline(
         tasks=tasks, datasets=datasets, user=user, pipeline_name="cognify_pipeline"
+    ):
+        pipeline_run_info = run_info
+
+    return pipeline_run_info
+
+
+async def run_cognify_as_background_process(tasks, user, datasets):
```
dexters1, line 59: Do we need this? What is it for?

dexters1: Can we move this complexity to cognee_pipeline? And have it able to transform all pipelines to work in the background or in a blocking way?

borisarzentar: `cognee_pipeline` is not a good abstraction, I can see from using it. Not flexible enough, I had to put a bunch of ifs now to handle different cases. So I think that we need one thing for `add` pipeline and separate one for other pipelines like `cognify`.

Vasilije1990: @borisarzentar @dexters1 do we merge this or keep in ambivalent state between ideal and not ideal. I'd appreciate suggestions on what in particular needs to be changed, and if nothing, then to merge

dexters1: This part can be kept as is for now regarding background processes, but I had issues with running this PR locally that I don't know if they were resolved. If they were I need to test it again to confirm it is working for me now

cognee/api/v1/cognify/routers/get_cognify_router.py
```diff
@@ -24,10 +33,109 @@ async def cognify(payload: CognifyPayloadDTO, user: User = Depends(get_authentic
...
+                    await websocket.close(code=WS_1000_NORMAL_CLOSURE)
+                    break
+            except WebSocketDisconnect:
+                remove_queue(pipeline_run_id)
+                break
+
     return router
+
+
+async def get_nodes_and_edges(data_points):
```
dexters1, line 94: Should we move this away from the router file?

cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py
```diff
@@ -310,14 +310,9 @@ async def search(
         if len(vector_list) == 0:
             return []
 
-        # Normalize vector distance and add this as score information to vector_list
-        normalized_values = normalize_distances(vector_list)
```
dexters1: We added this normalization because search was filtering out postgres results completely if the score was under 0.5 which caused a lot of problems with search

hajdul88: Yea I think this shouldn't be changed for now, at least we didn't decide to do this and then different vectordbs are not gonna work similarly.

cognee/api/v1/cognify/routers/get_cognify_router.py
borisarzentar: I have that in another PR, because I use the same function in multiple places. We can either merge it like this and I will change it with another PR, or copy the change here?

cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py
borisarzentar: Ok, then I will revert it. It was changed for crewai so I assumed it was fine.

PR #923, opened 2025-06-07
fix: Langfuse version limit

Vasilije1990:
## DCO Affirmation


dexters1:
We also need new poetry.lock and uv.lock files for these changes

dexters1:
> We also need new poetry.lock and uv.lock files for these changes

Will add them now

dexters1 (approved):

PR #962, opened 2025-06-10
fix:Extend postgres

Vasilije1990:
## DCO Affirmation


dexters1:
image
We haven't included psycopg2-binary to the postgres package intentionally due to this concern raised by the developers of the package themselves, but we can make a postgres-binary optional dependency to include it

pyproject.toml
```diff
@@ -73,6 +73,7 @@ postgres = [
     "psycopg2>=2.9.10,<3",
     "pgvector>=0.3.5,<0.4",
     "asyncpg==0.30.0",
+    "psycopg2-binary==2.9.10"
```
borisarzentar: This should be installed automatically by `psycopg2`. Was there some issue with it?

Vasilije1990: Yeah, people fail to run lib due to missing dependency. Check feedback channel on Discord + i have seen myself when we were onboarding an engineer for trial day

hajdul88: Yes I've also seen this so I can confirm it is an issue

dexters1 (approved):

PR #966, opened 2025-06-11
feat: migrate new UI to cognee

borisarzentar:
## DCO Affirmation


cognee/api/v1/add/routers/get_add_router.py
```diff
@@ -20,7 +20,7 @@ def get_add_router() -> APIRouter:
     @router.post("/", response_model=dict)
     async def add(
         data: List[UploadFile],
-        datasetName: str,
+        datasetName: Optional[str] = Form(default=None),
```
dexters1, line 23: If we want to have datasetName as optional as well I think we should say something like:
No DatasetName or datasetId provided

cognee/modules/users/authentication/get_auth_backend.py
```diff
@@ -1,40 +1,46 @@
...
-    bearer_transport = BearerTransport(tokenUrl="api/v1/auth/login")
+    transport = default_transport
 
     def get_jwt_strategy() -> JWTStrategy[models.UP, models.ID]:
+        from .default.default_jwt_strategy import DefaultJWTStrategy
+
         secret = os.getenv("FASTAPI_USERS_JWT_SECRET", "super_secret")
-        return CustomJWTStrategy(secret, lifetime_seconds=3600)
+
+        return DefaultJWTStrategy(secret, lifetime_seconds=3600)
```
dexters1: We are switching to a Cookie based auth with this change, which might cause backwards compatibility issues for our users. Does it makes sense to support both as an env parameter and have Bearer token be the default?

dexters1:
Can't get UI to run properly with docker compose

cognee/modules/users/authentication/get_auth_backend.py
borisarzentar: Good question. We can have both as well, side by side. Cookie is more secure for web based integrations, like our UI. But API based integrations will like Bearer token more.

cognee/api/v1/add/routers/get_add_router.py
borisarzentar: I checked now, we have that check already, so all good.

cognee/modules/users/authentication/get_auth_backend.py
borisarzentar: Added both, so cookie and bearer header both supported now.

cognee/infrastructure/databases/graph/kuzu/adapter.py
```diff
@@ -210,7 +215,7 @@ async def has_node(self, node_id: str) -> bool:
 
             - bool: True if the node exists, False otherwise.
         """
-        query_str = "MATCH (n:Node) WHERE n.id = $id RETURN COUNT(n) > 0"
+        query_str = "MATCH (n) WHERE n.id = $id RETURN COUNT(n) > 0"
```
hajdul88: Maybe its needed I dont know the context but why removing :Node filter?

cognee/infrastructure/databases/graph/kuzu/adapter.py
```diff
@@ -350,7 +355,7 @@ async def delete_node(self, node_id: str) -> None:
 
             - node_id (str): The identifier of the node to be deleted.
         """
-        query_str = "MATCH (n:Node) WHERE n.id = $id DETACH DELETE n"
+        query_str = "MATCH (n) WHERE n.id = $id DETACH DELETE n"
```
hajdul88: And basically same for all :Node removal

cognee/api/v1/cognify/routers/get_cognify_router.py
```diff
@@ -47,16 +65,33 @@ async def cognify(payload: CognifyPayloadDTO, user: User = Depends(get_authentic
     async def subscribe_to_cognify_info(websocket: WebSocket, pipeline_run_id: str):
         await websocket.accept()
 
-        auth_message = await websocket.receive_json()
+        access_token = websocket.cookies.get(os.getenv("AUTH_TOKEN_COOKIE_NAME", "auth_token"))
```
dexters1, line 67: Does this work with Bearer Tokens?

cognee/api/v1/datasets/routers/get_datasets_router.py
```diff
@@ -100,54 +141,32 @@ async def delete_data(
...
-        from cognee.infrastructure.databases.graph import get_graph_engine
-
         try:
-            graph_client = await get_graph_engine()
-            (nodes, edges) = await graph_client.get_graph_data()
+            from cognee.modules.data.methods import get_dataset
+
+            dataset = await get_dataset(user.id, dataset_id)
+
+            await set_database_global_context_variables(dataset.name, user.id)
```
dexters1: Maybe we should move this handling to the library level (as the rest of the context var handling is in the lib), directly in the get_formatted_graph_data in case permissions are enabled

cognee/api/v1/delete/routers/get_delete_router.py
```diff
@@ -14,7 +14,7 @@
 def get_delete_router() -> APIRouter:
     router = APIRouter()
 
-    @router.delete("/", response_model=None)
+    @router.delete("", response_model=None)
```
dexters1, line 17: We can add a check here that the user has 'delete' permission for the dataset

dexters1: Maybe on the library side? (delete function)

cognee/modules/search/methods/search.py
```diff
@@ -153,7 +160,10 @@ async def permissions_search(
     query = await log_query(query_text, query_type.value, user.id)
 
     # Find datasets user has read access for (if datasets are provided only return them. Provided user has read access)
-    search_datasets = await get_specific_user_permission_datasets(user.id, "read", dataset_ids)
+    try:
+        search_datasets = await get_specific_user_permission_datasets(user.id, "read", dataset_ids)
+    except PermissionDeniedError:
+        search_datasets = []
```
dexters1: Why add this exception handling? The user should know there are no datasets he has read access for instead of just not returning a result, also makes it hard for us to test and debug if the exception is handled

cognee/modules/users/get_fastapi_users.py
```diff
@@ -1,16 +1,21 @@
...
 
 @lru_cache
 def get_fastapi_users():
-    auth_backend = get_auth_backend()
+    api_auth_backend = get_api_auth_backend()
+    client_auth_backend = get_client_auth_backend()
 
-    fastapi_users = FastAPIUsers[User, uuid.UUID](get_user_manager, [auth_backend])
+    fastapi_users = FastAPIUsers[User, uuid.UUID](
+        get_user_manager, [api_auth_backend, client_auth_backend]
```
dexters1, line 18: Does this allow us to both use cookies and Bearer tokens now?

dexters1: The login endpoint doesn't return the Bearer token anymore

cognee/api/v1/cognify/routers/get_cognify_router.py
borisarzentar: It does not at the moment, we can enable it later.

cognee/modules/search/methods/search.py
borisarzentar: I can handle it in the router then. We don't want to show permission error in the UI, but return no results for search.

cognee/modules/users/get_fastapi_users.py
borisarzentar: It allows to use both, you can send token in a cookie, or send authentication header. I added authorization header now.

cognee/api/v1/delete/routers/get_delete_router.py
borisarzentar: Let's add this later, I don't want to expand score more.

cognee/api/v1/datasets/routers/get_datasets_router.py
borisarzentar: Moved

cognee/infrastructure/databases/graph/kuzu/adapter.py
borisarzentar: Reverted this

cognee/infrastructure/databases/graph/kuzu/adapter.py
borisarzentar: Reverted this

dexters1 (dismissed):

PR #978, opened 2025-06-12
Feature: Makes s3 pathway imports optional so cognee can run without s3fs

hajdul88:
## Description
Makes s3 pathway imports optional so cognee can run without s3fs

## DCO Affirmation


pyproject.toml
```diff
@@ -118,6 +115,7 @@ gui = [
     "qasync>=0.27.1,<0.28",
 ]
 graphiti = ["graphiti-core>=0.7.0,<0.8"]
+aws = ["s3fs[boto3]==2025.3.2"]
```
dexters1, line 120: Please add Note from above as well so we never forget this disaster that is s3fs with boto3 when not versioned together

hajdul88: True its better. will do thanks for pointing it out

hajdul88: Done

Vasilije1990 (approved):

PR #988, opened 2025-06-16
feat: s3 storage

borisarzentar:
## DCO Affirmation


cognee/tests/test_library.py
```diff
@@ -89,14 +90,21 @@ async def main():
...
+
+    assert not await file_storage.file_exists(file_path), (
+        "SQLite relational database is not deleted"
+    )
 
     from cognee.infrastructure.databases.graph import get_graph_config
 
     graph_config = get_graph_config()
-    assert not os.path.exists(graph_config.graph_file_path), "Networkx graph database is not empty"
+    assert not await file_storage.file_exists(os.path.basename(graph_config.graph_file_path)), (
```
hajdul88: This changes lately due to kuzu to default pr but probably its gonna be a conflict

cognee/api/v1/config/config.py
```diff
@@ -9,17 +9,18 @@
...
+        base_config.system_root_directory = system_root_directory
+
         databases_directory_path = os.path.join(system_root_directory, "databases")
 
         relational_config = get_relational_config()
         relational_config.db_path = databases_directory_path
-        LocalStorage.ensure_directory_exists(databases_directory_path)
 
         graph_config = get_graph_config()
         graph_config.graph_file_path = os.path.join(databases_directory_path, "cognee.graph")
```
hajdul88: I know now this part is not changed here, but will effect the S3 use case.  This is gonna change the graph_file_name env variable always to cognee.graph when a user is using S3 as data and root dirs. So maybe we could use the name from env here.

hajdul88 (commented):
These are the two things that I found in the code, the rest we discussed on slack. Kuzu + prune deleting S3, otherwise it looks good.

cognee/tests/test_library.py
borisarzentar: Yes, resolved.

cognee/api/v1/config/config.py
borisarzentar: Using now.

cognee/context_global_variables.py
```diff
@@ -40,28 +41,33 @@ async def set_database_global_context_variables(dataset: Union[str, UUID], user_
...
     dataset_database = await get_or_create_dataset_database(dataset, user)
 
-    # TODO: Find better location for database files
-    cognee_directory_path = str(
-        pathlib.Path(
-            os.path.join(pathlib.Path(__file__).parent, f".cognee_system/databases/{user.id}")
-        ).resolve()
+    base_config = get_base_config()
+
+    data_root_directory = os.path.join(base_config.data_root_directory, str(user.id))
```
dexters1: We can't do this, it won't work with the data deduplication mehcanism

dexters1: If multiple users upload the same file we save it only once and keep only one instance as deduplication

cognee/modules/pipelines/operations/pipeline.py
```diff
@@ -121,7 +121,7 @@ async def run_pipeline(
     check_dataset_name(dataset.name)
 
     # Will only be used if ENABLE_BACKEND_ACCESS_CONTROL is set to True
-    await set_database_global_context_variables(dataset.name, user.id)
+    await set_database_global_context_variables(dataset.id, user.id)
```
dexters1: needs to be dataset.id and dataset.owner_id (for conflict resolution)

borisarzentar: Changed it

cognee/context_global_variables.py
borisarzentar: We deduplicate by tenant now, or by user if tenant is not set.

PR #989, opened 2025-06-16
feat: Add alembic migration for permissions

dexters1:
## DCO Affirmation

### DIFF
--- a/alembic/versions/ab7e313804ae_permission_system_rework.py
+++ b/alembic/versions/ab7e313804ae_permission_system_rework.py
@@ -34,10 +34,26 @@ def _ensure_permission(conn, permission_name) -> str:
 
     if row is None:
         permission_id = uuid4()
-        # TODO: The Permission table might change in future version of Cognee,
-        #       if this happens this migration version won't work anymore. Not sure what to do
+
+        permissions_table = sa.Table(
+            "acls",
+            sa.MetaData(),
+            sa.Column("id", UUID, primary_key=True, default=uuid4),
+            sa.Column(
+                "created_at", sa.DateTime(timezone=True), default=lambda: datetime.now(timezone.utc)
+            ),
+            sa.Column(
+                "updated_at",
+                sa.DateTime(timezone=True),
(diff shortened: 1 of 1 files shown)


alembic/versions/ab7e313804ae_permission_system_rework.py
```diff
@@ -0,0 +1,97 @@
...
+    permission = conn.execute(
+        sa.select(Permission).filter(Permission.name == permission_name)
+    ).fetchone()
+
+    if permission is None:
+        permission = Permission(name=permission_name)
+
+    return {
+        "id": uuid4(),
+        "create_at": datetime.now(timezone.utc),
```
borisarzentar: `create_at` -> `created_at`

alembic/versions/ab7e313804ae_permission_system_rework.py
```diff
@@ -0,0 +1,97 @@
...
+        return postgresql.UUID(as_uuid=True)
+    # SQLite (and others): fall back to CHAR(36) – application inserts uuid4()
+    return sa.CHAR(36)
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+
+    # Recreate ACLs table with default permissions set to datasets instead of documents
+    op.drop_table("acls")
```
borisarzentar, line 101: Let's use `ACL.__table__` here.

alembic/versions/ab7e313804ae_permission_system_rework.py
```diff
@@ -0,0 +1,97 @@
...
+
+    uuid_type = _uuid_type()
+    op.create_table(
+        "acls",
```
borisarzentar, line 104: `ACL.__table__` here as well.

alembic/versions/ab7e313804ae_permission_system_rework.py
dexters1: We can't as ACL might be a different table in the future and this migration won't work then, I've manually recreated the current state of the ACL table instead

PR #1009, opened 2025-06-20
feat: Add dataset_id to pipeline run info and status

dexters1:
## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/databases/graph/config.py
+++ b/cognee/infrastructure/databases/graph/config.py
@@ -3,6 +3,8 @@
 import os
 from functools import lru_cache
 from pydantic_settings import BaseSettings, SettingsConfigDict
+import pydantic
+from pydantic import Field
 from cognee.shared.data_models import KnowledgeGraph
 from cognee.root_dir import get_absolute_path
 
@@ -29,18 +31,33 @@ class GraphConfig(BaseSettings):
     - model_config
     """
 
-    graph_filename: str = "cognee_graph"
-    graph_database_provider: str = "kuzu"
+    # Using Field we are able to dynamically load current GRAPH_DATABASE_PROVIDER value in the model validator part
+    # and determine default graph db file and path based on this parameter if no values are provided
+    graph_database_provider: str = Field("kuzu", env="GRAPH_DATABASE_PROVIDER")
+
     graph_database_url: str = ""
     graph_database_username: str = ""
     graph_database_password: str = ""
     graph_database_port: int = 123
-    graph_file_path: str = os.path.join(
-        os.path.join(get_absolute_path(".cognee_system"), "databases"), graph_filename
-    )
+    graph_file_path: str = ""
+    graph_filename: str = ""
     graph_model: object = KnowledgeGraph
     graph_topology: object = KnowledgeGraph
-    model_config = SettingsConfigDict(env_file=".env", extra="allow")
+    model_config = SettingsConfigDict(env_file=".env", extra="allow", populate_by_name=True)
+
+    # Model validator updates graph_filename and path dynamically after class creation based on current database provider
+    # If no specific graph_filename or path are provided
+    @pydantic.model_validator(mode="after")
+    def fill_derived(cls, values):
+        provider = values.graph_database_provider.lower()
+        # Set filename based on graph database provider if no filename is provided
+        if not values.graph_filename:
+            values.graph_filename = f"cognee_graph_{provider}"
+        # Set file path based on graph database provider if no file path is provided
+        if not values.graph_file_path:
+            base = os.path.join(get_absolute_path(".cognee_system"), "databases")
+            values.graph_file_path = os.path.join(base, values.graph_filename)
+        return values
 
     def to_dict(self) -> dict:
         """
--- a/cognee/infrastructure/databases/graph/kuzu/adapter.py
+++ b/cognee/infrastructure/databases/graph/kuzu/adapter.py
@@ -46,11 +46,7 @@ class KuzuAdapter(GraphDBInterface):
     def _initialize_connection(self) -> None:
         """Initialize the Kuzu database connection and schema."""
         try:
-            try:
-                os.makedirs(self.db_path, exist_ok=True)
-            except FileExistsError:
-                os.remove(self.db_path)
-                os.makedirs(self.db_path, exist_ok=True)
+            os.makedirs(self.db_path, exist_ok=True)
 
             self.db = Database(self.db_path)
             self.db.init_database()
@@ -79,7 +75,7 @@ class KuzuAdapter(GraphDBInterface):
             logger.debug("Kuzu database initialized successfully")
         except Exception as e:
             logger.error(f"Failed to initialize Kuzu database: {e}")
-            raise
+            raise e
 
     async def query(self, query: str, params: Optional[dict] = None) -> List[Tuple]:
         """
--- a/cognee/modules/users/permissions/methods/check_permission_on_dataset.py
+++ b/cognee/modules/users/permissions/methods/check_permission_on_dataset.py
@@ -1,14 +1,10 @@
 from cognee.shared.logging_utils import get_logger
 from cognee.modules.users.methods import get_default_user
 from uuid import UUID
-from sqlalchemy import select
-from sqlalchemy.orm import joinedload
 
-from cognee.modules.users.exceptions import PermissionDeniedError
-from cognee.infrastructure.databases.relational import get_relational_engine
+from cognee.modules.users.permissions.methods import get_specific_user_permission_datasets
 
 from ...models.User import User
-from ...models.ACL import ACL
 
 logger = get_logger()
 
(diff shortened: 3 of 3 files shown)


cognee/modules/pipelines/__init__.py
```diff
@@ -2,3 +2,4 @@
 from .operations.run_tasks import run_tasks
 from .operations.run_parallel import run_tasks_parallel
 from .operations.pipeline import cognee_pipeline
+from .operations.pipeline import merge_pipeline_run_info
```
borisarzentar: I wouldn't expose it from here, this is mostly high-level stuff. I think from `operations` is enough.

cognee/modules/pipelines/operations/get_pipeline_status.py
```diff
@@ -33,3 +33,34 @@ async def get_pipeline_status(dataset_ids: list[UUID], pipeline_name: str):
         pipeline_statuses = {str(run.dataset_id): run.status for run in runs}
 
         return pipeline_statuses
+
+
+async def get_pipeline_run(dataset_id: UUID, pipeline_name: str):
```
borisarzentar: Let's move it to its own file.

cognee/modules/pipelines/operations/pipeline.py
```diff
@@ -148,22 +156,45 @@ async def run_pipeline(
...
+        str(dataset_id) in task_status
+        and task_status[str(dataset_id)] == PipelineRunStatus.DATASET_PROCESSING_STARTED
+    ):
+        logger.info("Dataset %s is already being processed.", dataset_id)
+        pipeline_run = await get_pipeline_run(dataset_id, pipeline_name)
+        yield PipelineRunStarted(
+            pipeline_run_id=pipeline_run.pipeline_run_id,
+            datasets={dataset.name: dataset.id},
+            payload=data,
+        )
```
borisarzentar: You will need that `return` here as we discsussed.

cognee/modules/pipelines/operations/pipeline.py
```diff
@@ -148,22 +156,45 @@ async def run_pipeline(
...
+    elif (
+        str(dataset_id) in task_status
+        and task_status[str(dataset_id)] == PipelineRunStatus.DATASET_PROCESSING_COMPLETED
+    ):
+        logger.info("Dataset %s is already processed.", dataset_id)
+        pipeline_run = await get_pipeline_run(dataset_id, pipeline_name)
+        yield PipelineRunCompleted(
+            pipeline_run_id=pipeline_run.pipeline_run_id, datasets={dataset.name: dataset.id}
+        )
```
borisarzentar: You will need that `return` here as well.

cognee/modules/pipelines/operations/pipeline.py
```diff
@@ -148,22 +156,45 @@ async def run_pipeline(
...
-        if not isinstance(task, Task):
-            raise ValueError(f"Task {task} is not an instance of Task")
+        async for pipeline_run_info in pipeline_run:
+            yield pipeline_run_info
 
-    pipeline_run = run_tasks(tasks, dataset_id, data, user, pipeline_name, context=context)
 
-    async for pipeline_run_info in pipeline_run:
-        yield pipeline_run_info
+def merge_pipeline_run_info(old_run_info, new_run_info):
```
borisarzentar: This could be in `utils` directory.

cognee/modules/pipelines/operations/pipeline.py
```diff
@@ -148,22 +156,45 @@ async def run_pipeline(
...
+    if old_run_info:
+        updated_run_info = old_run_info
+        # Combine dataset dictionaries of run_infos
+        updated_run_info.datasets = old_run_info.datasets | new_run_info.datasets
```
borisarzentar: We don't want to merge status?

cognee/modules/pipelines/operations/run_tasks.py
```diff
@@ -86,6 +87,13 @@ async def run_tasks(
     if not user:
         user = get_default_user()
 
+    # Get Dataset object
+    db_engine = get_relational_engine()
```
borisarzentar, line 91: We already have `get_dataset` method, should we use it here?

dexters1: get_dataset method only gets datasets owned by current user, we now support work with datasets not owned by user as well. Would probably make sense to have two different functions now for this change 'get_dataset' and 'get_user_dataset' (if we even need this)

cognee/modules/pipelines/operations/pipeline.py
dexters1: We haven't agreed on what to return in the end, this is still just a draft. Whatever we return though it will be a breaking change for the endpoint for our users unfortunately (if they even used return value of these endpoints)

cognee/api/v1/cognify/code_graph_pipeline.py
```diff
@@ -65,16 +67,27 @@ async def run_code_graph_pipeline(repo_path, include_docs=False):
             ),
         ]
 
-    dataset_id = await get_unique_dataset_id("codebase", user)
+    dataset_name = "codebase"
```
borisarzentar, line 68: Should we set `repo_path` to be the `dataset_name`? That way it won't mix graph data between multiple repos.

cognee/api/v1/cognify/code_graph_pipeline.py
```diff
@@ -65,16 +67,27 @@ async def run_code_graph_pipeline(repo_path, include_docs=False):
...
+    # Create a Dataset instance
+    dataset = Dataset(
```
borisarzentar: Do we need to create instance? We create it for the purpose of reading the `id` from it, which we also supply to it.

cognee/api/v1/cognify/routers/get_cognify_router.py
```diff
@@ -56,7 +56,7 @@ async def cognify(payload: CognifyPayloadDTO, user: User = Depends(get_authentic
                 datasets, user, payload.graph_model, run_in_background=payload.run_in_background
             )
 
-            return cognify_run.model_dump()
+            return cognify_run
```
borisarzentar, line 59: Should we update the `response_model` as well? Right now it is `None`.

cognee/modules/pipelines/methods/get_pipeline_run_by_dataset.py
```diff
@@ -0,0 +1,36 @@
...
+            .filter(PipelineRun.dataset_id == dataset_id)
+            .filter(PipelineRun.pipeline_name == pipeline_name)
+            .subquery()
+        )
+
+        aliased_pipeline_run = aliased(PipelineRun, query)
+
+        latest_run = select(aliased_pipeline_run).filter(query.c.rn == 1)
+
+        run = (await session.execute(latest_run)).scalars().all()
```
borisarzentar: `.first()` instead of `all()` should return just one element or None if it doesn't exist.

cognee/api/v1/cognify/code_graph_pipeline.py
dexters1: Yes we do, as we need to read the dataset from the database later

cognee/api/v1/cognify/code_graph_pipeline.py
dexters1: Depends on what the plan for the CodeGraph is, if it will be in the SaaS it should def change a lot. I wouldn't make changes now until we know what we want from it

cognee/modules/pipelines/methods/get_pipeline_run_by_dataset.py
dexters1: True, will add it on Monday

cognee/api/v1/cognify/routers/get_cognify_router.py
dexters1: Yes definitely

cognee/api/v1/cognify/code_graph_pipeline.py
borisarzentar: But this doesn't create it in the db, or it does?

cognee/api/v1/cognify/code_graph_pipeline.py
borisarzentar: I agree. But with this change we can already cover both use cases.

cognee/api/v1/cognify/code_graph_pipeline.py
dexters1: this one creates it in the db: create_dataset

cognee/api/v1/cognify/code_graph_pipeline.py
dexters1: But what if people want to combine repos? Some people have code in multiple repos (like backend frontend etc.)

dexters1: I'd just leave it for now as is

dexters1: I think it needs more time and thought to be done properly

cognee/api/v1/cognify/code_graph_pipeline.py
borisarzentar: Yes, but we don't pass dataset instance to it. So it is a bit confusing why we create Dataset instance when we only read id from it, which we set before.

dexters1: True, I've removed pydantic instance creation and let it use the created dataset now

Vasilije1990 (approved):

PR #1016, opened 2025-06-24
Update README.md

siddharth0-dev:
## DCO Affirmation


Vasilije1990:
@siddharth0-dev thank you for the contribution. We do not encourage readme PRs, but feel free to open a ticket and we will change based on the request and if it makes sense

PR #1021, opened 2025-06-25
Feat: adds cursor developer rules demo

hajdul88:
## Description
Adds cursor developer rules demo and functionality to mcp server

## DCO Affirmation

### DIFF
--- a/cognee-mcp/src/server.py
+++ b/cognee-mcp/src/server.py
@@ -188,6 +188,58 @@ async def cognify(data: str, graph_model_file: str = None, graph_model_name: str
     ]
 
 
+@mcp.tool(
+    name="save_interaction", description="Logs user-agent interactions and query-answer pairs"
+)
+async def save_interaction(data: str) -> list:
+    """
+    Transform and save a user-agent interaction into structured knowledge.
+
+    Parameters
+    ----------
+    data : str
+        The input string containing user queries and corresponding agent answers.
+
+    Returns
+    -------
+    list
+        A list containing a single TextContent object with information about the background task launch.
+    """
+
+    async def save_user_agent_interaction(data: str) -> None:
+        """Build knowledge graph from the interaction data"""
+        with redirect_stdout(sys.stderr):
+            logger.info("Save interaction process starting.")
+
+            await cognee.add(data, node_set=["user_agent_interaction"])
+
+            try:
+                await cognee.cognify()
+                logger.info("Save interaction process finished.")
+            except Exception as e:
+                logger.error("Save  interaction process failed.")
+                raise ValueError(f"Failed to Save interaction: {str(e)}")
+
+    asyncio.create_task(
+        save_user_agent_interaction(
+            data=data,
+        )
+    )
+
+    log_file = get_log_file_location()
+    text = (
+        f"Background process launched to process the user-agent interaction.\n"
+        f"To check the current status, use the cognify_status tool or check the log file at: {log_file}"
+    )
+
+    return [
+        types.TextContent(
+            type="text",
+            text=text,
+        )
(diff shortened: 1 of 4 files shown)


dexters1:
Should we move all the MCP related code to be a part of the cognee-mcp subdirectory and not cognee core?

hajdul88:
> Should we move all the MCP related code to be a part of the cognee-mcp subdirectory and not cognee core?

This was a discussion we had with @borisarzentar, now I left it there because none of the other tools are implemented under the mcp directory. Ultimately this should go somewhere to community repo to a community mcp server or smthg like that. But since we plan to restructure this I don't think now it matters where we have this.

cognee-mcp/src/server.py
```diff
@@ -188,6 +194,64 @@ async def cognify_task(
...
+                await cognee.cognify()
+                logger.info("Save interaction process finished.")
+                logger.info("Generating associated rules from interaction data.")
+
+                await rule_associations(data=data, rules_nodeset_name="coding_agent_rules")
+
+                logger.info("Associated rules generated from interaction data.")
+
+            except Exception as e:
+                logger.error("Save  interaction process failed.")
```
borisarzentar:
```suggestion
                logger.error("Save interaction process failed.")
```

cognee-mcp/src/server.py
```diff
@@ -319,6 +383,41 @@ async def search_task(search_query: str, search_type: str) -> str:
     return [types.TextContent(type="text", text=search_results)]
 
 
+@mcp.tool()
+async def get_developer_rules() -> list:
+    """
+    Retrieve all developer rules that were generated based on previous interactions.
+
+    This tool queries the Cognee knowledge graph and returns a list of developer
+    rules..
```
borisarzentar:
```suggestion
    rules.
```

cognee-mcp/src/server.py
```diff
@@ -188,6 +194,64 @@ async def cognify_task(
...
+    data : str
+        The input string containing user queries and corresponding agent answers.
+
+    Returns
+    -------
+    list
+        A list containing a single TextContent object with information about the background task launch.
+    """
+
+    async def save_user_agent_interaction(data: str) -> None:
```
borisarzentar, line 215: Is this something that cursor will call and provide interactions between user and cursor?
We also collect the search history in our system, not sure if that can be useful.

cognee/modules/codingagents/coding_rule_associations.py
```diff
@@ -0,0 +1,120 @@
...
+                    )
+            except Exception as e:
+                logger.info(f"Warning: Skipping invalid rule due to error: {e}")
+    else:
+        logger.info("No valid origin_id or rules provided.")
+
+    return relationships
+
+
+async def rule_associations(data: str, rules_nodeset_name: str):
```
borisarzentar: Since it is adding data, maybe name it more explicit, like `add_rule_associations` or something like that.

cognee-mcp/src/server.py
hajdul88: Yes this it a tool that gets the actual user-agent interaction (basically what the user is chatting with cursor + the corresponding answer), cursor calls it automatically (if rule is defined) and then it saves it into cognee + associates developer rules from the actual conversation.

hajdul88: I think search history is different, this is more about user-agent interaction in cursor

cognee/modules/codingagents/coding_rule_associations.py
hajdul88: Done

Vasilije1990 (approved):

PR #1026, opened 2025-06-27
feat: set default database name based on provider

dexters1:
## Description
Default graph name will consider provider when being created

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/databases/graph/config.py
+++ b/cognee/infrastructure/databases/graph/config.py
@@ -1,6 +1,6 @@
 """This module contains the configuration for the graph database."""
 
-from pathlib import Path
+import os
 from functools import lru_cache
 from pydantic_settings import BaseSettings, SettingsConfigDict
 import pydantic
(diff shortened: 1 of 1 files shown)


cognee/infrastructure/databases/graph/config.py
```diff
@@ -29,18 +31,33 @@ class GraphConfig(BaseSettings):
     - model_config
     """
 
-    graph_filename: str = "cognee_graph"
-    graph_database_provider: str = "kuzu"
+    # Using Field we are able to dynamically load current GRAPH_DATABASE_PROVIDER value in the model validator part
+    # and determine default graph db file and path based on this parameter if no values are provided
+    graph_database_provider: str = Field("kuzu", env="GRAPH_DATABASE_PROVIDER")
```
borisarzentar, line 36: Shouldn't it automatically map `graph_database_provider` to `GRAPH_DATABASE_PROVIDER`?

borisarzentar (approved):

cognee/infrastructure/databases/graph/config.py
dexters1: I'm not sure, maybe, I need to try it out with the post-processing of names in this file. I've initially added this hoping it would map to the env. values right away so that we don't need post processing but that didn't work

PR #1045, opened 2025-07-02
feat: adding docstrings

Vasilije1990:
## DCO Affirmation

### DIFF
--- a/cognee/api/v1/cognify/cognify.py
+++ b/cognee/api/v1/cognify/cognify.py
@@ -41,17 +41,17 @@ async def cognify(
 ):
     """
     Transform ingested data into a structured knowledge graph.
-    
-    This is the core processing step in Cognee that converts raw text and documents 
-    into an intelligent knowledge graph. It analyzes content, extracts entities and 
+
+    This is the core processing step in Cognee that converts raw text and documents
+    into an intelligent knowledge graph. It analyzes content, extracts entities and
     relationships, and creates semantic connections for enhanced search and reasoning.
-    
+
     Prerequisites:
         - **LLM_API_KEY**: Must be configured (required for entity extraction and graph generation)
         - **Data Added**: Must have data previously added via `cognee.add()`
         - **Vector Database**: Must be accessible for embeddings storage
         - **Graph Database**: Must be accessible for relationship storage
-    
+
     Input Requirements:
         - **Datasets**: Must contain data previously added via `cognee.add()`
         - **Content Types**: Works with any text-extractable content including:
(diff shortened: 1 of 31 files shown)


cognee/api/v1/cognify/cognify.py
```diff
@@ -39,6 +39,151 @@ async def cognify(
...
+        7. **Content Summarization**: Creates hierarchical summaries for navigation
+
+    Graph Model Customization:
+        The `graph_model` parameter allows custom knowledge structures:
+        - **Default**: General-purpose KnowledgeGraph for any domain
+        - **Custom Models**: Domain-specific schemas (e.g., scientific papers, code analysis)
+        - **Ontology Integration**: Use `ontology_file_path` for predefined vocabularies
+
+    Args:
+        datasets: Dataset name(s) to process. Processes all available data if None.
```
dexters1: We also accept dataset UUID, will update this part

cognee/api/v1/search/search.py
```diff
@@ -20,6 +20,142 @@ async def search(
...
+            - Single dataset: "research_papers"
+            - Multiple datasets: ["docs", "reports", "analysis"]
+            - None: Search across all user datasets
+
+        dataset_ids: Alternative to datasets - use specific UUID identifiers.
+
+        system_prompt_path: Custom system prompt file for LLM-based search types.
+                          Defaults to "answer_simple_question.txt".
+
+        top_k: Maximum number of results to return (1-100).
```
dexters1: top_k does not have a top limit, especially for migrated relational databases top_k there can be in the thousands as there are no text chunks to occupy LLM context

cognee/api/v1/search/search.py
```diff
@@ -20,6 +20,142 @@ async def search(
...
+            Pre-generated hierarchical summaries of content.
+            Best for: Quick overviews, document abstracts, topic summaries.
+            Returns: Multi-level summaries from detailed to high-level.
+
+        **CODE**:
+            Code-specific search with syntax and semantic understanding.
+            Best for: Finding functions, classes, implementation patterns.
+            Returns: Structured code information with context and relationships.
+
+        **CYPHER**:
```
dexters1, line 69: Do we want to mention also the other search types like GRAPH_COMPLETION_COT (chain of thought). That one should be our most accurate search option and might be interesting to people

Vasilije1990: This is not docs

cognee/api/v1/search/search.py
Vasilije1990: Better give them smth sensible, since you don't want to overload memory

cognee/api/v1/cognify/cognify.py
Vasilije1990: Fair

cognee/api/v1/search/search.py
dexters1: Maybe just not mention any maximum number? You've warned them about results being noisy for higher values

Vasilije1990: Done

PR #1047, opened 2025-07-03
fix: Adds graceful handling quick fix for damaged pdf files

hajdul88:
## Description
fix: Adds graceful handling quick fix for damaged pdf files

## DCO Affirmation

### DIFF
--- a/cognee/modules/data/processing/document_types/PdfDocument.py
+++ b/cognee/modules/data/processing/document_types/PdfDocument.py
@@ -2,6 +2,9 @@ from pypdf import PdfReader
 from cognee.modules.chunking.Chunker import Chunker
 from .open_data_file import open_data_file
 from .Document import Document
+from cognee.shared.logging_utils import get_logger
+
+logger = get_logger("PDFDocument")
 
 
 class PdfDocument(Document):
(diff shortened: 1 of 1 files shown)


cognee/modules/data/processing/document_types/PdfDocument.py
```diff
@@ -2,19 +2,35 @@
...
 
     def read(self, chunker_cls: Chunker, max_chunk_size: int):
         with open_data_file(self.raw_data_location, mode="rb") as stream:
-            file = PdfReader(stream)
+            logger.info(f"Reading PDF:{self.raw_data_location}")
+            try:
+                file = PdfReader(stream, strict=False)
+            except Exception as e:
+                logger.warning(
+                    f"PyPDF couldn’t open PDF—skipping: {self.raw_data_location} with error: {e}"  # TODO: Once incremental pipeline implementation is done this has to save pdf errored state into metastore
```
dexters1: Comment to expand the TODO. We also need to return to the user information on which document processing was successful and which was not and not just in metastore

borisarzentar: We can have a custom error for this that we catch where we need to log that error. Don't send unnecessary data down the line, we will collect it at the surface.

cognee/modules/data/processing/document_types/PdfDocument.py
```diff
@@ -2,19 +2,35 @@
...
             def get_text():
-                for page in file.pages:
-                    page_text = page.extract_text()
-                    yield page_text
+                try:
+                    for page in file.pages:
+                        page_text = page.extract_text()
+                        yield page_text
+                except Exception as e:
+                    logger.warning(
```
borisarzentar: Same for this, custom error would be nice.

cognee/modules/data/processing/document_types/PdfDocument.py
hajdul88: @dexters1 I think if we have errored info in the metastore, based on that the user can see which files were errored when using the UI. For API yes maybe, I created task for this

hajdul88: @borisarzentar yes this was a quick fix for the modal run and I agree it is not nice as it is

hajdul88: Will see if tomorrow I have time to do it

cognee/modules/data/processing/document_types/PdfDocument.py
hajdul88: Also where do we want to catch the pypdf error?

hajdul88: Brought it til the extract chunks from docs

cognee/modules/data/processing/document_types/PdfDocument.py
hajdul88: Done

dexters1: I don't think it's enough to just log it as it will be lost in the logs, users won't notice. I think it needs to be part of the return value of Cognify to specify which data in a dataset was handled properly and which wasn't. We already have a dictionary per dataset_id in cognify we can add a new value in that dictionary for data status in there

It can be done as a post processing step in Cognify or as a last Task, but it should be done this way so users can have this information

hajdul88: Yea that can be also done, postprocessing step is smthg that we have to solve yes and I would also do it as post step. Anyhow, now its done on extract_chunks task level, later when we have the incremental done the exception can be passed onto cognify level.

PR #1068, opened 2025-07-08
feat: Weighted edges

Vasilije1990:
## DCO Affirmation


cognee/infrastructure/engine/models/Edge.py
```diff
@@ -0,0 +1,84 @@
...
+        # Mixed usage
+        has_items: (Edge(weight=0.5, weights={"confidence": 0.9}), list[Item])
+    """
+
+    weight: Optional[float] = None
+    weights: Optional[Dict[str, float]] = None
+    relationship_type: Optional[str] = None
+    properties: Optional[Dict[str, Any]] = None
+
+    def to_dict(self) -> Dict[str, Any]:
```
borisarzentar: We don't use these `to_dict()` methods, they are unnecessary because you can read attributes directly from the instance.

cognee/infrastructure/engine/models/Edge.py
```diff
@@ -0,0 +1,84 @@
...
+        has_items: (Edge(weight=0.5), list[Item])
+
+        # Multiple weights
+        has_items: (Edge(weights={"strength": 0.8, "confidence": 0.9, "importance": 0.7}), list[Item])
+
+        # Mixed usage
+        has_items: (Edge(weight=0.5, weights={"confidence": 0.9}), list[Item])
+    """
+
+    weight: Optional[float] = None
```
borisarzentar, line 23: We are introducing `weight` and `weights` as standard properties when `Edge` edge type is used?

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -21,44 +21,63 @@ async def get_graph_from_model(
...
     for field_name, field_value in data_point:
         if field_name == "metadata":
             continue
 
-        if isinstance(field_value, DataPoint):
+        # Check if field_value is a tuple with Edge metadata
+        edge_metadata = None
+        actual_field_value = field_value
+
+        if isinstance(field_value, tuple) and len(field_value) == 2:
```
borisarzentar: You can merge this if with one below, add here `and isinstance(first_element, Edge)`.
Then you can destructure `field_value` directly into `edge_metadata` and `actual_field_value`.

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -21,44 +21,63 @@ async def get_graph_from_model(
...
-                properties_to_visit.add(f"{field_name}.{index}")
+                properties_to_visit.append(
+                    (f"{field_name}.{index}", edge_metadata)
+                )  # Changed from add to append
+
 
             continue
 
-        data_point_properties[field_name] = field_value
+        # Only add to properties if it's not a tuple with Edge metadata
```
borisarzentar: It is a bit hard to follow. Let's add new checks after the old ones. So we first check if it is a single `DataPoint`, then list of `DataPoints`, then `tuple[Edge, DataPoint] and then `tuple[Edge, any]` and at the end we pass through only values that are not data points and don't have edge config.

cognee/infrastructure/engine/models/DataPoint.py
```diff
@@ -34,6 +34,8 @@ class DataPoint(BaseModel):
     - from_dict
     """
 
+    model_config = ConfigDict(arbitrary_types_allowed=True)
```
borisarzentar, line 37: We do this in order to be able to set `tuple` as a value of DataPoint attribute?

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -68,31 +87,53 @@ async def get_graph_from_model(
         nodes.append(SimpleDataPointModel(**data_point_properties))
         added_nodes[str(data_point.id)] = True
 
-    for field_name in properties_to_visit:
+    for property_item in properties_to_visit:
```
borisarzentar: I don't think we need to change `properties_to_visit` to be a list and put more than a `field_name`. You can anyway get `edge_metadata` from the `field_value` if it is a tuple. We get `field_value` below -> `field_value = getattr(data_point, field_name)`.

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -68,31 +87,53 @@ async def get_graph_from_model(
...
+            edge_properties = {
+                "source_node_id": data_point.id,
+                "target_node_id": field_value.id,
+                "relationship_name": field_name,
+                "updated_at": datetime.now(timezone.utc).strftime("%Y-%m-%d %H:%M:%S"),
+            }
+
+            # Add edge metadata if present
+            if edge_metadata:
+                edge_properties.update(edge_metadata.to_dict())
```
borisarzentar: We can use `dict` from pydantic model directly here, no need to `to_dict()` function.

examples/python/weighted_edges_example.py
```diff
@@ -0,0 +1,132 @@
...
+            ),
+            [item1],
+        ),
+        has_objects=[],
+        knows=[],
+    )
+
+    # Create relationships between people with multiple weights
+    person1.knows = (
+        Edge(
```
borisarzentar, line 95: Do we want to validate those edge properties in order to enforce them? If you create one edge with one set of weights and another edge with some other set of weights, not sure how would search handle that difference.

cognee/infrastructure/engine/models/Edge.py
Vasilije1990: Yeah. Can use only one if it makes sense. Wdyt?

cognee/infrastructure/engine/models/DataPoint.py
Vasilije1990: Yup

examples/python/weighted_edges_example.py
Vasilije1990: Would not that be on the user? I mean, deduping and cleaning when we don't know the intention is best left to usecase when building the graph + retriever

cognee/modules/graph/utils/get_graph_from_model.py
Vasilije1990: Done, please check

cognee/modules/graph/utils/get_graph_from_model.py
Vasilije1990: Done, looks kinda clunky. Please check

cognee/modules/graph/utils/get_graph_from_model.py
Vasilije1990: Done

cognee/modules/graph/utils/get_graph_from_model.py
Vasilije1990: Removed in base model, used base Pydantic features

cognee/infrastructure/engine/models/Edge.py
Vasilije1990: Fixed

cognee/infrastructure/engine/models/Edge.py
borisarzentar: It's ok, just thinking about the flexibility for the future. For other use cases users can extend `Edge` and add more properties.

cognee/infrastructure/engine/models/DataPoint.py
borisarzentar: Ok. Other way would be then to have a model that defines both instead of the tuple.
Something like `Edge(weight=1, relationship_name="contains", data=DataPoint)`, but that is a bit uglier. Let's keep it for now.

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -21,25 +21,24 @@ async def get_graph_from_model(
         "type": type(data_point).__name__,
     }
     excluded_properties = set()
-    properties_to_visit = set()
+    properties_to_visit = []
```
borisarzentar: Please revert this one too.

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -21,25 +21,24 @@ async def get_graph_from_model(
...
+        # 1. Check if it's a single DataPoint
         if isinstance(field_value, DataPoint):
             excluded_properties.add(field_name)
 
             property_key = str(data_point.id) + field_name + str(field_value.id)
-
             if property_key in visited_properties:
                 continue
 
-            visited_properties[property_key] = True
```
borisarzentar, line 38: You need this line `visited_properties[property_key] = True`, in order to prevent circular models resolving indefinitely.

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -49,15 +48,50 @@ async def get_graph_from_model(
 
             for index, item in enumerate(field_value):
                 property_key = str(data_point.id) + field_name + str(item.id)
-
                 if property_key in visited_properties:
                     continue
 
-                visited_properties[property_key] = True
```
borisarzentar, line 56: You need this line `visited_properties[property_key] = True`, in order to prevent circular models resolving indefinitely.

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -49,15 +48,50 @@ async def get_graph_from_model(
...
+            and len(field_value) == 2
+            and isinstance(field_value[0], Edge)
+            and isinstance(field_value[1], DataPoint)
+        ):
+            edge_metadata, actual_field_value = field_value
+            excluded_properties.add(field_name)
+
+            property_key = str(data_point.id) + field_name + str(actual_field_value.id)
+            if property_key in visited_properties:
+                continue
```
borisarzentar: Add `visited_properties[property_key] = True`

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -49,15 +48,50 @@ async def get_graph_from_model(
...
+            and isinstance(field_value[1][0], DataPoint)
+        ):
+            edge_metadata, actual_field_value = field_value
+            excluded_properties.add(field_name)
+
+            for index, item in enumerate(actual_field_value):
+                property_key = str(data_point.id) + field_name + str(item.id)
+                if property_key in visited_properties:
+                    continue
+
```
borisarzentar: add `visited_properties[property_key] = True`

cognee/modules/graph/utils/get_graph_from_model.py
Vasilije1990: It didn't work with a set

PR #1069, opened 2025-07-09
fix: Limit major dep version change

Vasilije1990:
## DCO Affirmation


borisarzentar:
@Vasilije1990 Please generate `uv.lock` for `cognee` and `cognee-mcp`.

Vasilije1990:
> @Vasilije1990 Please generate `uv.lock` for `cognee` and `cognee-mcp`.

Done, will monitor how updates did.

pyproject.toml
```diff
@@ -19,76 +19,76 @@ classifiers = [
...
+    "python-dotenv>=1.0.1,<2.0.0",
+    "pydantic>=2.10.5,<3.0.0",
     "pydantic-settings>=2.2.1,<3",
-    "typing_extensions==4.12.2",
-    "nltk==3.9.1",
-    "numpy>=1.26.4, <=2.1",
-    "pandas>=2.2.2",
+    "typing_extensions>=4.12.2,<5.0.0",
+    "nltk>=3.9.1,<4.0.0",
+    "numpy>=1.26.4, <=2.3.1",
```
dexters1, line 28: Any reason for numpy to not be <3.0.0?

cognee/api/v1/config/config.py
```diff
@@ -23,7 +23,13 @@ def system_root_directory(system_root_directory: str):
 
         graph_config = get_graph_config()
         graph_file_name = graph_config.graph_filename
-        graph_config.graph_file_path = os.path.join(databases_directory_path, graph_file_name)
+        # For Kuzu v0.11.0+, use single-file database with .kuzu extension
+        if graph_config.graph_database_provider.lower() == "kuzu":
+            graph_config.graph_file_path = os.path.join(
+                databases_directory_path, f"{graph_file_name}.kuzu"
```
dexters1, line 29: It has to have the .kuzu extension to work?

pyproject.toml
```diff
@@ -19,76 +19,76 @@ classifiers = [
...
     "sentry-sdk[fastapi]>=2.9.0,<3",
     "structlog>=25.2.0,<26",
-    "pympler>=1.1",
-    "onnxruntime<=1.21.1",
-    "pylance==0.22.0",
-    "kuzu==0.9.0"
+    "pympler>=1.1,<2.0.0",
+    "onnxruntime>=1.0.0,<2.0.0",
+    "pylance>=0.22.0,<1.0.0",
+    "kuzu (==0.11.0)"
```
dexters1, line 62: Do we want to support previous Kuzu versions?

cognee/api/v1/config/config.py
dexters1: It seems to only be a recommendation based on their blog:

 We recommend using the .kuzu or .kz extension to make these files easily identifiable, but you can use any file extension of your choice (or no extension at all).

dexters1: I suggest we remove adding of the .kuzu extension from our side it will make the database much easier to manage and a lot less if/esle statements

pyproject.toml
dexters1: Or do we instead maybe want to do a migration of the previous Kuzu databases from folder to file type if they exist?

dexters1: Or at least point our users on how this migration can be done on their blog

PR #1075, opened 2025-07-10
Incremental loading

dexters1:
## DCO Affirmation

### DIFF
--- a/cognee/api/v1/cognify/cognify.py
+++ b/cognee/api/v1/cognify/cognify.py
@@ -39,6 +39,7 @@ async def cognify(
     vector_db_config: dict = None,
     graph_db_config: dict = None,
     run_in_background: bool = False,
+    incremental_loading: bool = True,
 ):
     """
     Transform ingested data into a structured knowledge graph.
@@ -194,6 +195,7 @@ async def cognify(
             datasets=datasets,
             vector_db_config=vector_db_config,
             graph_db_config=graph_db_config,
+            incremental_loading=incremental_loading,
         )
     else:
         return await run_cognify_blocking(
@@ -202,6 +204,7 @@ async def cognify(
             datasets=datasets,
             vector_db_config=vector_db_config,
             graph_db_config=graph_db_config,
+            incremental_loading=incremental_loading,
         )
 
 
@@ -211,6 +214,7 @@ async def run_cognify_blocking(
     datasets,
     graph_db_config: dict = None,
     vector_db_config: dict = False,
+    incremental_loading: bool = True,
 ):
     total_run_info = {}
 
@@ -221,6 +225,7 @@ async def run_cognify_blocking(
         pipeline_name="cognify_pipeline",
         graph_db_config=graph_db_config,
         vector_db_config=vector_db_config,
+        incremental_loading=incremental_loading,
     ):
         if run_info.dataset_id:
             total_run_info[run_info.dataset_id] = run_info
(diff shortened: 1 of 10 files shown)


cognee/modules/ingestion/methods/open_data_file.py
```diff
@@ -0,0 +1,6 @@
+def open_data_file(file_path: str, s3fs):
+    if file_path.startswith("s3://"):
+        return s3fs.open(file_path, mode="rb")
+    else:
+        local_path = file_path.replace("file://", "")
+        return open(local_path, mode="rb")
```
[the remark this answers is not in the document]
dexters1: This issue is resolved by limiting access to local file locations in another PR

cognee/api/v1/cognify/code_graph_pipeline.py
```diff
@@ -79,7 +79,9 @@ async def run_code_graph_pipeline(repo_path, include_docs=False):
         async for run_status in non_code_pipeline_run:
             yield run_status
 
-    async for run_status in run_tasks(tasks, dataset.id, repo_path, user, "cognify_code_pipeline"):
+    async for run_status in run_tasks(
+        tasks, dataset.id, repo_path, user, "cognify_code_pipeline", incremental_loading=False
```
dexters1, line 83: Code Graph Pipeline doesn't work with Data items so incremental_loading can't be used for it.

If repo_path, info and content hash would be stored as Data we could enable it there as well

cognee/modules/pipelines/operations/run_tasks.py
```diff
@@ -79,20 +86,95 @@ async def run_tasks(
...
-                payload=result,
-            )
+        if not isinstance(data, list):
+            data = [data]
+
+        if incremental_loading:
+            data = await resolve_data_directories(data)
+
+        # TODO: Convert to async gather task instead of for loop (just make sure it can work there were some issues when async gathering datasets)
+        for data_item in data:
```
borisarzentar, line 100: Why we don't do this in `extract_chunks_from_documents` task? We update the token count of the Data there as well. We know when the document started being processed and when it finished. We can also wrap yield in a try/except block to know if the document processing errored and log that for the specific file or even chunk. That way we also won't mix logic from add into the `run_tasks` and make it more complex.

cognee/api/v1/cognify/cognify.py
```diff
@@ -194,6 +195,7 @@ class ScientificPaper(DataPoint):
             datasets=datasets,
             vector_db_config=vector_db_config,
             graph_db_config=graph_db_config,
+            incremental_loading=incremental_loading,
```
borisarzentar, line 198: Can we do this through environment variable? This feels like too much drilling to pass the flag.

cognee/modules/pipelines/operations/run_tasks.py
dexters1: Because there are task afterwards that might fail like Embedding, Summaries etc. it can't be done in the extract chunks part

cognee/api/v1/cognify/cognify.py
dexters1: Not really, we can also have it as an environment variable for pipelines that can support it, but the problem is code_graph_pipeline can't work at all with incremental_loading enabled to it must be explicitly disabled when used and this can't be done through env variables

PR #1081, opened 2025-07-13
fix: Add docstrings to routers

Vasilije1990:
## DCO Affirmation

### DIFF
--- a/cognee/api/v1/add/routers/get_add_router.py
+++ b/cognee/api/v1/add/routers/get_add_router.py
@@ -54,6 +54,8 @@ def get_add_router() -> APIRouter:
             PermissionDeniedError: If the user doesn't have permission to add to the dataset
 
         Note:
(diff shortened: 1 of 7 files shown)


cognee/api/v1/add/routers/get_add_router.py
```diff
@@ -25,7 +25,40 @@ async def add(
...
+                - Details about the processed data
+                - Any relevant metadata from the ingestion process
+
+        Raises:
+            ValueError: If neither datasetId nor datasetName is provided
+            HTTPException: If there's an error during the add operation
+            PermissionDeniedError: If the user doesn't have permission to add to the dataset
+
+        Note:
+            - GitHub repositories are cloned and all files are processed
```
dexters1, line 59: I will add a Note to let users know for add and cognify that to access datasets not owned by them the dataset_id must be used

dexters1 (approved):

PR #1082, opened 2025-07-13
feat: Loader separation

Vasilije1990:
## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/loaders/LoaderInterface.py
+++ b/cognee/infrastructure/loaders/LoaderInterface.py
@@ -1,0 +1,101 @@
+from abc import ABC, abstractmethod
+from typing import List
+from .models.LoaderResult import LoaderResult
+
+
+class LoaderInterface(ABC):
+    """
+    Base interface for all file loaders in cognee.
+
+    This interface follows cognee's established pattern for database adapters,
+    ensuring consistent behavior across all loader implementations.
+    """
+
+    @property
+    @abstractmethod
+    def supported_extensions(self) -> List[str]:
+        """
+        List of file extensions this loader supports.
+
+        Returns:
+            List of extensions including the dot (e.g., ['.txt', '.md'])
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def supported_mime_types(self) -> List[str]:
+        """
+        List of MIME types this loader supports.
+
+        Returns:
+            List of MIME type strings (e.g., ['text/plain', 'application/pdf'])
+        """
+        pass
+
+    @property
+    @abstractmethod
+    def loader_name(self) -> str:
+        """
+        Unique name identifier for this loader.
+
+        Returns:
+            String identifier used for registration and configuration
+        """
+        pass
+
+    @abstractmethod
+    def can_handle(self, file_path: str, mime_type: str = None) -> bool:
+        """
+        Check if this loader can handle the given file.
+
+        Args:
+            file_path: Path to the file to be processed
+            mime_type: Optional MIME type of the file
+
+        Returns:
+            True if this loader can process the file, False otherwise
+        """
+        pass
+
+    @abstractmethod
+    async def load(self, file_path: str, **kwargs) -> LoaderResult:
+        """
+        Load and process the file, returning standardized result.
+
+        Args:
+            file_path: Path to the file to be processed
+            **kwargs: Additional loader-specific configuration
+
+        Returns:
+            LoaderResult containing processed content and metadata
+
+        Raises:
+            Exception: If file cannot be processed
+        """
+        pass
+
+    def get_dependencies(self) -> List[str]:
+        """
+        Optional: Return list of required dependencies for this loader.
+
+        Returns:
+            List of package names with optional version specifications
+        """
+        return []
+
+    def validate_dependencies(self) -> bool:
+        """
+        Check if all required dependencies are available.
+
+        Returns:
+            True if all dependencies are installed, False otherwise
+        """
+        for dep in self.get_dependencies():
+            # Extract package name from version specification
+            package_name = dep.split(">=")[0].split("==")[0].split("<")[0]
+            try:
+                __import__(package_name)
+            except ImportError:
+                return False
+        return True
--- a/cognee/infrastructure/loaders/core/text_loader.py
+++ b/cognee/infrastructure/loaders/core/text_loader.py
@@ -1,0 +1,128 @@
+import os
+from typing import List
+from cognee.infrastructure.loaders.LoaderInterface import LoaderInterface
+from cognee.infrastructure.loaders.models.LoaderResult import LoaderResult, ContentType
+
+
+class TextLoader(LoaderInterface):
+    """
+    Core text file loader that handles basic text file formats.
+
+    This loader is always available and serves as the fallback for
+    text-based files when no specialized loader is available.
+    """
+
+    @property
+    def supported_extensions(self) -> List[str]:
+        """Supported text file extensions."""
+        return [".txt", ".md", ".csv", ".json", ".xml", ".yaml", ".yml", ".log"]
+
+    @property
+    def supported_mime_types(self) -> List[str]:
+        """Supported MIME types for text content."""
+        return [
+            "text/plain",
+            "text/markdown",
+            "text/csv",
+            "application/json",
+            "text/xml",
+            "application/xml",
+            "text/yaml",
+            "application/yaml",
+        ]
+
+    @property
+    def loader_name(self) -> str:
+        """Unique identifier for this loader."""
+        return "text_loader"
+
+    def can_handle(self, file_path: str, mime_type: str = None) -> bool:
+        """
+        Check if this loader can handle the given file.
+
+        Args:
+            file_path: Path to the file
+            mime_type: Optional MIME type
+
+        Returns:
+            True if file can be handled, False otherwise
+        """
+        # Check by extension
+        ext = os.path.splitext(file_path)[1].lower()
+        if ext in self.supported_extensions:
+            return True
+
+        # Check by MIME type
+        if mime_type and mime_type in self.supported_mime_types:
+            return True
+
+        # As fallback loader, can attempt to handle any text-like file
+        # This is useful when other loaders fail
+        try:
+            # Quick check if file appears to be text
+            with open(file_path, "rb") as f:
+                sample = f.read(512)
+                # Simple heuristic: if most bytes are printable, consider it text
+                if sample:
+                    try:
+                        sample.decode("utf-8")
+                        return True
+                    except UnicodeDecodeError:
+                        try:
+                            sample.decode("latin-1")
+                            return True
+                        except UnicodeDecodeError:
+                            pass
+        except (OSError, IOError):
+            pass
+
+        return False
+
+    async def load(self, file_path: str, encoding: str = "utf-8", **kwargs) -> LoaderResult:
+        """
+        Load and process the text file.
+
+        Args:
+            file_path: Path to the file to load
+            encoding: Text encoding to use (default: utf-8)
+            **kwargs: Additional configuration (unused)
+
+        Returns:
+            LoaderResult containing the file content and metadata
+
+        Raises:
+            FileNotFoundError: If file doesn't exist
+            UnicodeDecodeError: If file cannot be decoded with specified encoding
+            OSError: If file cannot be read
+        """
+        if not os.path.exists(file_path):
+            raise FileNotFoundError(f"File not found: {file_path}")
+
+        try:
+            with open(file_path, "r", encoding=encoding) as f:
+                content = f.read()
+        except UnicodeDecodeError:
+            # Try with fallback encoding
+            if encoding == "utf-8":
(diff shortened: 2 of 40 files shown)


LordCasser (changes requested):
It is recommended to use the `Path` type as an input parameter to allow users to explicitly understand whether the input represents a file path requiring processing or a plain string. Additionally, imposing reasonable constraints actually simplifies things for SDKs. For RESTful APIs or MCP implementations, the conversion and identification of file paths in string format should reside in the API interface implementation rather than within the SDK.

Vasilije1990:
> It is recommended to use the `Path` type as an input parameter to allow users to explicitly understand whether the input represents a file path requiring processing or a plain string. Additionally, imposing reasonable constraints actually simplifies things for SDKs. For RESTful APIs or MCP implementations, the conversion and identification of file paths in string format should reside in the API interface implementation rather than within the SDK.

Resolved

cognee/infrastructure/loaders/core/text_loader.py
```diff
@@ -0,0 +1,128 @@
...
+    Core text file loader that handles basic text file formats.
+
+    This loader is always available and serves as the fallback for
+    text-based files when no specialized loader is available.
+    """
+
+    @property
+    def supported_extensions(self) -> List[str]:
+        """Supported text file extensions."""
+        return [".txt", ".md", ".csv", ".json", ".xml", ".yaml", ".yml", ".log"]
```
dexters1, line 18: .csv files don't seem to be working well when just read and ingested directly into Cognee, maybe a custom loader/pipeline for the .csv type is worth making in the future. That can take into account each row and column information and create chunks with this information

cognee/tasks/ingestion/plugin_ingest_data.py
```diff
@@ -0,0 +1,324 @@
...
+        loader_config: Configuration for specific loaders
+
+    Returns:
+        List of Data objects that were ingested
+    """
+    if not user:
+        user = await get_default_user()
+
+    # Ensure NLTK data is downloaded (preserves automatic download behavior)
+    def ensure_nltk_data():
```
dexters1, line 58: Can we just handle nltk downloads in one place in the cognee_pipeline?

cognee/infrastructure/loaders/core/text_loader.py
LordCasser: I agree with that.

cognee/infrastructure/loaders/LoaderInterface.py
```diff
@@ -0,0 +1,101 @@
...
+
+        Returns:
+            LoaderResult containing processed content and metadata
+
+        Raises:
+            Exception: If file cannot be processed
+        """
+        pass
+
+    def get_dependencies(self) -> List[str]:
```
dexters1, line 78: Do we need the get_dependencies and validate_dependencies methods?

infrastructure/loaders/create_loader_engine.py
```diff
@@ -0,0 +1,49 @@
...
+    enable_dependency_validation: bool = True,
+) -> LoaderEngine:
+    """
+    Create loader engine with given configuration.
+
+    Follows cognee's pattern for engine creation functions used
+    in database adapters.
+
+    Args:
+        loader_directories: Directories to search for loader implementations
```
dexters1, line 20: How does this loader_directories mechanism work? Do we need it?

infrastructure/loaders/LoaderEngine.py
```diff
@@ -0,0 +1,237 @@
...
+                                attr = getattr(module, attr_name)
+                                if (
+                                    isinstance(attr, type)
+                                    and issubclass(attr, LoaderInterface)
+                                    and attr != LoaderInterface
+                                ):
+                                    # Instantiate and register the loader
+                                    try:
+                                        loader_instance = attr()
+                                        self.register_loader(loader_instance)
```
dexters1, line 196: Do we want to have automatic code import like this? Seems like a security issue

cognee/infrastructure/loaders/external/dlt_loader.py
```diff
@@ -0,0 +1,203 @@
...
+            import dlt
+        except ImportError as e:
+            raise ImportError(
+                "dlt is required for data loading. Install with: pip install dlt"
+            ) from e
+
+        try:
+            self.logger.info(f"Loading data with DLT: {file_path}")
+
+            file_ext = os.path.splitext(file_path)[1].lower()
```
dexters1, line 89: File extension doesn't have to be in file_path so this won't work, it can also allow users to upload .exe files by naming them .txt

cognee/infrastructure/loaders/external/dlt_loader.py
```diff
@@ -0,0 +1,203 @@
...
+            file_name = os.path.basename(file_path)
+            file_size = os.path.getsize(file_path)
+
+            # Determine source type if auto
+            if source_type == "auto":
```
dexters1, line 94: We don't need this, we can use file_ext directly if "auto"

cognee/infrastructure/loaders/external/dlt_loader.py
dexters1: There are ways to determine file type based on binary content (we do this already in places)

cognee/infrastructure/loaders/external/dlt_loader.py
```diff
@@ -0,0 +1,203 @@
...
+            # Load data based on source type
+            if source_type == "json":
+                content = self._load_json(file_path)
+            elif source_type == "jsonl":
+                content = self._load_jsonl(file_path)
+            elif source_type == "csv":
+                content = self._load_csv(file_path)
+            elif source_type == "parquet":
+                content = self._load_parquet(file_path)
+            elif source_type == "yaml":
```
dexters1, line 117: if we remove auto to use file_ext source types should be changed to check for file_ext instead

cognee/infrastructure/loaders/external/dlt_loader.py
```diff
@@ -0,0 +1,203 @@
...
+        except Exception as e:
+            self.logger.error(f"Failed to load data with DLT from {file_path}: {e}")
+            raise Exception(f"DLT data loading failed: {e}") from e
+
+    def _load_json(self, file_path: str) -> Dict[str, Any]:
+        """Load JSON file."""
+        import json
+
+        with open(file_path, "r", encoding="utf-8") as f:
+            return json.load(f)
```
dexters1, line 169: How is dlt used to load these json, jsonl, parquet and csv files? Do we need this dlt_loader?

cognee/infrastructure/loaders/external/unstructured_loader.py
```diff
@@ -0,0 +1,168 @@
...
+                            if len(element_text) > 100
+                            else element_text,
+                        }
+                    )
+
+            # Combine all text content
+            full_content = "\n\n".join(text_parts)
+
+            # Determine content type based on structure
+            content_type = ContentType.STRUCTURED if len(element_info) > 1 else ContentType.TEXT
```
dexters1, line 139: What does STRUCTURED mean in this case? 

In this case we should do what we did before with unstructured and return full_content as TEXT as there is no structure on how these elements were split into text_parts

cognee/modules/ingestion/data_types/BinaryData.py
```diff
@@ -30,9 +30,38 @@ def get_metadata(self):
 
     async def ensure_metadata(self):
         if self.metadata is None:
-            self.metadata = await get_file_metadata(self.data)
+            # Handle case where file might be closed
```
dexters1, line 33: Do we want to handle this case?

If File is not found exception should be raised

cognee/modules/ingestion/data_types/TextData.py
```diff
@@ -16,9 +16,12 @@ def __init__(self, data: BinaryIO):
...
 
     def get_identifier(self):
-        keywords = extract_keywords(self.data)
+        import hashlib
 
-        return "text/plain" + "_" + "|".join(keywords)
+        content_bytes = self.data.encode("utf-8")
+        content_hash = hashlib.md5(content_bytes).hexdigest()
+
+        return "text/plain" + "_" + content_hash
```
dexters1, line 24: Return value of get_identifier should just be content_hash, previously this function was never used as TextData became BinaryData, if it will be used now return value should just be the content_hash

cognee/modules/pipelines/operations/run_tasks.py
```diff
@@ -98,10 +98,9 @@ async def _run_tasks_data_item_incremental(
                 await session.execute(select(Data).filter(Data.id == data_id))
             ).scalar_one_or_none()
             if data_point:
-                if (
-                    data_point.pipeline_status.get(pipeline_name, {}).get(str(dataset.id))
-                    == DataItemStatus.DATA_ITEM_PROCESSING_COMPLETED
-                ):
+                if (data_point.pipeline_status or {}).get(pipeline_name, {}).get(
```
dexters1, line 101: Any reason for this change?

cognee/modules/pipelines/operations/run_tasks.py
```diff
@@ -133,11 +132,20 @@ async def _run_tasks_data_item_incremental(
                 data_point = (
                     await session.execute(select(Data).filter(Data.id == data_id))
                 ).scalar_one_or_none()
-                data_point.pipeline_status[pipeline_name] = {
-                    str(dataset.id): DataItemStatus.DATA_ITEM_PROCESSING_COMPLETED
-                }
-                await session.merge(data_point)
-                await session.commit()
+                if data_point is not None:
```
dexters1, line 135: Any reason for this change?

cognee/tasks/ingestion/adapters/loader_to_ingestion_adapter.py
```diff
@@ -0,0 +1,241 @@
...
+        """
+        # Always generate hash from content for consistency
+        import hashlib
+
+        content_bytes = self.loader_result.content.encode("utf-8")
+        content_hash = hashlib.md5(content_bytes).hexdigest()
+
+        # Add content type prefix for better identification
+        content_type = self.loader_result.content_type.value
+        return f"{content_type}_{content_hash}"
```
dexters1, line 40: Do not add content_type prefix, just return content_hash otherwise data id will not be mapped properly

cognee/tasks/ingestion/adapters/loader_to_ingestion_adapter.py
```diff
@@ -0,0 +1,241 @@
...
+        if "name" not in metadata:
+            if self.original_file_path:
+                metadata["name"] = os.path.basename(self.original_file_path)
+            else:
+                # Generate name from content hash
+                content_hash = self.get_identifier().split("_")[-1][:8]
+                ext = metadata.get("extension", ".txt")
+                metadata["name"] = f"content_{content_hash}{ext}"
+
+        if "content_hash" not in metadata:
```
dexters1, line 64: content_hash has to be consistent between all data types, loaders and etc. and should be present

cognee/tasks/ingestion/plugin_ingest_data.py
```diff
@@ -0,0 +1,324 @@
...
+            # Ensure metadata has all required fields with fallbacks
+            def get_metadata_field(metadata, field_name, default_value=""):
+                """Get metadata field with fallback handling."""
+                if field_name in metadata and metadata[field_name] is not None:
+                    return metadata[field_name]
+
+                logger.warning(f"Missing metadata field '{field_name}', using fallback")
+
+                # Provide fallbacks based on available information
+                if field_name == "name":
```
dexters1, line 186: There should be no fallback handling, if a fallback is necessary something was not done right in the loader

cognee/tasks/ingestion/resolve_data_directories.py
```diff
@@ -33,7 +35,26 @@ async def resolve_data_directories(
         )
 
     for item in data:
-        if isinstance(item, str):  # Check if the item is a path
+        if isinstance(item, Path):  # Path objects explicitly indicate file paths
```
dexters1, line 38: What if it's a string? We still accept string for file path

infrastructure/loaders/LoaderEngine.py
```diff
@@ -0,0 +1,237 @@
+import os
+import importlib.util
+from typing import Dict, List, Optional
```
dexters1, line 3: This infrastructure folder seems to be a duplicate of the infrastructure folder inside the cognee folder

cognee/tasks/ingestion/plugin_ingest_data.py
```diff
@@ -0,0 +1,324 @@
...
+        # Get existing dataset data for deduplication (preserve existing logic)
+        dataset_data: list[Data] = await get_dataset_data(dataset.id)
+        dataset_data_map = {str(data.id): True for data in dataset_data}
+
+        for data_item in data:
+            file_path = await save_data_item_to_storage(data_item)
+
+            # NEW: Use loader system or existing classification based on data type
+            try:
+                if loader_adapter.is_text_content(data_item):
```
dexters1, line 153: Will never be text content as text was saved to file in previous step

PR #1093, opened 2025-07-16
fix: add numpy higher version

Vasilije1990:
## DCO Affirmation


dexters1:
We should rebuild the lock files for this change, I'll do it now

PR #1095, opened 2025-07-16
fix: s3 file storage

borisarzentar:
## DCO Affirmation

### DIFF
--- a/cognee/tests/test_s3_file_storage.py
+++ b/cognee/tests/test_s3_file_storage.py
@@ -14,9 +14,9 @@ logger = get_logger()
 
 async def main():
     test_run_id = uuid4()
-    data_directory_path = f"s3://github-runner-cognee-tests/{test_run_id}/data"
+    data_directory_path = f"s3://cognee-storage-dev/{test_run_id}/data"
     cognee.config.data_root_directory(data_directory_path)
-    cognee_directory_path = f"s3://github-runner-cognee-tests/{test_run_id}/system"
+    cognee_directory_path = f"s3://cognee-storage-dev/{test_run_id}/system"
     cognee.config.system_root_directory(cognee_directory_path)
 
     await cognee.prune.prune_data()


cognee/api/v1/config/config.py
```diff
@@ -24,13 +24,7 @@ def system_root_directory(system_root_directory: str):
 
         graph_config = get_graph_config()
         graph_file_name = graph_config.graph_filename
-        # For Kuzu v0.11.0+, use single-file database with .kuzu extension
```
dexters1, line 27: Why is this removed?

cognee/infrastructure/databases/graph/config.py
```diff
@@ -50,19 +50,26 @@ class GraphConfig(BaseSettings):
     @pydantic.model_validator(mode="after")
     def fill_derived(cls, values):
         provider = values.graph_database_provider.lower()
-        # Set filename based on graph database provider if no filename is provided
+
+        # Set default filename if no filename is provided
         if not values.graph_filename:
-            values.graph_filename = f"cognee_graph_{provider}"
```
dexters1, line 55: Why was this changed?

borisarzentar: Because we anyway add ‘.kuzu’ below. If we have it here networkx will have _kuzu as well I think.

cognee/infrastructure/databases/graph/config.py
```diff
@@ -50,19 +50,26 @@ class GraphConfig(BaseSettings):
...
         # Set file path based on graph database provider if no file path is provided
         if not values.graph_file_path:
             base_config = get_base_config()
 
-            base = os.path.join(base_config.system_root_directory, "databases")
+            databases_directory_path = os.path.join(base_config.system_root_directory, "databases")
             # For Kuzu v0.11.0+, use single-file database with .kuzu extension
             if provider == "kuzu":
-                values.graph_file_path = os.path.join(base, f"{values.graph_filename}.kuzu")
+                values.graph_file_path = os.path.join(
```
dexters1: This should be removed

cognee/infrastructure/databases/graph/config.py
dexters1: It won't, I made it and tested it so it will be based on provider name

cognee/infrastructure/databases/utils/get_or_create_dataset_database.py
```diff
@@ -33,7 +33,7 @@ async def get_or_create_dataset_database(
     dataset_id = await get_unique_dataset_id(dataset, user)
 
     vector_db_name = f"{dataset_id}.lance.db"
-    graph_db_name = f"{dataset_id}.pkl"
+    graph_db_name = f"{dataset_id}.kuzu"
```
dexters1: This change will also affect anyone who is using our permissions system already, not sure what to do here

borisarzentar: Reverted it

PR #1096, opened 2025-07-16
fix: Remove milvus from core

Vasilije1990:
## DCO Affirmation


hao-phan-vns:
Hello, I noticed that Milvus was removed from the core in this recent update. I currently use Milvus in my workflow and wanted to understand the reasoning behind this decision. Could you please share more details about why it was removed and if there are any recommended alternatives or migration paths?

PR #1097, opened 2025-07-16
fix: Cog 1484 add logical test for entity extraction

Vasilije1990:
## DCO Affirmation


cognee/tests/test_knowledge_graph_quality.py
```diff
@@ -0,0 +1,300 @@
...
+            all_results_text += f" {' '.join(str(r) for r in result_list)}"
+
+    all_results_text = all_results_text.lower()
+
+    print("ENTITY ANALYSIS:")
+    print("-" * 40)
+    found_entities = []
+    missing_entities = []
+
+    for entity in expected_entities:
```
dexters1, line 172: We need to change how we check if entities are found. We should only use graph nodes and edges to determine if all expected entities are found. (Similar how it has been done for the s3 test)

All chunk text must be excluded from checking existing entities.

PR #1101, opened 2025-07-17
enhancement: Optimizing embedding calls in brute_force_search

thompsonp17:
@Vasilije1990

- Use query_vector instead of query_text in brute_force_search

## Description
[Here](https://github.com/topoteretes/cognee/blob/ef1aecd835b1a2044eb724197bbef77f6dee5d3c/cognee/modules/retrieval/utils/brute_force_triplet_search.py#L163) brute_force_search uses the vector engine to perform the same search — with the same query text — across multiple collections, making the embedding calls unnecessarily proportional to the number of collections being searched.

Since the [search](https://github.com/topoteretes/cognee/blob/ef1aecd835b1a2044eb724197bbef77f6dee5d3c/cognee/infrastructure/databases/vector/vector_db_interface.py#L85) interface is already designed to accept precomputed query vectors, I’m submitting an optimization to brute_force_search to take advantage of this.

If this is considered good practice, it might be worth implementing a direct query_vector argument in [map_vector_distances_to_graph_edges](https://github.com/topoteretes/cognee/blob/ef1aecd835b1a2044eb724197bbef77f6dee5d3c/cognee/modules/graph/cognee_graph/CogneeGraph.py#L135) , and using it both [here](https://github.com/topoteretes/cognee/blob/ef1aecd835b1a2044eb724197bbef77f6dee5d3c/cognee/modules/retrieval/utils/brute_force_triplet_search.py#L179) and in any future uses of map_vector_distances_to_graph_edges.

## DCO Affirmation


hajdul88 (approved):
Hi @thompsonp17!

Yes it is completely valid and valuable change. I will also add the suggested vector_embedding to the map_vector_distances_to_graph_edges although its a smaller issue, but still saves one embedding call.

Thanks for the valuable contribution, I ran CI/CD tests passed so I merge the PR!

thompsonp17:
I'm happy to help as much as I can. Thanks for the review, @hajdul88. 🫡

Issue #1084, opened 2025-07-14
Frontend hardcodes API base URL


oojacoboo:
Currently I'm running into an issue where the MCP server and the Cognee server (separate docker container needed here I guess) are both trying to bind to port `8000`.  This causes the MCP server to fail and exit.  The fix is to bind the main Cognee API server to another port.  I can do that.  But then the `cognee-frontend` has the API urls hardcoded... 

`http://localhost:8000/api/v1/datasets/`

The entire base URL should be an environment variable.

Vasilije1990:
@borisarzentar will look into this

Raj2604:
Hey I can fix that Please assign me this issue

Vasilije1990:
Assigned

Vasilije1990:
Merged to dev, will be a part of next release

PR #1104, opened 2025-07-18
Frontend hardcodes API base URL #1084

Raj2604:
## Description
This pull request refactors the frontend application to remove hardcoded backend API and WebSocket URLs. The Environment variables are stored at .env.template file which is created in PR. By this you can temporary solve the problem of binding to the same port 8000. You have to bind Cognee Api Server to different Port and then instead of changing the hardcoded urls you only need to change the Port Number in base url which is present in .env . Please let me know if you want to change Port Number of Cognee Api Server from 8000 to different Port Number which will resolve this issue permanently.

## DCO Affirmation

### DIFF
--- a/cognee-frontend/.env.template
+++ b/cognee-frontend/.env.template
@@ -1,0 +1,1 @@
+NEXT_PUBLIC_BACKEND_API_URL=http://localhost:8000/api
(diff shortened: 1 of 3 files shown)


Raj2604:
Fixes #1084

Vasilije1990 (approved):

PR #1110, opened 2025-07-19
changing telemetry to be standlone outside run_tasks

pc099:
## Description
- Decoupled telemetry logic from pipeline task execution functions.
- Telemetry handler is now injected and propagated, enabling easier testing and separation of concerns.
- Removed dependency on default user from database for telemetry.
- Updated run_tasks, run_tasks_with_telemetry, and run_tasks_base to support telemetry injection.

## DCO Affirmation

### DIFF
--- a/cognee/modules/pipelines/operations/run_tasks.py
+++ b/cognee/modules/pipelines/operations/run_tasks.py
@@ -56,6 +56,7 @@ async def run_tasks(
     user: User = None,
     pipeline_name: str = "unknown_pipeline",
     context: dict = None,
+    telemetry_handler=None,  # Add telemetry handler argument
 ):
     if not user:
         user = get_default_user()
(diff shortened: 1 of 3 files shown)


cognee/modules/pipelines/operations/run_tasks.py
```diff
@@ -56,6 +56,7 @@ async def run_tasks(
     user: User = None,
     pipeline_name: str = "unknown_pipeline",
     context: dict = None,
+    telemetry_handler=None,  # Add telemetry handler argument
```
borisarzentar, line 59: @pc099 What was the reason for this change?

pc099: @borisarzentar, I added this change to make telemetry_handler optional so that even if no value is passed it take none

dexters1:
@pc099 no need to decouple the telemetry, it can be disabled already through env variables by setting TELEMETRY_DISABLED=True

PR #1114, opened 2025-07-19
fix:Mcp improvements

Vasilije1990:
## DCO Affirmation


cognee-mcp/entrypoint.sh
```diff
@@ -43,19 +43,25 @@ if [ "$ENVIRONMENT" = "dev" ] || [ "$ENVIRONMENT" = "local" ]; then
         echo "Waiting for the debugger to attach..."
         if [ "$TRANSPORT_MODE" = "sse" ]; then
             exec python -m debugpy --wait-for-client --listen 0.0.0.0:5678 -m cognee --transport sse
+        elif [ "$TRANSPORT_MODE" = "http" ]; then
+            exec python -m debugpy --wait-for-client --listen 0.0.0.0:5678 -m cognee --transport http --host 0.0.0.0 --port 8000
```
oojacoboo: Why are ports being hardcoded?  They should be user-configurable as an ENV variable.

Vasilije1990: Good point, handled now

docker-compose.yml
```diff
@@ -1,3 +1,40 @@
+# Cognee Docker Compose Configuration
```
dexters1, line 1: Maybe remove these docs

PR #1115, opened 2025-07-19
chore: Cog 2354 add logging

Vasilije1990:
## DCO Affirmation


cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
```diff
@@ -870,34 +870,106 @@ async def get_graph_data(self):
...
+        import time
 
-        result = await self.query(query)
+        start_time = time.time()
 
-        nodes = [
-            (
-                record["properties"]["id"],
-                record["properties"],
+        try:
```
hajdul88, line 877: I dont wouldn't do the logging of this. This as I understand is doing some kind of a structural check on the retrieved nodes and edges and doesn' log anything. We cannot have missing properties in the graph, or at least it shouldn't be tested here at every retrieval. In this part (get_graph_data) I would measure the time it takes to retreive these, that part is useful.

cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
```diff
@@ -870,34 +870,106 @@ async def get_graph_data(self):
...
+            """
+            result = await self.query(query)
+
+            logger.debug(f"Raw edges query returned {len(result)} records")
+
+            edges = []
+            processed_edges = 0
+            failed_edges = 0
+
+            for record in result:
```
hajdul88, line 899: Same here, we theoretically cannot have edges where we dont have these. The time measurement is good tho.

cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
```diff
@@ -870,34 +870,106 @@ async def get_graph_data(self):
...
-        ]
 
-        return (nodes, edges)
+            retrieval_time = time.time() - start_time
+            logger.info(f"Full graph data retrieval completed in {retrieval_time:.2f} seconds")
+            logger.info(f"Retrieved {len(nodes)} nodes and {len(edges)} edges")
+
+            if len(nodes) == 0:
+                logger.warning("No nodes found in the graph database")
+            if len(edges) == 0 and len(nodes) > 0:
```
hajdul88: This is one way we can have isolated nodes, I would say here, no edges retrieved or smthg like that. We can have isolated nodes even if  we have edges retrieved.

cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
```diff
@@ -918,50 +990,117 @@ async def get_nodeset_subgraph(
...
+
+            # Process edges
+            edges = []
+            processed_edges = 0
+            failed_edges = 0
+
+            for r in raw_rels:
+                try:
+                    if (
+                        "properties" in r
```
hajdul88: Same here as before, I wouldn't log this, or at least I wouldn't do it at retrieval time. Each time we search we count how many edges doesn't have the excepted properties? Can stay like this tho but this cannot happen.

cognee/modules/graph/cognee_graph/CogneeGraph.py
```diff
@@ -64,103 +70,234 @@ async def project_graph_from_db(
...
+
+            logger.info(
+                f"Edge processing completed: {processed_edges} processed, {failed_edges} failed, {missing_node_errors} missing node errors"
+            )
+
+            # Final statistics
+            projection_time = time.time() - start_time
+            logger.info(f"Graph projection completed successfully in {projection_time:.2f} seconds")
+            logger.info(f"Final graph stats: {len(self.nodes)} nodes, {len(self.edges)} edges")
+            logger.debug(
```
hajdul88: I dont think graph density is important here but can stay.

cognee/modules/graph/cognee_graph/CogneeGraph.py
```diff
@@ -64,103 +70,234 @@ async def project_graph_from_db(
...
             for scored_result in scored_results:
                 node_id = str(scored_result.id)
                 score = scored_result.score
                 node = self.get_node(node_id)
                 if node:
                     node.add_attribute("vector_distance", score)
+                    mapped_nodes += 1
+                    logger.debug(f"Mapped vector distance {score} to node {node_id}")
+                else:
+                    logger.warning(f"Node {node_id} not found in graph for vector distance mapping")
```
hajdul88: This is not true, scored_results contains scored edge collection as well so for the edge ids we are gonna print that Node (id (id of an edge)) not found in graph for vector distances mapping.

cognee/modules/graph/cognee_graph/CogneeGraph.py
```diff
@@ -64,103 +70,234 @@ async def project_graph_from_db(
...
     async def map_vector_distances_to_graph_edges(self, vector_engine, query) -> None:
+        logger.info(
+            f"Starting vector distance mapping to graph edges for query: '{query[:100]}{'...' if len(query) > 100 else ''}'"
+        )
+
         try:
             query_vector = await vector_engine.embed_data([query])
             query_vector = query_vector[0]
             if query_vector is None or len(query_vector) == 0:
+                logger.error("Failed to generate query embedding for edge distance mapping")
```
hajdul88: This is gonna be changed in the following days

cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
hajdul88: Also now it is implemented only for neo4j, so I would just skip this logging ont he adapter level. (Time can stay that is useful)

cognee/modules/graph/cognee_graph/CogneeGraph.py
```diff
@@ -64,103 +70,234 @@ async def project_graph_from_db(
         node_type: Optional[Type] = None,
         node_name: Optional[List[str]] = None,
     ) -> None:
+        logger.info("Starting graph projection from database")
+        logger.debug(
```
hajdul88: I am not sure that it is worth logging every single edge and node even if its debug level. Can also slow down the processing when the graph is big.

cognee/modules/graph/cognee_graph/CogneeGraph.py
```diff
@@ -64,103 +70,234 @@ async def project_graph_from_db(
...
+                )
+
             elif len(memory_fragment_filter) == 0:
+                logger.info("Using full graph projection strategy")
                 nodes_data, edges_data = await adapter.get_graph_data()
                 if not nodes_data or not edges_data:
+                    logger.warning("Full graph projection returned empty results")
                     raise EntityNotFoundError(message="Empty graph projected from the database.")
+                logger.info(
+                    f"Full graph projection retrieved {len(nodes_data)} nodes and {len(edges_data)} edges"
```
hajdul88: I would keep logs that are like this, aggregated. This is useful

cognee/modules/graph/cognee_graph/CogneeGraph.py
```diff
@@ -27,17 +27,23 @@ def __init__(self, directed: bool = True):
         self.nodes = {}
         self.edges = []
         self.directed = directed
+        logger.debug(f"Initialized CogneeGraph with directed={directed}")
 
     def add_node(self, node: Node) -> None:
         if node.id not in self.nodes:
             self.nodes[node.id] = node
+            logger.debug(f"Added node {node.id} to graph")
```
hajdul88: I think it is too granular even if its on debug level but can slow down the processing and doesn't have added value imho.

cognee/modules/graph/cognee_graph/CogneeGraph.py
```diff
@@ -27,17 +27,23 @@ def __init__(self, directed: bool = True):
...
         else:
+            logger.warning(f"Attempted to add duplicate node {node.id}")
             raise EntityAlreadyExistsError(message=f"Node with id {node.id} already exists.")
 
     def add_edge(self, edge: Edge) -> None:
         self.edges.append(edge)
         edge.node1.add_skeleton_edge(edge)
         edge.node2.add_skeleton_edge(edge)
+        logger.debug(
```
hajdul88: Same here

cognee/modules/graph/cognee_graph/CogneeGraph.py
```diff
@@ -64,103 +70,234 @@ async def project_graph_from_db(
         node_type: Optional[Type] = None,
         node_name: Optional[List[str]] = None,
     ) -> None:
+        logger.info("Starting graph projection from database")
+        logger.debug(
+            f"Projection parameters: directed={directed}, node_dimension={node_dimension}, edge_dimension={edge_dimension}"
+        )
+        logger.debug(f"Node properties to project: {node_properties_to_project}")
```
hajdul88: I dont think this I would log. Its better to keep aggragated info, time, and node or edge in the case something fails. Otherwise this will dump the whole graph together with properties and everything.

cognee/modules/graph/cognee_graph/CogneeGraph.py
```diff
@@ -64,103 +70,234 @@ async def project_graph_from_db(
...
 
-                else:
-                    raise EntityNotFoundError(
-                        message=f"Edge references nonexistent nodes: {source_id} -> {target_id}"
-                    )
+            for source_id, target_id, relationship_type, properties in edges_data:
+                try:
+                    source_node = self.get_node(str(source_id))
+                    target_node = self.get_node(str(target_id))
+                    if source_node and target_node:
```
hajdul88: This makes sense here, this is the same code that was implemented on neo4j adapter level. Here its okay, there I think its not necessary.

cognee/modules/graph/cognee_graph/CogneeGraph.py
```diff
@@ -64,103 +70,234 @@ async def project_graph_from_db(
...
+                f"Processing category '{category}' with {len(scored_results)} scored results"
+            )
             for scored_result in scored_results:
                 node_id = str(scored_result.id)
                 score = scored_result.score
                 node = self.get_node(node_id)
                 if node:
                     node.add_attribute("vector_distance", score)
+                    mapped_nodes += 1
+                    logger.debug(f"Mapped vector distance {score} to node {node_id}")
```
hajdul88: Maybe this is again too detailed.

cognee/modules/graph/cognee_graph/CogneeGraph.py
```diff
@@ -170,9 +307,16 @@ async def calculate_top_triplet_importances(self, k: int) -> List:
...
 
             total_distance = source_distance + target_distance + edge_distance
+            processed_triplets += 1
 
             heapq.heappush(min_heap, (-total_distance, i, edge))
             if len(min_heap) > k:
                 heapq.heappop(min_heap)
 
-        return [edge for _, _, edge in sorted(min_heap)]
+        result = [edge for _, _, edge in sorted(min_heap)]
```
hajdul88: I wouldn't do this here, its a list comprehension just for the debug logging. We can keep the info logging but the result and debug logging is not needed imho. Also len(result) equals to k

cognee/modules/graph/methods/get_formatted_graph_data.py
```diff
@@ -1,41 +1,127 @@
...
-                        and value is not None
-                    },
-                },
-                nodes,
-            )
-        ),
-        "edges": list(
-            map(
-                lambda edge: {
+    logger.info(f"Starting graph data formatting for dataset_id={dataset_id}, user_id={user_id}")
```
hajdul88: I think this is used by the UI not by retrievers. @borisarzentar am I right?

hajdul88: I wouldn't add logging to this function its not the part of retriever.

cognee/modules/retrieval/chunks_retriever.py
```diff
@@ -23,6 +26,7 @@ def __init__(
         top_k: Optional[int] = 5,
     ):
         self.top_k = top_k
+        logger.info(f"Initialized ChunksRetriever with top_k={self.top_k}")
```
hajdul88: This is gonna print out all the retrievers we have because we have them in enum instantiated. Maybe we can delete this to avoid confusion for the users

cognee/modules/retrieval/code_retriever.py
```diff
@@ -32,29 +35,47 @@ def __init__(self, top_k: int = 3):
             "ClassDefinition_source_code",
             "FunctionDefinition_source_code",
         ]
+        logger.info(f"Initialized CodeRetriever with top_k={self.top_k}")
```
hajdul88: Same here

cognee/modules/retrieval/completion_retriever.py
```diff
@@ -70,22 +88,35 @@ async def get_completion(self, query: str, context: Optional[Any] = None) -> Any
...
+
+        try:
+            completion = await generate_completion(
+                query, context, self.user_prompt_path, self.system_prompt_path
+            )
+            logger.info("Completion generation successful")
+            return completion
+        except Exception as e:
+            logger.error(f"Error during completion generation: {str(e)}")
+            raise
```
hajdul88: What error do we raise here?

cognee/modules/retrieval/graph_completion_retriever.py
```diff
@@ -40,47 +40,67 @@ def __init__(
         self.top_k = top_k if top_k is not None else 5
         self.node_type = node_type
         self.node_name = node_name
+        logger.info(
```
hajdul88: Same here, this is gonna print out all the retreivers because of the enum

cognee/modules/retrieval/graph_completion_retriever.py
```diff
@@ -40,47 +40,67 @@ def __init__(
...
-            for info in nodes.values()
-        )
-        connection_section = "\n".join(
-            f"{nodes[edge.node1.id]['name']} --[{edge.attributes['relationship_type']}]--> {nodes[edge.node2.id]['name']}"
-            for edge in retrieved_edges
-        )
-        return f"Nodes:\n{node_section}\n\nConnections:\n{connection_section}"
+        logger.debug(f"Resolving {len(edges)} edges to text")
+
+        relationships_text = []
```
hajdul88: This is less optimal than the original one. Why do we want to change this?

cognee/modules/retrieval/insights_retriever.py
```diff
@@ -25,6 +28,9 @@ def __init__(self, exploration_levels: int = 1, top_k: int = 5):
         """Initialize retriever with exploration levels and search parameters."""
         self.exploration_levels = exploration_levels
         self.top_k = top_k
+        logger.info(
```
hajdul88: Same I would delete this because of the enum where we instantiate all the retrievers that we have.

cognee/modules/retrieval/utils/brute_force_triplet_search.py
```diff
@@ -59,22 +59,63 @@ async def get_memory_fragment(
...
             node_type=node_type,
             node_name=node_name,
         )
-    except EntityNotFoundError:
+
+        projection_time = time.time() - start_time
+        node_count = len(memory_fragment.nodes)
+        edge_count = len(memory_fragment.edges)
+
+        logger.info(f"Memory fragment projection completed in {projection_time:.2f} seconds")
```
hajdul88: We log this twice, once its:

 Graph projection completed successfully in 0.08 seconds [CogneeGraph]
2025-07-22T10:18:13.349345 [info     ] Final graph stats: 82 nodes, 113 edges [CogneeGraph]

other time its here:

Memory fragment projection completed in 0.08 seconds [cognee.shared.logging_utils]
2025-07-22T10:18:13.349526 [info     ] Memory fragment contains 82 nodes and 113 edges [cognee.shared.logging_utils]

cognee/modules/retrieval/utils/brute_force_triplet_search.py
```diff
@@ -59,22 +59,63 @@ async def get_memory_fragment(
...
+        logger.info(f"Memory fragment contains {node_count} nodes and {edge_count} edges")
+
+        if node_count == 0:
+            logger.warning("Memory fragment projection resulted in empty graph - no nodes found")
+        if edge_count == 0 and node_count > 0:
```
hajdul88: Same as in cognee graph, this is not necessarily true, we can have 10 nodes and 1 edge then 8 nodes are gonna be isolated.

cognee/modules/retrieval/utils/brute_force_triplet_search.py
```diff
@@ -59,22 +59,63 @@ async def get_memory_fragment(
...
+        logger.info("Continuing with empty memory fragment due to EntityNotFoundError")
+        # This is expected behavior - continue with empty fragment
         pass
+    except Exception as e:
+        logger.error(f"Unexpected error during memory fragment creation: {str(e)}")
+        logger.debug(f"Error type: {type(e).__name__}")
+        # Still return the fragment even if projection failed
+        pass
+
+    final_node_count = len(memory_fragment.nodes)
```
hajdul88: Again I think we log this twice, once in cogneeGraph, the other time here.

hajdul88 (commented):
In general I made a lot of comment mostly that debug logs are too detailed (imho), some parts are not necessary. In other cases we log things twice or the log is not actually true so the logic is faulty.

cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
```diff
@@ -870,34 +870,52 @@ async def get_graph_data(self):
...
+                        record["properties"]["source_node_id"],
+                        record["properties"]["target_node_id"],
+                        record["type"],
+                        record["properties"],
+                    )
+                )
+
+            retrieval_time = time.time() - start_time
+            logger.info(
+                f"Retrieved {len(nodes)} nodes and {len(edges)} edges in {retrieval_time:.2f} seconds"
```
hajdul88, line 911: Btw just found it out now during testing that this gets printed during ingestion too, probably when we look for existing edges and nodes. Its fine for now just good to know.

cognee/modules/retrieval/graph_completion_retriever.py
hajdul88: This part is changing the context that we give to the answer generation completely. I wouldn't change this. I compared the two, the old one looks reasonable, the new one contains continuous text in a None() string, so I would just revert this part. Maybe we can log the context itself or the length (which is already logged) and thats it

hajdul88 (commented):
Looks good, I would go back with the context generation to the original because the new one is changing the context completely, and inserts None into it, also it loses the nodes...edges structure

hajdul88 (dismissed):

PR #1117, opened 2025-07-21
Add generate_completion function to llm_utils using OpenAI API

Adityajl:
Summary
Added `generate_completion` function inside `cognee/modules/search/llm_utils.py` to handle prompt generation and OpenAI API responses.

 Changes
- Created `generate_completion()` with support for system/user prompt path (optional).
- Handles OpenAI async call and errors gracefully.

Motivation
Needed for enabling natural language query handling in the search flow.

Note
- Dependencies like `openai`, `pydantic_settings`, `structlog` required (already handled in dev setup).
- No tests written — to be added later.

Please review and let me know if anything needs changes!

### DIFF
--- a/cognee/modules/search/llm_utils.py
+++ b/cognee/modules/search/llm_utils.py
@@ -1,0 +1,20 @@
+# cognee/modules/search/llm_utils.py
+
+import openai
+
+# You can make this async, since get_completion uses `await`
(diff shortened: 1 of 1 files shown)


Vasilije1990:
@Adityajl why can't we use the openai llm adapter and the functions located there?

PR #1141, opened 2025-07-24
fix: UnstructuredDocument read method

borisarzentar:
## DCO Affirmation


cognee-frontend/Dockerfile
```diff
@@ -9,7 +9,7 @@ COPY package.json package-lock.json ./
 
 # Install any needed packages specified in package.json
 RUN npm ci
-# RUN npm rebuild lightningcss
+RUN npm rebuild lightningcss
```
dexters1, line 12: Do we want this as well?

dexters1 (approved):

cognee-frontend/Dockerfile
borisarzentar: Yes, because of x64 architecture.

Issue #1142, opened 2025-07-24
Support for SSH when running inside App Service - Web App for Containers


jaliyaudagedara:
Would this be something you'd like to have: `Support for SSH when running inside App Service - Web App for Containers`.

Right now, if we run `cognee/cognee:main` in `Azure App Service - Web App for Containers`, we can't SSH in. 
<img width="902" height="250" alt="Image" src="https://github.com/user-attachments/assets/16373e2e-16f9-4701-acfb-9cf1d3b6bde4" />

The Dockerfile doesn't seem to expose port `2222`.

[Custom Container: Enable SSH](https://learn.microsoft.com/en-us/azure/app-service/configure-custom-container?tabs=debian&pivots=container-linux#enable-ssh)

Vasilije1990:
Feel free to open a PR, otherwise we will add to the list

jaliyaudagedara:
@Vasilije1990 

Created a PR

Vasilije1990:
@jaliyaudagedara the PR did not meet our quality standards. I'll check with engineering next week on how to approach this one

jaliyaudagedara:
sure, it's copied from a sample from Microsoft, and not intended for public use.

Vasilije1990:
Fair, closing in that case

PR #1157, opened 2025-07-26
Enable SSH support.

jaliyaudagedara:
#1142 Enable SSH support.
 

## DCO Affirmation

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -49,6 +49,17 @@ RUN apt-get update && apt-get install -y \
     libpq5 \
     && rm -rf /var/lib/apt/lists/*
 
+# Install SSH Server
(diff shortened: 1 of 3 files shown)


borisarzentar:
@jaliyaudagedara What is the reason to add ssh server into the Docker?
If you want access to the container, you can do `docker exec -it <container_name> /bin/bash`.

dexters1:
Hey @jaliyaudagedara thanks for the contribution! But we don't want to enable ssh in the docker container

jaliyaudagedara:
Wanted to enable SSH, because when I was running in Azure, I wanted to access the container. If I am running locally, I can of course do `docker exec`.

All good!

PR #1212, opened 2025-08-06
Feature/cog 2672 dynamic multiple edges in datapoints

lxobr:
## Description

- Improved list handling, removed `.index` logic from `get_graph_from_model`, transitioned to fully datapoint-oriented processing
- Streamlined datapoint iteration by introducing `_datapoints_generator` with nested loops
- Generalized field processing to handle mixed lists: `[DataPoint, (Edge, DataPoint), (Edge, [DataPoint])]`, allowing dynamic multiple edges generation
- Small improvements and refactorings
- Added tests to `test_get_graph_from_model_flexible_edges()` covering weighted edges and dynamic multiple edges
- Created `dynamic_multiple_edges_example.py` demonstrating dynamic multiple edges

## DCO Affirmation


cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -176,65 +193,40 @@ async def get_graph_from_model(
...
-            # True single datapoint case
-            target_datapoint = field_value
+    # Process all relationships using generator
+    for target_datapoint, field_name, edge_metadata in _targets_generator(
+        data_point, properties_to_visit
+    ):
+        relationship_name = _get_relationship_key(field_name, edge_metadata)
 
         # Create edge if not already added
         edge_key = f"{data_point_id}{target_datapoint.id}{field_name}"
```
hajdul88: I know you didn't change this, but theoretically this can cause missing visits between:

datapoint_id = 1
target_datapoint_id = 23
field_name='something'

and

datapoint_id = 12
target_datapoint_id=3
field_name='something'


Because they will get the same edge key.

hajdul88: Maybe worth putting some separator between them.

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -4,43 +4,50 @@
...
-
-    # Handle tuple[Edge, list[DataPoint]]
+def _extract_field_data(field_value: Any) -> List[Tuple[Optional[Edge], List[DataPoint]]]:
+    """Extract edge metadata and datapoints from a field value."""
+    # Handle single DataPoint
+    if isinstance(field_value, DataPoint):
+        return [(None, [field_value])]
+
+    # Handle list - could contain DataPoints, edge tuples, or mixed
+    if isinstance(field_value, list) and len(field_value) > 0:
```
hajdul88, line 14: I dont know if we want to have the old one too that but this accepts these tuples only in a list but not a single tuple. Which is fine just its a change so top level non list tuples will get swallowed I thinkg

cognee/modules/graph/utils/get_graph_from_model.py
hajdul88: Actually it can even cause infinite loop so we definitely need separators

hajdul88 (changes requested):

cognee/modules/graph/utils/get_graph_from_model.py
lxobr: Good catch! I added separators.

cognee/modules/graph/utils/get_graph_from_model.py
lxobr: Actually, we handle tuples in the next lines of code. Please let me know if you think some changes are still needed.

Issue #1182, opened 2025-08-02
Fix Graph Visualization Access for Users with Read Permissions


Vasilije1990:
## Issue Description

Currently, only dataset owners can visualize knowledge graphs through the `/api/v1/datasets/{dataset_id}/graph` endpoint. Users who have been granted read permissions on datasets cannot access graph visualizations due to two issues in the access control implementation:

1. **Dataset Retrieval Restriction**: The `get_dataset()` function only returns datasets for owners
2. **Incorrect Database Context**: The `get_formatted_graph_data()` function sets database context with the requesting user's ID instead of the dataset owner's ID

This breaks the intended permission model where users with read access should be able to view graph visualizations.

## Current Behavior

When a user with read permissions tries to access a dataset's graph visualization:

1. **API Call**: `GET /api/v1/datasets/{dataset_id}/graph`
2. **Dataset Check**: `get_dataset(user.id, dataset_id)` returns `None` because user is not owner
3. **Result**: 404 error - "Dataset not found" even though user has read permissions

Even if the dataset retrieval were fixed, the database context would be set incorrectly, preventing access to the actual graph data.

## Expected Behavior

Users with read permissions should be able to:
- Access the graph visualization endpoint
- View the complete knowledge graph for datasets they have read access to
- See the same graph data as the dataset owner (subject to their read permissions)

### Phase 1: Update Dataset Retrieval Logic

Create a new function `get_dataset_with_permissions()` that checks read access:

### Phase 2: Fix Graph Data Context Setting

Update `get_formatted_graph_data()` to use dataset owner's ID for context:

### New Files
```
cognee/modules/data/methods/get_dataset_with_permissions.py
```

### Modified Files
```
cognee/modules/graph/methods/get_formatted_graph_data.py
cognee/api/v1/datasets/routers/get_datasets_router.py
cognee/modules/data/methods/__init__.py  # Add new import
```

### Test Files
```
cognee/tests/test_graph_visualization_permissions.py  # New integration tests
cognee/tests/test_dataset_permissions.py  # Update existing tests
```

EricXiao95:
Could you please assign this issue to me?

Vasilije1990:
@EricXiao95 done and thanks!

EricXiao95:
Hi @Vasilije1990 :

Thanks for taking a look at this. When you have a moment, could you please review my [PR](https://github.com/topoteretes/cognee/pull/1220)?

[...]

PR #1220, opened 2025-08-07
fix: graph visualization access for users with read permissions

EricXiao95:
## Description

Description
This PR fix graph visualization access for users with read permissions (https://github.com/topoteretes/cognee/issues/1182)

- Add permission checks for graph visualization endpoints to ensure users can only access datasets they have permission to view
- Create get_dataset_with_permissions method to validate user access before returning a dataset
- Remove redundant dataset existence validation in datasets router and delegate permission checking to graph data retrieval
- Add comprehensive test suite for graph visualization permissions covering owner access and permission granting scenarios
- Update get_formatted_graph_data() to use dataset owner's ID for context
## Testing
Tests can be run with:
```bash
pytest -s cognee/tests/test_graph_visualization_permissions.py
```

## DCO Affirmation

### DIFF
--- a/cognee/modules/data/methods/get_dataset_with_permissions.py
+++ b/cognee/modules/data/methods/get_dataset_with_permissions.py
@@ -1,0 +1,29 @@
+from typing import Optional
+from uuid import UUID
+from cognee.modules.users.exceptions import PermissionDeniedError
+from cognee.modules.users.permissions.methods import get_specific_user_permission_datasets
+from ..models import Dataset
+
+
+async def get_dataset_with_permissions(
+    user_id: UUID, dataset_id: UUID, permission_type="read"
+) -> Optional[Dataset]:
+    """
+    Get a specific dataset with permissions for a user.
+
+    Args:
+        user_id (UUID): user id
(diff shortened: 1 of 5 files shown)


cognee/modules/data/methods/get_dataset_with_permissions.py
```diff
@@ -0,0 +1,29 @@
+from typing import Optional
+from uuid import UUID
+from cognee.modules.users.exceptions import PermissionDeniedError
+from cognee.modules.users.permissions.methods import get_specific_user_permission_datasets
+from ..models import Dataset
+
+
+async def get_dataset_with_permissions(
```
dexters1: Can you rename the function to get_authorized_dataset? And could you let the exception happen for now instead of catching it if the user does not have permission to access the dataset.

Overall seems fine, will run the tests when changes are made

EricXiao95: Great suggestions. I've just pushed the changes. The function is now get_authorized_dataset and the exception is no longer caught.

Vasilije1990 (approved):

PR #1222, opened 2025-08-07
Lancedb async lock

dexters1:
## DCO Affirmation


cognee/infrastructure/databases/vector/lancedb/LanceDBAdapter.py
```diff
@@ -17,6 +17,8 @@
 from ..utils import normalize_distances
 from ..vector_db_interface import VectorDBInterface
 
+LANCEDB_LOCK = asyncio.Lock()
```
dexters1: Locks need to be part of database instance and not global locks

cognee/infrastructure/databases/vector/vector_db_async_lock.py
```diff
@@ -0,0 +1,4 @@
+import asyncio
+
+# A single lock shared by all coroutines
+VECTOR_DB_LOCK = asyncio.Lock()
```
dexters1: Remove this so it's not global but tied to vector db instance

dexters1: Or maybe using global context variables

dexters1:
Adding an async lock for database table creation should be added to all vector database adapter (community and core) a task should be made to add these as well after merge (before release)

PR #1238, opened 2025-08-12
Incremental loading migration

dexters1:
## Description
Add relational db migration for incremental loading, change incremental loading to work document per document instead of async together

## DCO Affirmation


hajdul88 (approved):
Gonna be slower but we discussed this.

PR #1240, opened 2025-08-12
Loader separation

dexters1:
## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/files/utils/guess_file_type.py
+++ b/cognee/infrastructure/files/utils/guess_file_type.py
@@ -1,7 +1,6 @@
 from typing import BinaryIO
 import filetype
 from .is_text_content import is_text_content
-from .get_data_file_path import get_data_file_path
 
 
 class FileTypeException(Exception):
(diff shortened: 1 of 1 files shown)


dexters1:
> Just one thing, as I understand this is a big change and we have to be sure that splitting and putting ingestion into add, doesn't cause issues with other systems that are not covered with tests. (distributed cognee, dreamify, modal eval entrypoints etc.)

Maybe we need to add basic tests for all of these?

hajdul88 (approved):
I tested it, we had back and forth rounds but now it seems everything is working.

Tested (with different files):
Locally
Local API
Local API with s3 

All good!

One note: With this dev we will have every file as TextDocuments in the graph for now.

Nice job LGTM

cognee/infrastructure/loaders/external/pypdf_loader.py
```diff
@@ -0,0 +1,96 @@
...
+
+                content_parts = []
+                page_texts = []
+
+                for page_num, page in enumerate(reader.pages, 1):
+                    try:
+                        page_text = page.extract_text()
+                        if page_text.strip():  # Only add non-empty pages
+                            page_texts.append(page_text)
+                            content_parts.append(f"Page {page_num}:\n{page_text}\n")
```
dexters1, line 78: Do we want to add this content_parts Page info during loading of pdf or just the raw pdf text? @Vasilije1990

cognee/infrastructure/files/utils/guess_file_type.py
```diff
@@ -125,6 +126,48 @@ def guess_file_type(file: BinaryIO) -> filetype.Type:
...
+        if file_name:
+            # Use existing function to transform file path to be OS usable
+            actual_file_path = get_data_file_path(file_name)
+
+            # Extract extension from the cleaned path
+            if "." in actual_file_path:
+                extension = actual_file_path.split(".")[-1].lower()
+
+                # Map common extensions to our custom file types
+                if extension == "txt":
```
dexters1: This is not a good idea, anyone can change extension of file to pretend to be a different file type it's not secure.

filetype.guess(file) derives the file type and mime type from actual file content

PR #1275, opened 2025-08-20
feat: make authentication optional

daukadolt:
## Description
Quick sanity check

https://github.com/user-attachments/assets/2c899472-2280-4f35-86d2-a3901131b919

### Testing through frontend

1. Verifying user can create new datasets, and that dataset exists in database

https://github.com/user-attachments/assets/a74ae246-6733-40c5-a63d-cd7c009f9e29

2. Verifying the dataset is owned by default user

https://github.com/user-attachments/assets/e7942428-df47-4902-be3d-9319dc5385cc

## DCO Affirmation


daukadolt:
Unit tests need better configuring - individually tests are passing, but running all together fails

cognee/modules/users/methods/get_conditional_authenticated_user.py
```diff
@@ -0,0 +1,42 @@
+import os
+from typing import Optional
+from fastapi import Depends, HTTPException
+from ..models import User
+from ..get_fastapi_users import get_fastapi_users
+from .get_default_user import get_default_user
+
+# Check environment variable to determine authentication requirement
+REQUIRE_AUTHENTICATION = os.getenv("REQUIRE_AUTHENTICATION", "false").lower() == "true"
```
dexters1: Make sure to also add this variable to the .env.template and our docs if we're introducing it. 

Also having this false by default will cause security issues to our existing users (and our SaaS if we don't update the value) since the authentication will be disabled for everyone, I'm not sure we should turn it off by default now due to this reason

Vasilije1990: @dexters1 based on logs from our docs chatbot, I can see that they don't really understand how to do the auth, which blocks them from even trying the tool.

I would argue that average user doesn't need auth, and this should represent it

dexters1: Discord users are using ENABLE_BACKEND_ACCESS_CONTROL, I've supported them myself about this feature, we need to make sure at least that if 

ENABLE_BACKEND_ACCESS_CONTROL=True

that authentication must be required, otherwise it won't work. If it's not enabled, authentication doesn't have to be required as different users don't have a point then. Might even make sense to just have authentication off unless ENABLE_BACKEND_ACCESS_CONTROL is set to True instead of another env variable

cognee/api/v1/cognify/routers/get_cognify_router.py
```diff
@@ -10,7 +10,7 @@
 from cognee.api.DTO import InDTO
 from cognee.modules.pipelines.methods import get_pipeline_run
 from cognee.modules.users.models import User
-from cognee.modules.users.methods import get_authenticated_user
+from cognee.modules.users.methods import get_conditional_authenticated_user
```
borisarzentar: Let's keep the `get_authenticated_user` method, but handle the logic inside of it.

cognee/modules/users/methods/get_authenticated_user.py
```diff
@@ -1,48 +0,0 @@
-from ..get_fastapi_users import get_fastapi_users
-
-
-fastapi_users = get_fastapi_users()
-
-get_authenticated_user = fastapi_users.current_user(active=True)
```
borisarzentar, line 6: We can decide here what to assign to `get_authenticated_user` based on the env variable. If auth is not required, we can always return default user and that should be it.

cognee/tests/unit/api/test_conditional_authentication_endpoints.py
```diff
@@ -0,0 +1,283 @@
...
+        # Should be a boolean value (the parsing logic works)
+        assert isinstance(REQUIRE_AUTHENTICATION, bool)
+
+        # In default environment, should be False
+        assert REQUIRE_AUTHENTICATION == False
+
+
+# Fixtures for reuse across test classes
+@pytest.fixture
+def mock_default_user():
```
borisarzentar: Is this already defined above?

cognee/tests/unit/api/test_conditional_authentication_endpoints.py
```diff
@@ -0,0 +1,283 @@
...
+
+        # Simulate authenticated request by directly testing the conditional function
+        from cognee.modules.users.methods.get_conditional_authenticated_user import (
+            get_conditional_authenticated_user,
+        )
+
+        async def test_logic():
+            # When user is provided (authenticated), should not call get_default_user
+            result = await get_conditional_authenticated_user(user=mock_authenticated_user)
+            assert result == mock_authenticated_user
```
borisarzentar: Does this test exists in the test file below as well? Looks like it is the same as one on line 39 in `test_conditional_authentication.py`

cognee/tests/unit/modules/users/test_conditional_authentication.py
```diff
@@ -0,0 +1,321 @@
...
+                # Test with user
+                result1 = await get_conditional_authenticated_user(user=mock_user)
+                assert result1 == mock_user
+
+                # Test with None
+                result2 = await get_conditional_authenticated_user(user=None)
+                assert result2 == mock_default_user
+
+                # Both should have user-like interface
+                assert hasattr(result1, "id")
```
borisarzentar: Can we check values when we are already testing it in detail?

cognee/tests/unit/api/test_conditional_authentication_endpoints.py
daukadolt: Good catch, must've missed it

cognee/api/client.py
```diff
@@ -110,7 +110,8 @@ def custom_openapi():
         },
     }
 
-    openapi_schema["security"] = [{"BearerAuth": []}, {"CookieAuth": []}]
```
borisarzentar, line 113: Let's do this conditionally, depending on the env var

cognee/modules/users/methods/get_authenticated_user.py
```diff
@@ -1,48 +1,45 @@
...
-#         raise HTTPException(status_code=401, detail="Token has expired")
-#     except jwt.InvalidTokenError:
-#         raise HTTPException(status_code=401, detail="Invalid token")
+if REQUIRE_AUTHENTICATION:
+    # When REQUIRE_AUTHENTICATION=true, enforce authentication (original behavior)
+    _auth_dependency = fastapi_users.current_user(active=True)
+else:
+    # When REQUIRE_AUTHENTICATION=false (default), make authentication optional
+    _auth_dependency = fastapi_users.current_user(
+        optional=True,  # Returns None instead of raising HTTPException(401)
```
borisarzentar: Can we just pass the `REQUIRE_AUTHENTICATION` as `optional` value?

cognee/modules/users/methods/get_authenticated_user.py
```diff
@@ -1,48 +1,45 @@
...
+    - If REQUIRE_AUTHENTICATION=true: Enforces authentication (raises 401 if not authenticated)
+    - If REQUIRE_AUTHENTICATION=false: Falls back to default user if not authenticated
+
+    Always returns a User object for consistent typing.
+    """
+    if user is None and not REQUIRE_AUTHENTICATION:
+        # When authentication is optional and user is None, use default user
+        try:
+            user = await get_default_user()
+        except Exception as e:
```
borisarzentar, line 37: We should log the error as well.

cognee/tests/unit/api/test_conditional_authentication_endpoints.py
daukadolt: True, they are identical. This one was initially created to be in the context of "add" endpoint, but they really are doing the same thing. Cleaning this one up

cognee/modules/users/methods/get_authenticated_user.py
daukadolt: That's a neat refactor. Updating

PR #1284, opened 2025-08-25
feat: migrate pipeline input validation to a layer

borisarzentar:
## DCO Affirmation

### DIFF
--- a/cognee/modules/pipelines/exceptions/tasks.py
+++ b/cognee/modules/pipelines/exceptions/tasks.py
@@ -1,0 +1,18 @@
+from fastapi import status
+from cognee.exceptions import CogneeValidationError
+
+
+class WrongTaskTypeError(CogneeValidationError):
+    """
+    Raised when the tasks argument is not a list of Task class instances.
+    """
+
+    def __init__(
+        self,
+        message: str = "tasks argument must be a list, containing Task class instances.",
+        name: str = "WrongTaskTypeError",
+        status_code=status.HTTP_400_BAD_REQUEST,
+    ):
+        self.message = message
+        self.name = name
+        self.status_code = status_code
--- a/cognee/modules/pipelines/layers/pipeline_tasks_validation.py
+++ b/cognee/modules/pipelines/layers/pipeline_tasks_validation.py
@@ -1,0 +1,13 @@
+from ..tasks.task import Task
+from ..exceptions.tasks import WrongTaskTypeError
+
+
+def pipeline_tasks_validation(tasks):
+    if not isinstance(tasks, list):
+        raise WrongTaskTypeError(f"tasks argument must be a list, got {type(tasks).__name__}.")
+
+    for task in tasks:
+        if not isinstance(task, Task):
+            raise WrongTaskTypeError(
+                f"tasks argument must be a list of Task class instances, got {type(task).__name__} in the list."
+            )
--- a/cognee/modules/pipelines/layers/__init__.py
+++ b/cognee/modules/pipelines/layers/__init__.py
@@ -1,0 +1,1 @@
+from .pipeline_tasks_validation import *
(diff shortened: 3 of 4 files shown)


cognee/modules/pipelines/layers/pipeline_tasks_validation.py
```diff
@@ -0,0 +1,13 @@
+from ..tasks.task import Task
+from ..exceptions.tasks import WrongTaskTypeError
+
+
+def pipeline_tasks_validation(tasks):
```
dexters1: Can we add typing to tasks and return value as None? Also a simple Docstring

borisarzentar: Why return None?

cognee/modules/pipelines/exceptions/tasks.py
```diff
@@ -0,0 +1,18 @@
...
+    Raised when the tasks argument is not a list of Task class instances.
+    """
+
+    def __init__(
+        self,
+        message: str = "tasks argument must be a list, containing Task class instances.",
+        name: str = "WrongTaskTypeError",
+        status_code=status.HTTP_400_BAD_REQUEST,
+    ):
+        self.message = message
```
hajdul88, line 16: This should call super too no? Something like

super().__init__(message=message, name=name, status_code=status_code)

optionally with log and log level params as well.

cognee/modules/pipelines/layers/pipeline_tasks_validation.py
dexters1: None is always returned if there is no return value. We don't have to specify this, but thought explicitly marking this could be helpful

cognee/modules/pipelines/exceptions/tasks.py
borisarzentar: I think it is the same if I do this or call super, as super class constructor will do what I did here. And log and log level can stay default, basically log as error.

cognee/modules/pipelines/layers/pipeline_tasks_validation.py
borisarzentar: There is a linter rule for the consistent return, where it is expected to return a value at the end if you returned  somewhere above in the code. If you don't use return, then it should not exist at the end as well. I follow that one, so I stay consistent with the rest of the code.

dexters1: Makes sense

cognee/modules/pipelines/exceptions/tasks.py
dexters1: I think the only difference is the log level handling and params in the super:

# Automatically log the exception details
        if log and (log_level == "ERROR"):
            logger.error(f"{self.name}: {self.message} (Status code: {self.status_code})")
        elif log and (log_level == "WARNING"):
            logger.warning(f"{self.name}: {self.message} (Status code: {self.status_code})")
        elif log and (log_level == "INFO"):
            logger.info(f"{self.name}: {self.message} (Status code: {self.status_code})")
        elif log and (log_level == "DEBUG"):
            logger.debug(f"{self.name}: {self.message} (Status code: {self.status_code})")
            
If log level is intended to be ERROR we probably don't need to use super, but might be good to be consistent everywhere. Idk don't have an opinion on this

dexters1: I just tried it out, it's the same other than the log level handling. Even the message handling as it's inherited from the super.

    def __str__(self):
        return f"{self.name}: {self.message} (Status code: {self.status_code})"

hajdul88: Yes, if its error its fine.

PR #1287, opened 2025-08-25
Cognee pipeline layers

dexters1:
## Description
Add dataset authorization layer
Add pipeline processing status layer

## DCO Affirmation


cognee/modules/pipelines/layers/authorized_user_datasets.py
```diff
@@ -0,0 +1,55 @@
...
+from cognee.modules.data.exceptions import DatasetNotFoundError
+from cognee.modules.data.methods import (
+    get_authorized_existing_datasets,
+    load_or_create_datasets,
+    check_dataset_name,
+)
+
+
+async def authorized_user_datasets(
+    datasets: Union[str, list[str], list[UUID]], user: User = None
```
hajdul88: Just typing, this can handle single UUID as well right? At least based on line 35 it can

cognee/modules/pipelines/layers/process_pipeline_check.py
```diff
@@ -0,0 +1,60 @@
...
+    PipelineRunCompleted,
+    PipelineRunStarted,
+)
+
+logger = get_logger(__name__)
+
+
+async def process_pipeline_check(
+    dataset: Dataset, data: list[Data], pipeline_name: str
+) -> [None, PipelineRunStarted, PipelineRunCompleted]:
```
hajdul88: Maybe Optional[Union[PipelineRunStarted, PipelineRunCompleted]]

cognee/modules/pipelines/layers/process_pipeline_check.py
```diff
@@ -0,0 +1,60 @@
...
+    # async with update_status_lock: TODO: Add UI lock to prevent multiple backend requests
+    if isinstance(dataset, Dataset):
+        task_status = await get_pipeline_status([dataset.id], pipeline_name)
+    else:
+        task_status = [
+            PipelineRunStatus.DATASET_PROCESSING_COMPLETED
+        ]  # TODO: this is a random assignment, find permanent solution
+
+    if str(dataset.id) in task_status:
+        if task_status[str(dataset.id)] == PipelineRunStatus.DATASET_PROCESSING_STARTED:
```
hajdul88, line 41: The else case assigns a list to task_status and here we access it like a dict. I mean probably never gonna happen but still. Am not 100% familiar what do we do in the case of the fallback (I assume we skip this whole if and just return None btu then why do we even need an else. Why not initialiting empty dict in beginning, but maybe am wrong.)

cognee/modules/pipelines/layers/authorized_user_datasets.py
```diff
@@ -0,0 +1,55 @@
...
+    if user is None:
+        user = await get_default_user()
+
+    # Convert datasets to list
+    if isinstance(datasets, str) or isinstance(datasets, UUID):
+        datasets = [datasets]
+
+    # Get datasets user wants write permissions for (verify user has permissions if datasets are provided as well)
+    # NOTE: If a user wants to write to a dataset he does not own it must be provided through UUID
+    existing_datasets = await get_authorized_existing_datasets(datasets, "write", user)
```
hajdul88: We get the authorized datasets here as well no? At least I have the feeling that the get_authorized_existing_datasets and the load_or_create_datasets could be merged.

hajdul88 (commented):
Had a couple of comments, but otherwise looks good

cognee/modules/pipelines/layers/authorized_user_datasets.py
dexters1: It can't, we use get_authorized_existing_datasets in different places for different reasons across Cognee. It shouldn't create a Dataset

cognee/modules/pipelines/layers/process_pipeline_check.py
dexters1: tbh I don't think the else does anything, was just scared to remove it. We probably just need something so the if bellow would not raise an error on an empty value, changed it to an empty dict let's see if tests pass

hajdul88 (approved):

borisarzentar (approved):

PR #1289, opened 2025-08-25
feat: migrate pipeline status reset to add pipeline

borisarzentar:
## DCO Affirmation

### DIFF
--- a/cognee/modules/pipelines/exceptions/tasks.py
+++ b/cognee/modules/pipelines/exceptions/tasks.py
@@ -1,0 +1,18 @@
+from fastapi import status
+from cognee.exceptions import CogneeValidationError
+
+
+class WrongTaskTypeError(CogneeValidationError):
+    """
+    Raised when the tasks argument is not a list of Task class instances.
+    """
+
+    def __init__(
+        self,
+        message: str = "tasks argument must be a list, containing Task class instances.",
+        name: str = "WrongTaskTypeError",
+        status_code=status.HTTP_400_BAD_REQUEST,
+    ):
+        self.message = message
+        self.name = name
+        self.status_code = status_code
--- a/cognee/modules/pipelines/layers/__init__.py
+++ b/cognee/modules/pipelines/layers/__init__.py
@@ -1,0 +1,1 @@
+from .validate_pipeline_tasks import validate_pipeline_tasks
--- a/cognee/modules/pipelines/layers/validate_pipeline_tasks.py
+++ b/cognee/modules/pipelines/layers/validate_pipeline_tasks.py
@@ -1,0 +1,20 @@
+from ..tasks.task import Task
+from ..exceptions.tasks import WrongTaskTypeError
+
+
+def validate_pipeline_tasks(tasks: list[Task]):
+    """
+    Validates the tasks argument to ensure it is a list of Task class instances.
+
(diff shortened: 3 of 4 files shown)


cognee/api/v1/add/add.py
```diff
@@ -140,11 +151,41 @@ async def add(
         Task(ingest_data, dataset_name, user, node_set, dataset_id, preferred_loaders),
     ]
 
+    await setup()
+
+    if not user:
+        user = await get_default_user()
+
+    if dataset_id:
```
dexters1: The get_authorized_existing_datasets function handles both name and UUID input already

cognee/api/v1/add/add.py
```diff
@@ -140,11 +151,41 @@ async def add(
...
+        authorized_dataset = await get_authorized_dataset_by_name(dataset_name, user, "write")
+
+        if not authorized_dataset:
+            authorized_dataset = await create_authorized_dataset(
+                dataset_name=dataset_name, user=user
+            )
+    else:
+        raise ValueError("Either dataset_id or dataset_name must be provided.")
+
+    if not authorized_dataset:
```
dexters1: Is there a way this can happen if we're creating a dataset just before? Maybe this kind of check should be part of the create function if necessary

cognee/modules/data/methods/create_authorized_dataset.py
```diff
@@ -0,0 +1,26 @@
...
+async def create_authorized_dataset(dataset_name: str, user: User) -> Dataset:
+    # Dataset id should be generated based on dataset_name and owner_id/user so multiple users can use the same dataset_name
+    dataset_id = await get_unique_dataset_id(dataset_name=dataset_name, user=user)
+    new_dataset = Dataset(id=dataset_id, name=dataset_name, data=[])
+    new_dataset.owner_id = user.id
+
+    db_engine = get_relational_engine()
+
+    async with db_engine.get_async_session() as session:
+        session.add(new_dataset)
```
dexters1: Can you use the create_dataset function as it will handle the situation if a dataset already exists and won't break?

cognee/modules/data/methods/create_authorized_dataset.py
```diff
@@ -0,0 +1,26 @@
+from cognee.infrastructure.databases.relational import get_relational_engine
+from cognee.modules.users.models import User
+from cognee.modules.data.models import Dataset
+from cognee.modules.users.permissions.methods import give_permission_on_dataset
+from cognee.modules.data.methods.get_unique_dataset_id import get_unique_dataset_id
+
+
+async def create_authorized_dataset(dataset_name: str, user: User) -> Dataset:
```
dexters1, line 8: If we'll use this function for creation of authorized datasets could you also use it in the load_or_create_datasets function so authorized datasets will all be created in one function?

cognee/modules/data/methods/create_authorized_dataset.py
borisarzentar: Using it now.

cognee/modules/data/methods/create_authorized_dataset.py
borisarzentar: Using it there as well now.

PR #1295, opened 2025-08-26
fix: Resolve issue with Windows path

dexters1:
## Description
Resolve issue with Windows path

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/files/utils/get_data_file_path.py
+++ b/cognee/infrastructure/files/utils/get_data_file_path.py
@@ -5,22 +5,13 @@ from urllib.parse import urlparse
 def get_data_file_path(file_path: str):
     # Check if this is a file URI BEFORE normalizing (which corrupts URIs)
     if file_path.startswith("file://"):
+        # Remove first occurrence of file:// prefix
+        pure_file_path = file_path.replace("file://", "", 1)
         # Normalize the file URI for Windows - replace backslashes with forward slashes
-        normalized_file_uri = os.path.normpath(file_path)
-
-        parsed_url = urlparse(normalized_file_uri)
-
-        # Convert URI path to file system path
-        if os.name == "nt":  # Windows
-            # Handle Windows drive letters correctly
-            fs_path = parsed_url.path
-            if fs_path.startswith("/") and len(fs_path) > 1 and fs_path[2] == ":":
-                fs_path = fs_path[1:]  # Remove leading slash for Windows drive paths
-        else:  # Unix-like systems
-            fs_path = parsed_url.path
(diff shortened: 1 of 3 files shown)


cognee/infrastructure/files/utils/get_data_file_path.py
```diff
@@ -5,19 +5,24 @@
 def get_data_file_path(file_path: str):
     # Check if this is a file URI BEFORE normalizing (which corrupts URIs)
     if file_path.startswith("file://"):
+        # Remove first occurrence of file:// prefix
+        pure_file_path = file_path.replace("file://", "", 1)
         # Normalize the file URI for Windows - replace backslashes with forward slashes
-        normalized_file_uri = os.path.normpath(file_path)
+        normalized_file_uri = os.path.normpath(pure_file_path)
 
-        parsed_url = urlparse(normalized_file_uri)
```
dexters1, line 11: @borisarzentar I've removed urlparse (as it didn't correctly handle windows path) and I'm just manually replacing the first instance of "file://" in the file_path, is this alright? Did we need urlparse for a reason I'm not aware?

.github/workflows/test_different_operating_systems.yml
```diff
@@ -0,0 +1,234 @@
...
+          LLM_ENDPOINT: ${{ secrets.LLM_ENDPOINT }}
+          LLM_API_KEY: ${{ secrets.LLM_API_KEY }}
+          LLM_API_VERSION: ${{ secrets.LLM_API_VERSION }}
+
+          EMBEDDING_PROVIDER: openai
+          EMBEDDING_MODEL: ${{ secrets.EMBEDDING_MODEL }}
+          EMBEDDING_ENDPOINT: ${{ secrets.EMBEDDING_ENDPOINT }}
+          EMBEDDING_API_KEY: ${{ secrets.EMBEDDING_API_KEY }}
+          EMBEDDING_API_VERSION: ${{ secrets.EMBEDDING_API_VERSION }}
+  run-integration-tests:
```
borisarzentar, line 76:
```suggestion

  run-integration-tests:
```

cognee/infrastructure/files/utils/get_data_file_path.py
borisarzentar: @dexters1 The windows part was written by Vasilije. Maybe in cases where there are multiple `/` characters between path chunks, like `file:///Users//blabla//directory//file`. But not sure if that is possible, because we usually pass file path that we generate.

dexters1: It's working for all file types and ingestion loaders currently so should be good

PR #1316, opened 2025-08-29
fix: Make metadata prune true by default

dexters1:
## DCO Affirmation


cognee/modules/data/deletion/prune_system.py
```diff
@@ -3,7 +3,7 @@
 from cognee.infrastructure.databases.relational import get_relational_engine
 
 
-async def prune_system(graph=True, vector=True, metadata=False):
+async def prune_system(graph=True, vector=True, metadata=True):
```
dexters1, line 6: Due to incremental loading if graph and vector DBs are deleted the metadata needs to be pruned as well

PR #1325, opened 2025-09-01
Revisit global upper bounds on dependencies (align with common OSS practice)

altafilius:
### Problem

The project currently caps most runtime dependencies with strict upper bounds. This diverges from common practice in widely used OSS libraries and causes resolver conflicts / blocked upgrades for downstream users (e.g., projects on newer minor versions such as OpenAI 2.x).

 **Examples from major projects:**
- **Flask** documents dependency requirements primarily as minimums (e.g., `Werkzeug ≥ 3.1, Jinja2 > 3.1.2, ItsDangerous ≥ 2.2, Click ≥ 8.1.3`), without broad upper caps. (https://flask.palletsprojects.com/en/stable/changes/)

- **FastAPI** uses selective upper bounds only where tight coupling warrants it (e.g., `starlette>=0.27.0,<0.28.0, temporary anyio<4.0.0`), rather than capping everything.  (https://github.com/tiangolo/fastapi/discussions/10248))

 ### Request

Adopt` a lighter, well-established policy:
 - Use lower bounds by default.
 - Add selective < next-major caps only when there’s proven tight coupling or recent breakage, with a brief rationale in comments/README.
 - Keep reproducibility via constraints/lock in CI, not via global caps exported to library users.
 - Add a small CI matrix (min + latest) for a few critical deps to enforce compatibility without broad caps.

>**Minimal unblock (now)**
> Relax the openai constraint to allow 2.x while keeping a next-major guard (e.g., openai>=2.0.0,<3.0.0, or >=1.80.1,<3.0.0 if 1.x must remain supported).

### Goal
Reduce friction for adopters while preserving stability—consistent with practices seen in major OSS projects (Flask’s lower-bounds approach; FastAPI’s targeted caps).

I’m happy to open a PR that implements the minimal change above and adds a brief “Dependency Versioning Policy” plus the CI checks.

### DIFF
--- a/.github/workflows/compatibility_matrix.yml
+++ b/.github/workflows/compatibility_matrix.yml
@@ -1,0 +1,87 @@
+name: CI
+
+on:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
+    branches:
+      - main
+      - master
+
+permissions:
+  contents: read
+
+jobs:
+  lint_and_compatibility:
+    name: Lint and Compatibility Matrix
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        python: ['3.10', '3.11', '3.12']
+        dep: ['lowest', 'latest']
+    steps:
+      - &checkout
+        name: Check out repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python ${{ matrix.python }}
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python }}
+
+      - &install_uv
+        name: Install uv
+        run: pip install uv
+
+      - name: Install dependencies
+        run: |
+          set -euo pipefail
+          if [ "${{ matrix.dep }}" = "lowest" ]; then
+            uv pip install --system -c constraints-lowest.txt -e .
+          else
+            uv pip install --system -e .
+          fi
+
+      - name: Run linting
+        uses: astral-sh/ruff-action@v2
+        with:
+          args: check .
+
+      - name: Show pip cache (for debugging)
+        run: echo "${PIP_CACHE_DIR:-$HOME/.cache/pip}"
+
+      - name: Install OpenAI dependency (latest only)
+        if: matrix.dep == 'latest'
+        run: |
+          set -euo pipefail
+          if uv pip install --system -e . 'openai>=2,<3'; then
+            true
+          else
+            echo '::notice::OpenAI 2.x not available; installing without constraint.'
+            uv pip install --system -e .
+          fi
+
+      - name: Run import smoke test
+        run: |
+          set -euo pipefail
+          python - <<'PY'
+          try:
+              import cognee
+              print({"import_ok": True})
+          except Exception as e:
+              print({"import_ok": False, "error": str(e)})
+              raise
+          PY
+
+      - name: Run tests (if present)
+        run: |
+          set -euo pipefail
+          if [ -d tests ] || ls -1 *.py 2>/dev/null | grep -qE '(^|/)test_'; then
+            uv pip install --system pytest pytest-asyncio pytest-cov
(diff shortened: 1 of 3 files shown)


Vasilije1990:
@altafilius we didn't have strict upper bounds, and then within a week, three dependencies broke and caused library not to work. I went a bit too hard  on the upper bounds and your suggestion seems reasonable.

Feel free to open a PR with associated CI matrix and we will review and merge!

Thank you for raising this

altifilius:
> _Apologies for the delayed response. While working on this, I encountered some unexpected issues due to the structure of the workflow, which prevented me from building as extensive a CI pipeline as I initially planned. Because of that, I opted for a simpler but stable version that still covers the key requirements_

### ci: add lowest+latest dependency matrix and OpenAI 2.x check

This PR revises the CI workflow to reflect the updated dependency management policy (PEP 440–style: lower bounds by default, only targeted upper bounds when necessary). The goal is to ensure continuous compatibility across both minimal and up-to-date dependency sets, improve reproducibility, and validate forward compatibility.

**Dependency matrix (lowest + latest):**

- Added a new matrix.dep with values lowest and latest.
- Lowest lane installs dependencies from constraints-lowest.txt, ensuring that the project always remains compatible with the minimum supported versions.
 - Latest lane installs without constraints, verifying that the project works with the most recent available packages.

**OpenAI 2.x compatibility check:**

- On latest lanes, the workflow explicitly attempts to install openai>=2,<3.
- If unavailable, the step falls back gracefully to an unconstrained install, ensuring CI continues while still probing for forward compatibility.

**Python version coverage:**

- The workflow now runs across Python 3.10, 3.11, and 3.12.
- Tests are executed in all matrix lanes to detect regressions early and ensure stability across supported Python versions.

**Linting:**

- Retained Ruff linting via ruff-action for consistent style enforcement throughout the codebase.

_I’m open to any suggestions or improvements that could make this workflow more robust or better aligned with the project’s long-term goals._

altifilius:
_@Vasilije1990 this is bump._

.github/workflows/compatibility_matrix.yml
```diff
@@ -0,0 +1,87 @@
...
+
+      - name: Install uv
+        run: pip install uv
+
+      - name: Install dependencies (lowest or latest)
+        run: |
+          set -euo pipefail
+          if [ "${{ matrix.dep }}" = "lowest" ]; then
+            if [ -f constraints-lowest.txt ]; then
+              echo '::notice::Using constraints-lowest.txt for lowest-deps lane.'
```
borisarzentar, line 41: @altafilius Where do we get this `constraints-lowest.txt` from?

.github/workflows/compatibility_matrix.yml
```diff
@@ -0,0 +1,87 @@
...
+            # latest lane + OpenAI 2.x probe in one go (no double install)
+            if uv pip install --system -e . 'openai>=2,<3'; then
+              :
+            else
+              echo '::notice::OpenAI 2.x not available; installing without constraint.'
+              uv pip install --system -e .
+            fi
+          fi
+
+      - name: Run linting
```
borisarzentar, line 59: We don't have to run linting here, it has its own workflow.

.github/workflows/compatibility_matrix.yml
```diff
@@ -0,0 +1,87 @@
...
+          python - <<'PY'
+          try:
+              import cognee
+              print({"import_ok": True})
+          except Exception as e:
+              print({"import_ok": False, "error": str(e)})
+              raise
+          PY
+
+      - name: Run tests (if present)
```
borisarzentar, line 79: We also don't have to run tests here, they have their own workflows.

Vasilije1990:
Resolved in another PR. Closing here, thanks for the contribution!

PR #1329, opened 2025-09-03
Memify pipeline

dexters1:
## Description
Memify pipeline and examples

## DCO Affirmation


cognee/api/v1/cognify/memify.py
```diff
@@ -0,0 +1,98 @@
+from typing import Union, Optional, List, Type, Any
```
Vasilije1990: Please move this to memify folder, also, add it to API

Vasilije1990: We are extending the core API

Vasilije1990 (approved):

PR #1336, opened 2025-09-05
Fix ChromaDBAdapter AttributeError

quark412:
## Description
Pretty self-explanatory! list_collections() is an attribute of the AsyncHttpClient, not of the ChromaDBAdapter, so line 541 should say client.list_collections() (as indeed line 556 correctly already does). This will fix the AttributeError I was getting.

This is my first ever PR so let me know if there is anything else I need to do :)

## DCO Affirmation


hemant2456:
Please assign the task

PR #1337, opened 2025-09-06
fix: Fix issue with path

Vasilije1990:
## DCO Affirmation

### DIFF
--- a/cognee/tasks/ingestion/data_item_to_text_file.py
+++ b/cognee/tasks/ingestion/data_item_to_text_file.py
@@ -75,9 +75,5 @@ async def data_item_to_text_file(
             else:
                 raise IngestionError(message="Local files are not accepted.")
 
-        # data has unsupported URL scheme or format
-        else:
-            raise IngestionError(message=f"Unsupported data source format: {data_item_path}")
-
     # data is not a supported type
     raise IngestionError(message=f"Data type not supported: {type(data_item_path)}")


cognee/tasks/ingestion/data_item_to_text_file.py
```diff
@@ -75,5 +75,16 @@ async def data_item_to_text_file(
             else:
                 raise IngestionError(message="Local files are not accepted.")
 
+        # data is a relative file path (e.g., "file.json", "data/file.txt")
+        else:
```
lxobr: Right now this `else` catches anything not handled above (including e.g. `http://`, `file://`).
Maybe we make this explicitly about relative paths like this:
`elif parsed_url.scheme == "" and not Path(data_item_path).is_absolute():`

Vasilije1990: Resolved

cognee/tasks/ingestion/data_item_to_text_file.py
```diff
@@ -75,5 +75,20 @@ async def data_item_to_text_file(
             else:
                 raise IngestionError(message="Local files are not accepted.")
 
+        # data is a relative file path (e.g., "file.json", "data/file.txt")
+        elif parsed_url.scheme == "" and not Path(data_item_path).is_absolute():
```
dexters1: This is not the right place to add this handler for relative paths. It should be added here:

save_data_item_to_storage

dexters1: I've pushed the fix to have it handle relative paths from save_data_item_to_storage

PR #1341, opened 2025-09-08
feat: implement combined context search

borisarzentar:
## DCO Affirmation


cognee/modules/retrieval/graph_completion_context_extension_retriever.py
```diff
@@ -41,15 +41,14 @@ def __init__(
             node_name=node_name,
             save_interaction=save_interaction,
             system_prompt=system_prompt,
-            only_context=only_context,
```
dexters1, line 44: Some of the retrievers like this one have lost the ability to return only_context after these changes, would be good to return this functionality to them. Users have specifically asked for the ability to only see raw data from all search types before sending them to the final LLM call

cognee/modules/retrieval/graph_completion_context_extension_retriever.py
```diff
@@ -117,19 +117,15 @@ async def get_completion(
...
             await self.save_qa(
-                question=query, answer=completion, context=context, triplets=triplets
+                question=query, answer=completion, context=context_text, triplets=triplets
             )
 
-        if self.only_context:
-            return [context]
-        else:
-            return [completion]
+        return completion
```
hajdul88, line 131: Here the same what Igor raised, from now on this can't return context

cognee/modules/retrieval/graph_completion_cot_retriever.py
```diff
@@ -133,10 +135,7 @@ async def get_completion(
...
             await self.save_qa(
-                question=query, answer=completion, context=context, triplets=triplets
+                question=query, answer=completion, context=context_text, triplets=triplets
             )
 
-        if self.only_context:
-            return [context]
-        else:
-            return [completion]
+        return completion
```
hajdul88, line 141: here as well

dexters1:
Do we have a way to remove duplicate information in the context between datasets from combined search? Are we ranking vector similarity results between datasets? What if one dataset has more valuable information compared to other datasets

borisarzentar:
> Do we have a way to remove duplicate information in the context between datasets from combined search? Are we ranking vector similarity results between datasets? What if one dataset has more valuable information compared to other datasets

We don't address those problems here. We just want to have this search enabled now for the visualization. We can later address those and improve the search. Better to have something than nothing.

cognee/modules/retrieval/graph_completion_context_extension_retriever.py
borisarzentar: There is a method for that, get_context. So if you want to use only the context you should call this method. Having get_completion as a proxy doesn't make sense. We also enabled to send external context to get_completion, so retriever doesn't have to get context twice. We handle `only_context` now a bit higher in the chain, so search API stays the same.

PR #1343, opened 2025-09-08
fix: Fix small import issue with fastembed

siillee:
## Description

An exception was not imported correctly, made fastembed not work. Now it works.

## DCO Affirmation


dexters1 (approved):

dexters1:
Does it make sense to add a simple fastembed CI/CD test as well? So this kind of thing can't happen again

siillee:
> Does it make sense to add a simple fastembed CI/CD test as well? So this kind of thing can't happen again

Makes sense, I can add a small test that runs one or two examples/tests with OpenAI LLM and Fastembed Embeddings, to make sure they work.

Issue #1356, opened 2025-09-09
Add OWL/RDF files as direct input


hande-k:
# OWL/RDF File to Knowledge Graph Pipeline

*NOTE: This issue is part of Contribute-to-Win. Please comment first to get assigned. Read the details [here](https://github.com/orgs/topoteretes/discussions/1361)*

## Problem Statement

Cognee currently supports ontology files (OWL/RDF) only as reference schemas for validating entities extracted from text documents through the `ontology_file_path` parameter in `cognee.cognify()`. 
However, there's no direct pipeline to ingest OWL/RDF files as primary data sources, extract their semantic structure, and convert them into searchable knowledge graphs with embeddings.

## Proposed Solution

Create a new, simple pipeline for OWL/RDF files:
1. **Load** - Add OWL/RDF file loader to ingest ontology files
2. **Process** - Extract semantic structure and generate embeddings  
3. **Search** - Enable all existing search types on ontology content

y-sudharshan:
@hande-k, I have opened PR #1360. Please review and let me know any changes

hande-k:
thanks for contributing @y-sudharshan  the team will soon review your PR.

Vattikuti-Manideep-Sitaram:
Hi @hande-k,  
I just wanted to check whether this issue is still open or has been completed.  
If it’s still open, I’d be happy to contribute and work on it. 🙂

hande-k:
hey @Vattikuti-Manideep-Sitaram thank you very much for picking up another issue already! It's assigned to you :)

Vattikuti-Manideep-Sitaram:
Hey hi @hande-k  sorry for the late reply. Was busy with some work just had a chance to look at the issue. Made some progress please check this PR.  Can you please let me know if any revision is needed the following code is tested in local using this following code snippet

PR: 
https://github.com/topoteretes/cognee/pull/1529
```python
import asyncio
from io import BytesIO
from cognee.api.v1.cognify.ontology_graph_pipeline import run_ontology_pipeline
from cognee.api.v1.search import search, SearchType


async def main():
    # Read ontology file in-memory
    with open("sample_ontology.owl", "rb") as f:
        ontology_bytes = BytesIO(f.read())

    print("🚀 Running ontology ingestion pipeline...\n")

    # Run the ontology pipeline directly on file bytes
    async for status in run_ontology_pipeline(ontology_bytes):
[...]

PR #1360, opened 2025-09-10
Feature/owl rdf pipeline

y-sudharshan:
This PR adds a complete pipeline for OWL/RDF ontology files:

Loader: Supports OWL/RDF file ingestion via [OntologyEngine and [OntologyResolver], with error handling.
Process: Extracts semantic structure and generates embeddings for ontology nodes and edges.
Search: Integrates ontology content and embeddings with all existing search types.
Docs: Updates [README.md] with usage instructions and examples.
Tests: Adds unit tests for loader, semantic extraction, embedding generation, and search integration.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,0 +1,251 @@
+## OWL/RDF Ontology Pipeline Integration
+
+Cognee now supports ingesting, processing, and searching OWL/RDF ontology files.
+
+### Usage
+
+1. **Load Ontology**
+  - Use `OntologyEngine.load_data('path/to/ontology.owl')` to ingest OWL/RDF files.
+
+2. **Semantic Extraction & Embeddings**
+  - Ontology nodes and edges are extracted and embeddings are generated automatically.
+
+3. **Search**
+  - All search types (summaries, insights, chunks, code, etc.) are enabled for ontology content and relationships.
+
+4. **Testing & Validation**
+  - Comprehensive unit and integration tests cover loader, semantic extraction, embedding generation, and search operations.
+
+### Example
+
+```python
+from cognee.tasks.graph.infer_data_ontology import OntologyEngine
+
+engine = OntologyEngine()
+ontology_data = await engine.load_data('ontology.owl')
+# ontology_data['nodes'], ontology_data['edges'], ontology_data['embeddings']
+```
+
+Refer to the documentation for advanced configuration and troubleshooting.
+<div align="center">
+  <a href="https://github.com/topoteretes/cognee">
+    <img src="https://raw.githubusercontent.com/topoteretes/cognee/refs/heads/dev/assets/cognee-logo-transparent.png" alt="Cognee Logo" height="60">
+  </a>
+
+  <br />
+
+  cognee - Memory for AI Agents in 5 lines of code
+
+  <p align="center">
+  <a href="https://www.youtube.com/watch?v=1bezuvLwJmw&t=2s">Demo</a>
+  .
+  <a href="https://cognee.ai">Learn more</a>
+  ·
+  <a href="https://discord.gg/NQPKmU5CCg">Join Discord</a>
+  ·
+  <a href="https://www.reddit.com/r/AIMemory/">Join r/AIMemory</a>
+  .
+  <a href="https://docs.cognee.ai/">Docs</a>
+  .
+  <a href="https://github.com/topoteretes/cognee-community">cognee community repo</a>
+  </p>
+
+
+  [![GitHub forks](https://img.shields.io/github/forks/topoteretes/cognee.svg?style=social&label=Fork&maxAge=2592000)](https://GitHub.com/topoteretes/cognee/network/)
+  [![GitHub stars](https://img.shields.io/github/stars/topoteretes/cognee.svg?style=social&label=Star&maxAge=2592000)](https://GitHub.com/topoteretes/cognee/stargazers/)
+  [![GitHub commits](https://badgen.net/github/commits/topoteretes/cognee)](https://GitHub.com/topoteretes/cognee/commit/)
+  [![Github tag](https://badgen.net/github/tag/topoteretes/cognee)](https://github.com/topoteretes/cognee/tags/)
+  [![Downloads](https://static.pepy.tech/badge/cognee)](https://pepy.tech/project/cognee)
+  [![License](https://img.shields.io/github/license/topoteretes/cognee?colorA=00C586&colorB=000000)](https://github.com/topoteretes/cognee/blob/main/LICENSE)
+  [![Contributors](https://img.shields.io/github/contributors/topoteretes/cognee?colorA=00C586&colorB=000000)](https://github.com/topoteretes/cognee/graphs/contributors)
+  <a href="https://github.com/sponsors/topoteretes"><img src="https://img.shields.io/badge/Sponsor-❤️-ff69b4.svg" alt="Sponsor"></a>
+
+<p>
+  <a href="https://www.producthunt.com/posts/cognee?embed=true&utm_source=badge-top-post-badge&utm_medium=badge&utm_souce=badge-cognee" target="_blank" style="display:inline-block; margin-right:10px;">
+    <img src="https://api.producthunt.com/widgets/embed-image/v1/top-post-badge.svg?post_id=946346&theme=light&period=daily&t=1744472480704" alt="cognee - Memory&#0032;for&#0032;AI&#0032;Agents&#0032;&#0032;in&#0032;5&#0032;lines&#0032;of&#0032;code | Product Hunt" width="250" height="54" />
+  </a>
+
+  <a href="https://trendshift.io/repositories/13955" target="_blank" style="display:inline-block;">
+    <img src="https://trendshift.io/api/badge/repositories/13955" alt="topoteretes%2Fcognee | Trendshift" width="250" height="55" />
+  </a>
+</p>
+
+
+
+**🚀 We launched Cogwit beta (Fully-hosted AI Memory): Sign up [here](https://platform.cognee.ai/)! 🚀**
+
+Build dynamic memory for Agents and replace RAG using scalable, modular ECL (Extract, Cognify, Load) pipelines.
+
+More on [use-cases](https://docs.cognee.ai/use-cases) and [evals](https://github.com/topoteretes/cognee/tree/main/evals)
+
+  <p align="center">
+  🌐 Available Languages
+  :
+  <!-- Keep these links. Translations will automatically update with the README. -->
+  <a href="https://www.readme-i18n.com/topoteretes/cognee?lang=de">Deutsch</a> |
+  <a href="https://www.readme-i18n.com/topoteretes/cognee?lang=es">Español</a> |
+  <a href="https://www.readme-i18n.com/topoteretes/cognee?lang=fr">français</a> |
+  <a href="https://www.readme-i18n.com/topoteretes/cognee?lang=ja">日本語</a> |
+  <a href="https://www.readme-i18n.com/topoteretes/cognee?lang=ko">한국어</a> |
+  <a href="https://www.readme-i18n.com/topoteretes/cognee?lang=pt">Português</a> |
+  <a href="https://www.readme-i18n.com/topoteretes/cognee?lang=ru">Русский</a> |
+  <a href="https://www.readme-i18n.com/topoteretes/cognee?lang=zh">中文</a>
+  </p>
+
+
+<div style="text-align: center">
+  <img src="https://raw.githubusercontent.com/topoteretes/cognee/refs/heads/main/assets/cognee_benefits.png" alt="Why cognee?" width="50%" />
+</div>
+</div>
+
+
+## Features
+
+- Interconnect and retrieve your past conversations, documents, images and audio transcriptions
+- Replaces RAG systems and reduces developer effort, and cost.
+- Load data to graph and vector databases using only Pydantic
+- Manipulate your data while ingesting from 30+ data sources
+
+## Get Started
+
+Get started quickly with a Google Colab  <a href="https://colab.research.google.com/drive/1jHbWVypDgCLwjE71GSXhRL3YxYhCZzG1?usp=sharing">notebook</a> , <a href="https://deepnote.com/workspace/cognee-382213d0-0444-4c89-8265-13770e333c02/project/cognee-demo-78ffacb9-5832-4611-bb1a-560386068b30/notebook/Notebook-1-75b24cda566d4c24ab348f7150792601?utm_source=share-modal&utm_medium=product-shared-content&utm_campaign=notebook&utm_content=78ffacb9-5832-4611-bb1a-560386068b30">Deepnote notebook</a> or  <a href="https://github.com/topoteretes/cognee/tree/main/cognee-starter-kit">starter repo</a>
+
+
+
+
+## Contributing
+Your contributions are at the core of making this a true open source project. Any contributions you make are **greatly appreciated**. See [`CONTRIBUTING.md`](CONTRIBUTING.md) for more information.
+
+
+
+
+
+## 📦 Installation
+
+You can install Cognee using either **pip**, **poetry**, **uv** or any other python package manager.
+
+Cognee supports Python 3.10 to 3.13
+
+### With pip
+
+```bash
+pip install cognee
+```
+
+## Local Cognee installation
+
+You can install the local Cognee repo using **uv**, **pip** and **poetry**.
+For local pip installation please make sure your pip version is above version 21.3.
+
+### with UV with all optional dependencies
+
+```bash
+uv sync --all-extras
+```
+
+## 💻 Basic Usage
+
+### Setup
+
+```
+import os
+os.environ["LLM_API_KEY"] = "YOUR OPENAI_API_KEY"
+
+```
+
+You can also set the variables by creating .env file, using our <a href="https://github.com/topoteretes/cognee/blob/main/.env.template">template.</a>
+To use different LLM providers, for more info check out our <a href="https://docs.cognee.ai">documentation</a>
+
+
+### Simple example
+
+This script will run the default pipeline:
+
+```python
+import cognee
+import asyncio
+
+
+async def main():
+    # Add text to cognee
+    await cognee.add("Natural language processing (NLP) is an interdisciplinary subfield of computer science and information retrieval.")
+
+    # Generate the knowledge graph
+    await cognee.cognify()
+
+    # Query the knowledge graph
+    results = await cognee.search("Tell me about NLP")
+
+    # Display the results
+    for result in results:
+        print(result)
+
+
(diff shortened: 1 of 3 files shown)


y-sudharshan:
Closes #1356

cognee/tasks/graph/infer_data_ontology.py
```diff
@@ -162,6 +160,49 @@ async def load_data(self, file_path: str) -> Union[List[Dict[str, Any]], Dict[st
...
+                llm = LLMGateway()
+                for category in ["classes", "individuals"]:
+                    for key, uri in resolver.lookup.get(category, {}).items():
+                        node_info = {"id": key, "uri": str(uri), "category": category}
+                        # Semantic extraction: get label and description if available
+                        node_info["label"] = key
+                        node_info["description"] = str(uri)
+                        # Generate embedding for node
+                        try:
+                            embedding = llm.generate_embedding(
```
hajdul88: There is no generate embedding method in LLMGateway, therefore embeddings are not generated and they always get None. 

In order to reach the similar logic as in cognify, you have to create DataPoints (similar to Entity or EntityType), parse ontology nodes into these datapoints and call the add_data_points method.

cognee/tasks/graph/infer_data_ontology.py
```diff
@@ -162,6 +160,49 @@ async def load_data(self, file_path: str) -> Union[List[Dict[str, Any]], Dict[st
...
+                            edge_embedding = llm.generate_embedding(text=edge[1])
+                        except Exception:
+                            edge_embedding = None
+                        edge_info["embedding"] = edge_embedding
+                        edges.append(edge_info)
+                # Store ontology data for search integration
+                self.ontology_nodes = nodes
+                self.ontology_edges = edges
+                self.ontology_embeddings = embeddings
+                return {"nodes": nodes, "edges": edges, "embeddings": embeddings}
```
hajdul88: This shouldn't be a separate task, and it does not make sense to return these. Instead you can call add_data_points once you have your datapoints objects created from Ontology nodes.

cognee/tasks/graph/infer_data_ontology.py
```diff
@@ -162,6 +160,49 @@ async def load_data(self, file_path: str) -> Union[List[Dict[str, Any]], Dict[st
...
+                        node_info = {"id": key, "uri": str(uri), "category": category}
+                        # Semantic extraction: get label and description if available
+                        node_info["label"] = key
+                        node_info["description"] = str(uri)
+                        # Generate embedding for node
+                        try:
+                            embedding = llm.generate_embedding(
+                                text=node_info["label"] + " " + node_info["description"]
+                            )
+                        except Exception:
```
hajdul88: Once the embedding is fixed, it is worth to do some batching for the embedding api calls, The current version would be calling embedding api separately for each node

cognee/tasks/graph/infer_data_ontology.py
hajdul88: Also you don't have to build this dict, as when the datapoint objects are ready to be ingested, you call add_data_points and thats it.

cognee/tasks/graph/infer_data_ontology.py
```diff
@@ -162,6 +160,49 @@ async def load_data(self, file_path: str) -> Union[List[Dict[str, Any]], Dict[st
...
+                        nodes.append(node_info)
+                for node in nodes:
+                    _, node_edges, _ = resolver.get_subgraph(
+                        node_name=node["id"], node_type=node["category"]
+                    )
+                    for edge in node_edges:
+                        edge_info = {"source": edge[0], "relation": edge[1], "target": edge[2]}
+                        # Generate embedding for edge relation
+                        try:
+                            edge_embedding = llm.generate_embedding(text=edge[1])
```
hajdul88: Same here, there is no generate embeddings. add_data_points should get the edges, there is a update_edge_collection boolean parameter, if its true, you get the edges from the graph and embeed them.

cognee/tasks/graph/infer_data_ontology.py
```diff
@@ -139,16 +139,14 @@ async def recursive_flatten(
 
     async def load_data(self, file_path: str) -> Union[List[Dict[str, Any]], Dict[str, Any]]:
```
hajdul88, line 140: Instead of extending the logic here, it would be better to create your own pipeline that is only doing this.

README.md
```diff
@@ -1,3 +1,32 @@
+## OWL/RDF Ontology Pipeline Integration
```
hajdul88, line 1: The readme doesn't have to be changed in this case, this functionality when it works will be included in our docs

cognee/tests/unit/modules/ontology/test_ontology_adapter.py
```diff
@@ -1,8 +1,73 @@
...
+            <owl:NamedIndividual rdf:about="http://example.org/test#Audi">
+                <rdf:type rdf:resource="http://example.org/test#Car"/>
+            </owl:NamedIndividual>
+        </rdf:RDF>"""
+    owl_file = tmp_path / "test.owl"
+    owl_file.write_text(owl_content)
+
+    engine = OntologyEngine()
+    asyncio.run(engine.load_data(str(owl_file)))
+    assert hasattr(engine, "ontology_nodes")
```
hajdul88: This doesn't test any content, just the dict structure. The name of the test would imply that its testing the search but this doesn't happen

cognee/tests/unit/modules/ontology/test_ontology_adapter.py
```diff
@@ -1,8 +1,73 @@
...
+                <rdf:type rdf:resource="http://example.org/test#Car"/>
+            </owl:NamedIndividual>
+        </rdf:RDF>"""
+    owl_file = tmp_path / "test.owl"
+    owl_file.write_text(owl_content)
+
+    engine = OntologyEngine()
+    data = asyncio.run(engine.load_data(str(owl_file)))
+    for node in data["nodes"]:
+        assert "embedding" in node
```
hajdul88: This is again testing the structure. If embeddings are None this passes.

cognee/tasks/graph/infer_data_ontology.py
```diff
@@ -162,6 +160,49 @@ async def load_data(self, file_path: str) -> Union[List[Dict[str, Any]], Dict[st
...
+                        node_info["label"] = key
+                        node_info["description"] = str(uri)
+                        # Generate embedding for node
+                        try:
+                            embedding = llm.generate_embedding(
+                                text=node_info["label"] + " " + node_info["description"]
+                            )
+                        except Exception:
+                            embedding = None
+                        node_info["embedding"] = embedding
```
hajdul88: The best is if you check how we ingest in cognify. The last task is add_datapoints if you put a break point to the beginning of add datapoints you will see what is the object that we pass there. 

Basically if you create a class that inherits datapoint (check cognee/modules/engine/models/Entity), you parse the ontology nodes there (and you add metadata: dict = {"index_fields": ["name"]})) (this means we want to embed the name property), then you call add_data_points with the created objects, the embeddings etc get created automatically.

hajdul88 (changes requested):
Hi @y-sudharshan!

Thanks for the contribution! I’ve fixed the linting and Ruff issues and changed the target branch to dev.

Regarding the code, I’ve left comments on sections that need to be rewritten. Right now the embeddings don’t work because the code calls a method that doesn’t exist. Also, nodes should be treated as datapoints and ingested via add_datapoints (see my comments).

This is promising work and would be a valuable addition. Please update the PR based on the feedback, and feel free to reach out if you have any questions or need more guidance. Keep up the great work!

cognee/tasks/graph/infer_data_ontology.py
y-sudharshan: @hajdul88 thanks for the review, will get back to you with the requested changes

cognee/modules/pipelines/operations/ontology_pipeline.py
```diff
@@ -0,0 +1,29 @@
...
+        self.resolver = OntologyResolver(ontology_file=ontology_file)
+        self.nodes = []
+        self.edges = []
+
+    async def ingest(self):
+        datapoints = []
+        for category in ["classes", "individuals"]:
+            for key, uri in self.resolver.lookup.get(category, {}).items():
+                datapoints.append(
+                    Entity(
```
hajdul88: Individuals are usually Entites, and Classes are Entity types. In this way everything will be ingested as Entity. Could you change this so that it gets the proper node types? Thanks!

cognee/infrastructure/llm/LLMGateway.py
```diff
@@ -4,6 +4,18 @@
 
 
 class LLMGateway:
+    @staticmethod
+    def generate_embedding(text: str) -> Optional[list]:
```
hajdul88: You don't need this. If you check we already have the embedding logic in the add_datapoints. If you use the add datapoint with the created Entities, will have its own vector collection called Entity_name.

cognee/tasks/graph/infer_data_ontology.py
```diff
@@ -162,6 +160,37 @@ async def load_data(self, file_path: str) -> Union[List[Dict[str, Any]], Dict[st
...
+                            description=str(uri),
+                            metadata={"category": category, "uri": str(uri)}
+                        )
+                        datapoints.append(datapoint)
+                        embeddings[str(uri)] = embedding
+                await add_data_points(datapoints, update_edge_collection=True)
+                self.ontology_nodes = [dp.dict() for dp in datapoints]
+                self.ontology_edges = []  # Edges handled by add_data_points
+                self.ontology_embeddings = embeddings
+                return {"nodes": self.ontology_nodes, "edges": self.ontology_edges, "embeddings": embeddings}
```
hajdul88: As I mentioned in my previous comment this task doesn't need to return this structure. The add_datapoints is saving and ingesting the nodes and the vector embeddings into the corresponding databases

cognee/tests/unit/modules/ontology/test_ontology_adapter.py
hajdul88: This comment still applies. This test is checking the structure but not the content and definitely not the search as the name would imply.

cognee/tests/unit/modules/ontology/test_ontology_adapter.py
```diff
@@ -1,8 +1,117 @@
...
+            </owl:NamedIndividual>
+        </rdf:RDF>"""
+    owl_file = tmp_path / "test.owl"
+    owl_file.write_text(owl_content)
+
+    engine = OntologyEngine()
+    data = asyncio.run(engine.load_data(str(owl_file)))
+    assert "nodes" in data
+    assert "edges" in data
+    assert "embeddings" in data
```
hajdul88: We dont need to create this structure. INstead you can wrrite a test that is checking if graph db has all the nodes and edges that are needed

and

collections with the correct created ontology related embeddings

cognee/tasks/graph/infer_data_ontology.py
```diff
@@ -162,6 +160,37 @@ async def load_data(self, file_path: str) -> Union[List[Dict[str, Any]], Dict[st
...
+                from cognee.modules.engine.models.Entity import Entity
+                from cognee.tasks.storage.add_data_points import add_data_points
+                from cognee.infrastructure.llm.LLMGateway import LLMGateway
+
+                resolver = OntologyResolver(ontology_file=file_path)
+                datapoints = []
+                embeddings = {}
+                embedding_fn = getattr(LLMGateway, "generate_embedding", None)
+                for category in ["classes", "individuals"]:
+                    for key, uri in resolver.lookup.get(category, {}).items():
```
hajdul88: As I mentioned in the previous review it is worth adding some batching here. This is gonna call the embedding api one by one

README.md
```diff
@@ -1,222 +0,0 @@
-<div align="center">
```
hajdul88, line 1: Please don't delete our README

cognee/tasks/graph/infer_data_ontology.py
```diff
@@ -162,6 +160,41 @@ async def load_data(self, file_path: str) -> Union[List[Dict[str, Any]], Dict[st
                     content = await f.read()
                     reader = csv.DictReader(content.splitlines())
                     return list(reader)
+            elif file_path.endswith(".owl") or file_path.endswith(".rdf"):
```
hajdul88, line 163: Now this whole logic is duplicated, you have it here in the infer_data_ontology and you also have this in the ontology_pipeline.py.

Please keep only the ontology pipeline, you do not have to change this infer_data_ontology file. Also the two logic is slightly different, here you implemented batching in the ontology pipeline you didn't.

cognee/tasks/graph/infer_data_ontology.py
```diff
@@ -162,6 +160,41 @@ async def load_data(self, file_path: str) -> Union[List[Dict[str, Any]], Dict[st
...
+                        description=str(uri),
+                        metadata={"index_fields": ["name"], "category": "individual", "uri": str(uri)}
+                    )
+                    datapoints.append(entity)
+                    texts_to_embed.append(f"{key} {str(uri)}")
+                # Batch embedding API placeholder
+                # Example: embeddings = await embedding_api.embed_batch(texts_to_embed)
+                # for dp, emb in zip(datapoints, embeddings):
+                #     dp.embedding = emb
+                await add_data_points(datapoints, update_edge_collection=True)
```
hajdul88: In this case you also do not create edges. please check this thread I prepared a detailed description on what and how should be done.

https://discord.com/channels/1120795297094832337/1417530176790724669


In cognee edges are added in two different ways. 

A:If datapoint contains another datapoint as property, add_datapoints will create an edge
B:you add edges manually with the add_edges methos. for a reference check extract_graph_from_data.py integrate_chunk_graphs method line 64

    if len(graph_edges) > 0:
        await graph_engine.add_edges(graph_edges)

cognee/modules/pipelines/operations/ontology_pipeline.py
```diff
@@ -0,0 +1,39 @@
...
+                Entity(
+                    name=key,
+                    description=str(uri),
+                    metadata={"category": "individual", "uri": str(uri)}
+                )
+            )
+        await add_data_points(datapoints, update_edge_collection=True)
+        self.nodes = [dp.dict() for dp in datapoints]
+        self.edges = []  # Edges handled by add_data_points
+        return {"nodes": self.nodes, "edges": self.edges}
```
hajdul88: I already raised this you dont need to return anything here

cognee/tests/unit/modules/ontology/test_ontology_adapter.py
```diff
@@ -1,8 +1,133 @@
...
+            <owl:NamedIndividual rdf:about="http://example.org/test#Audi">
+                <rdf:type rdf:resource="http://example.org/test#Car"/>
+            </owl:NamedIndividual>
+        </rdf:RDF>"""
+    owl_file = tmp_path / "test.owl"
+    owl_file.write_text(owl_content)
+
+    engine = OntologyEngine()
+    asyncio.run(engine.load_data(str(owl_file)))
+    resolver = OntologyResolver(str(owl_file))
```
hajdul88, line 124: This logic will change as we abstract the ontology adapter. The change is gonna be merged to dev in these days and can be found here: https://github.com/topoteretes/cognee/pull/1429

hande-k:
hey @y-sudharshan thanks a lot for your time and effort in this PR! please let us know if you have any questions - we are looking forward to the next iteration :)

README.md
hajdul88: Again, this was already raised in the first review and was never addressed. Please dont add anything to the readme

inspect_vector_db.py
```diff
@@ -0,0 +1,12 @@
+import asyncio
+from cognee.infrastructure.databases.vector import get_vector_engine
+
+async def print_vector_collections():
+    vector_engine = get_vector_engine()
+    collection_names = await vector_engine.get_collection_names()
+    print("Vector DB Collections:")
+    for name in collection_names:
```
hajdul88, line 8: What is the point of this method?

inspect_graph_db.py
```diff
@@ -0,0 +1,15 @@
+import asyncio
+from cognee.infrastructure.databases.graph import get_graph_engine
+
+async def print_graph():
+    graph_engine = await get_graph_engine()
+    nodes, edges = await graph_engine.get_graph_data()
+    print("Nodes:")
+    for node in nodes:
+        print(node)
```
hajdul88, line 9: What is the point of this method?

cognee/tasks/graph/infer_data_ontology.py
```diff
@@ -230,20 +233,22 @@ async def add_graph_ontology(self, file_path: str = None, documents: list = None
...
                     generate_node_id(edge.target_id),
                     edge.relationship_type,
-                    dict(
-                        source_node_id=generate_node_id(edge.source_id),
-                        target_node_id=generate_node_id(edge.target_id),
-                        relationship_name=edge.relationship_type,
-                        updated_at=datetime.now(timezone.utc).strftime("%Y-%m-%d %H:%M:%S"),
-                    ),
+                    {
+                        "source_node_id": generate_node_id(edge.source_id),
```
hajdul88, line 242: Why are you changing this? This has nothing to do with the PR or the ontology part

cognee/tasks/graph/infer_data_ontology.py
hajdul88: This wasn't addressed. You are still duplicating the logic. Also as I mentioned please don't include the logic in this file

cognee/modules/pipelines/operations/ontology_pipeline.py
```diff
@@ -0,0 +1,49 @@
...
+            datapoints.append(
+                Entity(
+                    name=key,
+                    description=str(uri),
+                    metadata={"category": "individual", "uri": str(uri)}
+                )
+            )
+            texts_to_embed.append(f"{key} {str(uri)}")
+        # Batch embedding
+        if texts_to_embed:
```
hajdul88, line 41: This is not ingesting the edges at all. I already raised this problem in my previous 2 reviews and wasn't addressed.

cognee/modules/pipelines/operations/ontology_pipeline.py
```diff
@@ -0,0 +1,49 @@
...
+from cognee.tasks.storage.add_data_points import add_data_points
+from cognee.infrastructure.databases.vector.get_vector_engine import get_vector_engine
+
+class OntologyPipeline:
+    """
+    Dedicated pipeline for OWL/RDF ontology ingestion and processing.
+    """
+    def __init__(self, ontology_file: str):
+        self.ontology_file = ontology_file
+        self.resolver = OntologyResolver(ontology_file=ontology_file)
```
hajdul88, line 13: Not implementing the changed ontology handling

PR #1364, opened 2025-09-10
Aws session token support - MSR97

dexters1:
## Description
PR from contributor MSR97

## DCO Affirmation


cognee/infrastructure/files/storage/S3FileStorage.py
```diff
@@ -21,10 +21,15 @@ class S3FileStorage(Storage):
     def __init__(self, storage_path: str):
         self.storage_path = storage_path
         s3_config = get_s3_config()
-        if s3_config.aws_access_key_id is not None and s3_config.aws_secret_access_key is not None:
+        if (
+            s3_config.aws_access_key_id is not None
+            and s3_config.aws_secret_access_key is not None
+            and s3_config.aws_session_token is not None
```
dexters1: token can be None here, it's not mandatory for S3 connection

dexters1:
PR needs to be tested if it will work with S3 when token is not provided

Vasilije1990:
@siillee can you test this?

dexters1:
> @siillee can you test this?

I think I tested it then after the comment and it was fine, but didn't want to merge it since we were about to do a release

PR #1450, opened 2025-09-23
refactor: Rework limit=0 for vector adapters

siillee:
## Description

Until now, limit=0 in vector search meant that there is no limit and we should return everything. This caused confusion and errors, so now it is reworked so that limit=None means no limit on the search. If someone puts limit=0, there will be no results returned, as it makes more sense and is less error prone.

## Type of Change

- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [ ] Other (please specify):

## Changes Made

- 
- 
- 

## Pre-submission Checklist

## DCO Affirmation


cognee/tests/unit/infrastructure/databases/vector/data/alice_in_wonderland.txt
```diff
@@ -0,0 +1,1772 @@
+CHAPTER I.
+Down the Rabbit-Hole
```
dexters1: Can we reuse the Alice example from the examples data folder? So we don't store two copies

dexters1: Actually let's not use this file at all for tests as it's big and expensive to process. Use something small like the NLP text from the test_data folder in the tests directory

siillee: I didn't want to make a nasty path to the Alice in examples, so I just put it closer, and was gonna ask about that. 

And okay, I'll use some smaller example, although I wanted a bigger one just to see if it will return all results when there is more than a couple of rows. 
Also, Alice is not the greatest example since it fails with Azure OpenAI because Azure flags it as inappropriate content. I forgot about this, and saw that tests failed on CI :)

EDIT: The extra copy of Alice is actually a mistake, I changed up the tests while working on this, so I forgot to delete it, it is actually not used anywhere, my bad.

Vasilije1990 (approved):

PR #1504, opened 2025-10-06
feat: Redis lock integration and Kuzu agentic access fix

hajdul88:
## Description
This PR introduces a shared locked mechanism in KuzuAdapter to avoid use case when multiple subprocesses from different environments are trying to use the same Kuzu adatabase.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [x] Performance improvement
- [ ] Other (please specify):

## Screenshots/Videos (if applicable)
None

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/tests/test_concurrent_subprocess_access.py
+++ b/cognee/tests/test_concurrent_subprocess_access.py
@@ -6,8 +6,16 @@ import subprocess
 
 from cognee.shared.logging_utils import get_logger
 
-
 logger = get_logger()
+
+"""
+Test: Redis-based Kùzu Locking Across Subprocesses
+
+This test ensures the Redis shared lock correctly serializes access to the Kùzu
+database when multiple subprocesses (writer/reader and cognify tasks) run in parallel.
+If this test fails, it indicates the locking mechanism is not properly handling
+concurrent subprocess access.
+"""
 
 
 async def concurrent_subprocess_access():


cognee/infrastructure/databases/graph/kuzu/adapter.py
```diff
@@ -225,12 +242,47 @@ def blocking_query():
...
+        if self.connection:
+            del self.connection
+            self.connection = None
+        if self.db:
+            del self.db
+            self.db = None
+        self._is_closed = True
+        logger.info("Kuzu database closed successfully")
+
+    def reopen(self):
```
borisarzentar, line 281: Why do we need `reopen`?

cognee/infrastructure/databases/cache/redis/RedisAdapter.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+class RedisAdapter(CacheDBInterface):
+    def __init__(self, host, port, lock_name, timeout=240, blocking_timeout=300):
+        super().__init__(host, port, lock_name)
+        self.redis = redis.Redis(host=host, port=port)
+        self.timeout = timeout
+        self.blocking_timeout = blocking_timeout
+
+    def acquire(self):
```
borisarzentar: Should we be more specific that it is lock, like `acquire_lock`?

cognee/infrastructure/databases/graph/kuzu/adapter.py
```diff
@@ -23,9 +23,14 @@
 from cognee.modules.storage.utils import JSONEncoder
 from cognee.modules.engine.utils.generate_timestamp_datapoint import date_to_int
 from cognee.tasks.temporal_graph.models import Timestamp
+from cognee.infrastructure.databases.cache.config import get_cache_config
 
 logger = get_logger()
 
+cache_config = get_cache_config()
+if cache_config.shared_kuzu_lock:
+    from cognee.infrastructure.databases.cache.get_cache_engine import get_cache_engine
```
borisarzentar, line 32: Should we import it down where we use it?

cognee/infrastructure/databases/graph/kuzu/adapter.py
hajdul88: Because otherwise we dont release the database connection and other subprocesses cannot access the kuzu file

cognee/infrastructure/databases/cache/redis/RedisAdapter.py
hajdul88: we can rename it, will do that

cognee/infrastructure/databases/graph/kuzu/adapter.py
hajdul88: yes we can, however we use it in query + as instance variable as well where I didn't want to have this because its often used + in different places. So I would leave it here, the config is for the whole adapter.

cognee/infrastructure/databases/cache/redis/RedisAdapter.py
hajdul88: Done

cognee/infrastructure/databases/graph/kuzu/adapter.py
hajdul88: discussed

cognee/tests/test_concurrent_subprocess_access.py
```diff
@@ -0,0 +1,76 @@
...
+import pathlib
+import subprocess
+
+from cognee.shared.logging_utils import get_logger
+
+
+logger = get_logger()
+
+
+async def concurrent_subprocess_access():
```
pazone, line 21: QQ: DIUC that this test runs 2 parallel subprocesses that print some text. In what case this test will fail? Is there the results validation? What is the best way to run this test locally?

hajdul88: If we run the same with default cognee without locks or with a wrong lock implementaton, this test fails because the two subprocesses are trying to access the same file. So the validation is that this doesn't fail. There is no result to validate as without this the whole run would fail. With the added setting (Caching=True, Shared_kuzu_lock=True) this should run as it is. Basically this setting allows cognee to set double locking mechanism on the kuzu connections.

Best way would be to run this test without and with the environmental variables that are added here in config. If you want to reproduce or test lmk I can help

pazone (approved):
LGTM from testing and CI perspective

cognee/tests/test_concurrent_subprocess_access.py
pazone: Thanks 👍

pazone: I'd add a short doc at the beginning if the test. Just to quickly understand what is this test about when it failed.

hajdul88: Done

borisarzentar (dismissed):

PR #1512, opened 2025-10-07
Feat/improve exception raising for better debugging

daukadolt:
## Description

When raising exceptions, in some places we lose the stack trace, or in some cases when we raise custom exceptions - we also lose any metadata from original exception that could help pinpoint the issue.

This PR updates exception throwing to use [`raise ... from`](https://docs.python.org/3/reference/simple_stmts.html#raise), which:
1. preserves stack trace
2. when we throw custom exception, indicates that our exception was caused by XYZ exception

_this is a scoped down version for upcoming release, rest of the changes are in #1518_

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/.github/workflows/test_s3_file_storage.yml
+++ b/.github/workflows/test_s3_file_storage.yml
@@ -6,6 +6,10 @@ on:
 permissions:
   contents: read
 
+env:
+  RUNTIME__LOG_LEVEL: ERROR
+  ENV: 'test'
+
 jobs:
   test-gemini:
     name: Run S3 File Storage Test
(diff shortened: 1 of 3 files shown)


.github/workflows/test_s3_file_storage.yml
```diff
@@ -6,6 +6,10 @@ on:
 permissions:
   contents: read
 
+env:
+  RUNTIME__LOG_LEVEL: ERROR
+  ENV: 'test'
```
dexters1: why is the ENV 'test'? shouldn't it be 'dev'?

daukadolt: it's dev everywhere, but it's a good question.

I had the opposite question - why do we use dev everywhere in CI?

Either is fine with me, but the reason I added it is `send_telemetry` was being run in S3 test.

dexters1 (approved):

PR #1534, opened 2025-10-12
feat: implement full delete feature

borisarzentar:
## Type of Change

- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

---

> [!NOTE]
> Adds coverage to ensure deletion helpers only remove targeted records.
> 
> - Adds tests for `delete_data_related_nodes` and `delete_data_related_edges` to confirm only entries with matching `data_id` are deleted
> - Adds tests for `delete_dataset_related_nodes` and `delete_dataset_related_edges` to confirm only entries with matching `dataset_id` are deleted
> - Verifies no-op behavior when tables are empty
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit fae9869a3f800e9b434fc62797b46e843bea5d66. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

## Summary by CodeRabbit

* **Tests**
  * Added unit tests for graph data deletion operations, verifying correct removal of target data-related nodes and edges while preserving unrelated data.
  * Added test coverage for dataset-specific deletion operations.
  * Added tests for edge case scenarios including operations on empty tables.

### DIFF
--- a/cognee/tests/unit/graph/test_delete_data_related_nodes_edges.py
+++ b/cognee/tests/unit/graph/test_delete_data_related_nodes_edges.py
@@ -1,0 +1,161 @@
+import uuid
+import pytest
+from sqlalchemy import select, func
+
+from cognee.infrastructure.databases.relational import (
+    create_db_and_tables,
+    get_async_session,
+)
+from cognee.modules.graph.models import Node, Edge
+from cognee.modules.graph.methods import (
+    delete_data_related_nodes,
+    delete_data_related_edges,
+)
+
+
+@pytest.mark.asyncio
+async def test_delete_data_related_nodes_removes_only_target_data():
+    await create_db_and_tables()
+
+    dataset_id = uuid.uuid4()
+    user_id = uuid.uuid4()
+    data_a = uuid.uuid4()
+    data_b = uuid.uuid4()
+
+    node_a1 = Node(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_a,
+        dataset_id=dataset_id,
+        label="A1",
+        type="TypeA",
+        indexed_fields=["text"],
+        attributes={"k": "v"},
+    )
+    node_a2 = Node(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_a,
+        dataset_id=dataset_id,
+        label="A2",
+        type="TypeA",
+        indexed_fields=["text"],
+        attributes={"k2": "v2"},
+    )
+    node_b1 = Node(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_b,
+        dataset_id=dataset_id,
+        label="B1",
+        type="TypeB",
+        indexed_fields=["text"],
+        attributes={"k3": "v3"},
+    )
+
+    async with get_async_session(auto_commit=True) as session:
+        session.add_all([node_a1, node_a2, node_b1])
+
+    await delete_data_related_nodes(data_a)
+
+    async with get_async_session() as session:
+        count_a = (
+            await session.scalar(select(func.count()).select_from(Node).where(Node.data_id == data_a))
+        )
+        count_b = (
+            await session.scalar(select(func.count()).select_from(Node).where(Node.data_id == data_b))
+        )
+
+        assert count_a == 0
+        assert count_b == 1
+
+
+@pytest.mark.asyncio
+async def test_delete_data_related_edges_removes_only_target_data():
+    await create_db_and_tables()
+
+    dataset_id = uuid.uuid4()
+    user_id = uuid.uuid4()
+    data_a = uuid.uuid4()
+    data_b = uuid.uuid4()
+
+    # Nodes for referencing
+    n1 = Node(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_a,
+        dataset_id=dataset_id,
+        label="N1",
+        type="TypeA",
+        indexed_fields=["text"],
+        attributes={},
+    )
+    n2 = Node(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_b,
+        dataset_id=dataset_id,
+        label="N2",
+        type="TypeB",
+        indexed_fields=["text"],
+        attributes={},
+    )
+
+    e_a = Edge(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_a,
+        dataset_id=dataset_id,
+        source_node_id=n1.id,
+        destination_node_id=n1.id,
+        relationship_name="REL_A",
+        label="LA",
+        attributes={},
+    )
+    e_b = Edge(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_b,
+        dataset_id=dataset_id,
+        source_node_id=n2.id,
+        destination_node_id=n2.id,
+        relationship_name="REL_B",
+        label="LB",
+        attributes={},
+    )
+
+    async with get_async_session(auto_commit=True) as session:
+        session.add_all([n1, n2, e_a, e_b])
+
+    await delete_data_related_edges(data_a)
+
+    async with get_async_session() as session:
+        count_a = (
+            await session.scalar(select(func.count()).select_from(Edge).where(Edge.data_id == data_a))
+        )
+        count_b = (
+            await session.scalar(select(func.count()).select_from(Edge).where(Edge.data_id == data_b))
+        )
+
+        assert count_a == 0
+        assert count_b == 1
+
+
+@pytest.mark.asyncio
+async def test_delete_data_related_nodes_edges_noop_on_empty_tables():
+    await create_db_and_tables()
+
+    data_x = uuid.uuid4()
+
+    # Should not raise
+    await delete_data_related_nodes(data_x)
+    await delete_data_related_edges(data_x)
+
+
--- a/cognee/tests/unit/graph/test_delete_dataset_related_nodes_edges.py
+++ b/cognee/tests/unit/graph/test_delete_dataset_related_nodes_edges.py
@@ -1,0 +1,169 @@
+import uuid
+import pytest
+from sqlalchemy import select, func
+
+from cognee.infrastructure.databases.relational import (
+    create_db_and_tables,
+    get_async_session,
+)
+from cognee.modules.graph.models import Node, Edge
+from cognee.modules.graph.methods import (
+    delete_dataset_related_nodes,
+    delete_dataset_related_edges,
+)
+
+
+@pytest.mark.asyncio
+async def test_delete_dataset_related_nodes_removes_only_target_dataset():
+    await create_db_and_tables()
+
+    dataset_a = uuid.uuid4()
+    dataset_b = uuid.uuid4()
+    user_id = uuid.uuid4()
+    data_id = uuid.uuid4()
+
+    node_a1 = Node(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_id,
+        dataset_id=dataset_a,
+        label="A1",
+        type="TypeA",
+        indexed_fields=["text"],
+        attributes={"k": "v"},
+    )
+    node_a2 = Node(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_id,
+        dataset_id=dataset_a,
+        label="A2",
+        type="TypeA",
+        indexed_fields=["text"],
+        attributes={"k2": "v2"},
+    )
+    node_b1 = Node(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_id,
+        dataset_id=dataset_b,
+        label="B1",
+        type="TypeB",
+        indexed_fields=["text"],
+        attributes={"k3": "v3"},
+    )
+
+    async with get_async_session(auto_commit=True) as session:
+        session.add_all([node_a1, node_a2, node_b1])
+
+    await delete_dataset_related_nodes(dataset_a)
+
+    async with get_async_session() as session:
+        count_a = (
+            await session.scalar(
+                select(func.count()).select_from(Node).where(Node.dataset_id == dataset_a)
+            )
+        )
+        count_b = (
+            await session.scalar(
+                select(func.count()).select_from(Node).where(Node.dataset_id == dataset_b)
+            )
+        )
+
+        assert count_a == 0
+        assert count_b == 1
+
+
+@pytest.mark.asyncio
+async def test_delete_dataset_related_edges_removes_only_target_dataset():
+    await create_db_and_tables()
+
+    dataset_a = uuid.uuid4()
+    dataset_b = uuid.uuid4()
+    user_id = uuid.uuid4()
+    data_id = uuid.uuid4()
+
+    # Create two nodes to reference in edges
+    n1 = Node(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_id,
+        dataset_id=dataset_a,
+        label="N1",
+        type="TypeA",
+        indexed_fields=["text"],
+        attributes={},
+    )
+    n2 = Node(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_id,
+        dataset_id=dataset_b,
+        label="N2",
+        type="TypeB",
+        indexed_fields=["text"],
+        attributes={},
+    )
+
+    e_a = Edge(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_id,
+        dataset_id=dataset_a,
+        source_node_id=n1.id,
+        destination_node_id=n1.id,
+        relationship_name="REL_A",
+        label="LA",
+        attributes={},
+    )
+    e_b = Edge(
+        id=uuid.uuid4(),
+        slug=uuid.uuid4(),
+        user_id=user_id,
+        data_id=data_id,
+        dataset_id=dataset_b,
+        source_node_id=n2.id,
+        destination_node_id=n2.id,
+        relationship_name="REL_B",
+        label="LB",
+        attributes={},
+    )
+
+    async with get_async_session(auto_commit=True) as session:
+        session.add_all([n1, n2, e_a, e_b])
+
+    await delete_dataset_related_edges(dataset_a)
+
+    async with get_async_session() as session:
+        count_a = (
+            await session.scalar(
+                select(func.count()).select_from(Edge).where(Edge.dataset_id == dataset_a)
+            )
+        )
+        count_b = (
+            await session.scalar(
+                select(func.count()).select_from(Edge).where(Edge.dataset_id == dataset_b)
+            )
+        )
+
+        assert count_a == 0
+        assert count_b == 1
+
+
+@pytest.mark.asyncio
+async def test_delete_dataset_related_nodes_edges_noop_on_empty_tables():
+    await create_db_and_tables()
+
+    dataset_x = uuid.uuid4()
+
+    # Should not raise
+    await delete_dataset_related_nodes(dataset_x)
+    await delete_dataset_related_edges(dataset_x)
+
+


cognee/tests/test_delete_default_graph.py
```diff
@@ -0,0 +1,78 @@
...
+
+    vector_engine = get_vector_engine()
+
+    assert not await vector_engine.has_collection("EdgeType_relationship_name")
+    assert not await vector_engine.has_collection("Entity_name")
+
+    await cognee.add(
+        "John works for Apple. He is also affiliated with a non-profit organization called 'Food for Hungry'"
+    )
+    await cognee.add("Marie works for Apple as well. She is a software engineer on MacOS project.")
```
dexters1: Can we mock LLM output / or data so that the graph/vector databases and the ledger will be deterministic?

dexters1:
If user1 (who is the owner of the dataset) adds Data to a dataset can User2 who has delete permission delete delete this Data? In a multi-user setting

cognee/api/v1/cognify/cognify.py
```diff
@@ -208,6 +209,9 @@ class ScientificPaper(DataPoint):
                 "ontology_config": {"ontology_resolver": get_default_ontology_resolver()}
             }
 
+    if user is None:
+        user = await get_default_user()
```
dexters1: What is the reason we are creating default user in the update, add and cognify again?

cognee-starter-kit/src/pipelines/low_level.py
```diff
@@ -221,18 +223,16 @@ async def execute_pipeline() -> None:
     await setup()
 
     # Get user and dataset
-    user = await get_default_user()
-    datasets = await load_or_create_datasets(["demo_dataset"], [], user)
-    dataset_id = datasets[0].id
+    user: User = await get_default_user()  # type: ignore
+    dataset = Dataset(id=uuid4(), name="demo_dataset")
```
dexters1: There is a function to create dataset_id we should use it here because most Cognee features rely on the Dataset ID being generated in that specific way: get_unique_dataset_id is the function name

.github/workflows/test_different_operating_systems.yml
```diff
@@ -174,8 +174,8 @@ jobs:
           cd dist
```
hajdul88, line 182: I will put general comments here:

- examples/low_level/pipeline.py is failing (I think it needs update)
- mcp server delete is failing (probably just param handling)

What we discussed in the meeting:
-Authentication is not handled (Igor will probably write that too)
-custom edges are not handled



Since its changing a lot I would merge this after the release so that we have more time to test + migration is needed for saas. This will break delete for existing users. Maybe we need to support old delete somehow (I dont know what is the solution here).

cognee/modules/pipelines/operations/pipeline.py
```diff
@@ -77,7 +78,7 @@ async def run_pipeline_per_dataset(
         return
 
     pipeline_run = run_tasks(
-        tasks, dataset.id, data, user, pipeline_name, context, incremental_loading
+        tasks, dataset, data, user, pipeline_name, context, incremental_loading
```
dexters1: Do we need this change? to send Dataset object instead of just UUID? It might break peoples low level pipeline implementations

cognee-starter-kit/src/pipelines/low_level.py
```diff
@@ -221,18 +223,16 @@ async def execute_pipeline() -> None:
...
+    data = load_default_payload()
 
     # Build and run pipeline
     tasks = [Task(ingest_payloads), Task(add_data_points)]
-    pipeline = run_tasks(tasks, dataset_id, None, user, "demo_pipeline")
+    pipeline = run_tasks(tasks, dataset, [data], user, "demo_pipeline")
```
hajdul88: Does this mean that users will have to pass this Data object in the case of the low_level pipeline? This is kinda changing the way how low level pipelines can be created. Does it work without it?

cognee/modules/graph/methods/delete_data_related_edges.py
```diff
@@ -0,0 +1,13 @@
...
+from sqlalchemy import delete, select
+from sqlalchemy.ext.asyncio import AsyncSession
+
+from cognee.infrastructure.databases.relational import with_async_session
+from cognee.modules.graph.models import Edge
+
+
+@with_async_session
+async def delete_data_related_edges(data_id: UUID, session: AsyncSession):
+    nodes = (await session.scalars(select(Edge).where(Edge.data_id == data_id))).all()
```
hajdul88: These are edges no? Just naming probably.

cognee/tasks/graph/extract_graph_from_data.py
```diff
@@ -85,19 +85,19 @@ async def integrate_chunk_graphs(
     )
 
     if len(graph_nodes) > 0:
-        await add_data_points(graph_nodes)
+        await add_data_points(graph_nodes, context)
 
     if len(graph_edges) > 0:
         await graph_engine.add_edges(graph_edges)
```
hajdul88: Custom edges are not handled in this way and they dont have context

cognee/tasks/storage/add_data_points.py
```diff
@@ -33,8 +37,15 @@ async def add_data_points(data_points: List[DataPoint]) -> List[DataPoint]:
         - Deduplicates nodes and edges across all results.
         - Updates the node index via `index_data_points`.
         - Inserts nodes and edges into the graph engine.
-        - Optionally updates the edge index via `index_graph_edges`.
     """
+    user: Optional[User] = None
```
hajdul88: so if a user doesn't provide the context the data will be saved as none none none origin. Then it will just not save the ledger right?

.github/workflows/test_different_operating_systems.yml
hajdul88: We agreed that after these are fixed we test the interface (delete_all, dataset, data) features with edge cases

hajdul88:
In general this changes a lot of things and can break cognee delete for existing users/cognee setups as it is.

cognee/tasks/storage/add_data_points.py
borisarzentar: Yeah, it will not be saved in the db.

cognee-starter-kit/src/pipelines/low_level.py
borisarzentar: Yes, it does. But we switched from using `run_tasks` to `run_pipeline` for low level pipelines, so this needs to be updated.

cognee/modules/pipelines/operations/pipeline.py
borisarzentar: Otherwise we need to query the dataset inside and in most cases we did that already outside. So we queried dataset at least two times in every API/SDK call. We also need to switch to `run_pipeline` for low level pipelines.

Vasilije1990:
This PR will not be merged until it is backward compatible and all issues raised by Igor and Laszlo have been addressed @borisarzentar 

The PR is too big and touches too many things without a good reason.

I would suggest slimming it down to base core delete + backward compatibility and opening PR for other changes separately.

cognee-starter-kit/src/pipelines/low_level.py
borisarzentar: I reverted it.

cognee/tests/test_delete_default_graph.py
borisarzentar: Changed this

borisarzentar:
> If user1 (who is the owner of the dataset) adds Data to a dataset can User2 who has delete permission delete delete this Data? In a multi-user setting

Added a test for that.

cognee/tests/test_delete_default_graph.py
```diff
@@ -0,0 +1,139 @@
+import os
```
hajdul88: I started testing it with the e2e tests you added. 

I changed it so I am not mocking the graph and added 4 documents. 

For now the delete_default_graph.py doesn't delete DocumentChunks or Entities or basically any node from vector collections. The graph nodes and edges get deleted correctly but the collection elements don't.

hajdul88: neo4j, pgvector setting

cognee/tests/test_delete_default_graph.py
```diff
@@ -0,0 +1,139 @@
...
+            edges=[
+                Edge(
+                    source_node_id="Marie",
+                    target_node_id="Apple",
+                    relationship_name="works_for",
+                ),
+                Edge(source_node_id="Marie", target_node_id="MacOS", relationship_name="works_on"),
+            ],
+        ),
+        SummarizedContent(summary="Summary of John's work.", description="Summary of John's work."),
```
hajdul88: I continued with the same example, now I added the two original texts but instead of mocking the LLM I let cognify create the graph.

-Custom edges are missing from the edgetype_relationship_name collection (I checked dev, and they are there so something is off with the ingestion on this branch)


-When I delete one document, edges from vector collection (like made_from, is_a) are getting deleted although they shouldn't be 

So basically we loose custom edges at ingestion time, and delete edges from collections that shouldn't be deleted

alembic/versions/84e5d08260d6_replace_graph_ledger_table_with_nodes_.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+    inspector = sa.inspect(conn)
+
+    table_names = inspector.get_table_names()
+
+    if "graph_relationship_ledger" in table_names:
+        op.drop_table("graph_relationship_ledger")
```
hajdul88: What happens with a knowledge graph in the following scenario?


- User has a built cognee system (KG + vector stored populated)
- We drop the relationship_ledger table
- I ingest one document that contains nodes that are already in the KG + vector stores
- I delete the document I ingested


In this case I believe old parts of the KG will be dropped as well no?

hajdul88 (commented):
I only tested the test_delete_default_graph, and found major discrepancies. I will write them ordered by importance (highest to lowest):

1. Cognify doesn't embed custom edges anymore. They are missing from the EdgeType_relationship_name vector collection. (Checked dev and its correct there)
2. When I ingest 2 file. delete one of them, edges like is_a, made_from, is_part_of are getting deleted from the EdgeType_relationship_name collection although they shouldn't.
3. When I call datasets.delete_data(dataset_id, added_data.id, user), nothing gets deleted from the vector collections (except the edges that shouldn't be deleted)


Additional comments questions:

How do we make sure that the new delete is backward compatible with the old solution? and what happens in the following case?


-User has a built cognee system (KG + vector stored populated)
-I ingest one document that contains nodes that are already in the KG + vector stores
-I delete the document I ingested
-In this case I believe old parts of the KG will be dropped as well no?


For now I only tested the test_delete_default_graph.py so basic delete. Lets discuss before we proceed with he review.

cognee/tests/test_delete_default_graph.py
borisarzentar: Fixed this and added tests for it.

cognee/tests/test_delete_default_graph.py
borisarzentar: This was a copy/paste mistake, fixed that.

borisarzentar:
> Started to check it again but tests are failing now + there is a conflict

Yeah, text was too long in one of the tests so Edge table failed. Changed that now and resolved the conflict. Everything else stayed the same basically.

hajdul88:
I think CI is still failing, and delete tests are the ones that are failing now

.github/workflows/e2e_tests.yml
```diff
@@ -9,6 +9,14 @@ on:
         required: false
         type: string
         default: '3.11.x'
+      supported-python-versions:
+        required: false
+        type: string
+        default: '["3.10.x", "3.12.x", "3.13.x"]'
```
dexters1: We don't test all version of python for all OS systems anymore as it was very expensive (we only test on Ubuntu), we've also moved different OS tests to the test_different_operating_systems.yml file

alembic/versions/84e5d08260d6_replace_graph_ledger_table_with_nodes_.py
```diff
@@ -0,0 +1,120 @@
...
+from typing import Sequence, Union
+from datetime import datetime, timezone
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision: str = "84e5d08260d6"
+down_revision: Union[str, None] = "211ab850ef3d"
```
dexters1: This is just a reminder for everyone, there is a chance the down revision might change due to other PRs and needs to be manually updated (before we merge)

.github/workflows/e2e_tests.yml
borisarzentar: Ok, I'm fine with one OS as well. Will just run on multiple python versions some of the tests.

cognee/modules/graph/methods/set_current_user.py
```diff
@@ -0,0 +1,8 @@
+import uuid
+from sqlalchemy import text
+from sqlalchemy.ext.asyncio import AsyncSession
+
+
+async def set_current_user(session: AsyncSession, user_id: uuid.UUID, local: bool = False):
```
dexters1: Should this be deleted?

cognee/tests/test_delete_default_graph_non_mocked.py
```diff
@@ -0,0 +1,119 @@
+import os
```
hajdul88: I think this test is not the part of the CI or at least I didn't see it called. I tried to run locally but it also failed for me in line 61.

If we dont need it lets delete it, if we do then lets put it into the CI

borisarzentar: Maybe I forgot to add it, will check

hajdul88: But I tried it with neo4j and pgvector without backend access control

borisarzentar: Added it, let's see what it says. I run it locally once, it passed.

hajdul88: Btw why is this failing without backend access control? (neo4j, pg) (and basically default graph test as well)

cognee/modules/graph/methods/set_current_user.py
borisarzentar: Deleted

cognee/tests/test_delete_default_graph_non_mocked.py
```diff
@@ -0,0 +1,119 @@
...
+        node_data = node[1]
+        collection_name = node_data["type"] + "_" + node_data["metadata"]["index_fields"][0]
+        if collection_name not in initial_nodes_by_vector_collection:
+            initial_nodes_by_vector_collection[collection_name] = []
+        initial_nodes_by_vector_collection[collection_name].append(node)
+
+    initial_node_ids = set([node[0] for node in initial_nodes])
+
+    user = await get_default_user()
+    await datasets.delete_data(dataset_id, johns_data_id, user)  # type: ignore
```
hajdul88: I ran this example. Deleted the first file so johns data. After that I inspected the nodes and edges rel tables. Nodes seems to be correct however, edges table contains edges from johns data. In my graph I have 11 edges in the edges rel table 14. So i think the ledger for edges is not updated correcly

hajdul88: If I think about it this can cause later a problem where we dont delete edges because of these phantom edges that stayed there.

cognee/tests/test_delete_default_graph_non_mocked.py
```diff
@@ -0,0 +1,119 @@
...
+
+    nodes, edges = await graph_engine.get_graph_data()
+    assert len(nodes) >= 9 and len(nodes) <= 11 and len(edges) >= 10 and len(edges) <= 12, (
+        "Nodes and edges are not deleted."
+    )
+    assert not any(
```
hajdul88: I again run this with the following settings:

ENABLE_BACKEND_ACCESS_CONTROL=False
Graphdb Neo4j
VectorDB: Pgvector


`data_directory_path = os.path.join(
        pathlib.Path(__file__).parent, ".data_storage/test_delete_default_graph_non_mocked"
    )
    cognee.config.data_root_directory(data_directory_path)

    cognee_directory_path = os.path.join(
        pathlib.Path(__file__).parent, ".cognee_system/test_delete_default_graph_non_mocked"
    )
    cognee.config.system_root_directory(cognee_directory_path)

    await setup()

    from cognee.modules.users.methods.create_default_user import create_user
    new_user = await create_user(
        email='smthg@something.com',
        password='my_pw_is_this',
        is_superuser=True,
        is_active=True,
        is_verified=True,
        auto_login=True,
    )

    add_result = await cognee.add(
        "John works for Apple. He is also affiliated with a non-profit organization called 'Food for Hungry'"
    )
    johns_data_id = add_result.data_ingestion_info[0]["data_id"]

    add_result = await cognee.add(
        "Marie works for Apple as well. She is a software engineer on MacOS project.",
        user=new_user
    )
    maries_data_id = add_result.data_ingestion_info[0]["data_id"]

    cognify_result: dict = await cognee.cognify()
    dataset_id = list(cognify_result.keys())[0]

    await cognee.cognify(user=new_user)

    default_user= await get_default_user()
    await datasets.delete_data(dataset_id, johns_data_id, default_user) `


This results in the following graph:

Image


Asked @dexters1 because I was not sure if this should work but he said it should.

hajdul88: nodes table is also inconsistent, for the above graph it has 9 nodes, edge table has 14 edges.

EdgeType_relationship_name vector collection has 5 elements is_a embedding is missing + have an additional develops embedding which is not in the graph.

Node related collections seem to be correct

cognee/tests/test_delete_default_graph_with_legacy_data_1.py
```diff
@@ -0,0 +1,382 @@
+import os
```
hajdul88: Without running this test, I did the following experiment to test backward compatibility:


1. Switched to dev branch
2. added and cognified the following sentence: "Bmw is a german carmanufacturer"
3. Inspected the graph -- BMW--produced in--germany -- is_a--country
4. Switched to the new delete branch
5. added and cognified the following sentence: "Germany is located next to the netherlands"
6. Inspected the graph --Germany--located_next_to--Netherlands (germany node is shared between the two documents)
6. Deleted the latest sentence about germany and netherlands
7. The node germany got deleted too, however it should have stayed there

I think the current solution is not backward compatible in this case as it deleted node that was the part of the previously ingested knowledge graph

cognee/tests/test_delete_default_graph_non_mocked.py
borisarzentar: @hajdul88 Tried this scenario with neo4j and pgvector, got these graphs initially and after delete:
Screenshot 2025-11-13 at 17 57 53
Screenshot 2025-11-13 at 17 57 58

hajdul88: I can check again, did you make some change + did you check the table inconsistencies?

hajdul88: Rerun it since its llm generated its not happening all the time, first time it didn't happen to me but second time did. I think the table inconsistencies are also there

hajdul88: Btw the picture you sent is also incorrect, if you check apple + person nodes shouldn't be deleted

cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
```diff
@@ -955,7 +955,22 @@ async def get_graph_data(self):
                 f"Retrieved {len(nodes)} nodes and {len(edges)} edges in {retrieval_time:.2f} seconds"
             )
 
-            return (nodes, edges)
+            return (
```
hajdul88: Why do we have to change this? Is it because of the delete test? I dont know but I am not sure we should change core adapter behavior because of one test, maybe then the test should be changed.

@dexters1 , @pazone wdyt?

dexters1: We should not change core adapter behavior in a breaking change if we don't have to. People are probably using this in their own implementations and it will break for them

cognee/api/v1/datasets/datasets.py
```diff
@@ -42,10 +55,60 @@ async def get_status(dataset_ids: list[UUID]) -> dict:
...
+            await delete_data_nodes_and_edges(dataset_id, data_id, user.id)
+
+        await delete_data(data)
+
+    @staticmethod
+    async def delete_all(user: Optional[User] = None):
+        if not user:
+            user = await get_default_user()
+
+        user_datasets = await get_authorized_existing_datasets([], "read", user)
```
dexters1: We should check "delete" permission here and not "read"
I'm also not sure if we should have this function in general as it's very unpredictable, it can unexpectedly delete half of an organizations datasets by accident since the user had the appropriate permissions
We should force specifying of dataset IDs when they want to be deleted through endpoints and I would remove ability to delete all datasets user has permission for (Similar platforms like Neo4j and Github always make you confirm if you want to delete something before you do it for a reason)

cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
borisarzentar: I can revert it, but it is inconsistent behaviour between the adapters. Kuzu is returning dict for metadata while neo4j returns string.

cognee/api/v1/datasets/datasets.py
borisarzentar: Changed to "delete", it was copy/paste mistake probably.
Confirming usually happens in the UI, after that API request is dispatched and there is no confirmation on the BE side. So if someone calls this, they are the ones deciding on it. Same as giving permissions, you can easily make a mistake and assign some permissions to a user that should not get it. We give freedom, but it comes with consequences.

cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
hajdul88: I would revert it for now if others agree on this, we are planning an adapter refactor with Lazar, am also working now around this region in another pr, I will check because I also want to reduce the number of queries here.

hajdul88: I just checked this in my PR, both of them look dict to me, with string it wouldn't even work I think. But maybe am missing smthg

borisarzentar: Reverted it. Metadata is saved as string, so when we retrieve the node we need to parse metadata in order to use it as dict.

hajdul88: Thanks will check it

cognee/tests/test_delete_default_graph_with_legacy_data_1.py
hajdul88: This is still not correct in the new review round

cognee/tests/test_delete_default_graph_non_mocked.py
hajdul88: Now the graph and the delete looks better, and it seems its deleting the subgraph correctly when enable backend access control is off.

However the inconsistency in the edge ledger table is still there, edge ledge table contains more edges than it is in the graph.

cognee/tests/test_delete_default_graph_non_mocked.py
hajdul88: This is still happening in the new review round

cognee/tests/test_delete_default_graph_non_mocked.py
hajdul88: This is still happening in the new pr review round

cognee/tests/test_delete_default_graph_non_mocked.py
hajdul88: I resolve this because the edge ledger part is in another comment

pazone:
As I understand, the `GitGuardian / GitGuardian Security Checks` failure is false positive

borisarzentar:
> I found a bug: When we create a DS with 2 files and delete it - the DS is being deleted, but the files remain hanging in the database

Checked this and looked at the files, everything is removed now.

cognee/tests/test_delete_default_graph_non_mocked.py
borisarzentar: Tried this and was following the data in databases, everything seems fine now. Tried with defaults and with pgvector/neo4j.

PR #1548, opened 2025-10-15
fix: search without prior cognify

daukadolt:
## Description

Running search when `cognee.add()` is called, but `cognee.cognify()` wasn't called yet goes through the whole search operation to throw a cryptic error:
```
Error during graph projection: EntityNotFoundError: Empty graph projected from the database. (Status code: 404)
```

## How to reproduce
modify `dynamic_steps_example.py` to not run cognify

## This PR

Checks graph before searching, and throws an informative exception to ensure cognify was run

| Logs Before | Logs After |
|--------------|------------|
| `Error during graph projection: EntityNotFoundError: Empty graph projected from the database. (Status code: 404)` | `2025-10-17T11:05:58.465315 [warning  ] Search attempt on an empty knowledge graph [cognee.shared.logging_utils]` |

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/tests/unit/api/test_search.py
+++ b/cognee/tests/unit/api/test_search.py
@@ -1,6 +1,5 @@
 import pytest
 import cognee
-from cognee.modules.data.exceptions import SearchOnEmptyGraphError
 
 
 @pytest.mark.asyncio
@@ -8,16 +7,15 @@ async def test_empty_search_raises_SearchOnEmptyGraphError_on_empty_graph():
     await cognee.prune.prune_data()
     await cognee.prune.prune_system(metadata=True)
     await cognee.add("Sample input")
-    with pytest.raises(SearchOnEmptyGraphError):
-        await cognee.search("Sample query")
+    result = await cognee.search("Sample query")
+    assert result == []
 
 
+@pytest.mark.asyncio
 async def test_empty_search_doesnt_raise_SearchOnEmptyGraphError():
     await cognee.prune.prune_data()
     await cognee.prune.prune_system(metadata=True)
     await cognee.add("Sample input")
     await cognee.cognify()
-    try:
-        await cognee.search("Sample query")
-    except SearchOnEmptyGraphError:
-        pytest.fail("Should not raise SearchOnEmptyGraphError when data was added and cognified")
+    result = await cognee.search("Sample query")
+    assert result != []


cognee/infrastructure/databases/graph/kuzu/adapter.py
```diff
@@ -185,6 +185,22 @@ async def pull_from_s3(self) -> None:
...
+        query = """
+        MATCH ()-[r]->()
+        RETURN COUNT(r);
+        """
+        query_result = await self.query(query)
+        return query_result[0][0]
+
+    async def count_nodes(self) -> int:
+        query = """
+        MATCH (n)
```
hajdul88, line 203: "MATCH (n) RETURN n LIMIT 1"

Maybe? Then you dont have to always count them. and I think its enough to have this for the nodes. if there are no nodes, you dont have edges either.

cognee/api/v1/search/search.py
```diff
@@ -175,6 +176,15 @@ async def search(
         if not datasets:
             raise DatasetNotFoundError(message="No datasets found.")
 
+    graph_engine = await get_graph_engine()
+    edges_count = await graph_engine.count_edges()
```
hajdul88: Also this shouldn't fail on other adapters where you dont have this.

cognee/infrastructure/databases/graph/kuzu/adapter.py
daukadolt: This'd help to check empty/not empty, but why not get count number when we can? ¯\_(ツ)_/¯

cognee/api/v1/search/search.py
```diff
@@ -175,6 +176,14 @@ async def search(
         if not datasets:
             raise DatasetNotFoundError(message="No datasets found.")
 
+    graph_engine = await get_graph_engine()
+    nodes_count = await graph_engine.count_nodes()
+
+    if nodes_count == 0:
```
daukadolt: add logs

cognee/infrastructure/databases/graph/kuzu/adapter.py
hajdul88: Because I think this one is O(1), and the other one is O(N) its slow for a big graph. The Count(n) I think scans all nodes while the other one stops early.

daukadolt: fair, but we'll need to refactor this method to be `is_empty() -> bool` instead then

daukadolt: if that sounds good I'll refactor the implementation

cognee/api/v1/search/search.py
```diff
@@ -175,6 +176,14 @@ async def search(
...
+        raise SearchOnEmptyGraphError(
```
hajdul88: Do we want to raise the exception? I feel its good but just to point it out. If the graph is empty the search will fail

cognee/infrastructure/databases/graph/kuzu/adapter.py
borisarzentar: Yes, let's do that, let's create `is_empty() -> bool` then.

cognee/api/v1/search/search.py
borisarzentar: Caught my attention as well. I wouldn't raise exception here because this is not wrong behaviour, having empty graph db is fine. We just need to make sure that when graph is **not** empty that we get nodes and edges for search graph projection.

daukadolt: no strong opinion here, I'm fine with returning an empty array

I'll replace this with logger warning and empty output

PR #1560, opened 2025-10-18
feat: Add multi-tenancy

dexters1:
## Description
Add multi-tenancy to Cognee

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/.github/workflows/weighted_edges_tests.yml
+++ b/.github/workflows/weighted_edges_tests.yml
@@ -2,7 +2,7 @@ name: Weighted Edges Tests
 
 on:
   push:
-    branches: [ main, weighted_edges ]
+    branches: [ main, dev, weighted_edges ]
     paths:
       - 'cognee/modules/graph/utils/get_graph_from_model.py'
       - 'cognee/infrastructure/engine/models/Edge.py'
@@ -10,7 +10,7 @@ on:
       - 'examples/python/weighted_edges_example.py'
       - '.github/workflows/weighted_edges_tests.yml'
   pull_request:
-    branches: [ main ]
+    branches: [ main, dev ]
     paths:
       - 'cognee/modules/graph/utils/get_graph_from_model.py'
       - 'cognee/infrastructure/engine/models/Edge.py'
@@ -32,7 +32,7 @@ jobs:
     env:
       LLM_PROVIDER: openai
       LLM_MODEL: gpt-5-mini
-      LLM_API_KEY: ${{ secrets.LLM_API_KEY }}
+      LLM_API_KEY: ${{ secrets.OPENAI_API_KEY }}
 
     steps:
       - name: Check out repository
@@ -67,14 +67,13 @@ jobs:
     env:
       LLM_PROVIDER: openai
       LLM_MODEL: gpt-5-mini
-      LLM_ENDPOINT: https://api.openai.com/v1/
-      LLM_API_KEY: ${{ secrets.LLM_API_KEY }}
+      LLM_ENDPOINT: https://api.openai.com/v1
+      LLM_API_KEY: ${{ secrets.OPENAI_API_KEY  }}
       LLM_API_VERSION: "2024-02-01"
-      EMBEDDING_PROVIDER: openai
-      EMBEDDING_MODEL: text-embedding-3-small
-      EMBEDDING_ENDPOINT: https://api.openai.com/v1/
-      EMBEDDING_API_KEY: ${{ secrets.LLM_API_KEY }}
-      EMBEDDING_API_VERSION: "2024-02-01"
+      EMBEDDING_MODEL: ${{ secrets.EMBEDDING_MODEL }}
+      EMBEDDING_ENDPOINT: ${{ secrets.EMBEDDING_ENDPOINT }}
+      EMBEDDING_API_KEY: ${{ secrets.EMBEDDING_API_KEY }}
+      EMBEDDING_API_VERSION: ${{ secrets.EMBEDDING_API_VERSION }}
     steps:
       - name: Check out repository
         uses: actions/checkout@v4
@@ -108,14 +107,14 @@ jobs:
     env:
       LLM_PROVIDER: openai
       LLM_MODEL: gpt-5-mini
-      LLM_ENDPOINT: https://api.openai.com/v1/
-      LLM_API_KEY: ${{ secrets.LLM_API_KEY }}
+      LLM_ENDPOINT: https://api.openai.com/v1
+      LLM_API_KEY: ${{ secrets.OPENAI_API_KEY  }}
       LLM_API_VERSION: "2024-02-01"
-      EMBEDDING_PROVIDER: openai
-      EMBEDDING_MODEL: text-embedding-3-small
-      EMBEDDING_ENDPOINT: https://api.openai.com/v1/
-      EMBEDDING_API_KEY: ${{ secrets.LLM_API_KEY }}
-      EMBEDDING_API_VERSION: "2024-02-01"
+      EMBEDDING_MODEL: ${{ secrets.EMBEDDING_MODEL }}
+      EMBEDDING_ENDPOINT: ${{ secrets.EMBEDDING_ENDPOINT }}
+      EMBEDDING_API_KEY: ${{ secrets.EMBEDDING_API_KEY }}
+      EMBEDDING_API_VERSION: ${{ secrets.EMBEDDING_API_VERSION }}
+
     steps:
       - name: Check out repository
         uses: actions/checkout@v4
--- a/cognee/infrastructure/databases/hybrid/neptune_analytics/NeptuneAnalyticsAdapter.py
+++ b/cognee/infrastructure/databases/hybrid/neptune_analytics/NeptuneAnalyticsAdapter.py
@@ -416,6 +416,15 @@ class NeptuneAnalyticsAdapter(NeptuneGraphDB, VectorDBInterface):
         self._client.query(f"MATCH (n :{self._VECTOR_NODE_LABEL}) DETACH DELETE n")
         pass
 
+    async def is_empty(self) -> bool:
+        query = """
+        MATCH (n)
+        RETURN true
+        LIMIT 1;
+        """
+        query_result = await self._client.query(query)
+        return len(query_result) == 0
+
     @staticmethod
     def _get_scored_result(
         item: dict, with_vector: bool = False, with_score: bool = False
--- a/cognee/modules/retrieval/cypher_search_retriever.py
+++ b/cognee/modules/retrieval/cypher_search_retriever.py
@@ -1,4 +1,6 @@
 from typing import Any, Optional
+from fastapi.encoders import jsonable_encoder
+
 from cognee.infrastructure.databases.graph import get_graph_engine
 from cognee.modules.retrieval.base_retriever import BaseRetriever
 from cognee.modules.retrieval.utils.completion import generate_completion
@@ -50,7 +52,7 @@ class CypherSearchRetriever(BaseRetriever):
                 logger.warning("Search attempt on an empty knowledge graph")
                 return []
 
-            result = await graph_engine.query(query)
+            result = jsonable_encoder(await graph_engine.query(query))
         except Exception as e:
             logger.error("Failed to execture cypher search retrieval: %s", str(e))
             raise CypherSearchError() from e


cognee/modules/users/tenants/methods/select_tenant.py
```diff
@@ -0,0 +1,61 @@
...
+        user_id: Id of the user.
+        tenant_id: Id of the tenant.
+
+    Returns:
+        None
+
+    """
+    db_engine = get_relational_engine()
+    async with db_engine.get_async_session() as session:
+        user = await get_user(user_id)
```
borisarzentar, line 30: Let's pass the `user` inside so we don't have to fetch it again.

cognee/modules/users/tenants/methods/select_tenant.py
```diff
@@ -0,0 +1,61 @@
...
+        result = await session.execute(
+            select(UserTenant)
+            .where(UserTenant.user_id == user_id)
+            .where(UserTenant.tenant_id == tenant_id)
+        )
+
+        try:
+            result = result.scalar_one()
+        except sqlalchemy.exc.NoResultFound as e:
+            raise TenantNotFoundError("User Tenant relationship not found.") from e
```
borisarzentar: Maybe something more clear for users, like "User is not part of the tenant. Add user to tenant first to be able to select it later."

cognee/modules/users/tenants/methods/create_tenant.py
```diff
@@ -34,6 +32,17 @@ async def create_tenant(tenant_name: str, user_id: UUID) -> UUID:
             user.tenant_id = tenant.id
```
borisarzentar: Does this mean that when user creates a tenant, that tenant will automatically be selected?

cognee/modules/users/tenants/methods/add_user_to_tenant.py
```diff
@@ -12,14 +15,19 @@
 )
 
 
-async def add_user_to_tenant(user_id: UUID, tenant_id: UUID, owner_id: UUID):
+async def add_user_to_tenant(
+    user_id: UUID, tenant_id: UUID, owner_id: UUID, set_active_tenant: Optional[bool] = True
```
borisarzentar: `set_active_tenant` sounds more like a function, I was a bit confused. Maybe `should_set_as_active_tenant` or something like that.

cognee/modules/data/models/Dataset.py
```diff
@@ -18,6 +18,7 @@ class Dataset(Base):
     updated_at = Column(DateTime(timezone=True), onupdate=lambda: datetime.now(timezone.utc))
 
     owner_id = Column(UUID, index=True)
+    tenant_id = Column(UUID, index=True, nullable=True)
```
borisarzentar, line 21: Why we need `tenant_id` in `Dataset`? We can control the access to a dataset through `acls` table.

dexters1: We need to it track ownership, we can't have the same database shared by two different tenants

dexters1: Also applies to Data, each Data instance in our SQL table needs to be owned by a single tenant to keep track of data usage in a tenant and etc.

cognee/modules/users/tenants/methods/add_user_to_tenant.py
dexters1: I can change it to set_as_active_tenant

cognee/modules/users/tenants/methods/select_tenant.py
dexters1: We should avoid passing User object or similar SQL objects. Because even if we pass them we have to update them from the state in the database regardless every time, so it's safer to send UUID. They always need to be up to date with what is in the database. Now since we were passing User objects in a lot of places I have issues with data like UserRole UserTenant mapping being overwritten with the old state by the User object when using multi-tenancy with SDK.

This is not noticeable in the backend since the User object is update in a request, but in SDK it's an issue.

cognee/modules/users/permissions/methods/get_all_user_permission_datasets.py
```diff
@@ -45,4 +39,10 @@ async def get_all_user_permission_datasets(user: User, permission_type: str) ->
...
         unique.setdefault(dataset.id, dataset)
 
-    return list(unique.values())
+    # Filter out dataset that aren't part of the current user's tenant
+    filtered_datasets = []
+    for dataset in list(unique.values()):
+        if dataset.tenant_id == user.tenant_id:
+            filtered_datasets.append(dataset)
+
+    return filtered_datasets
```
pazone, line 48: `return [dataset for dataset in unique.values() if dataset.tenant_id == user.tenant_id]`. would it do the same? 🤔

dexters1: Should be the same if unique.values() is changed to list(unique.values()) as it returns a set otherwise

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -226,4 +231,39 @@ async def create_tenant(tenant_name: str, user: User = Depends(get_authenticated
...
+        to organize users and resources in multi-tenant environments, providing
+        isolation and access control between different groups or organizations.
+
+        Sending a null/None value as tenant_id selects his default single user tenant
+
+        ## Request Parameters
+        - **tenant_id** (Union[UUID, None]): UUID of the tenant to select, If null/None is provided use the default single user tenant
+
+        ## Response
+        Returns a success message indicating the tenant was created.
```
pazone: Created?

cognee/modules/data/models/Dataset.py
pazone: This means that datasets can belong to only one tenant, right?

cognee/api/v1/permissions/routers/get_permissions_router.py
dexters1: Will change to selected, probably copied dosctring from the create tenant endpoint and missed updating it

alembic/versions/c946955da633_multi_tenant_support.py
```diff
@@ -0,0 +1,137 @@
...
+                user_tenants,
+                [
+                    {"user_id": user_id, "tenant_id": tenant_id, "created_at": _now()}
+                    for user_id, tenant_id in user_data
+                ],
+            )
+
+    tenant_id_column = _get_column(insp, "datasets", "tenant_id")
+    if not tenant_id_column:
+        op.add_column("datasets", sa.Column("tenant_id", sa.UUID(), nullable=True))
```
borisarzentar, line 110: Just to be clear, we want a dataset to be owned by user and by tenant? Can a dataset be owned by tenant but no user?

cognee/modules/data/methods/get_unique_data_id.py
```diff
@@ -0,0 +1,68 @@
...
+            tenant_id: UUID of the tenant for which data is being added
+
+        Returns:
+            UUID: Unique identifier for the data
+        """
+        # return UUID hash of file contents + owner id + tenant_id
+        return uuid5(NAMESPACE_OID, f"{data_identifier}{str(user.id)}{str(user.tenant_id)}")
+
+    # Get all possible data_id values
+    data_id = {
```
borisarzentar, line 52: Why have this in dict, why not simple variables?

cognee/modules/users/tenants/methods/select_tenant.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Returns:
+        None
+
+    """
+    db_engine = get_relational_engine()
+    async with db_engine.get_async_session() as session:
+        user = await get_user(user_id)
+        if tenant_id is None:
+            # If no tenant_id is provided set current Tenant to the single user-tenant
```
pazone, line 32: Is this behavior discussed and approved? It might be unexpected that `select_tenant` completely changes the access mode if the `tenant_id` is `None`. What do you think if we do nothing if `tenant_id` is `None` and add an explicit function `switch_to_single_user`?

cognee/modules/users/permissions/methods/get_all_user_permission_datasets.py
```diff
@@ -45,4 +39,10 @@ async def get_all_user_permission_datasets(user: User, permission_type: str) ->
         # If the dataset id key already exists, leave the dictionary unchanged.
         unique.setdefault(dataset.id, dataset)
 
-    return list(unique.values())
+    # Filter out dataset that aren't part of the current user's tenant
+    filtered_datasets = []
+    for dataset in list(unique.values()):
+        if dataset.tenant_id == user.tenant_id:
```
borisarzentar, line 45: Can we have this condition above in the code where we fetch tenant datasets? We will fetch less and won't need to filter it here after.

cognee/modules/users/tenants/methods/select_tenant.py
```diff
@@ -0,0 +1,62 @@
...
+
+from cognee.infrastructure.databases.relational import get_relational_engine
+from cognee.modules.users.methods.get_user import get_user
+from cognee.modules.users.models.UserTenant import UserTenant
+from cognee.modules.users.models.User import User
+from cognee.modules.users.permissions.methods import get_tenant
+from cognee.modules.users.exceptions import UserNotFoundError, TenantNotFoundError
+
+
+async def select_tenant(user_id: UUID, tenant_id: Union[UUID, None]) -> User:
```
borisarzentar, line 15: I'm thinking more and more that we should operate with `User`, `Tenant`, `Dataset`.... objects, instead of `id`s. When we send ids we often do the object fetching right away, which is unnecessary step in most scenarios.

alembic/versions/c946955da633_multi_tenant_support.py
dexters1: Yes we want to know who made it and which organization it belongs to

cognee/modules/data/methods/get_unique_data_id.py
dexters1: No specific reason, just liked to show that data_id can be two different values for backwards compatibility. Thought this made it more obvious

cognee/modules/users/tenants/methods/select_tenant.py
pazone: Is it possible to add a test that emulates the situation when we create tenants and then switch to the single-user mode?

cognee/modules/users/tenants/methods/select_tenant.py
dexters1: The problem is we either have to do everything with UUID or everything with the object. I tried changing select_tenant to return an object it brakes everything without and error and overwrites DB values.

I wanted to have the run_tasks/pipelines mechanism fetch and update user for tasks for this reason so:
1) Nothing will be overwritten
2) User object will be up to date at request time 

Having everything as an object is really complicated if someone is writing Cognee in the SDK as it's never up to date and will overwrite information in the DB without any error or warnings. Especially if they have two different long running threads/processes that can change user information. I would definitely avoid using objects whenever possible for this reason.

I spent hours debugging the SDK to see why permissions were not working properly to find out the User objects were overwriting what is actually in the database.

For now we have to use UUID as all of the functions are written to use UUID

cognee/modules/users/tenants/methods/select_tenant.py
dexters1: We can add another function if you think we need for switch_to_single_user_tenant, but as tenant_id is None is what is actually in the Database as a Tenant info for the User I would keep select_tenant when tenant_id is None to also select the single user mode.

For now single user mode can still share datasets with other users though, this feature was not removed.

cognee/modules/users/permissions/methods/get_all_user_permission_datasets.py
dexters1: I'll try it. I think that was the first thing I tried I just forgot why I didn't go through with that approach

cognee/modules/users/tenants/methods/select_tenant.py
pazone: My recommendation is to create the `switch_to_single_user_mode` that accepts User (or userID). Make the current `select_tenant` (that accepts None as `tenant_id`) private. And the public `select_tenant` should accept `tenant_id` only as UUID. The private `_select_tenant` will be used by the both `switch_to_single_user_mode` and `select_tenant` just with different parameters.
This is recommended fro clear separation of API and backend concerns. That approach decouples these 2 areas.

dexters1: Currently there is no actual single_user_mode, when tenant_id is None it's a global Tenant for all users where they can all share and read each others datasets like they are all part of one big tenant.

We have not yet officially decided if we want to have this be the default case if tenant_id is None or do we want to maybe actually make a single_user_mode where the user is the only one who can read/write his data.

For now I would keep this as is, as None is just a global Tenant and is in line with the single responsibility principal. If we ever decide to change course and have a real single_user_mode then we should do what you have suggested.

cognee/modules/users/permissions/methods/get_all_user_permission_datasets.py
dexters1: Ah yes, I remember, it's because when tenant_id is None. It won't be part of that list of tenant datasets so it needs to be filtered out later. We also don't fetch later we just filter out so it's not accessing the database.

borisarzentar: I don't understand. Which `tenant_id`? If it won't be part of the tenant datasets, why it needs to be filtered out later?

alembic/versions/c946955da633_multi_tenant_support.py
borisarzentar: Should we then set `tenant_id` to be `nullable=False`?

cognee/modules/data/methods/get_unique_data_id.py
borisarzentar: I don't know, it is obvious, but at the same time feels wrong getting those values from dict. Not sure, it is not a big deal.

alembic/versions/c946955da633_multi_tenant_support.py
dexters1: When tenant_id is None it means it's a global Tenant, users can share these datasets with other users in that case without restrictions

cognee/modules/users/permissions/methods/get_all_user_permission_datasets.py
dexters1: When tenant_id is None it means it's the global tenant where all users can share their own datasets (that are not part of a specific tenant, but the global tenant) with each other if they would like. These datasets need to be filtered out if a specific tenant_id is provided

cognee/modules/users/permissions/methods/get_all_user_permission_datasets.py
```diff
@@ -45,4 +39,10 @@ async def get_all_user_permission_datasets(user: User, permission_type: str) ->
         # If the dataset id key already exists, leave the dictionary unchanged.
         unique.setdefault(dataset.id, dataset)
 
-    return list(unique.values())
+    # Filter out dataset that aren't part of the current user's tenant
```
borisarzentar:
```suggestion
    # Filter out dataset that aren't part of the selected user's tenant
```

PR #1570, opened 2025-10-20
fix: search without prior cognify

daukadolt:

## Description

#1548 did not handle multi-tenant case, where graph db context is set further downstream.

This version adds empty graph logging/handling that works for multi-tenant case


daukadolt:
fyi only unit test that's failing is failing due to early return for empty graph (`The context provided does not include specific nodes or relations in the...`)

I'll keep it as is until review

Vasilije1990:
@daukadolt please check this, there is a conflict here

daukadolt:
rebasing

Issue #1348, opened 2025-09-08
CSV Ingestion Loader & Chunker for Cognee


hande-k:
*NOTE: This issue is part of Contribute-to-Win. Please comment first to get assigned. Read the details [here](https://github.com/orgs/topoteretes/discussions/1361)*

Create a dedicated CSV ingestion path with a custom loader and custom chunker that preserves row-column relationships in the produced chunks.

## Overview

Directly reading and ingesting .csv files into Cognee has not worked well. A purpose-built approach is needed so each chunk retains both the row data and the associated column context. This issue proposes building a custom CSV loader and a custom chunker (i.e., a CSV-specific loader or pipeline) to reliably transform tabular data into chunks that encode column names alongside the corresponding row values.

**Input:** .csv files intended for ingestion into cognee
**Output:** Chunks where each row’s values are explicitly tied to their columns, suitable for cognee ingestion

## Loader & Chunker Implementation Details

- Preserve column names and ensure every value is associated with its correct column.

- Chunk at row granularity so each CSV row yields chunk(s) that include column context.

Devashish2112:
please assign this to me

hande-k:
thanks for picking it up @Devashish2112

EricXiao95:
Hi, @hande-k. I'd like to work on this task.

hande-k:
hi @EricXiao95 happy to see you here! the issue is assigned to you :)

BHIMASAIKAUSHIK:
i can work on this

PR #1574, opened 2025-10-22
feat: csv ingestion & chunking

EricXiao95:
## Description

Create a dedicated CSV ingestion path with a custom loader and custom chunker that preserves row-column relationships in the produced chunks.  #1348

## Type of Change

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [x] Documentation update
- [x] Code refactoring
- [x] Performance improvement
- [x] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee-mcp/src/server.py
+++ b/cognee-mcp/src/server.py
@@ -194,7 +194,6 @@ async def cognify(
 
     Prerequisites:
         - **LLM_API_KEY**: Must be configured (required for entity extraction and graph generation)
-        - **Data Added**: Must have data previously added via `cognee.add()`
         - **Vector Database**: Must be accessible for embeddings storage
         - **Graph Database**: Must be accessible for relationship storage
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -156,7 +156,6 @@ Homepage = "https://www.cognee.ai"
 Repository = "https://github.com/topoteretes/cognee"
 
 [project.scripts]
-cognee = "cognee.cli._cognee:main"
 cognee-cli = "cognee.cli._cognee:main"
 
 [build-system]


cognee/infrastructure/files/utils/guess_file_type.py
```diff
@@ -134,3 +136,44 @@ def guess_file_type(file: BinaryIO) -> filetype.Type:
         raise FileTypeException(f"Unknown file detected: {file.name}.")
 
     return file_type
+
+
+class CsvFileType(filetype.Type):
```
borisarzentar: We use custom loaders now, we don't expand this file anymore.
You add a custom loader to cognee by using `use_loader` hook.

Vasilije1990: @EricXiao95 can you please check the comment? Would love to merge this one soon

EricXiao95: Sorry for the late reply. I'm currently on vacation. I'll solve this problem this week.

EricXiao95: > We use custom loaders now, we don't expand this file anymore. You add a custom loader to cognee by using `use_loader` hook.

image

-------------------------
I understand the suggestion about usin custom loaders, but I see that filetype.guess(file) isn't currently recognizing CSV files in the latest version of guess_file_type.py, so I want to add new code like the picture. What are your thoughts on this?
@borisarzentar

cognee/modules/chunking/CsvChunker.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+from cognee.tasks.chunks import chunk_by_row
+from cognee.modules.chunking.Chunker import Chunker
+from .models.DocumentChunk import DocumentChunk
+
+logger = get_logger()
+
+
+class CsvChunker(Chunker):
```
EricXiao95, line 11: Hi @dexters1 ,

Following up on our recent discussion on [Discord](https://discord.com/channels/1120795297094832337/1430494392338219059):

Based on that conversation, my idea is to remove both the CsvDocument and CsvChunker classes.

I wanted to confirm with you here: is this the correct approach? Please let me know if this sounds right before I proceed.

Thanks!

cognee/infrastructure/files/utils/guess_file_type.py
dexters1: The filetype.guess function should be properly handling csv and other file types already. Only .txt needed to be manually added as it has no byte signature to determine file type

EricXiao95: From my local test, filetype.guess returns None for CSV files, and [the supported types of filetype.guess](https://pypi.org/project/filetype/) don’t include text/csv. Because CSV is a special plain-text format, it also has no byte signature to determine file type. 

If there’s a recommended way to detect CSV using filetype.guess directly that I’m missing, I’m very happy to switch to that instead.

---------------
If it’s not recommended to modify guess_file_type, how about adding text/plain to CsvLoader’s supported_mime_types instead? @dexters1

dexters1: If it also can't be guessed we can do as you have initially suggested

cognee/modules/chunking/CsvChunker.py
dexters1: You can remove the CsvDocument, but we need the CsvChunker class, it should work with the text output of the CsvLoader and create chunks of CSV rows. Chunkers to be used can be manually specified in the cognify pipeline and the CsvChunker will need to be specified there to use the CsvLoader properly

cognee/infrastructure/files/utils/guess_file_type.py
EricXiao95: Got it. New code has been pushed. Please review.

PR #1695, opened 2025-10-29
refactor: Enable multi user mode by default if graph and vector db pr…

dexters1:
…oviders support it

## Description
Enable multi user mode by default for supported graph and vector DBs

## Type of Change

- [ ] Documentation update
- [x] Code refactoring
- [ ] Performance improvement
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/context_global_variables.py
+++ b/cognee/context_global_variables.py
@@ -24,7 +24,7 @@ async def set_session_user_context_variable(user):
     session_user.set(user)
 
 
-def check_multi_user_support():
+def multi_user_support_possible():
     graph_db_config = get_graph_context_config()
     vector_db_config = get_vectordb_context_config()
     return (
@@ -33,24 +33,21 @@ def check_multi_user_support():
     )
 
 
-def check_backend_access_control_mode():
+def backend_access_control_enabled():
     backend_access_control = os.environ.get("ENABLE_BACKEND_ACCESS_CONTROL", None)
     if backend_access_control is None:
         # If backend access control is not defined in environment variables,
         # enable it by default if graph and vector DBs can support it, otherwise disable it
-        return check_multi_user_support()
+        return multi_user_support_possible()
     elif backend_access_control.lower() == "true":
         # If enabled, ensure that the current graph and vector DBs can support it
-        multi_user_support = check_multi_user_support()
+        multi_user_support = multi_user_support_possible()
         if not multi_user_support:
             raise EnvironmentError(
                 "ENABLE_BACKEND_ACCESS_CONTROL is set to true but the current graph and/or vector databases do not support multi-user access control. Please use supported databases or disable backend access control."
             )
-        else:
-            return True
-    else:
-        # If explicitly disabled, return false
-        return False
+        return True
+    return False
 
 
 async def set_database_global_context_variables(dataset: Union[str, UUID], user_id: UUID):
@@ -74,7 +71,7 @@ async def set_database_global_context_variables(dataset: Union[str, UUID], user_
 
     base_config = get_base_config()
 
-    if not check_backend_access_control_mode():
+    if not backend_access_control_enabled():
         return
 
     user = await get_user(user_id)
--- a/cognee/modules/search/methods/search.py
+++ b/cognee/modules/search/methods/search.py
@@ -8,7 +8,7 @@ from cognee.infrastructure.databases.graph import get_graph_engine
 from cognee.shared.logging_utils import get_logger
 from cognee.shared.utils import send_telemetry
 from cognee.context_global_variables import set_database_global_context_variables
-from cognee.context_global_variables import check_backend_access_control_mode
+from cognee.context_global_variables import backend_access_control_enabled
 
 from cognee.modules.engine.models.node_set import NodeSet
 from cognee.modules.graph.cognee_graph.CogneeGraphElements import Edge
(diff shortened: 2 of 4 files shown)


cognee/context_global_variables.py
```diff
@@ -14,11 +16,50 @@
...
+    if (
+        graph_db_config["graph_database_provider"] in graph_dbs_with_multi_user_support
+        and vector_db_config["vector_db_provider"] in vector_dbs_with_multi_user_support
+    ):
+        return True
+    else:
+        return False
+
+
+def check_backend_access_control_mode():
```
dexters1: Not the best place to store these checks and variables but we can move them easily if anyone has a suggestion. Or we can rename the file as it's all about multi-user mode settings and variables

pazone: Tip: Could you name it a bit more deterministic: like `backend_access_control_enabled()`. It will help us better understand the codebase

cognee/context_global_variables.py
```diff
@@ -14,11 +16,50 @@
...
+    backend_access_control = os.environ.get("ENABLE_BACKEND_ACCESS_CONTROL", None)
+    if backend_access_control is None:
+        # If backend access control is not defined in environment variables,
+        # enable it by default if graph and vector DBs can support it, otherwise disable it
+        multi_user_support = check_multi_user_support()
+        if multi_user_support:
+            return True
+        else:
+            return False
```
pazone:
```suggestion
        return check_multi_user_support()
```

cognee/context_global_variables.py
```diff
@@ -14,11 +16,50 @@
...
+    elif backend_access_control.lower() == "true":
+        # If enabled, ensure that the current graph and vector DBs can support it
+        multi_user_support = check_multi_user_support()
+        if not multi_user_support:
+            raise EnvironmentError(
+                "ENABLE_BACKEND_ACCESS_CONTROL is set to true but the current graph and/or vector databases do not support multi-user access control. Please use supported databases or disable backend access control."
+            )
+        else:
```
pazone: Can we remove this `else: ` and just leave `return True`? In this case we either raise an exception or return true

cognee/context_global_variables.py
```diff
@@ -14,11 +16,50 @@
...
+            return True
+    else:
```
pazone: DIUC that this `else` can be also removed. `return False` is enough, right? 🤔

cognee/context_global_variables.py
```diff
@@ -14,11 +16,50 @@
...
+def check_multi_user_support():
+    graph_db_config = get_graph_context_config()
+    vector_db_config = get_vectordb_context_config()
+    if (
+        graph_db_config["graph_database_provider"] in graph_dbs_with_multi_user_support
+        and vector_db_config["vector_db_provider"] in vector_dbs_with_multi_user_support
+    ):
+        return True
+    else:
+        return False
```
pazone:
```suggestion
    return (
        graph_db_config["graph_database_provider"] in graph_dbs_with_multi_user_support
        and vector_db_config["vector_db_provider"] in vector_dbs_with_multi_user_support
    )
```

cognee/context_global_variables.py
pazone: Question (+warning): Might this be a breaking change in the configuration logic? For example, if a customer has kuzu, lanceDB and `ENABLE_BACKEND_ACCESS_CONTROL` is not set - it will be resolved to `true`. Is that expected?

cognee/context_global_variables.py
dexters1: We can, I prefer the more explicit writing of checks so it resembles switches in C/C++ with a default case, but most of the team doesn't like this style

cognee/context_global_variables.py
dexters1: This was requested to be the default setup, we are aware it will change the existing way of working for old users that haven't specified the setting. 

Also now these users who've worked with disabled backed access control, until they set the ENABLE_BACKEND_ACCESS_CONTROL to False they won't be able to access their existing data. Which is far from ideal.

If you have any suggestions on how we could have all new users have ENABLE_BACKEND_ACCESS_CONTROL set to True as default, but existing users have their settings as they are now would be happy to implement it

Otherwise we were thinking of making a statement on when the Cognee version with this change is released to let the users know about to change and how to have Cognee working as it was before for them (in case it was turned off) (Maybe a deprecation log message about the change as well for users that are not on Discord)

cognee/context_global_variables.py
dexters1: Yes, it was just a personal style choice

cognee/context_global_variables.py
```diff
@@ -14,11 +16,43 @@
...
 
+vector_dbs_with_multi_user_support = ["lancedb"]
+graph_dbs_with_multi_user_support = ["kuzu"]
+
 
 async def set_session_user_context_variable(user):
     session_user.set(user)
 
 
+def check_multi_user_support():
```
pazone: I'd recommend to rename this function to since it represent a flag/state. For example, `multi_user_support_available` or `multi_user_support_possible`

cognee/context_global_variables.py
pazone: If we want to keep the `ENABLE_BACKEND_ACCESS_CONTROL` by default when it's possible - it seems like this is the only way we can go. We must warn the customers about unexpected behavior when `ENABLE_BACKEND_ACCESS_CONTROL` is not set

pazone (approved):
LGTM. I'd appreciate another 👀

PR #1724, opened 2025-11-02
feat: add ontology endpoint in REST API

### LINKED ISSUE #1446: [Feature]: Ontology parameter in REST API
### Problem Statement

There doesn't seem to be a way to tell Cognify to use an Ontology through the REST API.

### Proposed Solution

Either take in the whole file through Cognify. Or have another API call that set's the Ontology file contents once with a named key, then when calling Cognify, we can pass in the named key of the Ontology to use.

### Alternatives Considered

_No response_

### Use Case

Through the SDK this is available but not in the REST API

### Implementation Ideas

_No response_

### Additional Context

_No response_

### Pre-submission Checklist

fahad-shoaib:
## Description

This PR resolves #1446 by adding support to upload ontology files and refer to them in the cognee POST request.

  Implementation Details:
  - New endpoint: POST /api/v1/ontologies for ontology file upload with a simple key parameter
  that can be referenced in POST cognify requests
  - File storage: Ontology files are stored in /tmp/ontologies/{user_id}/ with metadata
  management
  - New service: OntologyService created for file management and metadata handling
  - Resolver: RDFLibOntologyResolver modified to handle file-like objects.

## Type of Change

- [X] New feature (non-breaking change that adds functionality)
- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Other (please specify):

## Pre-submission Checklist

- [X] **I have tested my changes thoroughly before submitting this PR**
- [X] **This PR contains minimal changes necessary to address the issue/feature**
- [X] My code follows the project's coding standards and style guidelines
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if applicable)
- [X] All new and existing tests pass
- [X] I have searched existing PRs to ensure this change hasn't been submitted already
- [X] I have linked any relevant issues in the description
- [X] My commits have clear and descriptive messages

## DCO Affirmation

### DIFF
--- a/cognee/tests/unit/api/test_ontology_endpoint.py
+++ b/cognee/tests/unit/api/test_ontology_endpoint.py
@@ -104,7 +104,7 @@ def test_upload_ontology_unauthorized(mock_get_default_user, client, mock_defaul
 
 
 @patch.object(gau_mod, "get_default_user", new_callable=AsyncMock)
-async def test_upload_multiple_ontologies(mock_get_default_user, client, mock_default_user):
+def test_upload_multiple_ontologies(mock_get_default_user, client, mock_default_user):
     """Test uploading multiple ontology files in single request"""
     import io
 
@@ -132,7 +132,7 @@ async def test_upload_multiple_ontologies(mock_get_default_user, client, mock_de
 
 
 @patch.object(gau_mod, "get_default_user", new_callable=AsyncMock)
-async def test_upload_endpoint_accepts_arrays(mock_get_default_user, client, mock_default_user):
+def test_upload_endpoint_accepts_arrays(mock_get_default_user, client, mock_default_user):
     """Test that upload endpoint accepts array parameters"""
     import io
     import json
@@ -153,7 +153,7 @@ async def test_upload_endpoint_accepts_arrays(mock_get_default_user, client, moc
 
 
 @patch.object(gau_mod, "get_default_user", new_callable=AsyncMock)
-async def test_cognify_with_multiple_ontologies(mock_get_default_user, client, mock_default_user):
+def test_cognify_with_multiple_ontologies(mock_get_default_user, client, mock_default_user):
     """Test cognify endpoint accepts multiple ontology keys"""
     payload = {
         "datasets": ["test_dataset"],
@@ -168,7 +168,7 @@ async def test_cognify_with_multiple_ontologies(mock_get_default_user, client, m
 
 
 @patch.object(gau_mod, "get_default_user", new_callable=AsyncMock)
-async def test_complete_multifile_workflow(mock_get_default_user, client, mock_default_user):
+def test_complete_multifile_workflow(mock_get_default_user, client, mock_default_user):
     """Test complete workflow: upload multiple ontologies → cognify with multiple keys"""
     import io
     import json
@@ -218,7 +218,7 @@ async def test_complete_multifile_workflow(mock_get_default_user, client, mock_d
 
 
 @patch.object(gau_mod, "get_default_user", new_callable=AsyncMock)
-async def test_multifile_error_handling(mock_get_default_user, client, mock_default_user):
+def test_multifile_error_handling(mock_get_default_user, client, mock_default_user):
     """Test error handling for invalid multifile uploads"""
     import io
     import json
@@ -251,7 +251,7 @@ async def test_multifile_error_handling(mock_get_default_user, client, mock_defa
 
 
 @patch.object(gau_mod, "get_default_user", new_callable=AsyncMock)
-async def test_cognify_missing_ontology_key(mock_get_default_user, client, mock_default_user):
+def test_cognify_missing_ontology_key(mock_get_default_user, client, mock_default_user):
     """Test cognify with non-existent ontology key"""
     payload = {
         "datasets": ["test_dataset"],


cognee/api/v1/cognify/routers/get_cognify_router.py
```diff
@@ -68,6 +71,7 @@ async def cognify(payload: CognifyPayloadDTO, user: User = Depends(get_authentic
         - **dataset_ids** (Optional[List[UUID]]): List of existing dataset UUIDs to process. UUIDs allow processing of datasets not owned by the user (if permitted).
         - **run_in_background** (Optional[bool]): Whether to execute processing asynchronously. Defaults to False (blocking).
         - **custom_prompt** (Optional[str]): Custom prompt for entity extraction and graph generation. If provided, this prompt will be used instead of the default prompts for knowledge graph extraction.
+        - **ontology_key** (Optional[str]): Reference to a previously uploaded ontology file to use for knowledge graph construction.
```
hajdul88: Lately we added multifile ontology support to the system so we can attach multiple files at once. Could you extend the API interface so it can accept multiple ontologies? You can find an example for the config in the test_multifile_ontology_with_overlapping_entities unit test

cognee/tests/test_ontology_endpoint.py
```diff
@@ -0,0 +1,95 @@
+import pytest
```
hajdul88: I am not sure in this way these tests get triggered by the CI/CD. Please put them under tests/unit/api/ directory. Unit tests get triggered by basic_tests.yml where we run pytest on the tests/unit directory

cognee/tests/test_ontology_endpoint.py
```diff
@@ -0,0 +1,95 @@
...
+    response = client.post("/api/v1/ontologies", data={"ontology_key": "test"})
+    assert response.status_code == 400
+
+    # Missing key
+    response = client.post("/api/v1/ontologies", files={"ontology_file": ("test.owl", b"xml")})
+    assert response.status_code == 400
+
+
+@patch.object(gau_mod, "get_default_user", new_callable=AsyncMock)
+def test_upload_ontology_unauthorized(mock_get_default_user, client, mock_default_user):
```
hajdul88: Could you maybe e2e as well to make sure that the "happy" workflow is correct in a case of future changes. You can extend the test_cognee_server_start with an additional set of steps (adding file, adding ongoloty, cognifying) + maybe checking if ontology got attached to the cognify process.

fahad-shoaib: Got it. I’m extending the test with the full flow. Quick question though what’s the best way to confirm in the test that the ontology was actually attached to the cognify process? Should I check a specific flag, log entry, or graph property?

hajdul88: Cool, I think the simplest approach is to just check whether there are any nodes with ontology_valid=True in the graph. If an ontology node/subgraph gets connected, it will show up with that flag, so we can easily tell which subgraph comes from the ontology and which comes from the LLM.

fahad-shoaib:
Hi @hajdul88 , I've addressed your feedback by adding multifile support and improving tests - extended test_cognee_server_start.py and moved the ontology test under tests/unit/api. Ready for another review when you are.

cognee/tests/unit/api/test_ontology_endpoint.py
```diff
@@ -0,0 +1,264 @@
...
+    # The current system provides a default user when no explicit authentication is given
+    # This test verifies the system works with conditional authentication
+    assert response.status_code == 200
+    data = response.json()
+    assert data["uploaded_ontologies"][0]["ontology_key"] == unique_key
+    assert "uploaded_at" in data["uploaded_ontologies"][0]
+
+
+@patch.object(gau_mod, "get_default_user", new_callable=AsyncMock)
+async def test_upload_multiple_ontologies(mock_get_default_user, client, mock_default_user):
```
hajdul88: Please make the tests non async here, they dont need to be and in this way our CI/CD is gonna skip them.

These are the tests that are getting skipped:

test_ontology_endpoint.py::test_upload_multiple_ontologies
test_ontology_endpoint.py::test_upload_endpoint_accepts_arrays
test_ontology_endpoint.py::test_cognify_with_multiple_ontologies
test_ontology_endpoint.py::test_complete_multifile_workflow
test_ontology_endpoint.py::test_multifile_error_handling
test_ontology_endpoint.py::test_cognify_missing_ontology_key

cognee/tests/test_cognee_server_start.py
```diff
@@ -7,6 +7,7 @@
 from pathlib import Path
```
hajdul88, line 7: Due to the changes the test_cognee_server_start is failing in our CI pipeline with the following assert error:


FAIL: test_server_is_running (__main__.TestCogneeServerStart.test_server_is_running)
Test that the server is running and can accept connections.
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/cognee/cognee/./cognee/tests/test_cognee_server_start.py", line 116, in test_server_is_running
    self.assertEqual(ontology_response.status_code, 200)
AssertionError: 401 != 200

----------------------------------------------------------------------
Ran 1 test in 48.080s

hajdul88: Tried it locally with neo4j and pg but it seems it fails due to the added ontology related changes. We run the test with basic settings in CI so without defining the graph, vector and reldbs maybe this helps.

hajdul88: I've tried to add 2 ontologies locally, and now the ontology endpoint doesn't work for me. Got this error:

{
  "error": "Expecting value: line 1 column 1 (char 0)"
}

If you want to reproduce it, I added the examples/data/car_and_tech_companies.txt (so basically the text that we have in our ontology example) and tried to add our two owl examples from examples/python/ontology_input_example/. It is also not clear that in this case how the ontology name/description would work as we are adding multiple files at the same time.

fahad-shoaib: Thanks for pointing those out. I missed adding the auth headers in the POST ontology request, which is why the server startup test was failing and I have made the test functions non async. Hopefully the CI tests pass now :)

To add multiple (or single) ontologies, the curl request would be as follows:


curl -X POST "http://localhost:8000/api/v1/add" \
  -F "data=@examples/data/car_and_tech_companies.txt" \
  -F "datasetName=car_tech_companies"

curl --request POST "http://localhost:8000/api/v1/ontologies" \
  -F "ontology_file=@examples/python/ontology_input_example/basic_ontology.owl" \
  -F "ontology_file=@examples/python/ontology_input_example/enriched_medical_ontology_with_classes.owl" \
  -F 'ontology_key=["basic_ontology", "medical_ontology"]'

Then in the cognify process, the keys can be referenced:

curl --request POST "http://localhost:8000/api/v1/cognify" \
  --header 'Content-Type: application/json' \
  --data '{
    "datasets": ["car_tech_companies"],
    "ontology_key": ["basic_ontology", "medical_ontology"],
    "run_in_background": false
  }'

PR #1744, opened 2025-11-06
fix: Resolve issue with empty node set

dexters1:
## Description
Transform default node set value needed for FastAPI to default value needed by Cognee

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/api/v1/add/routers/get_add_router.py
+++ b/cognee/api/v1/add/routers/get_add_router.py
@@ -82,7 +82,9 @@ def get_add_router() -> APIRouter:
                 datasetName,
                 user=user,
                 dataset_id=datasetId,
-                node_set=node_set if node_set else None,
+                node_set=node_set
+                if node_set != [""]
+                else None,  # Transform default node_set endpoint value to None
             )
 
             if isinstance(add_run, PipelineRunErrored):


cognee/api/v1/add/routers/get_add_router.py
```diff
@@ -82,7 +82,9 @@ async def add(
                 datasetName,
                 user=user,
                 dataset_id=datasetId,
-                node_set=node_set if node_set else None,
+                node_set=node_set
```
pazone, line 85: Question: wouldn't it be more handy if we remove empty strings from the list? 
```
node_set = [s for s in node_set if s.strip()]
```

dexters1: I'm not sure tbh, for now I've just handled the default value scenario so if the default Swagger value is sent it will be converted to None.

The default value is always [""] so it would show properly in Swagger and we need it to be set to None.

I also think deeper down there was and maybe still is a direct check "if node_set is None:" (in which case [] would not be considered as None as None is an actual object in python, it would work if it was just checking "if node_set:" for example)

dexters1: I just checked there in fact is a "if node_set is not None:" which would not work with an empty list

PR #1775, opened 2025-11-11
Poc/simple vcrpy example

Geoff-Robin:
## Description
This is a demo of vcrpy to speedup tests.
I've used VCRpy here as a decorator here

## Type of Change
- [ ] Simple demo

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/tests/test_neo4j.py
+++ b/cognee/tests/test_neo4j.py
@@ -20,6 +20,17 @@ logger = get_logger()
     match_on=["method", "uri"],
 )
 async def main():
+    '''
+    VCR.py Configuration:
+        path (str): Path to the cassette file (.yaml) used to store and replay HTTP interactions.
+        record_mode (str): Controls recording behavior.
+            - "all": Record all requests, overwriting existing ones.
+            - "none": Replay only, never record new interactions.
+            - "new_episodes": Replay existing interactions and record new ones if missing.
+            - "once": Record only if cassette doesn’t exist, then replay thereafter.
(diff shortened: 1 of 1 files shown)


Geoff-Robin:
It seems to keep adding new interactions whenever the embedding model is being called. So there is a risk of the yaml file getting large but it can be avoided with better matching strategies.
"new_episodes" records and returns http interactions based on if it's recorded.
[https://vcrpy.readthedocs.io/en/latest/](https://vcrpy.readthedocs.io/en/latest/)
The docs are here.

Vasilije1990:
@pazone please check

pazone:
Hi @Geoff-Robin. Thanks you for the interesting idea! It turns integration tests into isolated unit tests with mocked environment. So we can run integration tests with VCR and we won't actually access the DB or backend. All requests are pre-recorded. 

I have a tiny concern about [Requests matching](https://vcrpy.readthedocs.io/en/latest/configuration.html#request-matching). Imagine a situation with API polling, e.g. progress bars and etc. For example, we process a large file and poll an endpoint like `/file_upload_progress?file_id=123`. It returns 10%, 40% and etc for the same request. The test moves forward when it receives a response with 100%. Would VCR cover the repetitive requests? 

For now, to integrate VCR we need to rework the testing approach. We need integration tests like `test_neo4j` without mocks to ensure the compatibility with Neo4j, for example to test new DB versions. I think it's possible to use VCR as a separate CI step/workflow that requires a bit different approach. I will let you know if we start to use VCR in a separate GH job. 👍

Issue #1784, opened 2025-11-13
[Feature]: Support to pass custom headers in openaiadapter


ketanjain3:
### Problem Statement

add support to pass custom headers in openaiadapter.

### Proposed Solution

Update OpenAIAdapter and its functions to pass headers for Litellm calls

### Use Case

for enterprise clients

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this feature hasn't been requested already
- [x] I have provided a clear problem statement and proposed solution
- [x] I have described my specific use case

memrrdc1974-ops:
It's time for repairs

siillee:
Hey @ketanjain3 , thanks for the request!

Could you please be more specific about what you had in mind with this feature request? Adding custom headers to these functions through **kwargs is easy enough, and makes sense, however they need to be propagated from add/cognify all the way through to the adapter. 

Is this what you had in mind? If not, could you elaborate more on the request? What do you want to achieve, why do you need this, etc.

Thanks!

PR #1802, opened 2025-11-17
feat: add support to pass custom parameters in llm adapters during cognify

siillee:
## Description

A user raised this issue/feature request: https://github.com/topoteretes/cognee/issues/1784
This PR implements this in a way, but I may have misunderstood the request. If I'm correct, I think it makes sense to support this, if our users need it.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

## Summary by CodeRabbit

* **New Features**
  * Enhanced the cognify API and LLM processing pipeline to support passing additional configuration parameters throughout the system, enabling greater customization flexibility for advanced use cases while maintaining full backward compatibility.


cognee/infrastructure/llm/structured_output_framework/litellm_instructor/llm/openai/adapter.py
```diff
@@ -108,7 +108,7 @@ def __init__(
         reraise=True,
     )
     async def acreate_structured_output(
-        self, text_input: str, system_prompt: str, response_model: Type[BaseModel]
+        self, text_input: str, system_prompt: str, response_model: Type[BaseModel], **kwargs
```
dexters1, line 115: These **kwargs can't be specified during Cognification and add time though, would be good to get feedback from user on what he needs and wants with the forwarding

siillee: I was thinking the same thing, my thoughts went to "maybe he wants to call the adapter directly for some reason", but will check with the user

cognee/infrastructure/llm/extraction/knowledge_graph/extract_content_graph.py
```diff
@@ -30,7 +30,7 @@ async def extract_content_graph(
         system_prompt = render_prompt(prompt_path, {}, base_directory=base_directory)
 
     content_graph = await LLMGateway.acreate_structured_output(
-        content, system_prompt, response_model
+        content, system_prompt, response_model, **kwargs
```
borisarzentar, line 33: A lot of argument drilling to enable this. Is there some other way to do this?

siillee: @borisarzentar Yes, I agree. This is still a draft, mostly because I wasn't sure what the user who suggested this wanted. Basically they made a feature request for passing custom headers to OpenAI adapter functions. `**kwargs` was the only way I could think of to do this, since litellm functions later on can handle these extra arguments. Although, I'm not aware of anything in python that prevents argument drilling in this case (except some form of context, similar to the ones in React, I know something like that exists, but not sure if it is the same use-case).

siillee:
I added these kwargs only for default tasks in cognify for now, and also all adapters we have, in order for them not to break.
In the end, if we want something like this, I think it's better to do it via kwargs rather than env variables. We already have a lot of options for customization via env variables (model, endpoint, temperature, max tokens, etc.), so adding another variable with a json/dict inside it doesn't make much sense. Plus, this way users can change their custom parameters between calls more easily.

If we want to have this, we should probably also add it to add and search as well. If we don't want/need this on the other hand, I can just close this PR.

Issue #1812, opened 2025-11-19
[Bug]:  Local llama.cpp + BAML structured output (GLM-4.6): backend instability under graph extraction load


QuantumPlayDev:
<img width="1914" height="928" alt="Image" src="https://github.com/user-attachments/assets/9a587730-cd88-495b-969b-4e100b6d0ca4" />

### Bug Description

Summary
I’m trying to run Cognee 100% locally with:

    • LLM: local llama-server (llama.cpp) exposing an OpenAI-compatible /v1 endpoint
    • Model: GLM-4.6 GGUF (Unsloth) for graph extraction and reasoning
    • Structured output: BAML (STRUCTURED_OUTPUT_FRAMEWORK=BAML)

Basic tests work, but when I run the full cognify() pipeline over a medium-sized docs corpus (OpenCV docs mirrored to Markdown), the llama.cpp back-end becomes unstable and eventually dies. At that point Cognee surfaces a BamlClientHttpError with status code 503.

I’m not sure whether I’m mis-configuring Cognee/BAML for local models, or if this is just too heavy a workload for a single llama.cpp instance, so I’m looking for guidance / best practices.

**Questions**

1. Are there recommended settings for using Cognee + BAML with a local OpenAI-compatible LLM? For example: expected context size, batch size, concurrency, or model size?
2.  Is BAML the recommended structured output backend for local models, or would you currently suggest using LiteLLM + Instructor instead for llama.cpp backends?    
3. Any knobs in Cognee to tune the concurrency / rate of acreate_structured_output calls?        ◦ Right now, extract_graph_from_data uses asyncio.gather(...) and seems to issue several concurrent calls per chunk; I’m not sure what an appropriate concurrency limit would be for a local 2-GPU llama.cpp server.
4. Should I be using a smaller model (7B/14B) for graph extraction and keep GLM-4.6 only for interactive chat?
5. If so, is there any guidance on what scale of model is sufficient for Cognee’s graph extraction tasks?

I’m happy to try alternative configurations (LiteLLM+Instructor, Ollama provider, smaller models, etc.) but would really appreciate some “blessed” reference settings for a fully local setup hitting an OpenAI-compatible LLM.

### Steps to Reproduce

Repro steps

1. Start llama-server with GLM-4.6 GGUF and --jinja (alias glm-4.6-reap) on http://127.0.0.1:8080/v1.
[...]

PR #1813, opened 2025-11-19
fix: Resolve issue with BAML rate limit handling

dexters1:
## Description
Add rate limit handling for BAML

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

## Summary by CodeRabbit

* **Improvements**
  * Centralized rate limiting for LLM and embedding requests to better manage API throughput.
  * Retry policies adjusted: longer initial backoff and reduced max retries to improve stability under rate limits.

* **Chores**
  * Refactored rate-limiting implementation from decorator-based to context-manager usage across services.

* **Tests**
  * Unit tests and mocks updated to reflect the new context-manager rate-limiting approach.


cognee/infrastructure/llm/structured_output_framework/baml/baml_src/extraction/acreate_structured_output.py
```diff
@@ -10,12 +18,17 @@
...
 )
 from cognee.infrastructure.llm.structured_output_framework.baml.baml_client import b
-from pydantic import BaseModel
-
+import logging
 
 logger = get_logger()
 
 
+@retry(
```
pazone, line 27: I have a couple of concerns about this approach. It feels like we are starting to add LLM-specific workarounds into our codebase and I’m not sure that’s ideal.
Are we sure exponential retries make sense for all cases with structured output? For example, if we send 1000 requests in one second - would all the failed ones get retried?
How do we know this retry strategy will work well with other LLMs?
Might be worth looking at other ways to handle rate limits too, like a reverse proxy or a message queue

dexters1: Yes, all failed requests would get retried if a 1000 different backend requests would be made, but we should use a proxy or a similar mechanism for a production scenario and not rely on this approach. This is just a last resort for a single user in a non production scenario where he doesn't need anything advanced and it will work fine in that scenario.

Btw it can be changed to handle wait differently, we can provide any function to handle the waiting, so we can make it work in a 1000 different request scenario with exponential backoff (if our wait function would work with a global variable for example). Might make sense to also update the wait to handle the 1000 different request scenario as well

Vasilije1990: @dexters1 please update it based on the feedback of Pavel.
I am worried that if we don't specify options in our docs, people will use this in production. Maybe add a ticket to handle and suggest best ways to run it in prod in Mintlify and pass to Lazar for review

pazone: @Vasilije1990 We proposed [an alternative way](https://github.com/topoteretes/cognee/issues/1812#issuecomment-3558273250). Also @dexters1 recommends LiteLLM as a proxy. Let's see if it solves the issue and then decide the future of this PR. Perhaps, we can avoid the rate limiter on our side

dexters1: I would this merge this PR to the dev branch as a fail safe for beginner developers. Would be good to expand the wait function to work globally as well when we get the time

pazone: I'm a bit concerned. We introduce exponential retries and expect it to solve the problem of the rate limiter. The initial problem is that local LLMs can't survive under a high RPM, and we are expecting them to recover by introducing retries (sending additional requests).

borisarzentar:
@pazone Should we close this one as the issue is also closed?

dexters1:
> @pazone Should we close this one as the issue is also closed?

We need this mechanism as well for casual users to handle when rate limit is hit on the Cognee side. We should enhance it to work on a global request level to not spam too many requests in parallel if they can't be handled and we can add optional RPM support. This can be achieved by specifying our own wait function in the tenacity requests wait criteria.

The reason an issue was raised about this only for BAML is because we don't have any retry mechanism on the BAML side and we have this already on the embedding and instructor side.

pazone:
Bringing more details to the PR

### What problem is attempted to solve: 
If a user runs an LLM locally with limited resources it might not handle the default requests rate and eventually start to respond with 500s or too many requests depending on the implementation. 

### What approaches have been discussed 
* Use LiteLLM proxy locally that has a rate limiter
* Use nginx proxy to limit RPM
* Implement the rate limiter on the Cognee side. Considered as a controversial solution, but might be helpful for the local try-outs

### What does this PR do
The PR introduces retries with exponential backoff. Retries failed requests to LLM with increasing delays.
I'm probably missing something. I can't understand how exponential retries may help us when the local LLM dies under the load.  

Adding @borisarzentar and @hajdul88 to review. 
Let's make sure that we solve the initial problem without introducing additional functionality that 100% helps.
IMHO, retrying requests to any throttled service will lead us to much bigger and obvious problems

dexters1:
> Bringing more details to the PR
> 
> ### What problem is attempted to solve:
> If a user runs an LLM locally with limited resources it might not handle the default requests rate and eventually start to respond with 500s or too many requests depending on the implementation.
> 
> ### What approaches have been discussed
> * Use LiteLLM proxy locally that has a rate limiter
> * Use nginx proxy to limit RPM
> * Implement the rate limiter on the Cognee side. Considered as a controversial solution, but might be helpful for the local try-outs
> 
> ### What does this PR do
> The PR introduces retries with exponential backoff. Retries failed requests to LLM with increasing delays. I'm probably missing something. I can't understand how exponential retries may help us when the local LLM dies under the load.
> 
> Adding @borisarzentar and @hajdul88 to review. Let's make sure that we solve the initial problem without introducing additional functionality that 100% helps.

It's not necessarily for when local LLMs die, it's

> Bringing more details to the PR
> 
> ### What problem is attempted to solve:
> If a user runs an LLM locally with limited resources it might not handle the default requests rate and eventually start to respond with 500s or too many requests depending on the implementation.
> 
> ### What approaches have been discussed
> * Use LiteLLM proxy locally that has a rate limiter
> * Use nginx proxy to limit RPM
> * Implement the rate limiter on the Cognee side. Considered as a controversial solution, but might be helpful for the local try-outs
> 
> ### What does this PR do
> The PR introduces retries with exponential backoff. Retries failed requests to LLM with increasing delays. I'm probably missing something. I can't understand how exponential retries may help us when the local LLM dies under the load.
> 
> Adding @borisarzentar and @hajdul88 to review. Let's make sure that we solve the initial problem without introducing additional functionality that 100% helps.

Exponential retries is not additional functionality, it's something we already have and use for instructor where we use it to handle OpenAI, local model and other cloud model calls. This PR just introduces the same mechanism to the BAML side as it was missing here, it's not here to help when a local LLM model dies, it's mostly here to handle when rate limit has been reached when using BAML for OpenAI and other cloud providers as otherwise it will break Cognee when rate limit would be reached. It might also help with local models as no one complained about local LLMs dying when using instructor but it's just a minor part of the mechanism.

For the local model use I will add optional simple RPM control on the Cognee side

PR #1876, opened 2025-12-08
feat: make pipeline processing cache optional

dexters1:
## Description
Make the pipeline cache mechanism optional, have it turned off by default but use it for add and cognify like it has been used until now

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Other (please specify):

## Pre-submission Checklist

- [ x I have searched existing PRs to ensure this change hasn't been submitted already

## DCO Affirmation

## Summary by CodeRabbit

* **New Features**
  * Introduced pipeline caching across ingestion, processing, and custom pipeline flows with per-run controls to enable or disable caching.
  * Added an option for incremental loading in custom pipeline runs.

* **Behavior Changes**
  * One pipeline path now explicitly bypasses caching by default to always re-run when invoked.
  * Disabling cache forces re-processing instead of early exit; cache reset still enables re-execution.

* **Tests**
  * Added tests validating caching, non-caching, and cache-reset re-execution behavior.

* **Chores**
  * Added CI job to run pipeline caching tests.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -97,6 +97,8 @@ DB_NAME=cognee_db
 
 # Default (local file-based)
 GRAPH_DATABASE_PROVIDER="kuzu"
+# Handler for multi-user access control mode, it handles how should the mapping/creation of separate DBs be handled per Cognee dataset
+GRAPH_DATASET_DATABASE_HANDLER="kuzu"
 
 # -- To switch to Remote Kuzu uncomment and fill these: -------------------------------------------------------------
 #GRAPH_DATABASE_PROVIDER="kuzu"
@@ -121,6 +123,8 @@ VECTOR_DB_PROVIDER="lancedb"
 # Not needed if a cloud vector database is not used
 VECTOR_DB_URL=
 VECTOR_DB_KEY=
+# Handler for multi-user access control mode, it handles how should the mapping/creation of separate DBs be handled per Cognee dataset
+VECTOR_DATASET_DATABASE_HANDLER="lancedb"
 
 ################################################################################
 # 🧩 Ontology resolver settings
--- a/.github/workflows/db_examples_tests.yml
+++ b/.github/workflows/db_examples_tests.yml
@@ -61,6 +61,7 @@ jobs:
       - name: Run Neo4j Example
         env:
           ENV: 'dev'
+          ENABLE_BACKEND_ACCESS_CONTROL: 'false'
           LLM_MODEL: ${{ secrets.LLM_MODEL }}
           LLM_ENDPOINT: ${{ secrets.LLM_ENDPOINT }}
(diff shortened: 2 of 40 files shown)


hajdul88 (dismissed):
LGTM, would be nice if someone else would also take a look

cognee/modules/pipelines/operations/pipeline.py
```diff
@@ -64,6 +69,7 @@ async def run_pipeline_per_dataset(
     data=None,
     pipeline_name: str = "custom_pipeline",
     context: dict = None,
+    pipeline_cache=False,
```
borisarzentar: `pipeline_cache` is a bit confusing, can we use the name that is more descriptive? Also, something like `use_pipeline_cache` associates more to a boolean. Since this is high-level API, we will have to live with this for quite some time.

cognee/modules/pipelines/operations/pipeline.py
```diff
@@ -77,8 +83,18 @@ async def run_pipeline_per_dataset(
...
         # return status information to async generator and finish execution
-        yield process_pipeline_status
-        return
+        if pipeline_cache:
+            # If pipeline caching is enabled we do not proceed with re-processing
+            yield process_pipeline_status
+            return
+        else:
+            # If pipeline caching is disabled we always return pipeline started information and proceed with re-processing
+            yield PipelineRunStarted(
```
borisarzentar, line 92: Won't then `PipelineRunStarted` be yielded two times? Once from here and second time from `run_tasks`.

dexters1: Yes, I'm not sure which yield we return as a response when running cognee in the background I assumed it was this one so returned started info so it can be forwarded to users

cognee/modules/pipelines/operations/pipeline.py
dexters1: I can rename it to use_pipeline_cache

cognee/modules/pipelines/operations/pipeline.py
borisarzentar: In that case `run_pipeline_as_background_process` will also yield started event. We could check in tests.

cognee/modules/pipelines/operations/pipeline.py
borisarzentar: Yeah, let's do that then.

cognee/modules/pipelines/operations/pipeline.py
dexters1: I just checked, it would break if it didn't return PipelineRunStarted here as we need to return the info when running Cognee in background

PR #2002, opened 2026-01-15
feat: enable batch queries in all graph completion retrievers

siillee:
## Description

Enabled batch querying for graph completion retrievers.

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

## Release Notes

* **New Features**
  * Added batch query processing capability to handle multiple queries simultaneously across retrieval modules.
  * Introduced query validation framework to ensure proper input formatting and error handling.

* **Refactor**
  * Restructured retrieval pipeline into three-stage process for improved modularity and extensibility.

* **Tests**
  * Enhanced test coverage for batch query scenarios and validation workflows.

### DIFF
--- a/cognee/modules/retrieval/graph_completion_context_extension_retriever.py
+++ b/cognee/modules/retrieval/graph_completion_context_extension_retriever.py
@@ -192,6 +192,8 @@ class GraphCompletionContextExtensionRetriever(GraphCompletionRetriever):
 
             round_idx += 1
 
+        if query:
+            return finished_queries_states[query].triplets
         return [finished_queries_states[batched_query].triplets for batched_query in query_batch]
 
     async def get_completion_from_context(
@@ -266,23 +268,3 @@ class GraphCompletionContextExtensionRetriever(GraphCompletionRetriever):
             )
 
         return completion if query_batch else [completion]
-
-    async def get_context_from_objects(
-        self,
-        query: Optional[str] = None,
-        query_batch: Optional[List[str]] = None,
-        retrieved_objects: List[Edge] | List[List[Edge]] = None,
-    ) -> Union[str, List[str]]:
-        triplets = retrieved_objects
-
-        if query:
-            query_batch = [query]
-
-        # Check if all triplets are empty, in case of batch queries
-        if not triplets or all(len(batched_triplets) == 0 for batched_triplets in triplets):
-            logger.warning("Empty context was provided to the completion")
-            return ["" for _ in query_batch]
-
-        return await asyncio.gather(
-            *[self.resolve_edges_to_text(batched_triplets) for batched_triplets in triplets]
-        )
--- a/cognee/modules/retrieval/graph_completion_cot_retriever.py
+++ b/cognee/modules/retrieval/graph_completion_cot_retriever.py
@@ -146,6 +146,8 @@ class GraphCompletionCotRetriever(GraphCompletionRetriever):
                 session_id=self.session_id,
             )
 
+        if query:
+            return triplets[0]
         return triplets
 
     async def _run_cot_completion(
@@ -362,26 +364,9 @@ class GraphCompletionCotRetriever(GraphCompletionRetriever):
             - List[str]: A list containing the generated answer to the user's query.
         """
 
-        if not retrieved_objects or all(len(triplet) == 0 for triplet in retrieved_objects):
+        if not retrieved_objects or (
+            query_batch and all(len(triplet) == 0 for triplet in retrieved_objects)
+        ):
             raise CogneeValidationError("No context retrieved to generate completion.")
         completion = self.completion
         return completion
-
-    async def get_context_from_objects(
-        self,
-        query: Optional[str] = None,
-        query_batch: Optional[List[str]] = None,
-        retrieved_objects=None,
-    ) -> Union[str, List[str]]:
-        triplets = retrieved_objects
-
-        if query:
-            query_batch = [query]
-
-        if not triplets or all(len(batched_triplets) == 0 for batched_triplets in triplets):
-            logger.warning("Empty context was provided to the completion")
-            return ["" for _ in query_batch]
-
-        return await asyncio.gather(
-            *[self.resolve_edges_to_text(batched_triplets) for batched_triplets in triplets]
-        )
--- a/cognee/tests/integration/retrieval/test_graph_completion_retriever_context_extension.py
+++ b/cognee/tests/integration/retrieval/test_graph_completion_retriever_context_extension.py
@@ -170,8 +170,8 @@ async def test_graph_completion_extension_context_simple(setup_test_environment_
 
     context = await retriever.get_context_from_objects(query=query, retrieved_objects=triplets)
 
-    assert "Mike Broski --[works_for]--> Canva" in context[0], "Failed to get Mike Broski"
-    assert "Christina Mayer --[works_for]--> Canva" in context[0], "Failed to get Christina Mayer"
+    assert "Mike Broski --[works_for]--> Canva" in context, "Failed to get Mike Broski"
+    assert "Christina Mayer --[works_for]--> Canva" in context, "Failed to get Christina Mayer"
 
     answer = await retriever.get_completion_from_context(
         query=query, retrieved_objects=triplets, context=context
@@ -193,9 +193,9 @@ async def test_graph_completion_extension_context_complex(setup_test_environment
 
     context = await retriever.get_context_from_objects(query=query, retrieved_objects=triplets)
 
-    assert "Mike Rodger --[works_for]--> Figma" in context[0], "Failed to get Mike Rodger"
-    assert "Ike Loma --[works_for]--> Figma" in context[0], "Failed to get Ike Loma"
-    assert "Jason Statham --[works_for]--> Figma" in context[0], "Failed to get Jason Statham"
+    assert "Mike Rodger --[works_for]--> Figma" in context, "Failed to get Mike Rodger"
+    assert "Ike Loma --[works_for]--> Figma" in context, "Failed to get Ike Loma"
+    assert "Jason Statham --[works_for]--> Figma" in context, "Failed to get Jason Statham"
 
     answer = await retriever.get_completion_from_context(
         query=query, retrieved_objects=triplets, context=context
@@ -216,7 +216,7 @@ async def test_get_graph_completion_extension_context_on_empty_graph(setup_test_
     triplets = await retriever.get_retrieved_objects(query)
 
     context = await retriever.get_context_from_objects(query=query, retrieved_objects=triplets)
-    assert context == [""], "Context should be empty on an empty graph"
+    assert context == "", "Context should be empty on an empty graph"
 
     answer = await retriever.get_completion_from_context(
         query=query, retrieved_objects=triplets, context=context
--- a/cognee/tests/integration/retrieval/test_graph_completion_retriever_cot.py
+++ b/cognee/tests/integration/retrieval/test_graph_completion_retriever_cot.py
@@ -163,8 +163,8 @@ async def test_graph_completion_cot_context_simple(setup_test_environment_simple
 
     context = await retriever.get_context_from_objects(query=query, retrieved_objects=triplets)
 
-    assert "Mike Broski --[works_for]--> Canva" in context[0], "Failed to get Mike Broski"
-    assert "Christina Mayer --[works_for]--> Canva" in context[0], "Failed to get Christina Mayer"
+    assert "Mike Broski --[works_for]--> Canva" in context, "Failed to get Mike Broski"
+    assert "Christina Mayer --[works_for]--> Canva" in context, "Failed to get Christina Mayer"
 
     answer = await retriever.get_completion_from_context(
         query=query, retrieved_objects=triplets, context=context
@@ -186,9 +186,9 @@ async def test_graph_completion_cot_context_complex(setup_test_environment_compl
 
     context = await retriever.get_context_from_objects(query=query, retrieved_objects=triplets)
 
-    assert "Mike Rodger --[works_for]--> Figma" in context[0], "Failed to get Mike Rodger"
-    assert "Ike Loma --[works_for]--> Figma" in context[0], "Failed to get Ike Loma"
-    assert "Jason Statham --[works_for]--> Figma" in context[0], "Failed to get Jason Statham"
+    assert "Mike Rodger --[works_for]--> Figma" in context, "Failed to get Mike Rodger"
+    assert "Ike Loma --[works_for]--> Figma" in context, "Failed to get Ike Loma"
+    assert "Jason Statham --[works_for]--> Figma" in context, "Failed to get Jason Statham"
 
     answer = await retriever.get_completion_from_context(
         query=query, retrieved_objects=triplets, context=context
@@ -209,7 +209,7 @@ async def test_get_graph_completion_cot_context_on_empty_graph(setup_test_enviro
     triplets = await retriever.get_retrieved_objects(query)
 
     context = await retriever.get_context_from_objects(query=query, retrieved_objects=triplets)
-    assert context == [""], "Context should be empty on an empty graph"
+    assert context == "", "Context should be empty on an empty graph"
 
     with pytest.raises(CogneeValidationError):
         await retriever.get_completion_from_context(


cognee/modules/retrieval/graph_completion_context_extension_retriever.py
```diff
@@ -56,8 +56,9 @@ def __init__(
 
     async def get_completion(
         self,
-        query: str,
-        context: Optional[List[Edge]] = None,
+        query: Optional[str] = None,
+        query_batch: Optional[List[str]] = None,
```
dexters1: Do we want to separate this as a new parameter or just allow query to be a str or a list of str?

siillee: We can do it either way I think. This is how it was done for brute_force_triplet_search, so in the interest of not accidentally breaking anything else now, I kept it similar.

cognee/modules/retrieval/graph_completion_context_extension_retriever.py
```diff
@@ -91,46 +92,102 @@ async def get_completion(
...
+                f"number of unique retrieved triplets for each query : {new_sizes}"
             )
 
             round_idx += 1
 
+        # Reset variables for the final generations. They contain the final state
+        # of triplets and contexts for each query, after all extension iterations.
+        query_batch = original_query_batch
+        context_texts = saved_context_texts if len(saved_context_texts) > 0 else context_texts
+        triplets = saved_triplets if len(saved_triplets) > 0 else triplets
```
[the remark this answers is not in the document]
siillee: This is a valid point, currently working on a fix.

cognee/modules/retrieval/graph_completion_retriever.py
```diff
@@ -139,7 +148,7 @@ async def get_context(self, query: str) -> List[Edge]:
             logger.warning("Search attempt on an empty knowledge graph")
             return []
 
-        triplets = await self.get_triplets(query)
+        triplets = await self.get_triplets(query, query_batch)
```
lxobr: I think it would be a good idea to enforce that exactly one of query or query_batch is provided (mirror brute_force_triplet_search.py lines 149-159).

cognee/modules/retrieval/graph_completion_retriever.py
```diff
@@ -139,7 +148,7 @@ async def get_context(self, query: str) -> List[Edge]:
...
 
         if len(triplets) == 0:
```
lxobr: This check will probably fails in batch mode where triplets is List[List[Edge]]. Maybe check based on mode: if query_batch is not None, check all(len(t) == 0 for t in triplets); otherwise check len(triplets) == 0?

cognee/modules/retrieval/graph_completion_retriever.py
lxobr: Please check whether this function needs updating in query_batch mode. If yes, please either update it, or create a follow up issue.

cognee/modules/retrieval/graph_completion_retriever.py
```diff
@@ -183,9 +193,16 @@ async def get_completion(
         triplets = context
 
         if triplets is None:
-            triplets = await self.get_context(query)
+            triplets = await self.get_context(query, query_batch)
 
-        context_text = await resolve_edges_to_text(triplets)
+        context_text = ""
+        context_texts = ""
```
lxobr: For naming consistency, I propose we use context_text_batch instead of context_texts.

cognee/modules/retrieval/graph_completion_retriever.py
```diff
@@ -183,9 +193,16 @@ async def get_completion(
...
+        if triplets and isinstance(triplets[0], list):
```
lxobr: This seems like a fragile way to detect mode (probably fails if triplets is empty or first element isn't a list). Instead, maybe check query_batch is not None to determine mode?

cognee/modules/retrieval/graph_completion_retriever.py
```diff
@@ -208,14 +225,29 @@ async def get_completion(
...
+                    *[
+                        generate_completion(
+                            query=query,
+                            context=context,
+                            user_prompt_path=self.user_prompt_path,
+                            system_prompt_path=self.system_prompt_path,
+                            system_prompt=self.system_prompt,
+                            response_model=response_model,
+                        )
+                        for query, context in zip(query_batch, context_texts)
```
lxobr: To avoid confusion, consider renaming this query variable, as it does not refer to the method's query variable.

lxobr (commented):
I reviewed the graph_completion_retriever and left comments there. Most of the same feedback should apply to the other two retrievers, too. Please apply the analogous changes.

Since the other two retrievers are closer to @hajdul88 's area, I would defer to their approval there. One note from my side, though: the extension retriever's methods are getting pretty bloated. It may be worth refactoring now, but I am fine with creating a follow up issue for that, too, as we want to merge this as soon as possible

cognee/modules/retrieval/graph_completion_retriever.py
siillee: Agreed, I somehow missed this, good catch.

cognee/modules/retrieval/graph_completion_retriever.py
siillee: I added the checks, but I think they are unnecessary in some places, particularly the context extension retriever. For now, we can leave it, and then rework it so it makes sense everywhere. I agree that it's maybe missing somewhere, I thought about this, and went the route with the least amount of code duplication.

siillee:
> I reviewed the graph_completion_retriever and left comments there. Most of the same feedback should apply to the other two retrievers, too. Please apply the analogous changes.
> 
> Since the other two retrievers are closer to @hajdul88 's area, I would defer to their approval there. One note from my side, though: the extension retriever's methods are getting pretty bloated. It may be worth refactoring now, but I am fine with creating a follow up issue for that, too, as we want to merge this as soon as possible

The context extension retriever is the most complex when it comes to these batched queries. This is a version with less code duplication, but maybe a bit more complex code. Essentially, we treat everything as a batch of queries, even when we have a single query. @hajdul88 would love your thoughts on this, I thought about making all retrievers act like this, but that's a refactor that can be done later.

cognee/modules/retrieval/graph_completion_retriever.py
hajdul88: Why do you think it is unnecessary in context extension retriever? I agree query and query_batch should be mututally exclusive so only one of them should be allowed at once.

cognee/modules/retrieval/graph_completion_retriever.py
```diff
@@ -139,17 +149,35 @@ async def get_context(self, query: str) -> List[Edge]:
...
+                    logger.warning(
+                        f"Empty context was provided to the completion for the query: {batched_query}"
+                    )
+            entity_nodes_batch = []
+            for batched_triplets in triplets:
+                entity_nodes_batch.append(get_entity_nodes_from_triplets(batched_triplets))
+
+            await asyncio.gather(
+                *[
+                    update_node_access_timestamps(batched_entity_nodes)
```
hajdul88: I am not a fan of updating the same node access timestamps paralell, is this feature enabled by default btw? My expectation is that this can cause deadlocks withing Neo4j when querying 100-200 queries at the same time. Maybe am wrong

cognee/modules/retrieval/graph_completion_retriever.py
hajdul88: I agree we should split the logic here based on the parameter (query of query_batch)

cognee/modules/retrieval/graph_completion_retriever.py
```diff
@@ -208,14 +248,29 @@ async def get_completion(
...
         else:
-            completion = await generate_completion(
-                query=query,
-                context=context_text,
-                user_prompt_path=self.user_prompt_path,
-                system_prompt_path=self.system_prompt_path,
-                system_prompt=self.system_prompt,
-                response_model=response_model,
-            )
+            if query_batch and len(query_batch) > 0:
```
hajdul88: if session_save is enabled, the multiquery search is not handled.

To enable it you have to set these envs:

CACHING=True
CACHE_BACKEND=redis

host redis locally and since this is not exposed in search.py, set the session user context var like this (This is done in search.py):

user = await get_default_user()
await set_session_user_context_variable(user)

hajdul88: For now if I pass multiple queries (query param is none), sessions and answers are generated incorrecly returning 1 answer generated without any context.

cognee/modules/retrieval/graph_completion_cot_retriever.py
```diff
@@ -202,6 +204,10 @@ async def get_completion(
 
             - List[str]: A list containing the generated answer to the user's query.
         """
+        query_validation = validate_queries(query, query_batch)
+        if not query_validation[0]:
```
hajdul88: Small thing but I am not sure this is a nice way to validate, now its returning a tuple and you are using the tuple positions for raising valueError

cognee/modules/retrieval/graph_completion_cot_retriever.py
```diff
@@ -213,19 +219,43 @@ async def get_completion(
...
+        if query_batch and len(query_batch) > 0:
+            if not context_batch:
+                # Having a list is necessary to zip through it
+                context_batch = []
+                for _ in query_batch:
+                    context_batch.append(None)
+
+            completion_batch = await asyncio.gather(
+                *[
+                    self._run_cot_completion(
```
hajdul88: This is not gonna be using the centralized graph projection, you are basically calling paralell get_context calls for each query alone.

If you go to graph_completion_cot_retriever.py line 120, there the get_context will be called with a single query which triggers the projection.

cognee/modules/retrieval/graph_completion_context_extension_retriever.py
```diff
@@ -89,54 +91,121 @@ async def get_completion(
             - List[str]: A list containing the generated answer based on the query and the
               extended context.
         """
-        triplets = context
+        # TODO: This may be unnecessary in this retriever, will check later
+        query_validation = validate_queries(query, query_batch)
```
hajdul88: Why would it be unnecessary? I think it is needed.

cognee/modules/retrieval/graph_completion_cot_retriever.py
hajdul88: @lxobr maybe COT stopping criteria can be removed and then they would call the get_contexts synchronized + they would do the same amount of rounds. What do you think?

cognee/modules/retrieval/graph_completion_cot_retriever.py
```diff
@@ -213,19 +219,43 @@ async def get_completion(
...
-            max_iter=max_iter,
-            response_model=response_model,
-        )
+        context_batch = context
+        completion_batch = []
+        if query_batch and len(query_batch) > 0:
+            if not context_batch:
+                # Having a list is necessary to zip through it
+                context_batch = []
+                for _ in query_batch:
```
hajdul88: Why do we need this here? wouldn't it be enough to have the same logic as in the graph completion retriever?

cognee/modules/retrieval/graph_completion_cot_retriever.py
lxobr: I think that's a quite good idea for now, though I would still treat it as a temporary solution. If we go this way, I would assume we would do COT without the stopping criteria  for batch searches. In any case, we should touch base async and create follow up issues for a longer term solution.

cognee/modules/retrieval/graph_completion_context_extension_retriever.py
siillee: At first I thought I needed it, then I thought maybe not for this retriever. In the interest of saving time now, I just added a comment for myself to check it later, and left the check in the code to be safe.

cognee/modules/retrieval/graph_completion_cot_retriever.py
siillee: I can change it to not use tuple positions, but just get the two returned values and check them.
Alternatively, I can just raise the error in the `validate_queries` helper function, but to me that seemed less nice, not sure.

cognee/modules/retrieval/graph_completion_context_extension_retriever.py
```diff
@@ -89,54 +91,121 @@ async def get_completion(
...
+            context_text_batch = await asyncio.gather(
+                *[self.resolve_edges_to_text(triplets) for triplets in triplets_batch]
+            )
+
+            new_sizes = [len(triplets) for triplets in triplets_batch]
+
+            for i, (batched_query, prev_size, new_size, triplets, context_text) in enumerate(
+                zip(query_batch, prev_sizes, new_sizes, triplets_batch, context_text_batch)
+            ):
+                finished_queries_data[query] = (triplets, context_text)
```
hajdul88: I am not sure if this is correct, I ran 2 queries and here the finished queries data I have None key values in the dict.

cognee/modules/retrieval/graph_completion_cot_retriever.py
hajdul88: I agree its a temporary solution. The context_extension have some semi done logic for threating the paralell branches differently and optionally stop some of them. Similar approach could be implemented in COT as well

cognee/modules/retrieval/graph_completion_retriever.py
siillee: I did not add batch queries to sessions and interaction saving yet, I can add it, and tests for it, if it is needed urgently now. I made a separate ticket for that, again in the interest of time.

cognee/modules/retrieval/graph_completion_retriever.py
siillee: Initially it seemed to me like I would just be adding duplicate checks, thus unnecessarily duplicating the code. Some time after leaving this comment I wasn't sure, hence my TODO comment in the context extension retriever.

cognee/modules/retrieval/graph_completion_retriever.py
siillee: Hm, maybe this is the cause. I think it's not enabled by default, one would need to set `ENABLE_LAST_ACCESSED` to true. 
And I agree, I didn't think much about this tbh, I will change this to be done sequentially, or something smarter.

cognee/modules/retrieval/graph_completion_cot_retriever.py
siillee: Was failing when context was empty, so zip was being skipped. I could just resolve the context to text, idk why this wasn't done before now that I think about it.

EDIT: since the logic for the batch queries in cot retriever needs changing, this will be deleted and not needed anyways.

cognee/modules/retrieval/graph_completion_retriever.py
```diff
@@ -139,17 +149,38 @@ async def get_context(self, query: str) -> List[Edge]:
...
-        # context = await self.resolve_edges_to_text(triplets)
+            # Remove duplicates and update node access, if it is enabled
+            for batched_entity_nodes in entity_nodes_batch:
+                # from itertools import chain
+                #
+                # flattened_entity_nodes = list(chain.from_iterable(entity_nodes_batch))
+                # entity_nodes = list(set(flattened_entity_nodes))
 
-        entity_nodes = get_entity_nodes_from_triplets(triplets)
+                await update_node_access_timestamps(batched_entity_nodes)
```
hajdul88: Question couldn't we just skip this if its disabled? Now we are iterating through the entity nodes batch just to check whether this feature is enabled for each nodes, which is not a nice thing to do.

cognee/modules/retrieval/graph_completion_cot_retriever.py
```diff
@@ -109,72 +132,167 @@ async def _run_cot_completion(
...
+            if context:
+                context = [context]
+
+        triplets_batch = context
+
+        # dict containing query -> QueryState key-value pairs
+        # For every query, we save necessary data so we can execute requests in parallel
+        query_state_tracker = {}
+        for batched_query in query_batch:
+            query_state_tracker[batched_query] = QueryState()
```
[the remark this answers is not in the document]
hajdul88: True, this is not solved

hajdul88 (changes requested):
It is much better than the previous one, changes are what we discussed +

- Handling same queries in a list
- Adding Cognee compatible exception handling
- Raise exception when session is enabled and query_batch is used
- Add followup question to the query state tracker
- Maybe separate QueryState, now we have two QueryState classes one for context extension other one for cot.

Checked the logic for both cot and context extension with multiquery looks okay, and tested the memory usage with 50 parallel questions in all the retrievers.

cognee/modules/retrieval/graph_completion_context_extension_retriever.py
```diff
@@ -153,18 +241,36 @@ async def get_completion(
...
+            # Make sure answers are returned for duplicate queries, in the order they were asked.
+            for batched_query, batched_completion in zip(
+                finished_queries_states.keys(), completion_batch
+            ):
+                finished_queries_states[batched_query].completion = batched_completion
+
+            for batched_query in query_batch:
+                result_completion_batch.append(finished_queries_states[batched_query].completion)
+
+        # TODO: Do batch queries for save interaction
```
hajdul88: When are we planning to add these? We talked about sessions that for multiquery its gonna throw exception for now, but what about save interaction? I feel this is not hard to add basically you have to just call it for each QA pair no? @lxobr what do you think?

siillee: I added an exception for that as well, Lazar mentioned a separate PR for that. It can be the highest priority, I just didn't add it to this PR as discussed.

Shouldn't be hard, it seemed more urgent to finish the regular stuff.

hajdul88: I see now, we are throwing errors for save interaction as well. It is fine for now but has to be a followup on this, the save_interaction is not hard. Sessions are more tricky

PR #2021, opened 2026-01-23
refactor: Refactor search/retrievers 

dexters1:
…e for selected retriever

## Description
Refactor Cognee retriever selection mechanism so it will instantiate only one retriever instance

## Type of Change

- [ ] Documentation update
- [ ] Code refactoring
- [ ] Performance improvement
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

## Summary by CodeRabbit

* **New Features**
  * Standardized SearchResultPayload for search responses.
  * New retriever factory and get_retriever_output entry; retriever_specific_config supported.

* **Refactor**
  * Retrieval pipeline reorganized to: fetch retrieved objects → build context → generate completion.
  * Retrievers now expose a three-step API and support optional session_id/response_model and iteration controls.

* **Removals**
  * Legacy no-access-control search path, legacy result preparer, base graph retriever abstraction, FEEDBACK enum, and an end-to-end feedback-enrichment test.

* **Tests**
  * Integration and unit tests updated to the new retrieval API and payload model.

### DIFF
--- a/cognee/tests/test_search_db.py
+++ b/cognee/tests/test_search_db.py
@@ -388,9 +388,11 @@ async def test_e2e_search_results_and_wrappers(e2e_state):
             )
             assert wrapper.get("dataset_id"), f"{name}: missing dataset_id in wrapper"
             assert wrapper.get("dataset_name") == "test_dataset"
-            result_payload = wrapper.get("search_result")
+            result_payload = wrapper.get("text_result")
         else:
-            result_payload = search_results[0]
+            entry = search_results[0]
+            assert isinstance(entry, dict), f"{name}: expected dict entries"
+            result_payload = entry.get("text_result")
 
         text_blob = str(result_payload)
         assert text_blob.strip()


cognee/modules/retrieval/chunks_retriever.py
```diff
@@ -28,7 +28,56 @@ def __init__(
...
+            - query (str): The query string to be used for generating a completion.
+            - retrieved_objects (Any): The retrieved objects to be used for generating a completion.
+            - context (Any): The context to be used for generating a completion.
+
+        Returns:
+        --------
+
+            - List[dict]: A list of payloads of found chunks.
+        """
+        # TODO: Do we want to generate a completion using LLM here?
```
hajdul88, line 51: As far as I know chunks retriever is only to get the closest chunks, this is basically the same as the get_retrieved_objects of our simple RAG endpoint.

hajdul88: Not in the scope of this PR but in long term maybe it would make sense to provide a only_objects setting in the search interface and then this retriever could be replaced. I am just thinking here.

cognee/modules/search/methods/get_search_type_retriever_instance.py
```diff
@@ -0,0 +1,252 @@
...
+    system_prompt = kwargs.get("system_prompt")
+    node_type = kwargs.get("node_type", NodeSet)
+    node_name = kwargs.get("node_name")
+    save_interaction = kwargs.get("save_interaction", False)
+    wide_search_top_k = kwargs.get("wide_search_top_k", 100)
+    triplet_distance_penalty = kwargs.get("triplet_distance_penalty", 3.5)
+    session_id = kwargs.get("session_id")
+
+    # Registry mapping search types to their corresponding retriever classes and input parameters
+    search_core_registry: dict[SearchType, Tuple[BaseRetriever, dict]] = {
```
hajdul88, line 66: type(BaseRetriever) is the correct typing I think

cognee/modules/retrieval/chunks_retriever.py
dexters1: an only_object option would reduce the need for a few of our retrievers

At the moment the verbose option returns all result types for a search retriever

cognee/modules/search/methods/get_search_type_retriever_instance.py
dexters1: We are returning a tuple where the first element is the class definition and the second one is it's parameters

We instance the object later with this info

cognee/modules/retrieval/base_retriever.py
```diff
@@ -1,22 +1,63 @@
 from abc import ABC, abstractmethod
-from typing import Any, Optional, Type, List
+from typing import Any, Optional, Type, List, Union
 
 
 class BaseRetriever(ABC):
```
lxobr, line 5: Right now the “full completion” flow (retrieve → context → completion) is orchestrated in get_retriever_output rather than on the retriever. I think it's iportant for the retrievers to be able to generate a completion from query without external orchestration. I’d find it cleaner if BaseRetriever had something like get_completion(query, ...)so a retriever can provide one full completion from a query without callers reimplementing the pipeline. And then get_retriever_output can wrap around that one.

cognee/modules/search/methods/get_search_type_retriever_instance.py
```diff
@@ -0,0 +1,252 @@
...
+)
+from cognee.modules.retrieval.graph_completion_cot_retriever import GraphCompletionCotRetriever
+from cognee.modules.retrieval.graph_completion_context_extension_retriever import (
+    GraphCompletionContextExtensionRetriever,
+)
+from cognee.modules.retrieval.cypher_search_retriever import CypherSearchRetriever
+from cognee.modules.retrieval.natural_language_retriever import NaturalLanguageRetriever
+
+
+async def get_search_type_retriever_instance(
```
lxobr, line 32: While we badly needed a correct function like this, it currently feels like it’s doing a few different things at once: resolving the search type, interpreting config, and wiring retriever constructors. That makes it a bit harder to follow, and I do worry about how easy it will be to maintain as retrievers evolve.

I would be in favour of moving the construction logic into the retrievers themselves (e.g. via a from_config classmethod), so this function mainly orchestrates selection and validation.
Not necessarily something for this PR, but since retriever instantiation is fairly central, I’d be curious to hear others’ thoughts from a maintainability / separation-of-concerns perspective.

cognee/modules/search/methods/get_retriever_output.py
```diff
@@ -0,0 +1,53 @@
...
+)
+from cognee.modules.search.types import SearchType
+from cognee.shared.logging_utils import get_logger
+
+logger = get_logger()
+
+
+async def get_retriever_output(query_type: SearchType, query_text: str, **kwargs):
+    graph_engine = await get_graph_engine()
+    is_empty = await graph_engine.is_empty()
```
lxobr, line 14: We do this check twice, and as noted in a related TODO it's not clear to me that we should be doing it at all, as we already have retrivers that never interact with the graph engine, and thus in theory don't need a non-empty graph database.

lxobr (commented):
I think this is a great step in the exactly right direction!
I left two high level comments about retriever instantiation and completion flow orchestration.

+1 on the scope concern raised by @hajdul88 , there’s no rush to change the public search interface in this PR. I’m not immediately convinced that retriever_specific_config is needed or better than simple kwargs. In any case, this is about the dev experience of using search, not the internal wiring, which looks solid here. That said, I’m open to a SearchConfig object for retrievers, but that feels like a separate follow-up change.

cognee/modules/search/methods/get_search_type_retriever_instance.py
hajdul88: That could clean up this part a bit so I agree in general.

cognee/modules/retrieval/base_retriever.py
hajdul88: In general, I do not have problem with orchestrating these three steps, IF retrievers are not used directly and we always use them through search. But having a get_retriever_output can be an in between solution.

cognee/modules/search/methods/get_retriever_output.py
hajdul88: True, theoretically we can have retrievers that don’t need a graph DB. But if we have something in the vector DB and the graph is empty, that implies an inconsistency imho. I think we shouldn’t have a state where we have data in the vector DB but nothing in the graph (if I’m correct).

hajdul88: Except if its some low level, specialized pipeline, so yes maybe in retrievers we shouldn't do this

cognee/modules/retrieval/base_retriever.py
dexters1: We can have retrievers have only one function that handles all three information and orchestration, but that makes retriever logic much more complex.

There will be scenarios like in only_context where there won't be a call to the completion interface for speed reasons that will need to be accounted for.

This refactor tries to enable all retriever work to be done through the search interface for this reason.

In general I feel like we really should use retrievers only through the search interface and not directly as search handles a lot of different important processes like logging, permissions and etc.

This is also the reason I've enabled forwarding of all possible retriever configurations and parameters through search with the retriever_specific_config kwargs argument.

cognee/modules/search/methods/get_retriever_output.py
dexters1: I have no idea, this mechanism was present already I just tried to enable it to work

dexters1:
> I think this is a great step in the exactly right direction! I left two high level comments about retriever instantiation and completion flow orchestration.
> 
> +1 on the scope concern raised by @hajdul88 , there’s no rush to change the public search interface in this PR. I’m not immediately convinced that retriever_specific_config is needed or better than simple kwargs. In any case, this is about the dev experience of using search, not the internal wiring, which looks solid here. That said, I’m open to a SearchConfig object for retrievers, but that feels like a separate follow-up change.

retriever_specific_config is a simple kwargs, it just has a name (that isn't kwargs) to explain when it should be used and for what (and that is not intended as a kwargs for all search related operation). When description will be added about it to the docstring it would explain it better and in a more meaningful way. Overall there is a need to allow specific parameter forwarding to retrievers do we do this with this kwargs or SearchConfig doesn't really matter, we can change this in any way you would like

cognee/modules/retrieval/base_retriever.py
lxobr: I don’t think this has to be an either/or.

Retrievers can own the full flow (retrieve → context → completion) and still be used through search. get_retriever_output / search can wrap the final output for logging, permissions, etc., without orchestrating the retriever’s internals.

My concern is forcing a single external orchestration model through get_retriever_output. The retriever should be able to decide how a query becomes a completion, rather than an external function coordinating retrieval, context, and completion. A BaseRetriever.get_completion(query, …) provides a clear contract, avoids re-implementing this pipeline outside the retriever, and keeps things backwards-compatible since get_completion already existed.

From client work so far, there’s cleaer interest in building custom retrievers by extending existing ones. A single “query → completion” entry point makes that easier and less error-prone.

Finally, this goes beyond a refactor: it changes the base retriever abstraction and implied flow, which is why I’m raising it so we’re explicit about the decision.

PR #2080, opened 2026-02-02
Fix missing raise statement in set_vector_db_config()

### LINKED ISSUE #2038: [Bug]: Missing raise statement in config.py causes silent validation failures
### Problem
Critical bug found in `cognee/api/v1/config/config.py` at line 169. An exception is created but NOT raised, causing silent failures in config validation.

**Current code (LINE 169):**
```python
InvalidConfigAttributeError(attribute=key)
```

**Expected code:**
```python
raise InvalidConfigAttributeError(attribute=key)
```

### Impact
- Invalid configuration attributes are silently ignored
- No error feedback to users when config is wrong
- Makes debugging configuration issues extremely difficult

### Location
- **File**: `cognee/api/v1/config/config.py`
- **Line**: 169
- **Function**: `set_vector_db_config()`

### Acceptance Criteria
- [ ] Add `raise` keyword before the exception
- [ ] Verify the fix with a test case that passes an invalid attribute
- [ ] Check if similar pattern exists elsewhere in the file

### Time Estimate
1-2 minutes

### Additional Context
This is inside the `set_vector_db_config()` method. The same pattern appears correctly (with `raise`) in other config setter methods, so this is likely a typo.

0xkems:
## Summary
- Adds missing `raise` keyword before `InvalidConfigAttributeError` in `set_vector_db_config()`
- Previously, the exception was constructed but not raised, causing silent failures

## Changes
- `cognee/api/v1/config/config.py`: Line 168 - Added `raise` keyword

## Test Plan
- [x] Verified the bug exists in current code
- [x] Confirmed the fix is consistent with other config setter methods in the same file

Fixes #2038

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved vector database configuration validation to properly detect and report invalid attributes with error messages, rather than silently ignoring them.


Vasilije1990:
@0xkems please read CONTRIBUTING.md and open a PR on the right branch + run tests and post screenshots

PR #2125, opened 2026-02-08
feat: add identity_fields to DataPoint for declarative node deduplication

Vasilije1990:
## Summary
- Adds `identity_fields` to `MetaData` TypedDict and `DataPoint.__init__`, enabling declarative deterministic UUID5 generation from specified field values
- Users can now set `metadata = {"index_fields": ["name"], "identity_fields": ["name"]}` on DataPoint subclasses to automatically deduplicate nodes by name
- Explicit `id` overrides still take precedence; models without `identity_fields` are unaffected (fully backward compatible)

## Changes
- **`cognee/infrastructure/engine/models/DataPoint.py`**: Added `identity_fields` to `MetaData`, modified `__init__` to generate deterministic UUID5 when `identity_fields` is set and no explicit `id` provided, added `_get_identity_fields()` and `_generate_identity_id()` helper methods
- **`cognee/tests/unit/infrastructure/engine/test_identity_fields.py`**: Comprehensive unit tests covering same-value deduplication, cross-type safety, explicit id override, missing field fallback, backward compat, multi-field keys, and string normalization
- **`examples/low_level/pipeline.py`**: Updated example models with `identity_fields` and simplified `ingest_files()` to demonstrate declarative deduplication (no more manual dict tracking)

## Test plan
- [ ] Run new unit tests: `pytest cognee/tests/unit/infrastructure/engine/test_identity_fields.py -v`
- [ ] Run existing unit tests to verify backward compatibility: `pytest cognee/tests/unit/`
- [ ] Run low-level pipeline example end-to-end: `python examples/low_level/pipeline.py`

## Summary by CodeRabbit

* **New Features**
  * DataPoint instances now support deterministic ID generation based on configurable `identity_fields`. When identity fields are specified, instances with identical values produce consistent IDs. The system falls back to random IDs when identity fields are undefined.
  * Added logging warnings for identity field configuration issues and missing values.

* **Refactor**
  * Reorganized ID generation utilities into a shared infrastructure module for better code reusability.

### DIFF
--- a/cognee/infrastructure/engine/models/DataPoint.py
+++ b/cognee/infrastructure/engine/models/DataPoint.py
@@ -1,10 +1,12 @@
 import logging
-from uuid import UUID, NAMESPACE_OID, uuid4, uuid5
+from uuid import UUID, uuid4
 from pydantic import BaseModel, Field, ConfigDict
 from datetime import datetime, timezone
 from typing import Any, Dict, List, Optional
 
 from typing_extensions import NotRequired, TypedDict
+
+from cognee.infrastructure.engine.utils.generate_node_id import generate_node_id
 
 logger = logging.getLogger(__name__)
 
@@ -118,10 +120,7 @@ class DataPoint(BaseModel):
             parts.append(str(value) if not isinstance(value, str) else value)
         joined = "|".join(parts)
         identity_string = f"{class_name}:{joined}"
-        # Same normalization as cognee.modules.engine.utils.generate_node_id
-        # (inlined to avoid circular import)
-        normalized = identity_string.lower().replace(" ", "_").replace("'", "")
-        return uuid5(NAMESPACE_OID, normalized)
+        return generate_node_id(identity_string)
 
     @classmethod
     def get_embeddable_data(self, data_point: "DataPoint"):
--- a/cognee/infrastructure/engine/utils/__init__.py
+++ b/cognee/infrastructure/engine/utils/__init__.py
@@ -1,1 +1,2 @@
 from .parse_id import parse_id
+from .generate_node_id import generate_node_id
--- a/cognee/infrastructure/engine/utils/generate_node_id.py
+++ b/cognee/infrastructure/engine/utils/generate_node_id.py
@@ -1,0 +1,5 @@
+from uuid import NAMESPACE_OID, UUID, uuid5
+
+
+def generate_node_id(node_id: str) -> UUID:
+    return uuid5(NAMESPACE_OID, node_id.lower().replace(" ", "_").replace("'", ""))
--- a/cognee/modules/engine/utils/generate_node_id.py
+++ b/cognee/modules/engine/utils/generate_node_id.py
@@ -1,5 +1,5 @@
-from uuid import NAMESPACE_OID, UUID, uuid5
+# Canonical implementation moved to cognee.infrastructure.engine.utils.generate_node_id
+# Re-exported here for backward compatibility.
+from cognee.infrastructure.engine.utils.generate_node_id import generate_node_id
 
-
-def generate_node_id(node_id: str) -> UUID:
-    return uuid5(NAMESPACE_OID, node_id.lower().replace(" ", "_").replace("'", ""))
+__all__ = ["generate_node_id"]


cognee/tests/unit/infrastructure/engine/test_identity_fields.py
```diff
@@ -0,0 +1,165 @@
...
+
+    def test_explicit_id_differs_from_generated(self):
+        explicit_id = uuid4()
+        p_explicit = PersonWithIdentity(id=explicit_id, name="John")
+        p_generated = PersonWithIdentity(name="John")
+        assert p_explicit.id != p_generated.id
+
+
+class TestMissingIdentityFieldFallback:
+    def test_missing_field_falls_back_to_uuid4(self):
```
lxobr: This test currently fails. Also not sure if the naming of the test and what is being tested are aligned.

examples/low_level/pipeline.py
```diff
@@ -47,41 +45,40 @@ class Data(BaseModel):
...
-                departments_data_points[person["department"]].employees.append(new_person)
+        departments = {
+            name: Department(name=name, employees=employees)
+            for name, employees in dept_employees.items()
+        }
 
-        # Create a single CompanyType node, so we connect all companies to it.
-        companyType = CompanyType()
+        # Create a single CompanyType node (deterministic ID via identity_fields)
+        company_type = CompanyType()
```
lxobr: I suppose this was not ideal in the original example, either. In any case, probably would be better to create one CompanyType data point outside the loop.

examples/low_level/pipeline.py
```diff
@@ -47,41 +45,40 @@ class Data(BaseModel):
...
-            else:
-                departments_data_points[person["department"]].employees.append(new_person)
+        departments = {
+            name: Department(name=name, employees=employees)
+            for name, employees in dept_employees.items()
+        }
 
-        # Create a single CompanyType node, so we connect all companies to it.
-        companyType = CompanyType()
+        # Create a single CompanyType node (deterministic ID via identity_fields)
```
lxobr: I don't think this will be true, as we will be falling back to UUID4. Can you please check?

cognee/infrastructure/engine/models/DataPoint.py
```diff
@@ -48,9 +49,49 @@ class DataPoint(BaseModel):
...
+    @classmethod
+    def _get_identity_fields(cls) -> Optional[list[str]]:
+        """Get identity_fields from the class's metadata field default, if defined."""
+        metadata_field = cls.model_fields.get("metadata")
+        if metadata_field is not None and metadata_field.default is not None:
+            return metadata_field.default.get("identity_fields")
+        return None
+
+    @staticmethod
+    def _generate_identity_id(
```
lxobr, line 102: The generate_node_id function in cognee/modules/engine/utils/generate_node_id.py already does something quite similar. It may be worthwhile to consolidate them.

cognee/infrastructure/engine/models/DataPoint.py
```diff
@@ -48,9 +49,49 @@ class DataPoint(BaseModel):
     belongs_to_set: Optional[List["DataPoint"]] = None
 
     def __init__(self, **data):
+        if "id" not in data:
```
lxobr: There may be a timing issue here, defaults from the DataPoint will get missed I think. Using CompanyType from the pipeline example, if a caller writes `CompanyType()` with no arguments, `data` is empty at this point  (`name` hasn't been set yet)  so the identity check silently fails and returns a random UUID4.

cognee/infrastructure/engine/models/DataPoint.py
```diff
@@ -48,9 +49,49 @@ class DataPoint(BaseModel):
...
+    ) -> Optional[UUID]:
+        """Generate a deterministic UUID5 from identity field values.
+
+        Returns None if any identity field is missing from data,
+        causing fallback to the default UUID4.
+        """
+        parts = []
+        for field_name in identity_fields:
+            if field_name not in data:
+                return None
```
lxobr, line 118: Here `None` is returned silently and the instance gets a random UUID4, we can consider logging a warning.

cognee/infrastructure/engine/models/DataPoint.py
```diff
@@ -48,9 +49,49 @@ class DataPoint(BaseModel):
...
+
         super().__init__(**data)
         object.__setattr__(self, "type", self.__class__.__name__)
 
+    @classmethod
+    def _get_identity_fields(cls) -> Optional[list[str]]:
+        """Get identity_fields from the class's metadata field default, if defined."""
+        metadata_field = cls.model_fields.get("metadata")
+        if metadata_field is not None and metadata_field.default is not None:
+            return metadata_field.default.get("identity_fields")
```
lxobr: We may run into inheritance problems here, if `identity_fields` is read from the class's own `metadata` default. If a subclass overrides `metadata` to add a new index field but forgets to copy over `identity_fields`. Maybe worth detecting a mismatch between a parent and a subclass.

cognee/tests/unit/infrastructure/engine/test_identity_fields.py
Vasilije1990: Fixed — renamed the test class to `TestDefaultFieldsIncludedInIdentity` and updated the assertions. Now that identity generation runs after `super().__init__()` using `self.model_dump()`, fields with defaults (like `age=0`) are correctly included. Added a test for different default overrides too.

examples/low_level/pipeline.py
Vasilije1990: Moved `CompanyType()` outside the loop. Good catch.

examples/low_level/pipeline.py
Vasilije1990: Confirmed — with the timing fix (identity generation now runs after `super().__init__()` using `self.model_dump()`), `CompanyType()` correctly gets a deterministic UUID5 since `name="Company"` is available from the Pydantic default. Tests pass.

cognee/infrastructure/engine/models/DataPoint.py
Vasilije1990: Good point. Couldn't import `generate_node_id` directly due to a circular import (`DataPoint` → `modules.engine.utils` → `modules.engine.models` → `DataPoint`), so the normalization logic is inlined with a comment pointing to the shared function. Same behavior: `lower() + space→underscore + apostrophe removal + uuid5(NAMESPACE_OID, ...)`.

cognee/infrastructure/engine/models/DataPoint.py
Vasilije1990: Fixed — moved identity generation to after `super().__init__()` and now uses `self.model_dump()` which includes all Pydantic defaults. `CompanyType()` now correctly resolves `name="Company"` for identity computation.

cognee/infrastructure/engine/models/DataPoint.py
Vasilije1990: Added `logger.warning()` when a missing field causes UUID4 fallback. Also added a warning when a subclass drops `identity_fields` that a parent defined (MRO walk).

cognee/infrastructure/engine/models/DataPoint.py
Vasilije1990: Added MRO walk in `_get_identity_fields()` — if a subclass overrides `metadata` but drops `identity_fields` that a parent had, it logs a warning. Added a test (`TestInheritanceWarning`) to verify.

cognee/infrastructure/engine/models/DataPoint.py
Vasilije1990: Updated — moved the canonical `generate_node_id` to `cognee/infrastructure/engine/utils/generate_node_id.py` (no circular import risk from DataPoint). DataPoint now imports it directly. The old location at `cognee/modules/engine/utils/generate_node_id.py` re-exports for backward compatibility.

PR #2126, opened 2026-02-08
feat: simplified pipeline API and declarative field annotations

Vasilije1990:
## Summary

- **Three-tier pipeline abstraction**: `flow()` for zero-config chaining, `@step` decorator for config, `Pipeline` builder for full control
- **Explicit type annotations**: `Pipe[T]`, `Ctx[T]`, `Drop` sentinel, `Batch`, `Stream` for declaring data flow roles
- **Field role markers**: `Embeddable()`, `LLMContext()`, `Dedup()` via `typing.Annotated` to make DataPoint field roles visible at definition time
- **identity_fields on DataPoint**: Declarative deterministic UUID5 generation for node deduplication
- **Ctx injection in run_tasks_base**: Annotation-based context injection with backward-compatible deprecation path for magic `context` parameter
- **ECL role comments**: Added EXTRACT/COGNIFY/LOAD role annotations to default cognify pipeline tasks
- **Legacy compatibility**: `from cognee.pipelines import Task, run_tasks` still works via lazy `__getattr__`

## Test plan

- [x] 34 new pipeline tests covering flow, @step, Pipeline builder, Ctx injection, Drop, dataset context manager, field annotations, legacy compat
- [ ] Verify existing integration tests still pass
- [ ] Run low-level pipeline example end-to-end

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added Claude plugin with agent memory and knowledge graph integration
  * Added API keys management interface
  * Added dataset sharing and permissions system
  * Added graph model creation and editing with validation
  * Added onboarding flow for new users
  * Added search with persistent history
  * Added local authentication support

* **Bug Fixes**
  * Improved error handling and user feedback for failed operations
  * Fixed navigation and routing consistency

* **Refactor**
  * Restructured authentication system
  * Reorganized component architecture with new providers
  * Updated data fetching mechanisms

* **Style**
  * Updated UI design tokens and color system
  * Refreshed button and form component styling
  * Enhanced visual consistency across the application

* **Chores**
  * Updated frontend dependencies (Mantine, Jest, testing libraries)
  * Updated environment configuration template with tiered sections

### DIFF
--- a/cognee/tests/integration/test_delete_edge_cases.py
+++ b/cognee/tests/integration/test_delete_edge_cases.py
@@ -121,13 +121,13 @@ async def test_last_accessed_updates_on_search():
         # Verify last_accessed was updated
         assert last_accessed_after is not None, "last_accessed should be set after search operation"
 
-        # Verify timestamp is recent (within last 5 seconds)
+        # Verify timestamp is recent (within last 30 seconds)
         if last_accessed_after.tzinfo is None:
             last_accessed_after = last_accessed_after.replace(tzinfo=timezone.utc)
 
         time_diff = (datetime.now(timezone.utc) - last_accessed_after).total_seconds()
-        assert time_diff < 8, (
-            f"last_accessed should be recent (within 8 seconds), but was {time_diff}s ago"
+        assert time_diff < 30, (
+            f"last_accessed should be recent (within 30 seconds), but was {time_diff}s ago"
         )
 
         logger.info(f"✅ last_accessed updated successfully (time_diff={time_diff:.2f}s)")


cognee/modules/pipelines/operations/run_tasks_base.py
```diff
@@ -1,16 +1,20 @@
...
 from cognee import __version__ as cognee_version
+from cognee.pipelines.types import inject_context_kwargs
 from cognee.modules.observability import (
     new_span,
     COGNEE_PIPELINE_TASK_NAME,
     COGNEE_RESULT_SUMMARY,
     COGNEE_RESULT_COUNT,
 )
 from cognee.infrastructure.engine import DataPoint
+from cognee.pipelines.types import get_ctx_param_name  # Direct import to avoid circular
```
lxobr: This looks like a leftover from a previous iteration and should be removed?

examples/demos/remember_recall_improve_example.py
```diff
@@ -0,0 +1,146 @@
...
+
+    await cognee.forget(everything=True)
+
+    # ----------------------------------------------------------------
+    # Part 1: Permanent memory -- remember() without session
+    # ----------------------------------------------------------------
+
+    # Ingest data directly into the knowledge graph.
+    print("--- Step 1: remember() -- permanent memory ---")
+    await cognee.remember(PERMANENT_TEXT, dataset_name=DATASET)
```
hajdul88, line 73: Triplet embedding appears to be triggered in Cognee.remember even when the triplet embedding environment variable is set to False. I'm not sure whether this is intentional or accidental.
If intentional, the triplet embedding control logic needs to be updated to reflect this behavior. If not, triplet embedding should not be called when the flag is disabled.

examples/demos/remember_recall_improve_example.py
```diff
@@ -0,0 +1,146 @@
...
+    import os
+
+    os.environ["CACHING"] = "true"
+    os.environ["CACHE_BACKEND"] = "fs"
+
+    from cognee.infrastructure.databases.cache.config import get_cache_config
+
+    get_cache_config.cache_clear()
+
+    await cognee.forget(everything=True)
```
hajdul88, line 65: await cognee.forget(everything=True) did not delete my previous redis session which was generated by another run.

examples/demos/remember_recall_improve_example.py
```diff
@@ -0,0 +1,146 @@
...
+    # ----------------------------------------------------------------
+
+    # Ingest data directly into the knowledge graph.
+    print("--- Step 1: remember() -- permanent memory ---")
+    await cognee.remember(PERMANENT_TEXT, dataset_name=DATASET)
+    print("  Data ingested into permanent graph.")
+
+    # Query the permanent graph
+    print("\n--- Step 2: recall() -- query permanent memory ---")
+    answer = await cognee.recall(
```
hajdul88, line 78: When running recall with the following environment variables set:
CACHING=True
CACHE_BACKEND=redis
Triggering cognee.recall via graph completion did not generate any entries in Redis. Previously, with these environment variables set, both the graph completion and QA-based retrievers would generate cache entries under default_session.

examples/demos/remember_recall_improve_example.py
```diff
@@ -0,0 +1,146 @@
...
+        create_db_and_tables,
+    )
+
+    await create_db_and_tables()
+
+    # Enable filesystem-based session caching (required for session_id and improve)
+    import os
+
+    os.environ["CACHING"] = "true"
+    os.environ["CACHE_BACKEND"] = "fs"
```
hajdul88, line 59: I set this to redis for the testing so the following comments are the results of testing the example with redis.

examples/demos/remember_recall_improve_example.py
hajdul88: Additionally, when I use neo4j + pgvector and backend access control is set to false I get the following error for the recall:

sqlalchemy.exc.ProgrammingError: (sqlalchemy.dialects.postgresql.asyncpg.ProgrammingError) <class 'asyncpg.exceptions.UndefinedTableError'>: relation "queries" does not exist
[SQL: INSERT INTO queries (id, text, query_type, user_id, created_at, updated_at) VALUES ($1::UUID, $2::VARCHAR, $3::VARCHAR, $4::UUID, $5::TIMESTAMP WITH TIME ZONE, $6::TIMESTAMP WITH TIME ZONE)]
[parameters: (UUID('304501d5-b3fa-479b-a2b3-372e0a429bb1'), 'What is the theory of general relativity?', 'GRAPH_COMPLETION', UUID('de6936f3-7896-4232-bb49-c0332befee1f'), datetime.datetime(2026, 4, 8, 8, 44, 20, 281899, tzinfo=datetime.timezone.utc), None)]
(Background on this error at: https://sqlalche.me/e/20/f405)

hajdul88: Basically now I got into a state when the whole flow is failing on the first question with the above error + I am getting this as well:

2026-04-08T08:50:10.402019 [info     ] Deleted Kuzu database files at /Users/laszlo/Documents/GitHub/cognee/cognee/.cognee_system/databases/de6936f3-7896-4232-bb49-c0332befee1f/b9e60e54-b5a7-5fa2-bc47-b959c8394475.pkl [cognee.shared.logging_utils]
2026-04-08T08:50:10.443278 [error    ] Local file which should exist can't be found. [cognee.shared.logging_utils]

hajdul88: I was unable to run and test this example beyond line 78. The forget function appears to have serious issues - it does not delete data from Postgres, does not delete data from Redis, and its behavior seems nondeterministic. In some runs I observed it deleting Neo4j data, but in many cases I had to either manually delete the data or fall back to the old prune methods.

cognee/infrastructure/session/session_manager.py
```diff
@@ -436,6 +448,52 @@ async def delete_qa(
...
+            pass
+        return ""
+
+    async def set_graph_context(
+        self, *, user_id: str, session_id: Optional[str] = None, context: str
+    ) -> None:
+        """Store (or overwrite) the graph knowledge snapshot for this session."""
+        if not self.is_available:
+            return
+        session_id = self._resolve_session_id(session_id)
```
hajdul88, line 492: Here are my findings and concerns regarding the current session memory implementation:

**Raw Text Storage**

If I understand correctly, sync_graph_to_sessions stores raw text representations of graph connections — essentially string serializations of edges in the format Alice —[knows]→ Bob. This raises a concern around custom nodes and edges: when a node carries more than just a name, this approach reduces rich data structures down to a simplified string representation, losing the fuller context of the original datapoints.

**Session Persistence Flow**

It is also unclear how sessions are persisted in the knowledge graph. bridge_sessions calls apply_feedback_weights, which operates on sessionQAEntries and mapped IDs to adjust weights this appears unrelated to remember based session memory. It also calls persist_session_in_knowledge_graph, which takes the sessionQAEntries, adds them to Cognee, and cognifies them. The overall flow is difficult to follow and the responsibilities of each step are not well defined.

**Information Loss**

As I understand it, the pipeline generates edge strings in the format node_name — edge_name — node_name, stores them in Redis or the filesystem as plain strings, and then — assuming the flow is working correctly — retrieves and cognifies them in a followup step (additionaly called). Introducing this intermediate session layer between the initial add step and the full graph cognification appears to result in significant information loss compared to the original graph structure, effectively flattening rich relational data into simplified string representations.

cognee/infrastructure/session/session_manager.py
```diff
@@ -179,6 +179,18 @@ async def generate_completion_with_session(
         resolved_session_id = self._resolve_session_id(session_id)
         conversation_history = await self._get_formatted_history(str(user_id), resolved_session_id)
 
+        # Prepend graph knowledge snapshot (from improve() sync) if available
+        graph_context = await self.get_graph_context(
+            user_id=str(user_id), session_id=resolved_session_id
+        )
+        if graph_context:
```
hajdul88, line 187: There is currently no limit on the amount of data retrieved from session memory. As the graph grows, every node-edge-node string representation gets ingested, which can result in polluted context. In extreme cases this may exhaust the context window of the model being used. Furthermore, injecting large amounts of unrelated context increases generation latency and reduces accuracy.

cognee/tasks/memify/sync_graph_to_session.py
```diff
@@ -0,0 +1,197 @@
...
+        logger.info("sync_graph_to_session: no new edges to sync")
+        return {"synced": 0}
+
+    # Merge: existing (older) + new (newer), then cap at max_lines
+    # keeping the tail (most recent) when over budget
+    existing = await sm.get_graph_context(user_id=user_id, session_id=session_id)
+    existing_lines = [line for line in existing.split("\n") if line] if existing else []
+    merged = existing_lines + new_lines
+    if len(merged) > max_lines:
+        merged = merged[-max_lines:]
```
hajdul88, line 206: It is worth questioning why a maximum line cap is being applied here. In addition to the information already lost through the node-edge-node string serialization, imposing a hard line limit risks introducing a second layer of information loss by potentially truncating context that would otherwise be relevant.

cognee/api/v2/improve/improve.py
```diff
@@ -0,0 +1,203 @@
...
+):
+    """Run feedback weights and session persistence pipelines."""
+
+    # Stage 1: apply feedback weights from session scores
+    from cognee.memify_pipelines.apply_feedback_weights import apply_feedback_weights_pipeline
+
+    dataset_name = dataset if isinstance(dataset, str) else "main_dataset"
+
+    try:
+        await apply_feedback_weights_pipeline(
```
hajdul88, line 167: This appears to be using retriever traces - specifically the nodes and edges accessed during retrieval. It is worth clarifying how this connects to the ingestion pipeline. If retrievers are not actually used in a given flow, it is unclear whether this will persist anything or have any meaningful effect at all. 

So this does not persist anything from the new edge string session layer.

cognee/api/v2/improve/improve.py
```diff
@@ -0,0 +1,203 @@
...
+    except Exception as e:
+        logger.warning("improve: feedback weights failed (non-fatal): %s", e)
+
+    # Stage 2: persist session Q&A into permanent graph
+    from cognee.memify_pipelines.persist_sessions_in_knowledge_graph import (
+        persist_sessions_in_knowledge_graph_pipeline,
+    )
+
+    try:
+        await persist_sessions_in_knowledge_graph_pipeline(
```
hajdul88, line 184: This step takes SessionQAEntries, adds them to Cognee, and cognifies the resulting text. If I understand correctly, the improve step takes the edge strings and cognifies them as plain text, which is questionable — as this would treat already-simplified string representations of graph edges as raw text input, potentially compounding the information loss discussed earlier.

cognee/api/v2/forget/forget.py
```diff
@@ -0,0 +1,120 @@
...
+        dataset_id,
+        user.id,
+    )
+    return {"data_id": str(data_id), "dataset_id": str(dataset_id), "status": "success"}
+
+
+async def _resolve_dataset_id(dataset_ref: Union[str, UUID], user) -> UUID:
+    """Resolve a dataset name or UUID to a UUID."""
+    if isinstance(dataset_ref, UUID):
+        return dataset_ref
```
dexters1: This return completely bypasses the permission system

cognee/api/v2/remember/routers/get_remember_router.py
```diff
@@ -0,0 +1,89 @@
...
+
+        if not datasetId and not datasetName:
+            raise ValueError("Either datasetId or datasetName must be provided.")
+
+        from cognee.api.v2.remember import remember as cognee_remember
+
+        try:
+            result = await cognee_remember(
+                data,
+                dataset_name=datasetName or "main_dataset",
```
dexters1: If someone just provides dataset_id, "main_dataset" should not be provided as dataset_name as the default

cognee/api/v2/remember/remember.py
```diff
@@ -0,0 +1,451 @@
...
+    incremental_loading: bool
+    data_per_batch: int
+    chunks_per_batch: int
+    user: object
+    vector_db_config: dict
+    graph_db_config: dict
+
+
+def _build_param_sets():
+    """Pre-compute which kwargs belong to add() vs cognify() vs both."""
```
dexters1: Why do we need this? It seems like it just complicates everything and is not readable

cognee/modules/pipelines/models/PipelineContext.py
```diff
@@ -0,0 +1,23 @@
+from dataclasses import dataclass
+from typing import Any, Optional
+
+
+@dataclass
+class PipelineContext:
```
dexters1, line 6: What if a custom pipeline wants to add new elements to the context? Moving this to a dataclass reduces mobility and flexibility and forces all pipelines to use same dataclass for context

This also breaks all existing custom pipelines

cognee/__init__.py
```diff
@@ -15,34 +15,67 @@
...
+    # Migrations
+    "run_migrations": "cognee.run_migrations",
+    # Tracing / Observability
+    "enable_tracing": "cognee.modules.observability.trace_context",
+    "disable_tracing": "cognee.modules.observability.trace_context",
+    "get_last_trace": "cognee.modules.observability.trace_context",
+    "get_all_traces": "cognee.modules.observability.trace_context",
+    "clear_traces": "cognee.modules.observability.trace_context",
+    "agent_memory": "cognee.modules.agent_memory",
+}
```
dexters1: SQLAlchemy relies on having all the tables for the relational DB fully imported at start time to be able to create the database properly, by switching everything to lazy loading we are now not creating some tables in the DB as SQLAlchemy is not aware of their existence so tests and some Cognee usage is failing.

Long-term fix for this would be to fix our migrations and have them create the whole database properly on their own and run migrations as first step of a pipeline or Cognee start

Short-term fix would be to make sure all DB models are fully loaded at __init__

dexters1:
Also running Cognee twice (PR version) causes an issue that won't show in CI/CD: 
image

When running the test_multi_tenancy.py test twice locally

Vasilije1990:
> Also running Cognee twice (PR version) causes an issue that won't show in CI/CD: image
> 
> When running the test_multi_tenancy.py test twice locally

Could not reproduce

PR #2164, opened 2026-02-12
CI: Retrieve full doc or neighbor chunks via MCP (Contributor PR rob-9)

Vasilije1990:
## Summary
- Fresh branch for CI/CD testing of contributor PR #2154 by @rob-9
- Adds get_document and get_chunk_neighbors MCP tools
- Enhances ChunksRetriever with optional strict enrichment and parent document metadata
- Comprehensive unit tests for enrichment scenarios

## Original PR
#2154

## Test plan
- [ ] CI/CD passes
- [ ] ChunksRetriever unit tests pass
- [ ] MCP server tools functional

### DIFF
--- a/cognee-mcp/src/server.py
+++ b/cognee-mcp/src/server.py
@@ -518,6 +518,343 @@ async def search(search_query: str, search_type: str, top_k: int = 10) -> list:
 
 
 @mcp.tool()
+@log_usage(function_name="MCP get_document", log_type="mcp_tool")
+async def get_document(document_id: str, include_metadata: bool = True) -> list:
+    """
+    Retrieve a complete document by its ID with all chunks and metadata.
+
+    Use this after a CHUNKS search to fetch the full source document that a chunk
+    belongs to. The document ID is found in the 'parent_document' field of enriched
+    chunk search results. Returns the document with all chunks in reading order.
+
+    Parameters
+    ----------
+    document_id : str
+        The unique identifier of the document to retrieve.
+        Obtained from chunk search results: result['parent_document']['id']
+
+    include_metadata : bool, optional
+        Whether to include document metadata (default: True).
+        Set to False for faster retrieval when metadata is not needed.
+
+    Returns
+    -------
+    list
+        A list containing a single TextContent object with JSON-formatted document data.
+
+        Response structure:
+        {
+            "document_id": str,
+            "name": str,
+            "type": str,
+            "chunk_count": int,
+            "chunks": [{"chunk_id": str, "chunk_index": int, "text": str, "word_count": int}],
+            "metadata": dict  (if include_metadata=True)
+        }
+
+        Returns error object if document not found:
+        {"error": str, "document_id": str}
+
+    Notes
+    -----
+    - Requires graph database to be accessible with Document nodes
+    - Chunks are returned sorted by chunk_index in reading order
+    - Document must have been processed via cognify to exist in graph
+    """
+
+    async def get_document_task(document_id: str, include_metadata: bool) -> str:
+        """
+        Internal task to retrieve document from graph database.
+
+        Parameters
+        ----------
+        document_id : str
+            Document identifier to retrieve
+        include_metadata : bool
+            Whether to include metadata in response
+
+        Returns
+        -------
+        str
+            JSON string containing document data or error information
+        """
+        # NOTE: MCP uses stdout to communicate, we must redirect all output
+        #       going to stdout ( like the print function ) to stderr.
+        with redirect_stdout(sys.stderr):
+            if cognee_client.use_api:
+                return json.dumps(
+                    {
+                        "error": "get_document is not supported in API mode",
+                        "document_id": document_id,
+                    },
+                    cls=JSONEncoder,
+                )
+
+            try:
+                from cognee.infrastructure.databases.graph import get_graph_engine
+
+                graph_engine = await get_graph_engine()
+
+                # query doc with all its chunks (flat cols)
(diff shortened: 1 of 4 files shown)


rob-9:
anything else needed from me here? @Vasilije1990

cognee-mcp/src/server.py
```diff
@@ -517,6 +517,343 @@ async def search_task(search_query: str, search_type: str, top_k: int) -> str:
...
+                    cls=JSONEncoder,
+                )
+
+    result = await get_document_task(document_id, include_metadata)
+    return [types.TextContent(type="text", text=result)]
+
+
+@mcp.tool()
+@log_usage(function_name="MCP get_chunk_neighbors", log_type="mcp_tool")
+async def get_chunk_neighbors(
```
R-omk, line 663: I have a concern about the current design of `get_chunk_neighbors`. The function always retrieves neighbors from **both directions** (before and after the target chunk), but there are use cases where we might want **unidirectional iteration**.

**Use Case:**
An agent finds a relevant chunk and wants to iteratively explore context in **one direction only** (e.g., continuing forward through a document). On each iteration, the agent would:
1. Get the next N chunks in one direction
2. Process them
3. Move to the next batch from the last chunk's position

**Current Limitation:**
With the current API, each call retrieves neighbors from both sides, which means:
- Unnecessary data transfer (fetching chunks we already have)
- No clean way to specify direction (forward-only or backward-only)

**Suggestion:**
Consider adding a `direction` parameter:
```python
async def get_chunk_neighbors(
    chunk_id: str,
    neighbor_count: int = 2,
    include_target: bool = True,
    direction: Literal["both", "forward", "backward"] = "both",  # NEW
) -> list:
```

This would enable efficient iterative context building where agents can traverse documents in a streaming fashion without re-fetching already processed chunks.

rob-9: agreed!

cognee-mcp/src/server.py
```diff
@@ -517,6 +517,343 @@ async def search_task(search_query: str, search_type: str, top_k: int) -> str:
     return [types.TextContent(type="text", text=search_results)]
 
 
+@mcp.tool()
+@log_usage(function_name="MCP get_document", log_type="mcp_tool")
+async def get_document(document_id: str, include_metadata: bool = True) -> list:
```
R-omk, line 522: The current implementation retrieves all chunks without any limit, which could cause context overload for large documents. I suggest adding a `max_chunks` parameter to cap the response size, along with `total_chunks` and `is_truncated` flags in the response so agents can detect when results are incomplete.

This would enable a clean pattern for handling large documents: agents call `get_document` with a reasonable limit, check `is_truncated`, and if true, continue fetching remaining chunks iteratively using `get_chunk_neighbors` with `direction="forward"` from the last received chunk. Without these indicators, agents have no way to know if they received the complete document or need to fetch more.

R-omk: After thinking about both methods, I realize they fundamentally retrieve chunks from a document with different parameters. Instead of maintaining two separate endpoints, we could unify them into a single flexible API:

```
get_chunks(
    document_id: str,
    start_chunk_id: str | None = None,  # None = start from beginning
    direction: Literal["forward", "backward"] = "forward",
    limit: int = 50
)
```

Response includes: `chunks`, `total_chunks`, `last_chunk_id`, `has_more`

This covers all use cases:
1. Full document - `start_chunk_id=None, direction="forward", limit=100`
2. Neighbors - `start_chunk_id=X, direction="both", limit=5`
3. Iterative reading - Use `last_chunk_id` from response as `start_chunk_id` for next call

The `last_chunk_id` in the response serves as the cursor for pagination, eliminating the need for a separate cursor field since chunks are already ordered by `chunk_index`.

This would reduce API surface area, eliminate duplication, and provide a more consistent pattern for all chunk retrieval scenarios.

The only thing is that it's not very clear how to set up the "both" mode here; this isn't about the classic cursor pattern.

rob-9: Agree with adding `max_chunks` for pagination. I'll have that implemented along with `direction`.

Regarding `get_document` and `get_chunk_neighbors`, I'd still lean towards keeping them separate. Two tools with clear names + purposes are easier for a model to pick between than one common tool where behavior depends on parameter combinations. 

Also, as you said, `direction="both"` doesn't fit the cursor model. Neighbor retrieval would be a windowing operation while document reading a sequential traversal. Merging them would make an awkward special case.

rob-9: Done.

FYI this branch was originally based on main, not dev. Rebasing onto dev to merge hits conflicts mainly bc dev refactored `ChunksRetriever` to use `get_unified_engine()` instead of separate `get_vector_engine()` / `get_graph_engine()`. 

My logic was written against the old pattern. Working on resolving this.

rob-9: Updated branch: [rob-9:feat/retrieve- full-document-or-neighbor-chunks](https://github.com/rob-9/cognee/tree/feat/retrieve-full-document-or-neighbor-chunks)

Vasilije1990: @rob-9 can you reopen PR? I can't seem to. Then I can pass it to CI

rob-9: @Vasilije1990 yeah, https://github.com/topoteretes/cognee/pull/2287

Vasilije1990:
Closing: this CI branch is superseded by the contributor's fresh PR #2287. No longer needed.

PR #2189, opened 2026-02-15
fix: Storage path resolution when multi-user mode is off

Vasilije1990:
## Summary

- **Decouple `REQUIRE_AUTHENTICATION` from `ENABLE_BACKEND_ACCESS_CONTROL`** in `get_authenticated_user.py` — the OR logic made it impossible to disable auth without also explicitly setting `ENABLE_BACKEND_ACCESS_CONTROL=false`
- **Always set `file_storage_config`** in `context_global_variables.py` even when backend access control is off, preventing stale context vars or fallback ambiguity from causing path mismatches
- **Update `.env.template` comment** to clarify what `REQUIRE_AUTHENTICATION=False` does (uses default user) instead of the misleading instruction to also disable backend access control

## Test plan

- [ ] Set `REQUIRE_AUTHENTICATION=False` and `ENABLE_BACKEND_ACCESS_CONTROL=False` in `.env`
- [ ] Start FastAPI server and upload a file via `POST /v1/add`
- [ ] Run cognify via `POST /v1/cognify` — verify no "Storage directory does not exist" error
- [ ] Verify `.data_storage/` contains files without UUID subdirectories
- [ ] Test with `ENABLE_BACKEND_ACCESS_CONTROL=True` to confirm multi-user mode still works as expected

## Summary by CodeRabbit

* **Chores**
  * Simplified authentication configuration so a single environment setting now determines whether authentication is required.
  * Ensured file storage configuration is consistently initialized even when backend access control is disabled.
  * Clarified the human-readable description for the authentication environment setting to reflect default user behavior for unauthenticated requests.


cognee/modules/users/methods/get_authenticated_user.py
```diff
@@ -10,10 +10,7 @@
 logger = get_logger("get_authenticated_user")
 
 # Check environment variable to determine authentication requirement
-REQUIRE_AUTHENTICATION = (
-    os.getenv("REQUIRE_AUTHENTICATION", "true").lower() == "true"
-    or os.environ.get("ENABLE_BACKEND_ACCESS_CONTROL", "true").lower() == "true"
-)
+REQUIRE_AUTHENTICATION = os.getenv("REQUIRE_AUTHENTICATION", "true").lower() == "true"
```
dexters1, line 13: If ENABLE_BACKEND_ACCESS_CONTROL is true authentication must be true as well or it won't work

.env.template
```diff
@@ -208,7 +208,7 @@ ALLOW_CYPHER_QUERY=True
 RAISE_INCREMENTAL_LOADING_ERRORS=True
 
 # When set to True, the Cognee backend will require authentication for requests to the API.
-# If you're disabling this, make sure to also disable ENABLE_BACKEND_ACCESS_CONTROL.
+# When set to False, a default user is used for all unauthenticated requests.
 REQUIRE_AUTHENTICATION=False
```
dexters1, line 223: Should be set to True by default as ENABLE_BACKEND_ACCESS_CONTROL is True by default now

dexters1:
Closing this PR as it seems to be working on the dev branch now

PR #2214, opened 2026-02-20
feat: Make db pool settings configurable from environment variables

dionb:
Add a pool_opts struct to the config passed through to sqlalchemy.create_engine to allow configuring the pool without having to edit code. 

## Acceptance Criteria
db connection pool parameters are configurable

## Type of Change
- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

## Summary by CodeRabbit

* **New Features**
  * Add customizable connection pool configuration for relational databases via JSON-based pool settings, enabling tuning of pool behavior at runtime.
* **Bug Fixes**
  * Improved provider handling with explicit error for unsupported database providers.
* **Documentation**
  * Updated environment template with examples and guidance for custom pool configuration and a caution about heavy workloads.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -100,6 +100,16 @@ DB_NAME=cognee_db
 #     DATABASE_CONNECT_ARGS='{"timeout": 60}'
 #DATABASE_CONNECT_ARGS='{}'
 
+# -- Advanced: Custom database connection pool arguments (optional) ---------------
+# Pass additional connection pool parameters as JSON. Useful for managing the number of connections and their lifetimes.
+# Examples:
+#   SqlAlchemy Defaults:
+#     DATABASE_POOL_ARGS='{"pool_size": 5, "max_overflow": 10, "pool_recycle": -1, "pool_timeout": 30}'
+#   For heavy and slow workloads (eg: using Cognee with large datasets):
+#     DATABASE_POOL_ARGS='{"pool_size": 20, "max_overflow": 20, "pool_recycle": 280, "pool_timeout": 280, "pool_pre_ping": true}'
+# 		Note: Be careful using so many connections in a multi-client environment as you can easily overwhelm a DB server this way.
+#DATABASE_POOL_ARGS='{}'
+
 ################################################################################
 # 🕸️ Graph Database settings
 ################################################################################


dionb:
@dexters1 and @siillee I've talked to @pazone about this, so he knows what to expect. CI passed, so it just needs one of you to check to make sure this makes sense to you.

cognee/infrastructure/databases/relational/sqlalchemy/SqlAlchemyAdapter.py
```diff
@@ -77,13 +77,11 @@ def __init__(self, connection_string: str, connect_args: dict = None):
                 connect_args={**{"timeout": 30}, **final_connect_args},
             )
         else:
+            if pool_args is None:
+                pool_args = {}
             self.engine = create_async_engine(
                 connection_string,
-                pool_size=20,
```
dexters1, line 82: Can we define default values in the pool for these arguments so we can keep track of the values more easily?

cognee/infrastructure/databases/relational/create_relational_engine.py
```diff
@@ -1,9 +1,9 @@
 from sqlalchemy import URL
 from .sqlalchemy.SqlAlchemyAdapter import SQLAlchemyAdapter
-from functools import lru_cache
+from functools import cache
 
 
-@lru_cache
+@cache
```
dexters1: Any reason for the change from lru_cache to cache?

Can we keep the naming uniform of the cache system across the project?

cognee/infrastructure/databases/relational/sqlalchemy/SqlAlchemyAdapter.py
dionb: SqlAlchemy already has good defaults for all of these values. Would it not add more confusion by setting our own defaults?

cognee/infrastructure/databases/relational/create_relational_engine.py
dionb: Using an LRU cache instead of the default one is (slightly) more expensive as the cache implementation is more complicated, and it can result in our db engine potentially getting garbage collected if there is a flood of writes to the cache in between db calls, which we don't want. This is meant to be a singleton, which an LRU cache does not correctly provide.

cognee/infrastructure/databases/relational/sqlalchemy/SqlAlchemyAdapter.py
dexters1: Can you tell me what are the defaults from SQLAlchemy? Without checking online it's a bit hard to access this information just by checking their code. (I would prefer to be able to see exact config from code)

We also need to change their defaults to support a big server way of working for Cognee. The default pool_size of 5 and max_overflow of 10 might not be enough to support multiple cognify requests at the same time as this can easily be running across 40 threads and we are caching our relational DB engine

cognee/infrastructure/databases/relational/create_relational_engine.py
dexters1: image

Cache is just lru_cache with max_size=None which means in that case we'd in theory could get a memory leak over a long time

If there are a flood of bad writes to the cache then we did something wrong and are not caching correctly or caching too much and we need to fix it

dionb: This isn't the right place to discuss this, so I'll revert this change and open a separate PR to fix handling of DB engine instances properly.

cognee/infrastructure/databases/relational/sqlalchemy/SqlAlchemyAdapter.py
dionb: I could tell you what the defaults are from memory, but only because I looked them up last week to work on this PR to make sure I wouldn't break things while working on this PR. Would comments with the defaults for the four we care about be acceptable?

The point of this PR is to make this configurable, so that we aren't using the same values everywhere, as that caused two days of outages two weeks ago. The default values that we choose should be tuned for the case where users/operators aren't going to think about it, which I think is people using the python libs or a container for experimentation. In this case I think the sqlalchemy defaults are going to be better than our production settings. What do you think?

PR #2230, opened 2026-02-24
fix: Persist pipeline_status cleanup when deleting a dataset

Vasilije1990:
## Summary

- When a dataset is deleted, `delete_dataset()` clears pipeline_status entries for the deleted dataset from related Data objects. However, the nested dict deletion (`del status[pipeline_name][dataset_id]`) was not detected by SQLAlchemy's `MutableDict` change tracking, which only monitors top-level dict operations. The cleanup ran in memory but was never persisted to the database on commit.
- Since dataset IDs are deterministic (`uuid5` of name + user_id + tenant_id), recreating a dataset with the same name produces the **same ID**. Without persisted cleanup, re-added data was incorrectly skipped by the incremental loading cache, which still saw the old "completed" pipeline_status entry.
- Fix: use `flag_modified(data_record, "pipeline_status")` from `sqlalchemy.orm.attributes` to explicitly mark the column as dirty after nested dict modifications.

## Root Cause

`Data.pipeline_status` is a `MutableDict.as_mutable(JSON)` column. `MutableDict` tracks top-level dict operations (`__setitem__`, `__delitem__`, `setdefault`, etc.) but **not** modifications to nested dicts. The deletion pattern:

```python
del data_record.pipeline_status[pipeline_name][dataset_id_str]
#    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^                               → MutableDict.__getitem__ (read, no change event)
#                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^    → plain dict.__delitem__ (no change event)
```

produces no change notification, so `session.commit()` sees no dirty columns and skips the UPDATE.

## Test plan

- [ ] Add data to a named dataset, cognify it, delete the dataset, re-add the same data to a dataset with the same name — verify the data is processed again (not skipped as cached)
- [ ] Verify Data.pipeline_status no longer contains the deleted dataset's ID after deletion
- [ ] Run existing unit/integration tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Deletions now respect dataset context and only fully remove items when no other dataset references remain.
  * Nested changes to data record status are reliably detected and persisted so pipeline/status updates are saved.
  * Batch deletion now correctly propagates dataset context to per-item deletion, preventing orphaned or partially-deleted records.

* **Tests**
  * Updated tests to validate dataset-scoped deletion behavior and file removal across datasets.

### DIFF
--- a/cognee/infrastructure/databases/relational/sqlalchemy/SqlAlchemyAdapter.py
+++ b/cognee/infrastructure/databases/relational/sqlalchemy/SqlAlchemyAdapter.py
@@ -81,11 +81,20 @@ class SQLAlchemyAdapter:
             if pool_args is None:
                 pool_args = {}
 
-            if "max_overflow" not in pool_args or pool_args["max_overflow"] is None:
+            if pool_args.get("pool_size") is None:
+                pool_args["pool_size"] = 20
+
+            if pool_args.get("max_overflow") is None:
                 pool_args["max_overflow"] = 20
 
-            if "pool_pre_ping" not in pool_args or pool_args["pool_pre_ping"] is None:
+            if pool_args.get("pool_pre_ping") is None:
                 pool_args["pool_pre_ping"] = True
+
+            if pool_args.get("pool_recycle") is None:
+                pool_args["pool_recycle"] = 280
+
+            if pool_args.get("pool_timeout") is None:
+                pool_args["pool_timeout"] = 280
 
             self.engine = create_async_engine(
                 connection_string,


dexters1:
We have a problem, when deleting a dataset the whole data object gets deleted as well even though it's used by other Datasets. Not sure if the problem happens due to changes in this PR, will check dev branch and write back

cognee/api/v1/datasets/datasets.py
```diff
@@ -84,7 +84,7 @@ async def empty_dataset(dataset_id: UUID, user: Optional[User] = None):
         # even if some fail.
         if dataset_data:
             results = await asyncio.gather(
-                *[delete_data(data) for data in dataset_data],
+                *[delete_data(data, dataset_id) for data in dataset_data],
                 return_exceptions=True,
             )
             deletion_errors = [r for r in results if isinstance(r, Exception)]
```
mdehsan873, line 90: Why we making n calls to db?
we can do something liks:
```suggestion
  data_ids = [data.id for data in dataset_data]

await session.execute(
    delete(DatasetData).where(
        DatasetData.dataset_id == dataset_id,
        DatasetData.data_id.in_(data_ids),
    )
)
```

This would reduce round trips and improve performance significantly for bigger datasets.

Is there a specific reason we’re handling deletions one-by-one instead of batching?

dexters1: You are correct, this in general is one of the improvement points we can make to the core codebase 

In multiple places we are not grouping write/delete operations to the DB

PR #2237, opened 2026-02-25
feat: make dlt resources first class citizens in cognee ingestion

siillee:
## Description

Added option to ingest dlt (re)sources via regular cognee pipeline. Currently works only for Postgres destination, and only takes schemas into account.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

## Summary by CodeRabbit

# Release Notes

* **New Features**
  * Added Data Load Tool (DLT) source ingestion for automatic database and CSV imports
  * Implemented relational data processing with foreign key relationship extraction
  * Enhanced graph building from structured database rows

* **Improvements**
  * Configurable row limits for ingestion via environment settings
  * Improved data classification and handling for structured sources
  * Content change detection for efficient data updates

* **Documentation**
  * Added DLT ingestion workflow examples with ontology integration

### DIFF
--- a/examples/custom_pipelines/agentic_reasoning_procurement_example.py
+++ b/examples/custom_pipelines/agentic_reasoning_procurement_example.py
@@ -168,7 +168,7 @@ async def run_procurement_example():
         for q in questions:
             print(f"Question: \n{q}")
             results = await procurement_system.search_memory(q, search_categories=[category])
-            top_answer = results[category][0]
+            top_answer = results[category][0]["search_result"][0]
             print(f"Answer: \n{top_answer}\n")
             research_notes[category].append({"question": q, "answer": top_answer})
 


examples/python/simple_example.py
```diff
@@ -46,6 +46,15 @@ async def main():
     await cognee.cognify()
     print("Cognify process complete.\n")
 
+    await cognee.add("""
+    Natural language processing (NLP) is an interdisciplinary
+    subfield of computer science and information retrieval.
+
+    Andrej is an expert in NLP.
```
siillee: Note to self: Remove this nonsense

cognee/tasks/ingestion/ingest_dlt_source.py
```diff
@@ -0,0 +1,396 @@
...
+        normalized = [(c.get("name"), c.get("type")) for c in schema_info]
+    elif isinstance(schema_info, dict):
+        # Postgres format: {"col_name": {"type": "TEXT", ...}, ...}
+        normalized = sorted((k, str(v)) for k, v in schema_info.items())
+    else:
+        normalized = str(schema_info)
+    return hashlib.md5(json.dumps(normalized, sort_keys=True).encode()).hexdigest()
+
+
+async def _read_rows_from_tables(
```
siillee, line 216: Question: Is this going to be a bottleneck performance-wise? Reading an entire database is an expensive operation.

siillee:
General comment here for myself and anyone interested:

Right now we read all the rows we ingest through dlt and send those rows down our pipelines (specifically add and cognify). This can be expensive for large datasets, and also the graph that is being generated doesn't look great. Sometimes there are nodes corresponding to tables, sometimes not; the way the nodes are connected is sometimes off, etc. Maybe the example I ran (`dlt_ingestion_example.py`) is not the best for testing this tho.

Another thing is that we need to handle when someone deletes some part of their data, and tries to ingest that piece again. We need a way to delete this from our databases, graphs, etc. DLT will handle the correct deletion from the dlt database, but then we need to have some deletion functionality in the cognify pipeline that ingests this data.

cognee/api/v1/add/add.py
```diff
@@ -11,13 +12,22 @@
...
 async def add(
-    data: Union[BinaryIO, list[BinaryIO], str, list[str], DataItem, list[DataItem]],
+    data: Union[
+        BinaryIO,
+        list[BinaryIO],
+        str,
+        list[str],
+        DataItem,
+        list[DataItem],
+        Any,  # DltResource, SourceFactory, or other dlt types
```
dexters1, line 29: Can we specify the DLT types? Any makes it confusing

cognee/api/v1/add/add.py
```diff
@@ -192,6 +203,15 @@ async def add(
         dataset_name=dataset_name, dataset_id=dataset_id, user=user
     )
 
+    # Expand DLT resources (and auto-detected CSV/connection strings) into
+    # standard DataItems before the pipeline sees them.
+    data = await resolve_dlt_sources(
```
dexters1, line 208: This should be a pipeline task

cognee/modules/pipelines/operations/run_tasks_data_item.py
```diff
@@ -64,12 +64,19 @@ async def run_tasks_data_item_incremental(
...
-        # Ingest data and add metadata
-        async with open_data_file(file_path) as file:
-            classified_data = ingestion.classify(file)
-            # data_id is the hash of file contents + owner id to avoid duplicate data
-            data_id = await ingestion.identify(classified_data, user)
+        # If the DataItem carries a stable data_id (e.g. from DLT), prefer it
+        # over the content-hash-based ID so lookups stay consistent.
+        from cognee.tasks.ingestion.data_item import DataItem as DataItemType
+
+        if isinstance(data_item, DataItemType) and data_item.data_id is not None:
```
dexters1, line 71: A DataItem can be a text document 

It is primarly created to allow users to define metadata and labels for their text documents

but they still need to be stored.

This should be changed to only not store data items in case they are a relational DB

cognee/modules/data/processing/document_types/DltRowDocument.py
```diff
@@ -0,0 +1,36 @@
+from cognee.modules.chunking.Chunker import Chunker
+from cognee.modules.engine.utils.generate_node_id import generate_node_id
+from .Document import Document
+
+
+class DltRowDocument(Document):
```
dexters1, line 6: Rows should not be treated as documents, all rows should be data points that point to the Database/schema document 

and all rows should be removed/updated based on the changes to the schema/database so there is a single source of origin

cognee/tasks/ingestion/extract_dlt_fk_edges.py
```diff
@@ -0,0 +1,242 @@
...
+    doc = getattr(data_point, "is_part_of", None)
+    if doc is None:
+        return False
+    from cognee.modules.data.processing.document_types import DltRowDocument
+
+    return isinstance(doc, DltRowDocument)
+
+
+async def extract_dlt_fk_edges(data_points: List[DataPoint]) -> List[DataPoint]:
+    """Create graph edges and schema nodes from DLT-sourced relational data.
```
dexters1, line 26: Why are we creating duplicate code? 

We already have methods to handle all database related node creation and edge mappings

dexters1: Also there is no column DataPoint support here

In case of multiple rows sharing the same column value there should be a relationship between them that is created based on the column

cognee/api/v1/cognify/cognify.py
```diff
@@ -324,6 +325,7 @@ async def get_default_tasks(  # TODO: Find out a better way to do this (Boris's
             embed_triplets=embed_triplets,
             task_config={"batch_size": chunks_per_batch},
         ),
+        Task(extract_dlt_fk_edges),
```
dexters1, line 328: Migrating a relational database should be its own unique set of tasks, they are not compatible with default cognify tasks

we can call the proper custom pipeline and set of tasks based on the document being processed

cognee/modules/data/processing/document_types/DltRowDocument.py
dexters1, line 6: Each row creates a new text file in Cognee with this way of working, this is not scalable at all

as there is a limit on how many files can be opened at the same time

Image

dexters1 (approved):

PR #2253, opened 2026-02-28
Fix #2191: sanitize embedding inputs before API call

### LINKED ISSUE #2191: [Bug]: embedding API returns 422 - invalid input
### Bug Description

Cognee sends invalid text (e.g. punctuation or single words from ontology Literals/blank nodes) to the embedding API, which returns 422.

### Ontology

Pick an TTL ontology that will be used during entity extraction (e.g. `cognee-integration-langgraph/cognee_mock_saas_entitlements_demo/ontologies/saas_finance_subset.ttl`).
Set ONTOLOGY_FILE_PATH to that file’s absolute path (e.g. in the same process that will call cognee.add / cognee.cognify).

### Data and cognify

Add a dataset that will be cognified with temporal cognify (e.g. an audit/event stream: JSONL or similar with timestamps).
Run cognee.cognify(datasets=["<that_dataset_name>"], user=<user>, temporal_cognify=True) (ontology must be active during this cognify).

### Trigger the failure

Let the pipeline run: ontology-based extraction runs, then vector indexing calls the embedding API with the list of texts to embed (e.g. entity names / index fields from the graph). Because of how ontology traversal (e.g. RDFLib) handles Literals/blank nodes, some of those “names” are invalid for the API (e.g. single punctuation like (, ), or words like the, with). The embedding request is sent with one or more of these invalid strings which lead to the embedding API to return 422 (validation error).

### Expected Behavior

Only valid, embeddable text (e.g. real entity names) is sent to the embedding API, or invalid values are filtered/sanitized before the call.

### Actual Behavior

The embedding API returns 422 (validation error) because it receives empty, punctuation-only, or otherwise invalid input.

### Environment

- OS: Ubuntu in Docker
- Python: 3.12
- Cognee:  0.5.1
- LLM provider: OpenAI (e.g. gpt-4o-mini / gpt-5-mini); 
embedding: OpenAI text-embedding-3-small
DB: relational: SQLite, vector: LanceDB, graph: Kuzu 

### Logs/Error Messages

```shell

```

### Suggested fix

Filter or sanitize embedding input (e.g. in cognee/infrastructure/databases/vector/embeddings/LiteLLMEmbeddingEngine.py in embed_text()) so empty, single-character, or punctuation-only strings are not sent to the API.

### Pre-submission Checklist

Aroool:
Summary
- Filters out empty / whitespace / punctuation-only strings before calling the embedding provider to prevent 422 validation errors.
- Preserves output alignment by returning zero-vectors for filtered inputs.

Tests
- python3 -m pytest cognee/tests/test_litellm_embedding_sanitize.py

Closes #2191

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved input validation for the embedding engine to gracefully handle empty and invalid strings while maintaining proper alignment with original inputs.
  * Fixed dimension handling when explicitly set or derived from responses.

* **Tests**
  * Added comprehensive unit tests for embedding sanitization and input/output alignment verification.


Aroool:
Thanks for the review!

I’ve addressed the feedback:

- Updated `_is_embeddable` to require at least two alphanumeric characters.
- Guarded against `dimensions=None` in the mock embedding path.
- Replaced direct `os.environ` mutation with `monkeypatch.setenv`.
- Strengthened the test to assert exact vector-to-index alignment.
- Moved the test to `cognee/tests/unit/` per project conventions.

All tests pass locally.

PR #2326, opened 2026-03-08
fix: use psycopg2-binary in MCP Docker build

Vasilije1990:
## Summary
- Switch `cognee[postgres]` to `cognee[postgres-binary]` in `cognee-mcp/pyproject.toml`
- Regenerate `cognee-mcp/uv.lock` to remove source-only `psycopg2` dependency

The MCP Docker image build has been failing consistently because `psycopg2` (source distribution) requires `libpq-dev` and `gcc` to compile, which are not present in the `python:3.12-slim-bookworm` base image used in the `uv` build stage. The `postgres-binary` extra uses pre-built `psycopg2-binary` wheels instead, which need no compilation.

## Test plan
- [ ] Verify the MCP Docker image builds successfully via the `build | Build and Push Cognee MCP Docker Image to dockerhub` workflow

## Summary by CodeRabbit

* **Chores**
  * Updated PostgreSQL driver dependencies to use binary distributions instead of source-based versions.

### DIFF
--- a/cognee-mcp/pyproject.toml
+++ b/cognee-mcp/pyproject.toml
@@ -8,7 +8,7 @@ requires-python = ">=3.10"
 dependencies = [
     # For local cognee repo usage remove comment bellow and add absolute path to cognee. Then run `uv sync --reinstall` in the mcp folder on local cognee changes.
     #"cognee[postgres,codegraph,gemini,huggingface,docs,neo4j] @ file:/Users/igorilic/Desktop/cognee",
-    "cognee[postgres,docs,neo4j]==0.5.3",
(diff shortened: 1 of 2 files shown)


dexters1:
@Vasilije1990 psycopg2-binary exists already as an optional dependency for MCP

It should not be the default dependency for MCP as it will just cause a different set of problems then

based on their docs: https://www.psycopg.org/docs/install.html#quick-install

specifically: The binary packages come with their own versions of a few C libraries, among which libpq and libssl, which will be used regardless of other libraries available on the client

This is why it is only recommended for use for beginners and not in a prod environment

PR #2348, opened 2026-03-10
fix: graph extraction name field

kura120:
## Description

While integrating Cognee into a local AI memory system with Ollama, I noticed `cognify()` wasn't extracting any semantic graph nodes with `llama3.1:8b`. Querying with `SearchType.GRAPH_COMPLETION` came back with a single generated string instead of graph entities, no error, no warning.

Inspecting the raw graph showed only internal pipeline nodes (`DocumentChunk`, `TextDocument`, `TextSummary`) with empty `name` fields. No entities, no relationships.

Root cause: `generate_graph_prompt.txt` never asks the LLM for a `name` field, but the `KnowledgeGraph`/`Node` Pydantic schema requires it. Weaker models follow the prompt literally and skip it, so Pydantic drops every extracted node silently. Stronger models like `qwen2.5:7b` seem to fill it in anyway, which is probably why this wasn't caught before.

The fix patches the immediate mismatch. But while reproducing this I noticed something worth flagging separately: there's no model capability check anywhere in the pipeline. When graph extraction fails for any reason, the pipeline swallows it and the caller gets back a plausible looking response with no way to know extraction never happened. In this case it returned the `TextSummary` text, a model-generated summary of the input, which looks like a valid result but isn't. Not raising this as part of this PR, just flagging it since the reproduction makes it pretty visible.

## Reproduction

**Model:** `llama3.1:8b` via Ollama (fails without fix, passes with fix)  
**Model:** `qwen2.5:7b` via Ollama (passes with and without fix — infers `name` from context)  
**Cognee version:** 0.5.3  
**Python:** 3.13.7, Windows 11

**Step 1 — run cognify:**
```python
import asyncio
import cognee

async def main():
    await cognee.add("The Sena project uses LadybugDB as its graph database.")
    await cognee.cognify()

asyncio.run(main())
```

**Step 2 — inspect raw graph:**
```python
import asyncio
from cognee.infrastructure.databases.graph import get_graph_engine

async def main():
    engine = await get_graph_engine()
    nodes, edges = await engine.get_graph_data()
    for n in nodes:
        print(n[1].get('type'), '|', n[1].get('name'))

asyncio.run(main())
```

## Acceptance Criteria

* `cognify()` with `llama3.1:8b` extracts semantic entity nodes with populated `name` fields
* Raw graph inspection shows `Entity` and `EntityType` nodes alongside internal scaffolding
* `qwen2.5:7b` behavior is unchanged
* No breaking changes for existing cloud model integrations

## Type of Change

## Screenshots
before

^^ Before the fix, GRAPH_COMPLETION just returned the TextSummary node text - a model-generated summary of the input, not actual graph entities. No error, nothing to indicate extraction failed.

after

## Pre-submission Checklist

## DCO Affirmation

## Summary by CodeRabbit

* **New Features**
  * Nodes now automatically populate name fields from their identifier when not explicitly provided, ensuring all entities have descriptive identifiers.

* **Documentation**
  * Updated guidance for node naming to use complete, human-readable names.

### DIFF
--- a/cognee/shared/data_models.py
+++ b/cognee/shared/data_models.py
@@ -45,7 +45,7 @@ else:
         """Node in a knowledge graph."""
 
         id: str
-        name: str
+        name: str = ""
         type: str
         description: str
         


Vasilije1990:
Nice catch! Merging

Issue #2305, opened 2026-03-08
Move dev-only dependencies (pympler, pylance, datamodel-code-generator) to dev extras


Vasilije1990:
## Problem / Background

Three packages in `pyproject.toml` are listed in core runtime dependencies but are only used for development/build-time tasks:

1. **`pympler`** (line ~48) — Memory profiling library, not used in production code paths
2. **`pylance`** (line ~50) — Type checker tool, not a runtime dependency
3. **`datamodel-code-generator`** (line ~66) — Code generation tool, build-time only

This unnecessarily bloats the install for end users.

## Goal

Move these packages to the `[dev]` extras group so they're only installed when explicitly requested.

## Proposed Solution

1. Remove the three packages from `[project.dependencies]` in `pyproject.toml`
2. Add them to the `[project.optional-dependencies] dev` section
3. Verify no production import paths depend on them

To check usage:
```bash
rg "import pympler|from pympler" cognee/ --type py
rg "import pylance|from pylance" cognee/ --type py
rg "import datamodel_code_generator|from datamodel_code_generator" cognee/ --type py
```

## Acceptance Criteria

- [ ] The three packages are moved from core deps to `[dev]` extras
- [ ] `uv pip install -e .` succeeds without these packages
- [ ] `uv pip install -e ".[dev]"` includes them
- [ ] No production import paths break
- [ ] CI still works (it should already install dev extras)

## Dependencies / Blockers

None.

## Relevant Links / Resources

- `pyproject.toml` dependency sections

initcatcher:
Hi! I'd like to take on this issue.

nehaaprasad:
Hello! @Vasilije1990 
please check out this PR!

nightcityblade:
Hi, I'd like to work on this. I'll submit a PR shortly.

vivan070707-ctrl:
Hi, I’d like to work on this issue. Can you assign it to me?

PR #2366, opened 2026-03-10
chore(deps): move pympler and pylance to dev extras

nehaaprasad:
fix : #2305

### summary

- Move dev-only deps out of core runtime.

### Changes

- Removed pympler and pylance from [project.dependencies].
- Added pympler and pylance to [project.optional-dependencies].dev.
- Kept datamodel-code-generator as a core dependency since it is used at API runtime.

### Testing
uv pip install -e .
uv pip install -e ".[dev]"

## Release Notes

* **Chores**
  * Updated dependency configuration to streamline the runtime environment and optimize installation for end-users.


Vasilije1990:
Would not merge

PR #2383, opened 2026-03-11
fix: neo4j adapter fails explicitly on incomplete credentials

omar-y-abdi:
k


Vasilije1990:
Can you explain what the problem is exactly and show how it fails to anon auth if something is wrong?

PR #2409, opened 2026-03-14
fix(neo4j): raise ValueError on incomplete credentials instead of silent fallback

### LINKED ISSUE #2307: Neo4j adapter should fail explicitly on incomplete credentials instead of silent fallback
## Problem / Background

The Neo4j adapter in `cognee/infrastructure/databases/graph/neo4j_driver/adapter.py` (lines ~59-82) logs a warning when credentials are incomplete but silently falls back to anonymous authentication. This can lead to confusing runtime errors later when operations fail due to missing permissions, rather than a clear startup error.

## Goal

Fail fast with a clear error message when Neo4j credentials are incomplete, rather than silently degrading to anonymous auth.

## Proposed Solution

Replace the silent fallback with an explicit error:

```python
if not username or not password:
    raise ValueError(
        "Neo4j credentials incomplete. Set GRAPH_DATABASE_USERNAME and "
        "GRAPH_DATABASE_PASSWORD in your .env file, or configure them programmatically."
    )
```

If anonymous auth is a legitimate use case, make it opt-in with a clear config flag (e.g., `GRAPH_DATABASE_ALLOW_ANONYMOUS=true`).

## Acceptance Criteria

- [ ] Incomplete Neo4j credentials raise a clear, actionable error message
- [ ] The error message tells the user exactly which env vars to set
- [ ] If anonymous auth is intentionally supported, it requires explicit opt-in
- [ ] Existing tests pass
- [ ] `ruff check .` passes

## Dependencies / Blockers

None.

## Relevant Links / Resources

- File: `cognee/infrastructure/databases/graph/neo4j_driver/adapter.py`

LeC-D:
## Summary

Fixes #2307.

The Neo4j adapter was silently falling back to anonymous auth when only one of username/password was provided, logging a warning but continuing. This leads to confusing permission errors at runtime rather than a clear startup failure.

## Changes

Replaced the silent fallback with an explicit `ValueError` that identifies the missing credential and tells the user exactly how to fix it:

```python
raise ValueError(
    f"Neo4j credentials incomplete: {missing} is not set. "
    "Provide both GRAPH_DATABASE_USERNAME and GRAPH_DATABASE_PASSWORD in your "
    ".env file, or configure them programmatically. "
    "If anonymous authentication is intentional, leave both values empty."
)
```

Anonymous auth (both values empty) is still fully supported — only partial credentials are rejected.

## Checklist
- [x] Incomplete credentials raise a clear, actionable `ValueError`
- [x] Error message identifies the specific missing variable
- [x] Anonymous auth (both empty) still works
- [x] Existing behavior for valid credentials unchanged

## Summary by CodeRabbit

* **Bug Fixes**
  * Neo4j authentication now enforces a strict requirement: both username and password must be provided together, or neither. Partial credentials will raise an error with a clear message.

### DIFF
--- a/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
+++ b/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
@@ -63,13 +63,19 @@ class Neo4jAdapter(GraphDBInterface):
         graph_database_name: Optional[str] = None,
         driver: Optional[Any] = None,
     ):
-        # Only use auth if both username and password are provided
+        # Only use auth if both username and password are provided.
+        # Partial credentials are an error — fail fast with a clear message.
         auth = None
         if graph_database_username and graph_database_password:
             auth = (graph_database_username, graph_database_password)
         elif graph_database_username or graph_database_password:
-            logger = get_logger(__name__)
-            logger.warning("Neo4j credentials incomplete – falling back to anonymous connection.")
+            missing = "GRAPH_DATABASE_PASSWORD" if graph_database_username else "GRAPH_DATABASE_USERNAME"
+            raise ValueError(
+                f"Neo4j credentials incomplete: {missing} is not set. "
+                "Provide both GRAPH_DATABASE_USERNAME and GRAPH_DATABASE_PASSWORD in your "
+                ".env file, or configure them programmatically. "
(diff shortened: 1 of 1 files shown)


LeC-D:
Closing in favour of #2411 which correctly targets the `dev` branch as per CONTRIBUTING.md.

PR #2418, opened 2026-03-15
fix(mcp): cognify_status works in API mode via REST endpoint

acsezen:
## Summary

- `cognify_status` MCP tool fails with `sqlite3.OperationalError: unable to open database file` when running in API mode (Docker Compose setup with separate backend)
- Root cause: the function calls local DB functions (`get_default_user()`, `get_unique_dataset_id()`) which require a SQLite database that doesn't exist in the MCP container (API mode skips migrations)
- Fix: in API mode, resolve dataset ID via `cognee_client.list_datasets()` (HTTP) and call `GET /api/v1/datasets/status` instead of raising `NotImplementedError`

## Changes

**`cognee-mcp/src/cognee_client.py`**:
- `get_pipeline_status()`: replaced `NotImplementedError` in API mode with actual HTTP call to `GET /api/v1/datasets/status`

**`cognee-mcp/src/server.py`**:
- `cognify_status()`: in API mode, uses `cognee_client.list_datasets()` to resolve dataset IDs via HTTP instead of calling local `get_default_user()` + `get_unique_dataset_id()` which hit the non-existent local SQLite DB

Direct (standalone) mode behavior is unchanged.

## Known limitation

The `pipeline_name` parameter is not used in API mode because `GET /api/v1/datasets/status` returns dataset-level status, not pipeline-specific status. In direct mode, `get_pipeline_status()` filters by pipeline name. This matches the current backend API capability — a pipeline-specific status endpoint would need to be added separately.

## Test plan

- [x] Tested in API mode Docker Compose setup — returns `{'dataset_id': 'DATASET_PROCESSING_COMPLETED'}` instead of SQLite error
- [x] Verified standalone (direct) mode still works unchanged
- [ ] Unit tests (none exist for this function currently)

## DCO Affirmation

## Summary by CodeRabbit

* **Bug Fixes**
  * Pipeline status checking now works in API mode: the app queries the backend and returns pipeline status instead of failing; API-mode returns empty status when no matching dataset is found and reports errors consistently.

* **Documentation**
  * README updated to clarify supported operations and API-mode limitations, specifying which status-tracking features are available in API mode.

### DIFF
--- a/cognee-mcp/src/server.py
+++ b/cognee-mcp/src/server.py
@@ -859,10 +859,11 @@ async def cognify_status():
                     UUID(d["id"]) for d in datasets if d.get("name") == "main_dataset"
                 ]
                 if not dataset_ids:
-                    return [types.TextContent(type="text", text="{}")]
-                status = await cognee_client.get_pipeline_status(
-                    dataset_ids, "cognify_pipeline"
-                )
+                    status = "{}"
+                else:
+                    status = await cognee_client.get_pipeline_status(
+                        dataset_ids, "cognify_pipeline"
+                    )
             else:
                 # Direct mode: use local cognee functions
                 from cognee.modules.data.methods.get_unique_dataset_id import get_unique_dataset_id


Vasilije1990:
@acsezen can you please check why test was changed and there seems to be a few changes that are not relevant or needed. Would be good to streamline this to just base changes to facilitate the fixes

PR #2441, opened 2026-03-17
fix: remove gpt-5 constraints in order to make it work with LM Studio…

malahmen:
This version will now work with the LAN LM studio server, where the models required are identified in the request.

## Summary by CodeRabbit

* **Refactor**
  * Standardized model initialization so mode settings are applied consistently across all model types. This reduces variance in behavior, improves reliability when switching modes, and streamlines internal logic to make responses more predictable and stable for end users.


malahmen:
wrong repo. ignore or delete this crap.

PR #2470, opened 2026-03-23
fix(llm): pass custom endpoint config and harden embedding fallback

shortslab-maker:
## Summary
- pass `endpoint` and `api_version` through the custom LLM adapter path
- harden embedding retrieval when providers return empty vectors
- retry embedding requests without `dimensions` when unsupported or when empty vectors are returned
- add regression tests for custom adapter wiring and embedding fallback behavior

## Root Cause
The custom provider path created `GenericAPIAdapter` without forwarding the configured endpoint and API version.
Separately, vector search code assumed at least one embedding was always returned and indexed into the first vector directly.

## Validation
- targeted unit tests for custom LLM adapter configuration
- targeted unit tests for empty-embedding and no-dimensions fallback behavior
- manual MCP `CHUNKS` call returned successfully without MCP error

## Release Notes

* **New Features**
  * Enhanced CORS configuration to support expanded local network access

* **Bug Fixes**
  * Improved embedding provider error handling with explicit validation for empty embeddings
  * Added intelligent retry logic for embedding requests with parameter incompatibilities

* **Tests**
  * Added test coverage for embedding engine edge cases
  * Added test coverage for custom LLM provider configuration

* **Chores**
  * Updated database dependency configuration

### DIFF
--- a/cognee/infrastructure/databases/vector/chromadb/ChromaDBAdapter.py
+++ b/cognee/infrastructure/databases/vector/chromadb/ChromaDBAdapter.py
@@ -11,7 +11,7 @@ from cognee.infrastructure.engine import DataPoint
 from cognee.infrastructure.engine.utils import parse_id
 from cognee.infrastructure.databases.vector.exceptions import CollectionNotFoundError
 from cognee.infrastructure.databases.vector.models.ScoredResult import ScoredResult
-from cognee.infrastructure.databases.exceptions import MissingQueryParameterError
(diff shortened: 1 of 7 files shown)


Vasilije1990:
Closing: this PR has merge conflicts and no follow-up activity. If the custom endpoint config and embedding fallback fix is still needed, a fresh PR against current main would be welcome. Thank you for the contribution!

PR #2486, opened 2026-03-26
fix: coerce empty port string to None in create_graph_engine

soichisumi:
## Problem

When `GRAPH_DATABASE_PORT` is unset or empty, `get_graph_context_config()` returns `graph_database_port=""` (empty string). This value is passed directly to community adapters (via `supported_databases` dict), which may call `int(port)`, causing:

```
ValueError: invalid literal for int() with base 10: ''
```

This affects FalkorDB and potentially other community adapters that expect a numeric port.

## Fix

Coerce empty string and `None` to `None` before passing to adapters in `_create_graph_engine()`. This lets each adapter apply its own default port value (e.g. FalkorDB defaults to 6379).

## Impact

- Only affects the `supported_databases` code path (community adapters)
- Built-in adapters (Neo4j, Kuzu, Neptune) are unaffected as they don't use the port parameter directly
- No breaking changes — adapters already need to handle `None` port gracefully

## Summary by CodeRabbit

* **Bug Fixes**
  * Normalized empty or missing graph database port values so they’re treated consistently, preventing redundant cache entries and improving connection reliability.

### DIFF
--- a/cognee/infrastructure/databases/graph/get_graph_engine.py
+++ b/cognee/infrastructure/databases/graph/get_graph_engine.py
@@ -39,6 +39,13 @@ def create_graph_engine(
     Wrapper function to call create graph engine with caching.
     For a detailed description, see _create_graph_engine.
     """
+    # Normalize empty/None port before the lru_cache boundary to avoid
+    # duplicate cache entries for "" vs None.
+    normalized_port = (
+        None
+        if graph_database_port in ("", None)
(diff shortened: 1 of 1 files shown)


dexters1 (changes requested):
@soichisumi thank you for raising the issue, we have to resolve it in a different way though. @siillee can you add a ticket for yourself to resolve this when you get time?

PR #2487, opened 2026-03-26
fix: add EdgeType nodes to graph DB for GRAPH_COMPLETION search

soichisumi:
## Problem

`index_graph_edges()` creates `EdgeType` datapoints and calls `index_data_points()`, but some graph adapters (specifically FalkorDB via `cognee-community-hybrid-adapter-falkor`) don't persist nodes through `index_data_points()`. This results in EdgeType entries existing without vector embeddings in the graph database.

The consequence is that `GRAPH_COMPLETION` triplet search fails silently because it cannot find EdgeType vectors to match against query embeddings.

## Fix

Add an explicit `graph_engine.add_nodes(edge_type_datapoints)` call before `index_data_points()` in `index_graph_edges()`. This ensures EdgeType nodes are persisted with their vector embeddings in the graph database regardless of the adapter's `index_data_points` implementation.

## Impact

- Fixes `GRAPH_COMPLETION` search returning empty results on FalkorDB
- No impact on adapters where `index_data_points` already persists nodes (e.g. Kuzu, LanceDB) — the `add_nodes` call is idempotent
- The additional `add_nodes` call only runs when there are edge type datapoints to persist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved persistence of edge-type data to the graph database for more reliable storage and access across different adapters.
  * Ensured indexing proceeds only after edge data is persisted, reducing data loss or inconsistency.
  * Unified and clearer error reporting for failures during edge retrieval/transform/persist/index operations, aiding troubleshooting.

### DIFF
--- a/cognee/tasks/storage/index_graph_edges.py
+++ b/cognee/tasks/storage/index_graph_edges.py
@@ -53,7 +53,8 @@ async def index_graph_edges(
     Steps:
     1. Initialize the graph engine if needed and retrieve edge data.
     2. Transform edge data into EdgeType datapoints.
-    3. Index the EdgeType datapoints using the standard indexing function.
+    3. Persist EdgeType nodes in the graph database for triplet search.
+    4. Index the EdgeType datapoints using the standard indexing function.
 
     Raises:
         RuntimeError: If initialization of the graph engine fails.
@@ -61,6 +62,7 @@ async def index_graph_edges(
     Returns:
         None
     """
+    graph_engine = None
     try:
         if edges_data is None:
             graph_engine = await get_graph_engine()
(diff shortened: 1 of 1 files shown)


soichisumi:
## Update: v0.5.5 compatibility

I noticed that `index_graph_edges.py` was significantly rewritten in v0.5.5 ([relevant changes](https://github.com/topoteretes/cognee/compare/v0.5.3...v0.5.5)):

1. **`create_edge_type_datapoints()`** was extracted as a standalone helper function
2. **`vector_engine` parameter** was added to `index_graph_edges()` signature
3. **`index_data_points()`** now receives `vector_engine=vector_engine` kwarg

This PR's diff is based on the v0.5.3 code and no longer applies cleanly to `main` (v0.5.5+).

### The root cause still exists in v0.5.5

The v0.5.5 `index_graph_edges` calls `index_data_points(edge_type_datapoints, vector_engine=vector_engine)`, but for adapters where `vector_engine.index_data_points()` doesn't persist nodes to the graph DB (e.g., FalkorDB hybrid adapter), EdgeType nodes still won't have vector embeddings in the graph.

The fix remains the same: call `graph_engine.add_nodes(edge_type_datapoints)` before `index_data_points()`. I'll rebase this PR onto `main` to match the current code.

soichisumi:
Correction: I checked and this PR already applies cleanly to current `main` (v0.5.5) — no rebase needed. The diff correctly targets the rewritten `index_graph_edges.py` with `create_edge_type_datapoints()` and `vector_engine` parameter. Status is MERGEABLE with no conflicts.

soichisumi:
Thanks for the feedback! I looked into this further, and I believe this gap affects nearly all adapter configurations, not just FalkorDB.                                  
                                     
  In [add_data_points.py (L106-107),](https://github.com/topoteretes/cognee/blob/5469622dabc528d59921be711ade5bcb379889e7/cognee/tasks/storage/add_data_points.py#L106-L107) regular nodes go through two steps:
```                                                               
  await graph_engine.add_nodes(nodes)                            # persist to graph DB
  await index_data_points(nodes, vector_engine=vector_engine)    # embed in vector engine
```   


But index_graph_edges (L75-76) only calls index_data_points for EdgeType — the  add_nodes step is missing.
Since the core task calls the vector engine's index_data_points, only Neptune Analytics (whose hybrid create_data_points MERGEs into the graph as a side effect) ends up with EdgeType nodes in the graph DB. 


Every other configuration — PGVector, ChromaDB, LanceDB, QDrant, DuckDB, and yes FalkorDB — is missing them, which breaks GRAPH_COMPLETION triplet search.

This PR just applies the same two-step pattern to EdgeType that regular nodes already use. add_nodes is MERGE-based (upsert), so it's idempotent on Neptune. I think one fix in core is cleaner than patching each adapter individually — what do you think? ( If I'm misreading the codebase, happy to be corrected! ) 
@lxobr

lxobr:
Thanks for the follow up research @soichisumi! Will look into it and get back to you tomorrow.

soichisumi:
Thank you!

We could also move this to cognee-community.
I was just checking because I thought it might be cleaner to implement it here.

Vasilije1990:
@soichisumi good job. I will merge this and we will check internally later on if we can optimize.

PR #2506, opened 2026-03-27
Feature/cog 4463 graph aware embeddings implement minimal graphdb in postgres

DanielNScott:
## Description
Implementation of a minimal graph DB in postgres

## Acceptance Criteria
- new adapter implements all base class methods except query()

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

## Summary by CodeRabbit

* **New Features**
  * Added PostgreSQL as a supported graph database backend.

* **Documentation**
  * Added PostgreSQL configuration examples and documented behavioral limitations versus other graph backends.

* **Tests**
  * Added unit and end-to-end tests for the PostgreSQL graph backend.
  * CI workflow updated with a new Postgres test job and a reusable python-version input.

* **Bug Fixes / Behavior**
  * Retrieval and temporal-awareness flows now fail fast with clear errors for the PostgreSQL backend; fixed a log-message typo.


DanielNScott:
@dexters1 @lxobr Igor, this is the minimal postgres graph database backend. The other PR is a more complex hybrid adapter on top of it to deal with integrating graph and vector queries.

cognee/infrastructure/databases/graph/postgres/adapter.py
```diff
@@ -0,0 +1,696 @@
...
+        params = {}
+        for i, filter_dict in enumerate(attribute_filters):
+            for attr, values in filter_dict.items():
+                param = f"filt_{i}_{attr}"
+                where_parts.append(f"n.{attr} = ANY(:{param})")
+                params[param] = values
+
+        where_clause = " AND ".join(where_parts)
+
+        async with self._session() as session:
```
lxobr, line 503: Should we check if where_clause is empty at some point? Otherwise the query below will fail.

cognee/infrastructure/databases/graph/postgres/adapter.py
```diff
@@ -0,0 +1,696 @@
...
+    async def add_edges(
+        self, edges: Union[List[Tuple[str, str, str, Optional[Dict[str, Any]]]], List]
+    ) -> None:
+        if not edges:
+            return
+
+        now = datetime.now(timezone.utc)
+
+        # Normalize edge tuples into dicts matching table columns
+        values = []
```
lxobr: Nit: `values = []` shadows the module-level `from sqlalchemy import ... values ...`.

cognee/infrastructure/databases/graph/postgres/adapter.py
```diff
@@ -0,0 +1,696 @@
...
+            core, extra_json = self._split_core_and_extra(props)
+            rows.append({**core, "properties": extra_json, "now": now})
+
+        # Remap 'now' into the column names the table expects
+        values = [
+            {
+                "id": r["id"],
+                "name": r["name"],
+                "type": r["type"],
+                "properties": json.loads(r["properties"]),
```
lxobr: Minor: It seems that `_split_core_and_extra` returns a JSON string, and this line immediately calls `json.loads` on it, which feels like a redundant serialize → deserialize.

cognee/infrastructure/databases/graph/postgres/adapter.py
```diff
@@ -0,0 +1,696 @@
...
+            "The Postgres graph backend does not support raw Cypher queries. "
+            "Use a graph-native backend (Neo4j, Kuzu) for raw query support, "
+            "or use the typed adapter methods (add_nodes, get_neighbors, etc.)."
+        )
+
+    # ------------------------------------------------------------------
+    # GraphDBInterface: node operations
+    # ------------------------------------------------------------------
+
+    async def is_empty(self) -> bool:
```
lxobr, line 73: Every call to `is_empty()` runs `CREATE TABLE IF NOT EXISTS` against the database, even when the tables already exist. Maybe we can add a `self._initialized = False` flag, set it to `True` after the first successful `initialize()`, and skip the call if it's already `True`?

cognee/infrastructure/databases/graph/postgres/adapter.py
```diff
@@ -0,0 +1,696 @@
...
+            edges = []
+            for row in edge_result.fetchall():
+                props = {}
+                if row[3]:
+                    props = row[3] if isinstance(row[3], dict) else json.loads(row[3])
+                edges.append((row[0], row[1], row[2], props))
+
+            return nodes, edges
+
+    async def get_nodeset_subgraph(
```
lxobr, line 542: We have 4 db calls here, and it seems that some of them can be combined in a CTE? Maybe worth doing for efficiency.

cognee/infrastructure/databases/graph/postgres/adapter.py
```diff
@@ -0,0 +1,696 @@
...
+                    tgt.update(row[9] if isinstance(row[9], dict) else json.loads(row[9]))
+
+                connections.append((src, edge, tgt))
+            return connections
+
+    # ------------------------------------------------------------------
+    # GraphDBInterface: graph-wide reads
+    # ------------------------------------------------------------------
+
+    async def get_graph_data(
```
lxobr, line 431: Would it make sense to gather these two database calls...?

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: The problem with the flag solution is that it can become stale (incorrect) when pruning or deletion drop graph tables (which I encountered, although I don't remember the exact context). The current pattern is defensive, and if the table exists, it's a no-op anyway, so I would argue that we should keep it. But I can look into it further if this is something we want to nail down.

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Yeah, this should be streamlined.

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Will modify.

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Definitely.

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Looks like it, good catch.

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Yes, agreed.

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Fixed

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Fixed

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Addressed -- the serialize --> deserialize loop is the correct way to use JSONEncoder given, given that postgres wants a python dictionary. As I understand it, the alternative is duplicating the JSONEncoder logic for handling serialization, which would be brittle (remove JSONEncoder as the source of truth on this, would require updating any time JSONEncoder was updated). I think another possibility would be introducing a function standalone function which makes use of JSONEncoder as necessary, or does in-place conversion for this setting, but the current change is the minimal solution, as I understand it. The indirection wasn't necessary though, so I removed that.

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Fixed

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Fixed

cognee/infrastructure/databases/graph/get_graph_engine.py
```diff
@@ -121,6 +121,15 @@ def _create_graph_engine(
...
         )
 
+    elif graph_database_provider == "postgres":
+        from .postgres.adapter import PostgresAdapter
+        from cognee.infrastructure.databases.relational.get_relational_engine import (
+            get_relational_engine,
+        )
+
+        relational_engine = get_relational_engine()
+        return PostgresAdapter(relational_engine=relational_engine)
```
dexters1: @DanielNScott The relational_engine does not have to be postgres, would be good to enable other relational engines as well. This also removes the ability for users to seperate their relational DB Postgres instance from their graph DB Postgres instance (this might be done for security reasons). Check how configuration has been handled for PGVector adapter (Our postgres vector adapter)

cognee/infrastructure/databases/graph/postgres/adapter.py
```diff
@@ -0,0 +1,798 @@
...
+
+
+class PostgresAdapter(GraphDBInterface):
+    """Graph-as-tables adapter backed by Postgres, accessed via SQLAlchemy async sessions."""
+
+    _ALLOWED_FILTER_ATTRS = {"id", "name", "type"}
+
+    def __init__(self, relational_engine: Any) -> None:
+        """Accept an existing SQLAlchemyAdapter (shared with the relational layer)."""
+        self.engine = relational_engine
```
dexters1: Let's not assume that the relational layer is shared and that it is Postgres

cognee/infrastructure/databases/graph/postgres/adapter.py
```diff
@@ -0,0 +1,798 @@
...
+        """Create tables and indexes if they do not exist."""
+        async with self.engine.engine.begin() as conn:
+            await conn.run_sync(_meta.create_all, checkfirst=True)
+
+    async def _ensure_initialized(self) -> None:
+        """Re-run initialize() if tables were dropped (e.g. by prune_system).
+
+        Uses CREATE IF NOT EXISTS, so this is cheap and idempotent.
+        """
+        await self.initialize()
```
dexters1: There is no need for this function as it just call initialize

cognee/infrastructure/databases/graph/postgres/adapter.py
```diff
@@ -0,0 +1,798 @@
...
+    _meta,
+    Column("id", String, primary_key=True),
+    Column("name", String),
+    Column("type", String),
+    Column("properties", JSONB),
+    Column("created_at", DateTime(timezone=True), server_default=func.now()),
+    Column("updated_at", DateTime(timezone=True), server_default=func.now()),
+)
+
+_edge_table = Table(
```
dexters1: Can we move the models, indexes and metadata classes to a new file? To keep code clean. We do this pattern in general when model definitions are not dynamically generated

cognee/infrastructure/databases/graph/postgres/adapter.py
```diff
@@ -0,0 +1,798 @@
...
+            yield session
+
+    def _serialize_properties(self, props: Dict[str, Any]) -> str:
+        """Serialize a dict to a JSON string, handling datetimes and UUIDs."""
+        return json.dumps(props, cls=JSONEncoder)
+
+    def _parse_node_row(self, row) -> Dict[str, Any]:
+        """Convert a (id, name, type, properties) row to a merged dict."""
+        data = {"id": row.id, "name": row.name, "type": row.type}
+        if row.properties:
```
dexters1: Is there a situation where a row will not have properties defined? If not we should raise an error and not silently proceed

cognee/infrastructure/databases/graph/postgres/adapter.py
```diff
@@ -0,0 +1,798 @@
...
+
+    async def add_nodes(self, nodes: Union[List[Tuple[str, Dict]], List[DataPoint]]) -> None:
+        """Add multiple nodes via batch upsert.
+
+        Parameters:
+        -----------
+            nodes: A list of (id, properties) tuples or DataPoint instances.
+        """
+        if not nodes:
+            return
```
dexters1, line 110: Should we raise an error if no nodes? Can there be a function call without nodes defined or defined as Null?

cognee/infrastructure/databases/graph/postgres/adapter.py
```diff
@@ -0,0 +1,798 @@
...
+        "source_id", String,
+        ForeignKey("graph_node.id", ondelete="CASCADE"),
+        primary_key=True, nullable=False,
+    ),
+    Column(
+        "target_id", String,
+        ForeignKey("graph_node.id", ondelete="CASCADE"),
+        primary_key=True, nullable=False,
+    ),
+    Column("relationship_name", String, primary_key=True, nullable=False),
```
dexters1: None of these can be a primary_key (source_id, target_id, relationship_name) as a primary_key has to be unique so a duplicate relationship name, source or target will break execution. A new id column is needed for the Edge to represent the edge as a primary_key that has to be unique

cognee/infrastructure/databases/graph/get_graph_engine.py
DanielNScott: Yes, good point, fixed.

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Fixed.

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Good catch, it was left over from an earlier version in which I experimented with additional logic there. Thank you. Fixed.

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Changed to `if row.properties is not None:` since empty dictionaries evaluate to false. (So None is the only danger.) I don't think this case should come up, but now should be safe.

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Yes, now in a tables.py file alongside. I didn't see a strong convention on this (PGVector internalizes, others store tables elsewhere). LMK if something else would be preferable.

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: I think not. This is the pattern kuzu and neptune use, and neo4j doesn't check at all, dispatching to the DB anyway. Incidentally, `if not None:` also evaluates to true, so this would work if None is passed explicitly... but that shouldn't happen. `nodes` is a required argument.

cognee/infrastructure/databases/graph/postgres/adapter.py
DanielNScott: Ok, I looked into this, and this is the pattern for the other adapters, as I understand it. Specifically, the other adapters all use merges on the same triple. Using an id column will turn the graph into a multigraph  (repeat edges of the same type between the same nodes), which is ruled out by the merge logic elsewhere. Using this as the primary key just reproduces existing logic.

Here are the other snippets:

Neo4j:
```
        query = dedent(
            f"""\
            MATCH (from_node :`{BASE_LABEL}`{{id: $from_node}}),
                  (to_node :`{BASE_LABEL}`{{id: $to_node}})
            MERGE (from_node)-[r:`{relationship_name}`]->(to_node)
            ON CREATE SET r += $properties, r.updated_at = timestamp()
            ON MATCH SET r += $properties, r.updated_at = timestamp()
            RETURN r
            """
        )

```
Kuzu:
```
        query = """
            MATCH (from:Node), (to:Node)
            WHERE from.id = $from_id AND to.id = $to_id
            MERGE (from)-[r:EDGE {
                relationship_name: $relationship_name
            }]->(to)
            ON CREATE SET
                r.created_at = timestamp($created_at),
                r.updated_at = timestamp($updated_at),
                r.properties = $properties
            ON MATCH SET
                r.updated_at = timestamp($updated_at),
                r.properties = $properties
        """
```

Neptune:
```
            query = f"""
            MATCH (source:{self._GRAPH_NODE_LABEL})
            WHERE id(source) = $source_id
            MATCH (target:{self._GRAPH_NODE_LABEL})
            WHERE id(target) = $target_id
            MERGE (source)-[r:{relationship_name}]->(target)
            ON CREATE SET r = $properties, r.updated_at = timestamp()
            ON MATCH SET r = $properties, r.updated_at = timestamp()
            RETURN r
            """
```

So I think this is the right approach. LMK if I'm misunderstanding something though.

cognee/tests/e2e/postgres/test_graphdb_shared.py
```diff
@@ -0,0 +1,144 @@
...
+        vector_engine = get_vector_engine()
+        random_node = (
+            await vector_engine.search("Entity_name", "Quantum computer", include_payload=True)
+        )[0]
+        random_node_name = random_node.payload["text"]
+
+        # Test GRAPH_COMPLETION search (exercises graph adapter)
+        search_results = await cognee.search(
+            query_type=SearchType.GRAPH_COMPLETION, query_text=random_node_name
+        )
```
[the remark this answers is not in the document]
DanielNScott: Not actually an issue - test doesn't use query().

cognee/tests/e2e/postgres/test_graphdb_shared.py
```diff
@@ -0,0 +1,144 @@
...
+
+    # Set up isolated directories for this test run
+    base = pathlib.Path(__file__).parent
+    data_dir = str((base / f".data_storage/test_{provider}").resolve())
+    system_dir = str((base / f".cognee_system/test_{provider}").resolve())
+
+    try:
+        cognee.config.set_graph_database_provider(provider)
+        cognee.config.data_root_directory(data_dir)
+        cognee.config.system_root_directory(system_dir)
```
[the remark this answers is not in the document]
DanielNScott: Not a problem, because this is a stand-alone script (test directory contains pre existing mixture of pytest and non-pytest...). Fixed anyway.

DanielNScott:
@dexters1 See last few commit messages for some important updates since your review. Thanks!

cognee/infrastructure/databases/graph/postgres/adapter.py
dexters1: Yes, you are correct. We should not have repeat edges. I was under the impression that the syntax indicated that each column was a primary key, was not aware that SQLAlchemy under the hood transformed that into one composite primary key for the three columns, the syntax for that is different in raw SQL. So all good after all.

cognee/infrastructure/databases/graph/postgres/adapter.py
dexters1: Thanks! For PGVector the models are internalized because they are fully dynamic and depend on the function input

cognee/tests/e2e/postgres/test_graphdb_postgres.py
```diff
@@ -0,0 +1,18 @@
+"""End-to-end graph database test using the Postgres backend.
+
+Requires:
+  - A running Postgres instance
```
dexters1, line 4: Can we move unit tests that require a running Postgres instance to integration tests?

Unit tests have to mock DBs and DB communication

dexters1:
@DanielNScott can you format your code with ruff? Seems the checks for this are failing: https://github.com/topoteretes/cognee/actions/runs/24124924222/job/70387847664?pr=2506

dexters1:
@DanielNScott @lxobr do we want to also enable multiple users / multiple datasets to have their own postgres graphs in this PR like we do with PGVector? Or is this a later step?

Adding a DatasetDatabase handler like we have for PGVector (and registering it as an option): https://github.com/topoteretes/cognee/blob/b51dcce1d273d47ce864cd6c5e44a7a82f7f8dce/cognee/infrastructure/databases/vector/pgvector/PGVectorDatasetDatabaseHandler.py#L12

but for graph databases would enable this. E.G how we handle it for graphs with Kuzu:
https://github.com/topoteretes/cognee/blob/b51dcce1d273d47ce864cd6c5e44a7a82f7f8dce/cognee/infrastructure/databases/graph/kuzu/KuzuDatasetDatabaseHandler.py#L12

Ideally the PGVector and PostgresGraph implementations could use the same database if the postgres instance is shared

cognee/tests/e2e/postgres/test_graphdb_postgres.py
DanielNScott: Yes, got confused about which branch this was already fixed on. Fixed, thanks.

DanielNScott:
@dexters1 @lxobr 

As I understand it, the test failures above are a result of the PR coming from a branch outside of the topoteretes/cognee repository.

The build environment in `.github/workflows/test_suites.yml` shows the following:

Pull origin case testing:
```yml
  build-ci-env:
    name: Build CI Environment
    runs-on: ubuntu-latest
    if: >-
      github.event_name == 'push' || github.event_name == 'workflow_dispatch' ||
      (github.event_name == 'pull_request' &&
       github.event.pull_request.head.repo.full_name == github.repository)
```
The if is not satisfied, then the build-ci-env actions substance is skipped, and as a result, secret inheritance below (which proceeds because nothing was cancelled and the pre-test block immediately above it was successful) fails. 

```yml
  basic-tests:
    name: Basic Tests
    uses: ./.github/workflows/basic_tests.yml
    needs: [ pre-test, build-ci-env ]
    if: ${{ !cancelled() && needs.pre-test.result == 'success' }}
    with:
      ci-image: ${{ needs.build-ci-env.outputs.ci-image || '' }}
    secrets: inherit
 ```

Other test groups have the same pattern. 

I don't have the ability to fork within Cognee (no write access), which is why I forked this to my GH account in the first place, not realizing this was going to be an issue. So IIUC we need to either change my permissions in topoteretes/cognee or have one of you set up a branch there. Then we need to migrate this content there and PR. Unless there is something you guys can do from upstream.

LMK how you'd like me to proceed.

PR #2530, opened 2026-03-31
fix: use dict spread to avoid shallow copy mutation in index_data_points

### LINKED ISSUE #2529: [Bug]index_data_points: shallow copy of metadata dict causes only first index_field to be embedded
## Title

`index_data_points`: shallow copy of `metadata` dict causes only first `index_field` to be embedded

## Description

When a custom `DataPoint` defines multiple `index_fields` (e.g., `["problem", "conclusion", "follow_up"]`), only the first field gets properly embedded. The remaining collections are created but contain embeddings and text from the first field.

### Root Cause

In `cognee/tasks/storage/index_data_points.py`, lines 36-48:

```python
for field_name in data_point.metadata["index_fields"]:  # iterates over original list
    # ...
    indexed_data_point = data_point.model_copy()  # shallow copy
    indexed_data_point.metadata["index_fields"] = [field_name]  # mutates the ORIGINAL dict!
```

`model_copy()` (Pydantic v2) performs a **shallow copy** of dict fields. Since `indexed_data_point.metadata` and `data_point.metadata` point to the **same dict object**, the assignment `indexed_data_point.metadata["index_fields"] = [field_name]` mutates the original `data_point.metadata["index_fields"]` from `["problem", "conclusion", "follow_up"]` to `["problem"]`.

This causes the `for` loop to terminate after the first iteration, as the list it's iterating over has been truncated to a single element.

### Reproduction

```python
from cognee.tasks.storage import add_data_points
import cognee

class MyCase(DataPoint):
    problem: str = ""
    conclusion: str = ""
    metadata: dict = {"index_fields": ["problem", "conclusion"]}

async def main():

    from cognee.modules.engine.operations.setup import setup
    await setup()

    cases = [
        MyCase(problem="PROBLEM_AAA", conclusion="CONCLUSION_BBB"),
    ]
    await add_data_points(cases)

    # Check: both collections have the same content (should be different)
    from cognee.infrastructure.databases.vector import get_vector_engine
    engine = get_vector_engine()
    conn = await engine.get_connection()
    for coll in ["MyCase_problem", "MyCase_conclusion"]:
        table = await conn.open_table(coll)
        data = await table.to_arrow()
        text = data.column("payload")[0].as_py().get("text", "")
        print(f"{coll}: {text}")
        # Both print "CONCLUSION_BBB" — should be "PROBLEM_AAA" and "CONCLUSION_BBB" respectively

import asyncio
asyncio.run(main())
```

### Minimal verification of the shallow copy issue

```python

class TestDP(DataPoint):
    a: str = ""
    b: str = ""
    metadata: dict = {"index_fields": ["a", "b"]}

dp = TestDP(a="AAA", b="BBB")
clone = dp.model_copy()
clone.metadata["index_fields"] = ["b"]

print(dp.metadata["index_fields"])
# Output: ["b"]  — original object was mutated!
# Expected: ["a", "b"]
```

### Suggested Fix

Two-line change in `cognee/tasks/storage/index_data_points.py`:

```diff
-        for field_name in data_point.metadata["index_fields"]:
+        for field_name in list(data_point.metadata["index_fields"]):
             # ...
             indexed_data_point = data_point.model_copy()
+            indexed_data_point.metadata = dict(data_point.metadata)
             indexed_data_point.metadata["index_fields"] = [field_name]
```

1. `list(...)` creates a copy of the field names list before iteration, preventing the loop from being truncated.
2. `dict(...)` creates a shallow copy of the metadata dict, preventing mutation of the original DataPoint's metadata.

Using `model_copy(deep=True)` would also work but is unnecessarily expensive since only the metadata dict needs isolation.

### Environment

- cognee version: 0.5.5
- Python: 3.12
- Pydantic: v2
- Vector DB: LanceDB

### Impact

Any user defining a custom DataPoint with multiple `index_fields` will silently get incorrect embeddings — all vector collections will contain the same embeddings from the first field only. This makes multi-field semantic search ineffective.

nightcityblade:
## Description

When a `DataPoint` has multiple `index_fields`, the `for` loop in `index_data_points()` only processes the first field. This happens because `model_copy()` performs a shallow copy, so `indexed_data_point.metadata` and `data_point.metadata` reference the same dict. Setting `indexed_data_point.metadata["index_fields"] = [field_name]` mutates the original, truncating the iteration list.

The fix replaces the direct dict mutation with a dict spread (`{**data_point.metadata, "index_fields": [field_name]}`), creating a new dict for each copy so the original is never modified.

## Acceptance Criteria

* Multiple `index_fields` are all processed and embedded correctly
* Original `data_point.metadata["index_fields"]` is not mutated during iteration

## Type of Change
- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots
N/A — single-line logic fix verified via manual simulation.

## Pre-submission Checklist

## DCO Affirmation

Fixes #2529

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed metadata preservation during data point indexing to ensure all metadata fields are properly retained while correctly managing index field information.

### DIFF
--- a/cognee/tasks/storage/index_data_points.py
+++ b/cognee/tasks/storage/index_data_points.py
@@ -43,7 +43,7 @@ async def index_data_points(data_points: list[DataPoint]):
                 data_points_by_type[type_name][field_name] = []
 
             indexed_data_point = data_point.model_copy()
-            indexed_data_point.metadata["index_fields"] = [field_name]
+            indexed_data_point.metadata = {**data_point.metadata, "index_fields": [field_name]}
             data_points_by_type[type_name][field_name].append(indexed_data_point)
 
     batch_size = vector_engine.embedding_engine.get_batch_size()


dexters1 (changes requested):
PR is made towards wrong branch. Check CONTRIBUTING.md

https://github.com/topoteretes/cognee/blob/main/CONTRIBUTING.md

nightcityblade:
Thanks @dexters1! Updated the base branch to `dev` as per CONTRIBUTING.md. Sorry about that!

nightcityblade:
Thanks for the feedback! The PR currently targets `dev` as the base branch, which aligns with CONTRIBUTING.md's guidance to branch from `dev`. Could you clarify which branch I should target instead? Happy to update it.

lxobr:
Thanks for the PR. I agree this change avoids a shallow-copy metadata mutation issue, but I don’t think we should expand support for multiple `index_fields` right now.

The confusing part is that `index_fields` is a list, but the retrieval flow does not fully support it as “multiple searchable fields.” Indexing may create one vector index per field, but graph retrieval later collapses matches for the same node into a single score. So for example, if a node matches on both `name` and `description`, we do not have clear behavior for how those scores should be combined.

Because of that, this fixes one symptom but makes multi-field indexing look more supported than it is. For now I’d rather treat `index_fields` as effectively single-field, and revisit multiple fields separately when we define the retrieval/scoring behavior properly.

nightcityblade:
Thanks, that makes sense. I agree the broader multi-field retrieval/scoring behavior should be designed separately. I’m happy to close this PR instead of pushing the current symptom-level fix further if you’d prefer to keep `index_fields` effectively single-field for now.

PR #2549, opened 2026-04-04
fix: disable search history by default to limit the potential impact of table size growth

Vasilije1990:
## Summary
Fixes unbounded growth of `queries` and `results` tables in the relational DB (SQLite/Postgres). Reported case: 28 GB SQLite file in 9 days at ~4700 queries/day.

- **FIFO eviction**: caps rows per table at `COGNEE_MAX_SEARCH_HISTORY` (default 10,000, set to 0 for unlimited)
- **Batched**: eviction runs every ~1000 writes (1/10th of cap), not on every search call
- **SQLite VACUUM**: optional auto-VACUUM after eviction to reclaim disk space (`COGNEE_SEARCH_HISTORY_VACUUM=true` by default)
- Wired into `log_result()` and `log_query()` with zero impact when cap not reached

New env vars documented in `.env.template`.

## Test plan
- [ ] CI passes
- [ ] Run 100+ searches, verify `queries` and `results` tables don't exceed cap
- [ ] Verify `COGNEE_MAX_SEARCH_HISTORY=0` disables eviction
- [ ] Verify SQLite file size decreases after VACUUM

## Summary by CodeRabbit

* **New Features**
  * Added an optional (commented) configuration to enable/disable search history logging.
  * Search logging now records only user-visible text from results (not full payloads), reducing persisted detail.

* **Chores**
  * Improved database performance by adding indexes on search history timestamps and optimizing user-related lookups.
  * When logging is disabled, search events avoid persistence to the database.

### DIFF
--- a/cognee/alembic/versions/b2c3d4e5f6a7_add_search_history_indexes.py
+++ b/cognee/alembic/versions/b2c3d4e5f6a7_add_search_history_indexes.py
@@ -1,0 +1,23 @@
+"""add created_at index to queries and results tables
+
+Revision ID: b2c3d4e5f6a7
+Revises:
+Create Date: 2026-04-04
+"""
+
+from alembic import op
+
+revision = "b2c3d4e5f6a7"
+down_revision = None
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_index("ix_queries_created_at", "queries", ["created_at"])
+    op.create_index("ix_results_created_at", "results", ["created_at"])
+
+
+def downgrade() -> None:
+    op.drop_index("ix_results_created_at", "results")
+    op.drop_index("ix_queries_created_at", "queries")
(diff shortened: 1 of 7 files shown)


cognee/alembic/versions/b2c3d4e5f6a7_add_search_history_indexes.py
```diff
@@ -0,0 +1,23 @@
...
+from alembic import op
+
+revision = "b2c3d4e5f6a7"
+down_revision = None
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_index("ix_queries_created_at", "queries", ["created_at"])
```
dexters1, line 17: We might need to have a check if this index exists before trying to make it otherwise migrations could fail. (Same applies for dropping an index)

cognee/modules/search/operations/log_query.py
```diff
@@ -1,20 +1,24 @@
+import os
 from uuid import UUID
 from cognee.infrastructure.databases.relational import get_relational_engine
 from ..models.Query import Query
 
+_LOG_ENABLED = os.getenv("COGNEE_LOG_SEARCH_HISTORY", "true").lower() in ("true", "1", "yes")
```
dexters1, line 6: Should it be disabled by default? PR title mentions that it should

PR #2579, opened 2026-04-09
fix: fix issues in sqlite and postgres when dataset name is reused

siillee:
## Description

This PR fixes an issue we had in the following execution flow:
`cognee.add(dataset_name="test") -> delete_dataset(test) -> cognee.add(dataset_name="test")`

When we tried to add documents to the same dataset name as a previously deleted dataset, we had connection errors in postgres. This was coming from stale engine and connection pools towards the deleted dataset.

Similar issue with SQLite, the foreign key enforcement is off by default, so it caused an issue where the entries in the `dataset_database` were not deleted properly, causing old (deleted) datasets to return even after deletion.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

## Summary by CodeRabbit

* **Bug Fixes**
  * Centralized vector DB engine initialization to improve connection reliability and pooling.
  * Guaranteed disposal/cleanup of maintenance DB connections to prevent resource leaks during create/delete operations.
  * Enforced SQLite foreign keys during delete operations so related records are correctly honored and removed.


cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py
```diff
@@ -66,15 +66,15 @@ def __init__(
         db_name2 = make_url(self.db_uri).database
         if backend_access_control_enabled() and (db_name1 != db_name2):
             # If backend access control create new instances of engine and sessionmaker
-            self.engine = create_async_engine(self.db_uri)
+            self.engine = create_async_engine(self.db_uri, pool_pre_ping=True)
```
dexters1: Are we not forwarding all postgres configuration here? This might be another issue we need to resolve

cognee/infrastructure/databases/vector/pgvector/create_db_and_tables.py
```diff
@@ -75,17 +108,46 @@ async def delete_pg_database(dataset_database: DatasetDatabase):
...
+                ),
+                {"db": dataset_database.vector_database_name},
+            )
+            await connection.execute(
+                text(f'DROP DATABASE "{dataset_database.vector_database_name}";')
+            )
+        finally:
+            await connection.close()
+    finally:
+        await maintenance_engine.dispose()
```
dexters1, line 99: What does dispose do? Disposing of the maintenance engine while another DB uses it could be a problem

siillee: This was added as just an extra cleaning step. It closes the engine's connection pool and resets the pool state.
`maintenance_engine` is used when creating a new dataset database, and deleting it, and with each operation a new engine is created (it is not cached). Therefore, I think there won't be issues with any usage overlaps.

cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py
siillee: Now we are using the `pool_args` from the `relational_db_config` for PGVector as well, in cases where we are creating a new engine for PGVector.

siillee: Turns out `pool_args` was all that was missing, I removed the new dispose I added, since it was unnecessary in the end.

cognee/infrastructure/databases/vector/pgvector/PGVectorDatasetDatabaseHandler.py
```diff
@@ -74,3 +77,7 @@ async def delete_dataset(cls, dataset_database: DatasetDatabase):
 
         # Drop entire database
         await delete_pg_database(dataset_database)
+
+        # Reset cached metadata from the vector adapter
+        vector_engine = get_vector_engine()
```
dexters1: I'm not sure proper dataset context is set to be able to do this here

dexters1 (dismissed):

dexters1 (approved):

PR #2597, opened 2026-04-11
fix: V2 remember endpoint returns 500 due to unserializable response

Vasilije1990:
## Summary
- The `POST /v2/remember` endpoint returned a `RememberResult` object directly, but the `response_model=dict` declaration caused FastAPI's response validation to reject it with `ResponseValidationError`
- The pipeline itself succeeded — only the serialization step failed, causing a misleading 500
- Added `RememberResult.to_dict()` for reusable serialization and call it in the router

## Root cause
`get_remember_router.py` line 84 returned the raw `RememberResult` object. FastAPI tried to validate it against `response_model=dict` and raised:
```
fastapi.exceptions.ResponseValidationError: 1 validation error:
  {'type': 'dict_type', 'loc': ('response',), 'msg': 'Input should be a valid dictionary',
   'input': RememberResult(status='completed', ...)}
```

## Test plan
- [ ] `curl -X POST /api/v2/remember -F 'data=@test.txt' -F 'datasetName=test'` returns JSON dict with `status`, `dataset_name`, `dataset_id`, etc.
- [ ] Background mode (`run_in_background=true`) also returns a valid dict
- [ ] Session mode (`session_id=...`) also returns a valid dict

## Summary by CodeRabbit

* **Refactor**
  * Updated workspace creation to return unavailable status in local mode.
  * Restructured remember endpoint response serialization to return JSON-encodable dicts.

* **Chores**
  * Updated default sign-in email.
  * Removed tenant user management API routes.
  * Removed unused pod context caching logic.


cognee/api/v2/remember/routers/get_remember_router.py
```diff
@@ -81,7 +81,7 @@ async def remember(
                 chunks_per_batch=chunks_per_batch,
             )
 
-            return result
+            return result.to_dict()
```
dexters1: jsonable_encoder using this function should be a cleaner way to do it, but we need to test

Issue #2607, opened 2026-04-12
HTTP/2 stream stalls cause 60s timeouts on sequential Anthropic API calls


MDGrey33:
# Bug Report: HTTP/2 stream stalls cause 60-second timeouts on sequential Anthropic API calls

**Repository:** topoteretes/cognee
**Component:** `cognee/infrastructure/llm/structured_output_framework/litellm_instructor/llm/anthropic/adapter.py`
**Severity:** High — makes `cognee.search()` unreliable; roughly 40% of calls timeout in typical usage

---

## Environment

- cognee: 0.5.7
- anthropic SDK: 0.94.0
- instructor: 1.14.4
- httpx: (default version installed with anthropic SDK)
- macOS Darwin 25.3.0 (Apple Silicon M3 Max)
- Python: 3.12

---

## Reproduction Steps

1. Configure cognee with Anthropic as LLM provider (`LLM_PROVIDER=anthropic`, `LLM_MODEL=claude-haiku-4-5-20251001`)
2. Run cognify on a document set so the graph is populated
3. Execute 5 sequential `cognee.search()` calls with `query_type=GRAPH_COMPLETION`:

```python
import asyncio
import cognee
from cognee.api.v1.search.search import SearchType

async def run_searches():
    queries = [
        "Who is Evika Silina?",
        "What political party does she lead?",
        "What is her role in government?",
        "What are her key policy positions?",
        "Who are her political allies?",
    ]
    for q in queries:
        print(f"Searching: {q}")
        result = await cognee.search(q, query_type=SearchType.GRAPH_COMPLETION)
        print(f"  -> {len(result)} results")

asyncio.run(run_searches())
```

**Expected:** All 5 searches complete in 1–3 seconds each.

**Actual:** Exactly 2 of 5 searches hang for 60.0–60.4 seconds before timing out. The specific queries that fail rotate between runs — it is not always the same 2 queries.

---

## Root Cause

The `AnthropicAdapter` constructor in `adapter.py` calls `anthropic.AsyncAnthropic()` without specifying an `http_client`:

```python
# Current code (simplified)
self.aclient = instructor.from_anthropic(
    anthropic.AsyncAnthropic(api_key=self.api_key)
)
```

[...]

PR #2609, opened 2026-04-12
fix(llm): disable HTTP/2 on Anthropic adapter to prevent stream stalls

Vasilije1990:
## Summary
- Passes `http_client=httpx.AsyncClient(http2=False)` to `AsyncAnthropic()` in the Anthropic adapter constructor
- Fixes HTTP/2 stream multiplexing stalls that cause ~40% of sequential Anthropic API calls to timeout at 60 seconds
- `httpx` is already a transitive dependency of the `anthropic` SDK — no new dependencies

## Root cause
`get_llm_client()` creates a new `AnthropicAdapter` (and thus a new `AsyncAnthropic` + httpx connection pool) on every call. The HTTP/2 multiplexing state left by previous pools interferes with new connections on the same TCP session, causing exactly 2 out of every 5 sequential requests to stall for 60s (the default httpx timeout).

## Test plan
- [ ] Configure `LLM_PROVIDER=anthropic` with a valid API key
- [ ] Run 5 sequential `cognee.search()` calls with `query_type=GRAPH_COMPLETION`
- [ ] Verify all 5 complete in 1–3s each with no 60s timeouts

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Anthropic LLM adapter HTTP connection handling for enhanced stability and compatibility.

### DIFF
--- a/cognee/infrastructure/llm/structured_output_framework/litellm_instructor/llm/anthropic/adapter.py
+++ b/cognee/infrastructure/llm/structured_output_framework/litellm_instructor/llm/anthropic/adapter.py
@@ -1,6 +1,7 @@
 import logging
 from typing import Any, Dict, Type, Optional
 from pydantic import BaseModel
+import httpx
 import litellm
 import instructor
 import anthropic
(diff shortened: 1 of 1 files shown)


dexters1 (changes requested):
Do we want to disable HTTP2?

https://github.com/topoteretes/cognee/issues/2607

User who raised issue mentioned caching would resolve issue without the need for this change

PR #2613, opened 2026-04-13
fix(llm): respect user's LLM_MAX_COMPLETION_TOKENS setting

Vasilije1990:
## Summary
- Uses `min(model_max, user_max)` instead of unconditionally preferring litellm's `model_cost["max_tokens"]`
- User's `LLM_MAX_COMPLETION_TOKENS` now acts as a ceiling; litellm's value only guards against exceeding the model's hard limit
- Fixes silent config override that caused Anthropic SDK to raise `ValueError("Streaming is required for operations that may take longer than 10 minutes")`

## Root cause
For `claude-haiku-4-5-20251001`, litellm's `model_cost` reports `max_tokens = 64000`. The previous logic used this unconditionally, ignoring the user's `.env` setting (e.g., `LLM_MAX_COMPLETION_TOKENS=4096`). The Anthropic SDK then computed a 1800-second timeout from the inflated token count, triggering its streaming-required safety check.

## Test plan
- [ ] Set `LLM_MAX_COMPLETION_TOKENS=4096` in `.env` with Anthropic provider
- [ ] Run `cognee.cognify()` and `cognee.search()` — should succeed without streaming error
- [ ] Verify that without the env setting, litellm's model max is still used as the default
- [ ] Verify that setting a value higher than the model max gets capped to the model max

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed token limit computation to consistently respect user-configured maximum completion token settings as the primary ceiling, with model-specific limits applied as additional constraints when available.

### DIFF
--- a/cognee/infrastructure/llm/structured_output_framework/litellm_instructor/llm/get_llm_client.py
+++ b/cognee/infrastructure/llm/structured_output_framework/litellm_instructor/llm/get_llm_client.py
@@ -64,11 +64,12 @@ def get_llm_client(raise_api_key_error: bool = True):
     )  # imported here to avoid circular imports
 
     model_max_completion_tokens = get_model_max_completion_tokens(llm_config.llm_model)
-    max_completion_tokens = (
-        model_max_completion_tokens
-        if model_max_completion_tokens
-        else llm_config.llm_max_completion_tokens
-    )
+    user_max = llm_config.llm_max_completion_tokens
+    if model_max_completion_tokens is not None:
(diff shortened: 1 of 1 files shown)


dexters1 (changes requested):
There is a big problem here:

# Use the lower of the model's hard limit and the user's configured ceiling
max_completion_tokens = min(model_max_completion_tokens, user_max)

user_max has a default value of

llm_max_completion_tokens: int = 16384

Meaning there will be a lot of situation where it will be taken as the minimum when it shouldn't as users did not define any value and the model can handle a bigger input

PR #2618, opened 2026-04-13
fix(cognee-mcp): return all dataset results for GRAPH_COMPLETION search

### LINKED ISSUE #2617: cognee-mcp: GRAPH_COMPLETION search discards all but first dataset result
## Bug Description

The `search` tool in `cognee-mcp/src/server.py` only returns the first `SearchResult` from `cognee.search()` for `GRAPH_COMPLETION` and `RAG_COMPLETION` query types, silently discarding results from all other matched datasets.

When a user searches across multiple datasets (e.g., `datasets="project-a,project-b"`), `cognee.search()` returns a `List[SearchResult]` with one entry per dataset. The current code returns only `search_results[0]`, so the user only ever sees results from the first (alphabetically earliest) dataset.

## Affected Code

`cognee-mcp/src/server.py`, inside `search_task()`:

**API mode** (~line 571):
```python
if (search_type.upper() in ["GRAPH_COMPLETION", "RAG_COMPLETION"]
    and len(search_results) > 0):
    return str(search_results[0])  # <-- discards results[1:]
```

**Direct mode** (~line 582):
```python
elif (search_type.upper() == "GRAPH_COMPLETION"
      or search_type.upper() == "RAG_COMPLETION"):
    return str(search_results[0])  # <-- also crashes with IndexError if empty
```

## Additional Issue

The direct mode path has **no empty-list guard** — if `search_results` is `[]`, it crashes with `IndexError: list index out of range`. The API mode path checks `len > 0` but silently returns nothing for empty results.

## Expected Behavior

All dataset results should be combined and returned to the user. Each result should be labeled with its dataset name so users can distinguish which dataset each answer came from.

## Reproduction

1. Index 2+ datasets into cognee via the MCP `cognify` tool
2. Call `search` with `search_type="GRAPH_COMPLETION"` and `datasets="dataset-a,dataset-b"`
3. **Actual**: Only `dataset-a`'s result is returned
4. **Expected**: Results from both datasets are returned

## Proposed Fix

Iterate all `SearchResult` objects and combine them with dataset name labels. Handle both dict results (from `_backwards_compatible_search_results()`) and object results:

```python
elif search_type.upper() in ("GRAPH_COMPLETION", "RAG_COMPLETION"):
    if isinstance(search_results, list) and len(search_results) > 0:
        combined = []
        for sr in search_results:
            if isinstance(sr, dict):
                ds_name = sr.get('dataset_name', 'unknown')
                sr_content = sr.get('search_result', str(sr))
            elif hasattr(sr, 'dataset_name') and hasattr(sr, 'search_result'):
                ds_name = sr.dataset_name or "unknown"
                sr_content = sr.search_result
            else:
                combined.append(str(sr))
                continue
            if isinstance(sr_content, list):
                for item in sr_content:
                    combined.append(f"[{ds_name}] {item}")
            else:
                combined.append(f"[{ds_name}] {sr_content}")
        return "\n\n".join(combined)
    return str(search_results)
```

I'll submit a PR with this fix shortly.

raynasr:
## Summary

- Fixes GRAPH_COMPLETION and RAG_COMPLETION search returning only `search_results[0]`, discarding results from all other matched datasets
- Adds empty-list guard to direct mode path (previously crashed with `IndexError` on empty results)
- Extracts `_combine_completion_results()` helper shared by both API and direct mode

Fixes #2617

## Problem

When a user searches across multiple datasets (e.g., `datasets="project-a,project-b"`), `cognee.search()` correctly returns a `List[SearchResult]` with one entry per dataset. However, the MCP server's `search_task()` function returns only the first element:

```python
# API mode (line ~639):
return str(search_results[0])  # discards results[1:]

# Direct mode (line ~651):
return str(search_results[0])  # also crashes if empty
```

This means users only ever see results from a single dataset (typically the alphabetically first one), making multi-dataset search effectively broken.

## Fix

A `_combine_completion_results()` helper iterates all `SearchResult` entries, labels each with its dataset name (e.g., `[my-dataset] answer text`), and joins them with double newlines. It handles:
- **Dict results** (produced by `_backwards_compatible_search_results()`) — accesses via `.get()`
- **SearchResult Pydantic model instances** — accesses via attribute
- **Empty results** — returns `str(results)` instead of crashing

## Test plan

- [x] Verified with 12 indexed datasets, cross-dataset GRAPH_COMPLETION queries return labeled results from all requested datasets
- [x] Single-dataset scoped queries still work correctly
- [x] Empty result lists no longer crash
- [ ] Upstream CI passes

## Summary by CodeRabbit

* **Bug Fixes**
  * GRAPH_COMPLETION and RAG_COMPLETION searches now return aggregated results from all datasets instead of only the first dataset's results. Dataset names are included with each entry for improved clarity.


Vasilije1990:
@raynasr please rebase and open against dev. We don't merge to main without testing

raynasr:
Closing in favor of a new PR rebased onto `dev` per maintainer feedback.

PR #2625, opened 2026-04-14
feat(cognee-mcp): support batch file-path ingestion in cognify tool

### LINKED ISSUE #2624: cognee-mcp: cognify tool should support batch file-path ingestion
## Feature Request

The `cognify` MCP tool accepts a single `data: str` parameter, which gets passed to `cognee.add()`. While `cognee.add()` natively supports file paths, directory paths, URLs, and lists of mixed inputs, the MCP tool only processes one item per call.

This creates a performance problem when indexing multiple files into the same dataset: each `cognify()` call bundles `add()` + `cognify()`, so indexing N files means N full pipeline runs on the growing dataset — O(n²) work.

## Affected Code

`cognee-mcp/src/server.py`, inside `cognify_task()`:

```python
await cognee_client.add(data, dataset_name=dataset_name)
```

This always passes `data` as a single string. If the caller wants to add multiple files to a dataset before running the pipeline, there's no way to do it through the MCP interface.

## Proposed Solution

Detect if `data` is a JSON-encoded array of strings. If so, call `add()` for each item before running `cognify()` once. This is fully backwards-compatible — non-JSON strings behave exactly as before.

```python
# Support JSON array of items (file paths, text) for batch add
items = [data]
try:
    parsed = json.loads(data)
    if isinstance(parsed, list):
        items = parsed
        logger.info(f"Batch mode: adding {len(items)} items to '{dataset_name}'")
except (json.JSONDecodeError, TypeError):
    pass

for item in items:
    await cognee_client.add(item, dataset_name=dataset_name)
```

### Why this matters

- **File-path ingestion**: `cognee.add()` auto-detects file paths and uses format-aware parsers (PDF, DOCX, code files, etc.) — much better chunking than raw text concatenation
- **Batch efficiency**: Add 18 files to a dataset, run the pipeline once instead of 18 times
- **Backwards-compatible**: Plain strings pass through unchanged; only JSON arrays trigger batch mode

### Verified behavior

Tested with the Docker-based cognee-mcp server (direct mode):
- Single file path (e.g., `/app/codebase/project/server.py`) → cognee reads the file natively ✓
- JSON array of file paths → each file added separately, pipeline runs once ✓
- Plain text string → unchanged behavior ✓
- 12 datasets reindexed successfully using batch file-path mode ✓

I'll submit a PR with this fix.

raynasr:
## Summary

- The `cognify` MCP tool now detects JSON-encoded arrays in the `data` parameter
- Calls `cognee_client.add()` for each item before running a single `cognify()` pipeline pass
- Enables efficient multi-file ingestion (file paths, text, URLs) into a dataset with one pipeline run instead of N separate runs
- Plain string inputs are unchanged — fully backwards-compatible

Fixes #2624

## Problem

The `cognify` tool bundles `add()` + `cognify()` in a single call. When indexing N files into the same dataset, each call re-runs the full pipeline on the growing dataset — O(n²) work. There's no way to batch multiple `add()` calls before a single `cognify()` through the MCP interface.

Meanwhile, `cognee.add()` natively supports file paths and lists, but the MCP tool doesn't leverage this.

## Solution

Before the `add()` call in `cognify_task()`, try to parse `data` as a JSON array. If it is one, iterate and call `add()` per item. Otherwise, pass the string as-is (backwards-compatible).

Input validation:
- Empty arrays are rejected with `ValueError`
- Each item must be a non-empty string
- Only `json.JSONDecodeError` is caught (not `TypeError`)

## Test plan

- [x] Single file path (`/path/to/file.py`) — cognee reads the file natively via its file loader
- [x] JSON array of file paths — each file added separately, pipeline runs once
- [x] Plain text string — unchanged behavior (backwards-compatible)
- [x] Empty array `[]` — raises ValueError, does not run pipeline
- [x] 12 datasets (71 files total) reindexed successfully using batch file-path mode
- [x] Search quality verified on reindexed data (CHUNKS + GRAPH_COMPLETION)

## Summary by CodeRabbit

* **New Features**
  * Added batch processing: system now auto-detects list-formatted submissions, processes each item individually, logs item count, and preserves backward compatibility with single-item submissions.

* **Bug Fixes**
  * Added input validation to reject empty/whitespace items and fallback to single-item handling when input isn't valid JSON or not a list.

### DIFF
--- a/cognee-mcp/src/server.py
+++ b/cognee-mcp/src/server.py
@@ -347,7 +347,18 @@ async def cognify(
 
                     graph_model = load_class(graph_model_file, graph_model_name)
 
-            await cognee_client.add(data, dataset_name=dataset_name)
+            # Support JSON array of items (file paths, text, URLs) for batch add
+            items = [data]
+            try:
+                parsed = json.loads(data)
+                if isinstance(parsed, list):
+                    items = parsed
+                    logger.info(f"Batch mode: adding {len(items)} items to '{dataset_name}'")
+            except (json.JSONDecodeError, TypeError):
+                pass
+
+            for item in items:
+                await cognee_client.add(item, dataset_name=dataset_name)
 
             try:
                 await cognee_client.cognify(
--- a/cognee-mcp/src/server.py
+++ b/cognee-mcp/src/server.py
@@ -352,12 +352,17 @@ async def cognify(
             try:
                 parsed = json.loads(data)
                 if isinstance(parsed, list):
+                    if not parsed:
+                        raise ValueError("Batch input cannot be an empty JSON array.")
                     items = parsed
                     logger.info(f"Batch mode: adding {len(items)} items to '{dataset_name}'")
-            except (json.JSONDecodeError, TypeError):
+            except json.JSONDecodeError:
+                # Plain text / non-JSON input: keep backward-compatible single-item flow
                 pass
 
             for item in items:
+                if not isinstance(item, str) or not item.strip():
+                    raise ValueError(f"Invalid batch item (must be a non-empty string): {item!r}")
                 await cognee_client.add(item, dataset_name=dataset_name)
 
             try:


siillee (changes requested):
Hey @raynasr, thanks a lot for your contirbution!

Imo, this feels more like a workaround. The idea is great, but it would be better if `cognify` (and `remember`, as part of our new API) took as parameter a list of strings or otherwise items, rather than just one string type. This would make it more clear during mcp use as well, both for the llm and for the user.

Also, we can then add all the data in one `add` function call, no need to loop over them.

Vasilije1990:
Thanks for adding batch file-path ingestion. I am closing this as superseded by #2737: https://github.com/topoteretes/cognee/pull/2737

#2737 includes the batch cognify behavior, targets `dev`, validates every JSON-array item before launching the background task, preserves inline text inputs, and adds focused MCP tests for the batch tool flow. Any remaining MCP cognify follow-up should happen on #2737.

PR #2637, opened 2026-04-15
fix: vector db migrations run for all datasets. fixes to lancedb migration

siillee:
## Description

This PR fixes how we call vector database migrations. Now we call them only on startup, not during operations like remember, cognify, etc. We now also make sure the migrations are done on correct databases, i.e. all dataset databases, not just the default one (e.g. cognee.lancedb).

This PR also fixes the LanceDB migration. We compared collection payloads to cognee data models (DocumentChunk, Entity, etc.), but we store data in the form of IndexSchema models, which are defined in every adapter. Now this is fixed, and the correct schemas are compared.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

## Summary by CodeRabbit

* **Refactor**
  * Simplified vector migration flow to run per-dataset from relational state; removed the one-time local migration guard and runtime subclass discovery. Startup migrations now retry after initializing the relational database on failure.

* **New Features**
  * Added async helper to list dataset databases for migration orchestration.
  * Migration summaries now include dataset and provider details.

* **Bug Fixes**
  * Migrations abort on row validation failures and preserve existing data; improved payload default inference.

* **Tests**
  * New unit tests for per-dataset iteration, failure handling, and payload-default behavior.


dexters1:
@ArmagedonFlamer can you verify:
1. Migrations work when updating from older Cognee version to new one
2. You no longer have issues you had?
3. Search adding data to LanceDB works

Vasilije1990:
This addresses the fact that we are effectively using Cognee often as a CLI or using it in ways where the server won't be available. How does it address it?

siillee:
> This addresses the fact that we are effectively using Cognee often as a CLI or using it in ways where the server won't be available. How does it address it?

I don't really understand the question, but let me try and answer what I think was the point.

The vector db migrations are called in the same way as the relational db migrations. When running Cognee only from the SDK directly, the migrations do not happen automatically. They are only called at startup of the backend server.

We can change this if necessary, but then we would need a guard that will make the migrations optional, because in case of the cloud, we have multiple pods running Cognee, and if all of them ran the migrations at the same time, it would lead to trouble.

PR #2652, opened 2026-04-16
fix: keep belongs_to_set consistent across dataset cognify and deletion

lkzppm:
## Summary

Shared DataPoints (deterministic UUID5 ids) are meant to stay consistent across datasets when users re-cognify the same content or delete one of the datasets. Two symmetric bugs broke that consistency:

**Upsert side — tags were dropped or overwritten:**

- **PGVector** (`PGVectorAdapter.create_data_points`) used `ON CONFLICT (id) DO NOTHING`. The second dataset's rows were silently discarded and its `belongs_to_set` tag was never stored.
- **LanceDB** (`create_data_points`) used `merge_insert("id").when_matched_update_all()`, which replaces the whole row, overwriting `belongs_to_set`.
- **Neo4j** (`add_node` / `add_nodes`) used `SET n += properties`, which overwrites `belongs_to_set` with the incoming list (last-writer-wins). The `belongs_to_set` edges to NodeSet nodes are already additive, so this was "cosmetic" but left the node property inconsistent with the edges.

**Delete side — orphaned tags were never cleaned up:**

When a dataset is deleted, `delete_from_graph_and_vector` hard-deletes uniquely-owned nodes (including the dataset's NodeSet node), but leaves `belongs_to_set` on shared rows pointing at a NodeSet that no longer exists. Search-time NodeSet filtering then still matches a dataset that's been removed.

## What this PR does

1. **PGVector merge on conflict** — `ON CONFLICT (id) DO UPDATE` with a `jsonb_set` expression that unions `{existing, incoming}` belongs_to_set arrays via `jsonb_agg(DISTINCT ...)`. Dedupes `pgvector_data_points` by id in Python before the INSERT to avoid the "cannot affect row a second time" Postgres error that `DO NOTHING` used to silently swallow.
2. **LanceDB merge on conflict** — before constructing the `LanceDataPoint` batch, query the table for existing rows with matching ids, capture their `belongs_to_set`, and union those prior tags into each incoming payload so `merge_insert().when_matched_update_all()` writes the merged list.
3. **Neo4j merge on merge** — in both `add_node` and `add_nodes`, capture the existing `belongs_to_set` via `apoc.coll.toSet(...)` before the `SET n += properties` overwrite, then re-apply the union. Nodes in the same batch with the same id are deduped in Python first (same reason as PGVector — UNWIND over a duplicated id recomputes the union twice and drops a tag from the first pass).
4. **New primitive `remove_belongs_to_set_tags(tags)`** on all three adapters (default no-op on `VectorDBInterface` / `GraphDBInterface` so Kuzu, Chroma, Qdrant, Weaviate, Milvus, Neptune, Falkor are unchanged):
   - **PGVector** enumerates its vector collections (tables whose name is PascalCase, which disambiguates them from lowercase relational tables that share the schema), issues a `jsonb_agg` UPDATE that strips the tags from each row's `belongs_to_set`, then `DELETE`s rows whose `belongs_to_set` is now empty.
   - **LanceDB** iterates `list_tables`, skips tables whose payload struct doesn't have a `belongs_to_set` field, then per matched row either rewrites (delete + re-add; `merge_insert` silently no-ops for dicts with nested payload structs) or deletes outright if the array empties.
   - **Neo4j** runs one Cypher `SET n.belongs_to_set = [x IN n.belongs_to_set WHERE NOT x IN $tags]` over every matching `__Node__`.
5. **Automatic detag on dataset deletion** — `delete_from_graph_and_vector` now inspects `unique_nodes` (the rows about to be hard-deleted) for NodeSet-typed entries, captures their labels, and after the hard-delete pass calls `graph_engine.remove_belongs_to_set_tags(...)` + `vector_engine.remove_belongs_to_set_tags(...)`. This is gated on the NodeSet actually being exclusive to the deleted dataset (i.e., in `unique_nodes`): if another dataset still references it, we leave the tag alone so that other dataset's filtering keeps working.

## Scope note

Scope intentionally covers the three adapters affected by the user-reported bug. Kuzu / Chroma / Qdrant / Weaviate / Milvus / Neptune / Falkor inherit the no-op default — they're not regressed, but they also don't get detag semantics in this PR. Happy to extend if maintainers want.

## Test plan

- [x] `uv run pytest cognee/tests/unit/infrastructure/databases/vector/test_belongs_to_set_merge.py` — 5 cases: merge across upserts, dedup, first-insert, detag strips+deletes, detag is no-op for empty input.
- [x] `uv run pytest cognee/tests/unit/modules/graph/test_delete_detag_nodeset.py` — orchestration test using mocked graph/vector engines verifies that `delete_from_graph_and_vector` calls the detag primitives with the right tag list when a NodeSet is in the hard-delete set, and skips them otherwise.
- [x] `uv run pytest cognee/tests/unit/infrastructure/databases/` — 117 pass / 2 skipped (pgvector + chromadb adapters skip without their extras installed locally).
- [x] `uv run ruff check` / `ruff format` / pre-commit hooks green.
- [ ] Integration: `cognee/tests/test_pgvector.py` and `cognee/tests/test_neo4j.py` with `docker compose --profile postgres --profile neo4j up -d`. The existing pgvector test exercises multi-dataset cognify with shared files, which is exactly the setup the fixes target — pending a run with live containers.

## Summary by CodeRabbit

* **Improvements**
  * Upsert behavior now merges and deduplicates dataset membership tags ("belongs_to_set") instead of overwriting them across vector and graph backends.
* **New Features**
  * Added a global detagging operation to remove specified "belongs_to_set" tags across graph and vector stores; delete workflow now invokes detagging when relevant.
* **Tests**
  * Added unit tests covering tag accumulation, deduplication, and detagging orchestration.

### DIFF
--- a/cognee/infrastructure/databases/graph/graph_db_interface.py
+++ b/cognee/infrastructure/databases/graph/graph_db_interface.py
@@ -109,6 +109,16 @@ class GraphDBInterface(ABC):
         """
         raise NotImplementedError
 
+    async def remove_belongs_to_set_tags(self, tags: List[str]) -> None:
+        """
+        Remove the given tag names from every node's `belongs_to_set` property
+        array. Keeps the property consistent with the additive
+        `belongs_to_set` edges after a NodeSet or its containing dataset is
+        deleted. Default no-op; adapters that store a list property (Neo4j,
+        Neptune) override this.
+        """
+        return None
+
     @abstractmethod
     async def get_node(self, node_id: str) -> Optional[NodeData]:
         """
--- a/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
+++ b/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
@@ -200,10 +200,18 @@ class Neo4jAdapter(GraphDBInterface):
         """
         serialized_properties = self.serialize_properties(node.model_dump())
 
+        # Capture the existing belongs_to_set before `+=` overwrites it, then
+        # write back the union so a DataPoint cognified into multiple datasets
+        # retains every dataset tag on its node property (edges are already
+        # additive; this keeps the property consistent with them).
         query = dedent(
             f"""MERGE (node: `{BASE_LABEL}`{{id: $node_id}})
-                ON CREATE SET node += $properties, node.updated_at = timestamp()
-                ON MATCH SET node += $properties, node.updated_at = timestamp()
+                WITH node, apoc.coll.toSet(
+                    coalesce(node.belongs_to_set, [])
+                    + coalesce($properties.belongs_to_set, [])
+                ) AS merged_belongs_to_set
+                SET node += $properties, node.updated_at = timestamp()
+                SET node.belongs_to_set = merged_belongs_to_set
                 WITH node, $node_label AS label
                 CALL apoc.create.addLabels(node, [label]) YIELD node AS labeledNode
                 RETURN ID(labeledNode) AS internal_id, labeledNode.id AS nodeId"""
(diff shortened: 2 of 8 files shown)


lkzppm:
Superseded by #2653 — branch was renamed from `fix/belongs-to-set-append` to `fix/belongs-to-set-consistency` to reflect the broader scope (merge on upsert **and** detag on deletion). Renaming via the GitHub API auto-closed this PR; the same commits plus new integration tests are in #2653.

PR #2653, opened 2026-04-16
fix: keep belongs_to_set consistent across dataset cognify and deletion

### LINKED ISSUE #2732: Shared data: deleting from one dataset wipes the same data from other datasets that share it
## Summary

When the same data item is added to multiple datasets, deleting it from one dataset also removes the corresponding nodes / edges / vectors from the **other** datasets that still reference it. The other datasets also lose their NodeSet / `belongs_to_set` membership for the shared content, breaking dataset-scoped search.

## Reproduction

1. Create two datasets `alfa` and `beta`.
2. Add file `x.md` to **both** datasets and a second file `y.md` to **only** `alfa`.
3. Run `cognify` on both datasets.
4. Delete `x.md` from `alfa`.
5. Inspect the graph and vector stores for `beta`.

**Expected:** only the `(alfa, x)` ownership and the `belongs_to_set: ["alfa"]` membership for shared nodes are removed. `beta` still has the full graph for `x`.

**Actual:** the shared nodes are hard-deleted globally; `beta`'s NodeSet node `beta` itself still exists but the shared rows lose all `belongs_to_set` tags (or get deleted outright), and `belongs_to_set` graph edges to the surviving NodeSets become stale.

## Root causes (multiple, symmetric)

**Hard-delete query was scoped only by `data_id`, not the `(dataset_id, data_id)` pair.** So a slug owned by both `(alfa, x)` and `(beta, x)` was treated as orphaned the moment `(alfa, x)` was removed from the ledger.

**Per-dataset deletion never reconciled `belongs_to_set` on shared rows.** Even after pair-scoping, a node co-owned by `alfa` and `beta` was left tagged `["alfa", "beta"]` after the dataset-`alfa` NodeSet was deleted, so search filtering kept matching the removed dataset.

**Adapter-level scoping issues compounded this:**
- pgvector: a single data-modifying CTE that did `UPDATE … RETURNING` then `DELETE` from the same CTE silently no-op'd because the outer `DELETE` shared the snapshot of the `UPDATE` branch and couldn't observe the row-version changes.
- Neo4j: stripping the `belongs_to_set` property left orphaned `belongs_to_set` graph edges to the surviving NodeSet nodes — graph and property went out of sync.
- LanceDB: a TOCTOU race in the prefetch path could lose merged tags when two datasets cognified the same content concurrently.

## Fix

PR #2653 addresses the bug by:

1. Scoping the hard-delete query to `(dataset_id, data_id)` so co-owned slugs are not dropped.
2. After the hard-delete, identifying which slugs lost their last anchor in the deleted dataset and calling a new `remove_belongs_to_set_tags(tags, node_ids=...)` adapter primitive scoped to those exact ids, so other datasets' tags are untouched.
3. Splitting the pgvector detag into a three-statement `prepare → update → delete` flow that escapes the snapshot-isolation trap.
4. Extending the Neo4j detag to delete stale `belongs_to_set` edges in the same call so the graph stays in sync with the property.
5. Closing the LanceDB prefetch race so concurrent cognifies on shared data don't lose tags.

Backed by new unit tests plus live-container integration tests for pgvector and Neo4j.

## Affected versions

Reproducible on `dev` (and `1.0.3`). Bug exists for all setups that use shared data across datasets without per-dataset isolated databases (i.e., pgvector + Neo4j, where `is_multi_user_support_possible()` is False and a single global graph/vector store is shared across users and datasets).

## Linked PR

Fixed by #2653.

lkzppm:
Fixes #2732

## Summary

Shared DataPoints (deterministic UUID5 ids) are meant to stay consistent across datasets when users re-cognify the same content or delete one of the datasets. Two symmetric bugs broke that consistency:

**Upsert side — tags were dropped or overwritten:**

- **PGVector** (`PGVectorAdapter.create_data_points`) used `ON CONFLICT (id) DO NOTHING`. The second dataset's rows were silently discarded and its `belongs_to_set` tag was never stored.
- **LanceDB** (`create_data_points`) used `merge_insert("id").when_matched_update_all()`, which replaces the whole row, overwriting `belongs_to_set`.
- **Neo4j** (`add_node` / `add_nodes`) used `SET n += properties`, which overwrites `belongs_to_set` with the incoming list (last-writer-wins). The `belongs_to_set` edges to NodeSet nodes are already additive, so this was "cosmetic" but left the node property inconsistent with the edges.

**Delete side — orphaned tags were never cleaned up:**

When a dataset is deleted, `delete_from_graph_and_vector` hard-deletes uniquely-owned nodes (including the dataset's NodeSet node), but leaves `belongs_to_set` on shared rows pointing at a NodeSet that no longer exists. Search-time NodeSet filtering then still matches a dataset that's been removed.

## What this PR does

1. **PGVector merge on conflict** — `ON CONFLICT (id) DO UPDATE` with a `jsonb_set` expression that unions `{existing, incoming}` belongs_to_set arrays via `jsonb_agg(DISTINCT ...)`. Dedupes `pgvector_data_points` by id in Python before the INSERT to avoid the "cannot affect row a second time" Postgres error that `DO NOTHING` used to silently swallow.
2. **LanceDB merge on conflict** — before constructing the `LanceDataPoint` batch, query the table for existing rows with matching ids, capture their `belongs_to_set`, and union those prior tags into each incoming payload so `merge_insert().when_matched_update_all()` writes the merged list.
3. **Neo4j merge on merge** — in both `add_node` and `add_nodes`, capture the existing `belongs_to_set` via `apoc.coll.toSet(...)` before the `SET n += properties` overwrite, then re-apply the union. Nodes in the same batch with the same id are deduped in Python first (same reason as PGVector — UNWIND over a duplicated id recomputes the union twice and drops a tag from the first pass).
4. **New primitive `remove_belongs_to_set_tags(tags)`** on all three adapters (default no-op on `VectorDBInterface` / `GraphDBInterface` so Kuzu, Chroma, Qdrant, Weaviate, Milvus, Neptune, Falkor are unchanged):
   - **PGVector** enumerates its vector collections (tables whose name is PascalCase — parsed from Postgres's `schema.table` qualified names — to disambiguate them from lowercase relational tables that share the schema), issues a `jsonb_agg` UPDATE that strips the tags from each row's `belongs_to_set`, then `DELETE`s rows whose `belongs_to_set` is now empty. Each table runs in its own transaction so a failure on one non-vector table can't roll back updates already committed on others.
   - **LanceDB** iterates `list_tables`, skips tables whose payload struct doesn't have a `belongs_to_set` field, then per matched row either rewrites (delete + re-add; `merge_insert` silently no-ops for dicts with nested payload structs) or deletes outright if the array empties.
   - **Neo4j** runs one Cypher `SET n.belongs_to_set = [x IN n.belongs_to_set WHERE NOT x IN $tags]` over every matching `__Node__`.
5. **Automatic detag on dataset deletion** — `delete_from_graph_and_vector` now inspects `unique_nodes` (the rows about to be hard-deleted) for NodeSet-typed entries, captures their labels, and after the hard-delete pass calls `graph_engine.remove_belongs_to_set_tags(...)` + `vector_engine.remove_belongs_to_set_tags(...)`. This is gated on the NodeSet actually being exclusive to the deleted dataset (i.e., in `unique_nodes`): if another dataset still references it, we leave the tag alone so that other dataset's filtering keeps working.

## Scope note

Scope intentionally covers the three adapters affected by the user-reported bug. Kuzu / Chroma / Qdrant / Weaviate / Milvus / Neptune / Falkor inherit the no-op default — they're not regressed, but they also don't get detag semantics in this PR. Happy to extend if maintainers want.

## Test plan

- [x] **Unit tests** (`uv run pytest cognee/tests/unit/infrastructure/databases/vector/ cognee/tests/unit/modules/graph/`) — 90 passed / 1 skipped. Five new LanceDB adapter tests: merge across upserts, dedup on repeat, first-insert path, detag strips+deletes, detag is no-op on empty input. Two orchestration tests verify `delete_from_graph_and_vector` calls `remove_belongs_to_set_tags` with the right tag list when NodeSet nodes are in the hard-delete set, and skips the call otherwise.
- [x] **Integration tests against live containers** (`cognee/tests/integration/test_belongs_to_set_pgvector.py` + `test_belongs_to_set_neo4j.py`) — **8/8 pass** against `docker compose --profile postgres --profile neo4j up -d`. Tests use a deterministic stub `embed_text` returning a fixed 3-D vector so they consume **no LLM / embedding API calls**. Gated on `TEST_PGVECTOR_URL` / `TEST_NEO4J_URL` env vars so they skip cleanly in environments without containers. Coverage:
    - pgvector: merge across upserts, in-batch dedup, detag strips + deletes orphans, detag ignores non-vector tables.
    - neo4j: merge across calls, in-batch duplicate-id dedup, detag strips property, detag no-op on empty input.
- [x] `uv run ruff check` / `ruff format` / pre-commit hooks green.

## Summary by CodeRabbit

* **New Features**
  * Automated detagging: when dataset/node sets are removed, surviving graph and vector items are cleaned of those tags; storage adapters can opt in to support this.

* **Bug Fixes**
  * Upserts and batch inserts now reliably merge and preserve dataset tags across repeated updates and duplicate entries.
  * Detag operations remove specified tags and delete items that become tag-empty; empty-input detag is a no-op.

* **Tests**
  * New unit and integration tests validating tag-merge and detag behaviors across graph and vector backends.


lkzppm:
Tagging suggested reviewers from CodeRabbit: @dexters1 @lxobr @hajdul88 — would appreciate your eyes when you have a moment. (I don't have permission to request reviews directly from a fork.)

lkzppm:
Addressed the bot reviews in `704c3b3c`:

**GitGuardian — hardcoded secret in `test_belongs_to_set_neo4j.py`.** Fixed. That was the docker-compose default password echoed in the docstring's example. Docstring now just points at the env vars without a literal value.

**CodeRabbit — PGVector in-batch dedup drops tags.** Real bug, fixed. The dedup now unions `belongs_to_set` across duplicates before collapsing, mirroring `Neo4jAdapter.add_nodes`. Added live-pgvector integration coverage (`test_create_data_points_merges_tags_across_in_batch_duplicates`) — passes.

**CodeRabbit — LanceDB `remove_belongs_to_set_tags` loads whole collection.** Fixed. Now pushes `array_has_any(payload.belongs_to_set, [...])` into LanceDB (same predicate pattern used in `search()`), only materializing candidate rows. Tags are escaped the same way `search()` escapes them.

**CodeRabbit — LanceDB delete+add risks data loss with debug-level failure log.** Escalated: re-add failure now logs at WARNING with the affected ids and re-raises so the caller sees it instead of silently leaving the collection short of rows. I considered adding a "restore from `rows_to_update`" compensator, but the in-memory rows are already the *post-detag* versions — there's no pre-detag snapshot to restore, and a retry of the same `add()` call that just failed won't succeed for a different reason. Escalate + raise is the honest behavior.

**CodeRabbit — PGVector detag shares one session across tables.** This comment is on stale code. In commit `8ae671f1` I moved `async with self.get_async_session()` *inside* the `for table_name in candidate_tables` loop (see line ~597 of `PGVectorAdapter.py`); each table runs in its own transaction, so a failure on one doesn't roll back the others.

**Test runs after the changes:**
- Unit (`cognee/tests/unit/infrastructure/databases/vector/ + modules/graph/`): 7/7 pass.
- Integration against live pgvector+neo4j containers: 9/9 pass (added the new in-batch tag-merge case).

lkzppm:
Addressed the fresh round of reviews in the force-push (previous history replaced by commits `0a3efb21` → `411e37d1`):

**GitGuardian — hardcoded password.** History rewritten via interactive rebase. `grep -R pleaseletmein` across the PR's commit range (`upstream/dev..HEAD`) now returns zero matches; the offending `8ae671f1` is no longer reachable from the branch tip. The orphaned SHA may still resolve via direct URL until GitHub's background GC sweeps it — that's a GitHub-side artifact of fork object retention, not a current-history secret.

**CodeRabbit — LanceDB `create_data_points` in-batch dedup drops tags.** Real bug (same class as the PG/Neo4j ones already fixed). Now merges `belongs_to_set` across duplicates before collapsing. Added a unit test (`test_belongs_to_set_merges_tags_across_in_batch_duplicates`) that fails on the old dict-collapse.

**CodeRabbit — LanceDB `remove_belongs_to_set_tags` re-adds raw dicts.** Fixed. New helper `_coerce_rows_to_typed_payload` resolves the collection's payload schema via `_resolve_collection_payload_schema` and validates every row through `schema_model.model_validate(...).model_dump()` before `collection.add()`. When the schema can't be resolved, falls back to the original dicts (logs at debug).

**CodeRabbit — LanceDB existing-row lookup uses `tuple` repr.** Fixed. `create_data_points`' existing-row WHERE now escapes ids (`replace("'", "''")`) and builds the IN list explicitly (`id IN ('a', 'b')`), mirroring how `search()` handles node-name literals.

**CodeRabbit — PGVector comment claiming `schema.table` is inaccurate.** Reworded to describe what `get_table_names()` actually returns: raw SQLAlchemy reflection keys that may be schema-qualified on Postgres and plain on SQLite; the `_table_only` helper strips any leading `schema.` prefix before the PascalCase filter.

**CodeRabbit — pgvector integration `_fresh_adapter` is `async` but doesn't await.** Fixed. Now plain `def`; all call sites dropped the `await`.

**CodeRabbit — `test_remove_belongs_to_set_tags_ignores_non_vector_tables` doesn't actually assert anything.** Rewritten. Now creates both a snake_case relational-style table and a PascalCase table without a `payload` column, runs detag, and asserts both tables' rows survive intact.

**Docstring coverage.** Pushed to 81.4% (158/194 across the changed files) by adding one-liners to the helpers I introduced and a handful of pre-existing undocumented methods/classes in the touched files. Should clear the 80% threshold on the next re-scan.

**Tests.** Locally: 8/8 unit + 9/9 live-container integration (pgvector + neo4j) green.

lkzppm:
Addressed in `c709a7a4`:

**CodeRabbit — LanceDB `remove_belongs_to_set_tags` N round-trips on deletes.** Fixed. Both buckets (orphaned rows + rows-to-update) now collapse into a single `collection.delete("id IN ('a','b',...)")` call per collection. Ids are the same UUID strings cognee already uses as `retrieve()` keys, so no escaping is required.

**CodeRabbit — pgvector integration tests leaking empty tables across CI runs.** Fixed. New helper `_drop_collection(adapter, name)` issues `DROP TABLE IF EXISTS \"{name}\"` using the adapter's own async session, and every test's `finally` calls it after deleting data points. Verified: after a full local run, `SELECT tablename FROM pg_tables WHERE tablename LIKE 'IntegTaggedPoint_%'` returns zero rows.

**Docstring coverage.** Now **99.5%** (193/194 on the diff's touched files) — well past the 80% threshold. Added one-line docstrings to the pre-existing adapter methods (`create_collection`, `create_data_points`, `retrieve`, `search`, `batch_search`, `delete_data_points`, `prune`, `create_vector_index`, `index_data_points`, etc.) that were missing them in the three changed files, plus the two LanceDB internal probe classes (`MigrationLanceDataPoint`, `SchemaProbeDataPoint`) and the Neo4j `feedback_weights` getters/setters.

**Tests.** Locally: 8/8 unit + 9/9 live-container integration (pgvector + neo4j).

lkzppm:
Addressed the latest review in \`9106182b\`:

**CodeRabbit — LanceDB prefetch race against concurrent \`create_data_points\`.** Fixed. The prefetch of \`existing_belongs_to_set\` now runs inside the same \`VECTOR_DB_LOCK\` section as \`merge_insert\`, so another upsert can't slide in between our read and our write and clobber tags we never saw. The embedding call stays outside the lock (it's network-bound to the embedding service and doesn't touch the collection). The migration fallback path remains outside the lock since \`_migrate_collection_schema\` re-acquires \`VECTOR_DB_LOCK\` internally — letting the lock release before the migration call avoids a deadlock.

**CodeRabbit — PGVector detag over-deletes empty \`belongs_to_set\` rows.** Fixed. Replaced the separate UPDATE + DELETE with a single CTE:

\`\`\`sql
WITH updated AS (
    UPDATE \"{table}\" SET payload = jsonb_set(…) WHERE … RETURNING id, payload
)
DELETE FROM \"{table}\"
WHERE id IN (SELECT id FROM updated WHERE payload->'belongs_to_set' = '[]'::jsonb)
\`\`\`

The \`RETURNING\` clause reflects the post-UPDATE payload, so the outer DELETE only targets rows whose \`belongs_to_set\` became empty *as a result of this detag*. Pre-existing rows that already had \`belongs_to_set = []\` (unrelated to the current tag removal) are no longer touched.

**CodeRabbit — \`graph_db_interface.py\` docstring mentions Neptune but only Neo4j overrides.** Fixed. Docstring now notes Neo4j as the sole current override and leaves the door open for other adapters without naming one that doesn't exist.

**CodeRabbit — PGVector \`_table_only\` called 3× per iteration.** Fixed. Converted the list comprehension into an explicit loop that computes \`_table_only(name)\` once per row and reuses the cached value for the truthiness + PascalCase checks.

**CodeRabbit — \`test_create_data_points_dedupes_duplicate_ids_in_batch\` name is misleading.** Fixed. Renamed to \`test_create_data_points_tolerates_exact_duplicates_in_batch\` with a docstring that clarifies this case only checks the \"cannot affect row a second time\" tolerance, and points readers at \`test_create_data_points_merges_tags_across_in_batch_duplicates\` for the distinct-tag union case.

**Tests.** Unit locally: 8/8 pass. Integration (pgvector + neo4j) can't be run this cycle because my local Docker daemon is down — the logic changes are conservative (the LanceDB change only reorders work under the existing lock; the CTE strictly narrows the set of rows that can be deleted), and the previous integration suite already green on the unchanged surface. I'll rerun them end-to-end once the containers are back up.

lkzppm:
@coderabbitai review

lkzppm:
@siillee Could you review the PR?

Vasilije1990:
Closing, caused issues, had to revert merge

PR #2663, opened 2026-04-17
fix(graph): sanitize null bytes before relational upserts

### LINKED ISSUE #2612: [Bug]: add_data_points crashes with asyncpg.CharacterNotInRepertoireError on documents containing null bytes (0x00)
## Bug Description

When ingesting text or scraped-HTML documents that contain null bytes (`\x00`), the `add_data_points` task in the cognify pipeline crashes with `asyncpg.exceptions.CharacterNotInRepertoireError: invalid byte sequence for encoding \"UTF8\": 0x00` while bulk-inserting into the `edges` table on the relational store (PostgreSQL). Cognee gets pinned at 100% CPU retrying the insert in an infinite loop until the upstream HTTP client gives up.

## Steps to Reproduce

1. Run cognee with `DB_PROVIDER=postgres` (confirmed on `cognee/cognee:latest`, cognee v0.5.8).
2. `await cognee.add(\"Some text with a \x00 null byte inside\", dataset_name=\"ltm\")`
3. `await cognee.cognify()`

Alternatively, upload a scraped HTML file (e.g. a saved government style guide page) — many HTML parsers leave stray 0x00s as padding/separators.

## Expected Behavior

Cognee's text loader / data point writer should sanitize null bytes (and ideally invalid UTF-8 sequences) before handing content to the relational store. PostgreSQL TEXT columns strictly disallow `0x00` regardless of encoding, so this will always fail on untrusted input.

## Actual Behavior

```
sqlalchemy.exc.DBAPIError: (sqlalchemy.dialects.postgresql.asyncpg.Error)
  <class 'asyncpg.exceptions.CharacterNotInRepertoireError'>:
  invalid byte sequence for encoding \"UTF8\": 0x00
[SQL: INSERT INTO edges (id, slug, user_id, data_id, dataset_id, source_node_id,
     destination_node_id, relationship_name, label, attributes, created_at)
     VALUES (...), (...), ... (multi-row insert with up to ~121 rows)]
```

The exception is raised repeatedly (retry loop) without Cognee giving up, so the whole cognify call hangs and consumes CPU until the caller times out.

## Environment

- Cognee: v0.5.8 (`cognee/cognee:latest`)
- Python: 3.12.13
- DB_PROVIDER: postgres (remote, multi-tenant)
- GRAPH_DATABASE_PROVIDER: kuzu
- VECTOR_DB_PROVIDER: lancedb
- Docker deployment

## Related

- Same class of bug filed in LightRAG: HKUDS/LightRAG#884 (also RAG + asyncpg + null bytes).
- PostgreSQL's stance: https://www.postgresql.org/message-id/2186784.1704815308@sss.pgh.pa.us — `0x00` is rejected unconditionally in TEXT/VARCHAR, regardless of client encoding.

## Suggested Fix

In the text extraction/chunking path (likely `cognee/tasks/chunks/` or wherever `add_data_points` materializes edges/nodes to the relational store), strip `\x00` and replace invalid UTF-8 sequences before persisting. Minimal fix:

\`\`\`python
def _sanitize_text(text: str) -> str:
    return text.encode(\"utf-8\", errors=\"replace\").replace(b\"\x00\", b\"\").decode(\"utf-8\")
\`\`\`

Happy to submit a PR if you'd like. Reproducing input available on request.

## Workaround (for others hitting this)

Reject or sanitize files at the application layer before calling `cognee.add()`. We now reject textual files containing `0x00` with a clear error message in our upload frontend so users can re-export cleanly from the source.

txhno:
Closes #2612

## Summary
- strip `\x00` recursively from node and edge payloads before relational upserts
- sanitize `relationship_name`, labels, and JSON attributes headed into Postgres
- add focused unit coverage for generic payloads and `contains` edge text

## Testing
- `uv sync --extra dev`
- `uv run python -m pytest cognee/tests/unit/modules/graph/test_relational_upserts.py`
- `uv run python -m pytest cognee/tests/unit/tasks/storage/test_add_data_points.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Stripped embedded null characters and normalized binary/invalid byte sequences from graph node and edge data prior to storage.

* **New Features**
  * Automatic recursive sanitization applied to node and edge payloads, labels, identifiers, slugs and attributes.

* **Tests**
  * Added unit and async tests validating recursive sanitization across nested structures and verifying upsert payloads and commit behavior.


txhno:
Addressed the review feedback in 1e3ef6a2.

- added a docstring plus `bytes`/`bytearray` decoding in `sanitize_relational_payload()` so invalid UTF-8 is normalized safely before Postgres writes
- removed the redundant non-`contains` label sanitization pass in `upsert_edges()` while keeping the `contains` branch sanitized
- extended the node upsert test coverage to assert sanitized `type` and `indexed_fields`, and added bytes sanitization coverage

Re-tested with:
- `uv run python -m ruff check cognee/modules/graph/methods/sanitize_relational_payload.py cognee/modules/graph/methods/upsert_edges.py cognee/tests/unit/modules/graph/test_relational_upserts.py`
- `uv run python -m pytest cognee/tests/unit/modules/graph/test_relational_upserts.py cognee/tests/unit/tasks/storage/test_add_data_points.py`

txhno:
Pushed a small follow-up for the Ruff nit in `b78d9203`.

- extracted the long expected `relationship_name` string into a local variable in `test_relational_upserts.py`
- reused that value in the payload assertions to keep the line length under Ruff's limit

Re-checked with:
- `uv run python -m ruff check cognee/tests/unit/modules/graph/test_relational_upserts.py`

dexters1:
Thank you for your contribution! I am running our full CI on your PR now here: 

https://github.com/topoteretes/cognee/pull/2665/checks

If there are no issues I will merge your PR through that branch (as I had to format the PR) and your contributions will be part of the commit history 🙏

PR #2674, opened 2026-04-19
fix: dynamic parameter forwarding and collision-resistant naming in Neo4jAuraDevDatasetDatabaseHandler

### LINKED ISSUE #2354: Refactor `Neo4jAuraDevDatasetDatabaseHandler` to allow dynamic parameter forwarding and better instance naming
*(Created as part of the codebase TODO/FIXME audit #2316)*

In `cognee/infrastructure/databases/graph/neo4j_driver/Neo4jAuraDevDatasetDatabaseHandler.py`, the `create_dataset()` method currently uses a hardcoded `payload` dictionary for creating Neo4j Aura instances via the REST API. 

There are two adjacent `TODO` comments here that need addressing:
1. Allow `**kwargs` parameter forwarding so different datasets can use different configurations (e.g. region, memory, aura tier) when spinning up the Neo4j Aura environment.
2. Find a better naming strategy for the `graph_db_name` instead of blindly slicing `dataset_id[0:29]` to fit the 30-character limit, which could potentially cause naming collisions if the first 29 characters of multiple UUIDs are identical.

**Acceptance Criteria:**
- Update the `create_dataset` method signature to safely accept and merge `**kwargs` into the API request payload.
- Implement a more robust hashing or base62 encoding strategy for the dataset name to guarantee uniqueness within the 30-character Neo4j API limit.
- Ensure existing integration tests for creating Neo4j handler instances still pass.

octo-patch:
## Summary

Addresses the two TODOs left in `Neo4jAuraDevDatasetDatabaseHandler.create_dataset()` noted in issue #2354:

- **Dynamic parameter forwarding**: Add `**kwargs` to `create_dataset()` so callers can override any Neo4j Aura instance configuration field (`version`, `region`, `memory`, `type`, `cloud_provider`) without subclassing. Previously all fields were hardcoded with no override path.

- **Collision-resistant instance naming**: Replace the raw UUID slice `f"{dataset_id}"[:29]` with a SHA-256 hash truncated to 29 hex characters. The original approach risked name collisions for deterministic UUIDs sharing long common prefixes, and also allowed hyphens which Neo4j Aura may reject. The SHA-256 approach provides 116 bits of effective uniqueness from a 29-character hex string.

- **Updated docstring**: Documents all supported `**kwargs` keys and their defaults.

## Test plan

- [ ] Existing callers of `create_dataset(dataset_id, user)` continue to work unchanged (backward-compatible `**kwargs` addition)
- [ ] Callers can override any payload field: `create_dataset(dataset_id, user, region="us-east-1", memory="2GB")`
- [ ] Generated instance name is always 29 hex characters (no hyphens, fits within Neo4j Aura's 30-char limit)
- [ ] Same `dataset_id` always produces the same name (deterministic hash)

Fixes #2354

## Summary by CodeRabbit

* **New Features**
  * Dataset creation handlers now accept optional override parameters for instance settings.

* **Improvements**
  * Aura instance names are generated deterministically to meet naming constraints.
  * Interfaces/docstrings clarify that handlers may accept and validate implementation-specific overrides.

* **Bug Fixes**
  * Validation added: dataset ID required; unknown or explicitly null overrides are rejected.

* **Tests**
  * Test handlers updated to accept and validate the new override parameter.

### DIFF
--- a/cognee/infrastructure/databases/dataset_database_handler/dataset_database_handler_interface.py
+++ b/cognee/infrastructure/databases/dataset_database_handler/dataset_database_handler_interface.py
@@ -1,4 +1,4 @@
-from typing import Optional
+from typing import Any, Optional
 from uuid import UUID
 from abc import ABC, abstractmethod
 
@@ -9,7 +9,9 @@ from cognee.modules.users.models.DatasetDatabase import DatasetDatabase
 class DatasetDatabaseHandlerInterface(ABC):
     @classmethod
     @abstractmethod
-    async def create_dataset(cls, dataset_id: Optional[UUID], user: Optional[User]) -> dict:
+    async def create_dataset(
+        cls, dataset_id: Optional[UUID], user: Optional[User], **kwargs: Any
+    ) -> dict:
         """
         Return a dictionary with database connection/resolution info for a graph or vector database for the given dataset.
         Function can auto handle deploying of the actual database if needed, but is not necessary.
@@ -28,6 +30,9 @@ class DatasetDatabaseHandlerInterface(ABC):
         Args:
             dataset_id: UUID of the dataset if needed by the database creation logic
             user: User object if needed by the database creation logic
+            **kwargs: Implementation-specific overrides for the database creation request.
+                Concrete handlers should document and validate the keys they accept and
+                raise ``ValueError`` for unknown keys.
         Returns:
             dict: Connection info for the created graph or vector database instance.
         """
--- a/cognee/infrastructure/databases/graph/kuzu/KuzuDatasetDatabaseHandler.py
+++ b/cognee/infrastructure/databases/graph/kuzu/KuzuDatasetDatabaseHandler.py
@@ -1,6 +1,6 @@
 import os
 from uuid import UUID
-from typing import Optional
+from typing import Any, Optional
 
 from cognee.infrastructure.databases.graph.get_graph_engine import create_graph_engine
 from cognee.base_config import get_base_config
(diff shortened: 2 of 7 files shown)


octo-patch:
Thanks for the careful review @coderabbitai! I've pushed 077c5a8b1 to address all three actionable items:

1. **Interface contract update** — `DatasetDatabaseHandlerInterface.create_dataset()` now declares `**kwargs: Any` so the abstract signature matches subclasses that accept implementation-specific overrides; the docstring documents that concrete handlers must validate the keys they accept.
2. **`dataset_id=None` guard** — Now raises `ValueError` before the hash is computed. Hashing `None` would otherwise silently collapse every `None`-id dataset onto the same Aura instance name, defeating the collision-resistant naming this PR introduced.
3. **Allowed-keys validation** — Replaced the silent `kwargs.get(...)` fallback with an explicit `_ALLOWED_AURA_OVERRIDE_KEYS = {version, region, memory, type, cloud_provider}` set. Unknown keys (e.g. typos like `regoin`) and explicit `None` values now raise `ValueError` so callers learn immediately instead of receiving a default payload silently.

Ready for another look.

octo-patch:
Thanks for the LSP catch @coderabbitai! Pushed [a8fd34d](https://github.com/octo-patch/cognee/commit/a8fd34df4) which propagates the new `**kwargs: Any` through every concrete handler that implements `DatasetDatabaseHandlerInterface.create_dataset` and the `VectorDBInterface` protocol method:

- `KuzuDatasetDatabaseHandler.create_dataset`
- `LanceDBDatasetDatabaseHandler.create_dataset`
- `PGVectorDatasetDatabaseHandler.create_dataset`
- `VectorDBInterface.create_dataset` (Protocol)

Each concrete handler raises `ValueError("<Handler>.create_dataset does not accept overrides; got unsupported keys: ...")` when any override keys are supplied, matching the validation pattern Neo4jAuraDev already uses. Callers that pass nothing through the abstract interface keep the existing behavior; passing overrides now fails fast with a descriptive message instead of a confusing `TypeError`.

octo-patch:
Thanks for the additional review @coderabbitai! Pushed [64fee4a](https://github.com/octo-patch/cognee/commit/64fee4a16) which addresses both the outside-diff blocker and the docstring nitpicks:

1. **Test handler signatures** — `LanceDBTestDatasetDatabaseHandler.create_dataset` and `KuzuTestDatasetDatabaseHandler.create_dataset` in `cognee/tests/test_dataset_database_handler.py` now accept `**kwargs: Any` (with `Any` imported from `typing`) and reject non-empty overrides via the same `ValueError` pattern used by the production handlers, so they stay LSP-compliant with the widened `DatasetDatabaseHandlerInterface` contract.
2. **LanceDB / PGVector docstrings** — added method docstrings on both handlers' `create_dataset` methods, mirroring the Kuzu handler style and explicitly noting that no override kwargs are accepted (`ValueError` on unsupported keys).
3. **`vector_db_interface.py` docstring** — added the missing blank line between the `**kwargs` description and the `Returns:` block to match the rest of the file's docstring style.

Ran `ruff check` and `ruff format --check` on all four files locally — all pass.

dexters1:
Closing PR as it doesn't resolve core issue, added kwargs can never  be used or set properly.

PR #2676, opened 2026-04-19
feat(search): agentic retriever with skills and tools

Vasilije1990:
## Summary

Unifies memory retrieval, skills, and tool calls inside a single `cognee.search()` entry point. Skills and Tools are first-class `DataPoint`s, so they flow through the existing graph + vector pipeline and inherit the existing ACL path — no parallel permission system.

```python
await cognee.remember("skills/")                              # ingest SKILL.md files
await cognee.search("why are customers leaving?",
                    skills=["churn-analyst"])                  # by name
await cognee.search("why are customers leaving?",
                    skills_auto_retrieve=True, skills_top_k=3) # by similarity
```

## What's included

- **New `SearchType.AGENTIC_COMPLETION`** + auto-upgrade: `search()` routes to the new retriever when `skills`/`tools`/`skills_auto_retrieve` are passed alongside any `GRAPH_COMPLETION` family type. No breaking change to existing callers.
- **`AgenticRetriever`** extends `GraphCompletionRetriever` with a ReAct-style tool-use loop emitting structured `AgentStep` output. Skill-declared tools are intersected with the ambient tool set (skills cannot escalate beyond user permissions).
- **`Skill` and `Tool` DataPoints** in `cognee/modules/engine/models/`, using the `Annotated[str, Embeddable()/Dedup()]` convention so they are both vector-indexed and deduplicated by UUID5.
- **SKILL.md loader** (Anthropic Agent Skills format) + `add_skills(path)` helper. `cognee.remember()` auto-dispatches SKILL.md paths/dirs to the skill ingest path and persists them via `add_data_points` (graph + vector).
- **Skill vector auto-retrieve**: `skills_auto_retrieve=True` runs cosine similarity over the `Skill_description` collection and merges hits with any explicit skills (deduplicated by id, filtered by `dataset_scope`).
- **Permission enforcement** via `get_authorized_existing_datasets(...)` — the same function the search API already uses. Single gate for memory reads and tool execution.
- **Built-in tools**: `load_skill` (progressive disclosure primitive — skill bodies load only when the LLM asks for them) and `memory_search` (callable triplet retrieval mid-loop).

## Design anchors (prior art)

- **Anthropic Agent Skills** — SKILL.md format + progressive disclosure (description always in the catalog, procedure body on demand).
- **LangChain toolkits** — tools live in the graph, scoped per dataset; future SQL toolset will follow the `list_tables`/`get_schema`/`check_query`/`run_query` decomposition.
- **Voyager-style skill library** — skills are vector-retrievable, so they compose with cognify's extraction automatically.
- **LangGraph pre-selection** — manifest is computed at turn boundary, preserving prompt caching.

## Test plan

- [x] `pytest cognee/tests/unit/modules/search/` — 25 pass
- [x] `pytest cognee/tests/unit/modules/retrieval/` — 226 pass
- [x] `pytest cognee/tests/unit/api/` — 116 pass
- [x] Offline smoke: Skill/Tool DataPoint identity, SKILL.md parser, built-in registration, skill/tool intersection (with `load_skill` always preserved), `ToolScopeError` on out-of-scope call, `load_skill` via `execute_tool` with `active_skills_var`
- [x] End-to-end: ingest 3 SKILL.md files → `resolve_skills(["churn-analyst"])` resolves by name → auto-retrieve ranks correctly across 3 distinct query topics → combined explicit + auto dedupes
- [ ] LLM-backed integration test (deferred; requires API key in CI)

## Out of scope (deferred)

- `SourceConnection` + SQL passthrough toolset (persist source DB config at ingest time, four-tool LangChain decomposition with `sqlglot` AST filter + readonly role).
- MCP server wrapper for agentic search.

## Summary by CodeRabbit

* **New Features**
  * Agentic search: iterative agents that call tools, run multi-step reasoning, and return stepwise or final answers.
  * Skills framework: create, ingest (SKILL.md), auto-retrieve, and surface reusable skills with new search parameters.
  * Built-in tools: load-skill and on-demand memory search available during agent runs.
  * Tool execution & registry: centralized execution with dataset scoping, permissions, and builtin tool registration.
  * Agent prompts: new system/user templates guiding agent behavior.


cognee/api/v1/remember/remember.py
```diff
@@ -696,6 +696,24 @@ async def _remember_inner(
     # writes, even when the API server was never started.
     await _ensure_migrations_run()
 
+    # Auto-dispatch: a SKILL.md file or directory of SKILL.md files is persisted
+    # as Skill DataPoints rather than run through the chunk+extract pipeline.
+    from cognee.modules.tools import add_skills, looks_like_skill_source
+
+    if looks_like_skill_source(data):
```
hajdul88, line 703: I believe this bypasses user resolution and dataset resolution + setup. So basically the following can happen:

1. A user calls remember("/path/to/SKILL.md", dataset_name="customer_a")
2. Remember() detects that this is a skill source and returns through this shortcut before it resolves the authorized dataset/user and before it enters the normal add() pipeline.
3. add_skills() stores the parsed Skill objects via add_data_points() with no ctx, so add_data_points.py (line 93) never runs the (upsert_nodes / upsert_edges).
4. Later, the same user runs agentic search against customer_an and expects that skill to be available, but skill resolution is dataset-scoped and this ingested skill was never attached to that dataset in the normal way.
5.remember() said it succeeded, but the skill may be invisible to searches in that dataset, and dataset lifecycle operations like status/cleanup won’t treat it as belonging to that dataset.

hajdul88: Additionally:
In enable_backend_access_control = False, I believe the new skill path is not dataset-scoped. Search runs without per-dataset DB context when backend access control is disabled, and resolve_skills() only filters by skill.dataset_scope. But ingested Skill objects never set dataset_scope, so None is effectively treated as “global.” That means a skill stored for dataset A can be auto-retrieved or name-resolved while searching dataset B.

cognee/modules/tools/loaders/skill_md_loader.py
```diff
@@ -0,0 +1,92 @@
...
+def load_skill_from_md(path: Path) -> Skill:
+    """Read a SKILL.md file from disk and return a Skill DataPoint."""
+    text = Path(path).read_text(encoding="utf-8")
+    return parse_skill_md(text)
+
+
+def load_skills_from_directory(root: Path) -> List[Skill]:
+    """Recursively load every SKILL.md under root."""
+    root_path = Path(root)
+    return [load_skill_from_md(p) for p in sorted(root_path.rglob("SKILL.md"))]
```
hajdul88: There’s a mismatch between skill-directory detection and loading. looks_like_skill_source() treats a directory containing either SKILL.md or skill.md as a skill source, but load_skills_from_directory() only loads SKILL.md. As a result, a directory containing only lowercase skill.md files can be detected as a skill directory and still load zero skills.

cognee/modules/retrieval/agentic_retriever.py
```diff
@@ -0,0 +1,278 @@
...
+        declared: set = set()
+        for skill in skills:
+            declared.update(skill.declared_tools or [])
+
+        if declared:
+            # Always expose load_skill so the model can reach procedure bodies.
+            declared.add("load_skill")
+            all_tools = [t for t in all_tools if t.name in declared]
+
+        return all_tools
```
hajdul88, line 141: Coderabbit finding is correct, I believe this can fallback to all tools. I dont know if that was intentional

cognee/modules/tools/resolve_skills.py
```diff
@@ -0,0 +1,186 @@
...
+        graph_engine = await get_graph_engine()
+    except Exception:
+        return None
+
+    get_nodeset = getattr(graph_engine, "get_nodeset_subgraph", None)
+    if get_nodeset is None:
+        return None
+
+    try:
+        nodes_and_edges = await get_nodeset(node_type=Skill, node_name=[name])
```
hajdul88, line 150: The new graph-backed tool lookup appears to be reading get_nodeset_subgraph() with the wrong return shape. That graph API returns (nodes, edges), and resolve_skills() already unpacks it that way, but _query_tool_nodes() iterates the raw return value directly as if it were a flat node list. In practice that means persisted custom Tool datapoints are likely never coerced into Tool objects, so only built-ins show up in the active tool manifest.

cognee/modules/retrieval/agentic_retriever.py
```diff
@@ -0,0 +1,278 @@
...
+
+        context = (
+            f"# Memory\n{memory_text or '(empty)'}\n\n"
+            f"# Available skills\n{skills_catalog}\n\n"
+            f"# Available tools\n{tools_manifest}"
+        )
+        self._cached_context = context
+        return context
+
+    async def get_completion_from_context(
```
hajdul88, line 173: The new agentic path bypasses the existing graph-completion session manager flow. GraphCompletionRetriever normally uses generate_completion_with_session(...) for single-query cached searches, but AgenticRetriever.get_completion_from_context() short-circuits into the tool loop and never hits that path. Because the factory also upgrades existing GRAPH_COMPLETION-family searches implicitly when skills / tools / auto-retrieve are present, enabling agentic behavior also silently drops session-aware completion behavior for those searches.

This means basically sessions are not gonna work, if its intentional its fine, it not I think it is worth addressing it.

hajdul88 (changes requested):
I’ve left comments on the PR covering the most critical issues. The CodeQL findings should also be treated as critical, since they expose serious vulnerabilities that would be especially risky in a SaaS context, particularly if the retriever and its parameters are exposed through the public API.
Beyond the security findings, there are substantial problems with the current implementation and integration of the retriever. In its current state, I would not recommend merging this PR, naturally the final decision lies with you. 
Once the intended behavior and exact specification are clearly defined, I think the right next step is either a substantial redesign or a fresh implementation. I don’t think incremental patching of the current version is the best path, because the issues are structural rather than isolated.

PR #2679, opened 2026-04-19
feat(sessions): unified memory entries, session lifecycle + dashboard endpoints

Vasilije1990:
## Summary

- Makes `remember()` / `recall()` polymorphic via a typed `MemoryEntry` union (`QAEntry` / `TraceEntry` / `FeedbackEntry`) so rich shapes dispatch to the matching `SessionManager` method instead of flattening through `add_qa(answer=blob)`.
- Adds a narrow `session_records` table for lifecycle + aggregate counters (tokens / cost / status / timestamps). Rich content stays in the session cache — this is a metrics + lifecycle layer, not a duplicate.
- New `/api/v1/sessions` router backs a dashboard: list, aggregate stats (spend / tokens / agent time / success rate), cost-by-model, detail.
- **No sweeper process.** The `abandoned` status is computed at read time (`CASE WHEN status='running' AND last_activity_at < :threshold THEN 'abandoned' ELSE status END`) with a Python-computed threshold passed as a bound param, portable across SQLite/Postgres.
- Per-session concurrency: `asyncio.Lock` registry keyed by `(session_id, op)` serializes `update_qa` / `add_feedback` / `delete_qa`. A separate improve-lock prevents duplicate improve runs when auto-improve, idle-watcher, and SessionEnd all fire.
- Token/cost accumulation via a `ContextVar` scope + `LLMGateway` post-hook. Char-based estimate with a small per-model price table — approximate but unblocks the dashboard numbers. Plumbing exact `response.usage` deeper through litellm is a follow-up.

## Key interface changes (backward-compatible)

| Change | Shape |
|---|---|
| `cognee.remember(entry, session_id=...)` now accepts `MemoryEntry` | `TraceEntry(origin_function=..., status=..., method_params=..., method_return_value=...)` etc. |
| `cognee.recall(query, scope=...)` | `scope ∈ {auto, graph, session, trace, graph_context, all}` or list |
| `RememberResult.entry_id` / `.entry_type` | Populated on typed writes so callers can attach feedback to a freshly stored QA |
| `POST /api/v1/remember/entry` (JSON) | Discriminated-union body; legacy multipart `/api/v1/remember` unchanged |
| `POST /api/v1/recall` | Now routes through `cognee.recall` (was `cognee.search`); accepts `session_id` + `scope` |
| `GET /api/v1/sessions` | List with filters + pagination |
| `GET /api/v1/sessions/stats?range=24h\|7d\|30d\|all` | Dashboard aggregates |
| `GET /api/v1/sessions/cost-by-model?range=...` | Grouped by `last_model` |
| `GET /api/v1/sessions/{session_id}` | Detail: record + recent QAs + recent trace steps |

## Migration

One Alembic revision (`c3d4e5f6a7b8_add_session_records_table`) creates the new table off current head `b2c3d4e5f6a7`. No backfill — existing session caches keep working; rows are created lazily on the next write per session.

## Test plan

- [x] Typed `remember(QAEntry/TraceEntry/FeedbackEntry, session_id=...)` returns `entry_id`; verified QA/trace/feedback chain (attach feedback to returned `qa_id`).
- [x] `recall(scope=["session","trace","graph_context"])` returns results correctly tagged by `_source`.
- [x] `GET /api/v1/sessions`, `/stats`, `/cost-by-model`, `/{session_id}` return 200 on both default user and agent logins.
- [x] `SESSION_ABANDON_AFTER_SECONDS=2` flips stored `running` sessions to effective `abandoned` 3s after last activity, with no process touching the row.
- [x] `error_count` accumulates atomically when writing `status='error'` trace steps.
- [ ] Distributed-improve-lock race under multi-worker deploy (in-process asyncio.Lock is single-worker only; SQL-advisory fallback deferred).
- [ ] Exact `response.usage` from litellm instead of char-estimate (deferred — plumbing change deeper in the stack).

## Follow-ups (not in this PR)

- Frontend dashboard consuming the four endpoints (screenshot reference matches the data contract as shipped).
- Exact token accounting via `response.usage` from the litellm/instructor client.
- SQL-advisory fallback on `session_locks` for multi-worker deployments.

## Summary by CodeRabbit

* **New Features**
  * Typed memory entries (QA/Trace/Feedback) with normalization and package exports; new typed remember-entry API and cloud client method.
  * Sessions API: listing, detail, dashboard stats, and per-model cost breakdown; frontend adds Sessions UI, deep-linking, stats, and Activity & Memory feed.
  * Recall API supports flexible multi-source scope and merged results; per-session locking and usage tracking integrated end-to-end.
  * Automatic LLM usage recording and session lifecycle tracking.

* **Bug Fixes**
  * Pipeline-runs endpoint returns empty list on denied dataset permissions (unless explicitly requested).


cognee/api/v1/improve/improve.py
```diff
@@ -119,16 +119,45 @@ async def improve(
...
+                    feedback_alpha=feedback_alpha,
+                    run_in_background=run_in_background,
+                )
+                stages_run.extend(["feedback_weights", "persist_sessions"])
+            except Exception:
+                if acquired_lock_for:
+                    from cognee.infrastructure.locks import release_improve_lock
+
+                    await release_improve_lock(acquired_lock_for)
+                raise
```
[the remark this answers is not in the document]
hajdul88: This is valid, I see the release at the end but still, maybe consider releasing the lock in finally, its cleaner

cognee/api/v1/improve/improve.py
```diff
@@ -225,6 +272,47 @@ async def _bridge_sessions(
         logger.warning("improve: session persistence failed (non-fatal): %s", e)
 
 
+async def _persist_session_traces(
```
hajdul88, line 275: Its minor but maybe wrapper is not needed

cognee/api/v1/recall/recall.py
```diff
@@ -51,36 +52,126 @@ def _tokenize(text: str) -> set[str]:
...
+      We pick the candidate most likely to have real entries — the
+      one with ``dataset_id`` populated wins over an empty row.
+
+    * Falls back to ``caller_user_id`` when nothing is in
+      ``session_records`` yet so behaviour matches the pre-visibility
+      state.
+    """
+    if not session_id:
+        return caller_user_id
+    try:
```
hajdul88, line 90: With this different users, maybe even different tenants will be able to an access and memify each others data from sessions. Is this something that we are okay with or was this the goal? (We are checking this together with Igor)

cognee/api/v1/recall/recall.py
```diff
@@ -119,13 +208,120 @@ async def _search_session(
...
+
+
 async def recall(
     query_text: str,
     query_type: Optional[SearchType] = None,
     *,
     datasets: Optional[list[str]] = None,
     top_k: int = 10,
     auto_route: bool = True,
+    scope: Optional[Union[str, list[str]]] = None,
```
hajdul88, line 324: Igor: Its not intuitive and very complicated, it should be just different endpoints/functions. This is a new concept people don't know and it will be hard to maintain.

cognee/api/v1/recall/recall.py
hajdul88: Sharing sessions could make sense, but maybe should be handled in a different way

cognee/api/v1/recall/recall.py
```diff
@@ -161,16 +357,42 @@ async def recall(
...
+            sources = ["session", "graph"]
+            auto_fallthrough = False  # both contribute
+        else:
+            sources = ["graph"]
+            auto_fallthrough = False
+    else:
+        sources = resolved_scope
+        auto_fallthrough = False
+
+    span_scope = ",".join(sources)
```
hajdul88, line 388: Is the scope a string separated by commas? Maybe it is worth reconsidering the scope data structure and make it more robust.

hajdul88 (commented):
We did the review together with Igor and added comments to the relevant parts. Coderabbit issues seem mostly correct so would be worth considering addressing those too.

dexters1:
Scope feature needs to be reconsidered:
1. Do we want the feature? Does not seem intuitive and makes much more sense to seperate functions/endpoints. Non tech users can use UI + ClaudeCode to handle it
2. If we want it parsing strings is not a good way to handle it

Sharing permissions needs to have a permission system behind it, otherwise all users across tenants can read each other sessions this is not properly handled and the work there needs to be refactored.

There are lock issues with how everything is handled

Issue #2489, opened 2026-03-26
perf: parallelize DB operations in add_data_points


soichisumi:
## Problem

`add_data_points()` in `cognee/tasks/storage/add_data_points.py` runs 6 sequential database operations:

1. `get_graph_from_model()` — Extract nodes and edges from each DataPoint
2. `graph_engine.add_nodes()` — Persist nodes to graph DB
3. `graph_engine.add_edges()` — Persist edges to graph DB
4. `index_data_points()` — Create vector indexes
5. `graph_engine.has_edges()` — Check for existing edges
6. `index_graph_edges()` — Index edge types with vectors

Several of these can run concurrently:

### Phase 1: Extract (sequential, CPU-bound)
- `get_graph_from_model()` for all DataPoints

### Phase 2: Persist (parallel, I/O-bound)
- `graph_engine.add_nodes()` ‖ `graph_engine.add_edges()` ‖ `index_data_points()`

### Phase 3: Edge indexing (parallel, I/O-bound)
- `graph_engine.has_edges()` → `index_graph_edges()`

## Impact

On FalkorDB with ~100 nodes and ~200 edges, this reduces `add_data_points` time by approximately 40% in our testing, as the graph DB operations are network-bound.

## Implementation Notes

- `add_nodes` and `add_edges` are independent and can safely run concurrently
- `index_data_points` can run concurrently with graph operations
- `has_edges` depends on edges being added first, so it must wait for Phase 2
- `index_graph_edges` depends on edge data, so it must wait for Phase 2

soichisumi:
## Update: v0.5.5 partially addressed this

In v0.5.5, `add_data_points()` now uses `asyncio.gather()` for the initial `get_graph_from_model()` extraction phase:

```python
results = await asyncio.gather(
    *[get_graph_from_model(dp, ...) for dp in data_points]
)
```

This addresses **Phase 1** from the original proposal.

### Still sequential in v0.5.5

The post-dedup database operations remain sequential:

```python
await graph_engine.add_nodes(nodes)          # Phase 2 — could parallelize
await index_data_points(nodes, ...)          # Phase 2
await upsert_nodes(...)                      # Phase 2
await upsert_edges(...)                      # Phase 2
await graph_engine.add_edges(edges)          # Phase 3 — could parallelize
await index_graph_edges(edges, ...)          # Phase 3
```

The proposed parallelization into phases is still valid:
[...]

PR #2688, opened 2026-04-21
refactor: parallelize db operations in add_data_points

siillee:
## Description

This PR introduces changes suggested in Issue #2489. Graph and vector database operations now run in parallel during add_data_points.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

## Summary by CodeRabbit

* **Refactor**
  * Improved data point storage performance by processing graph insertion and vector indexing concurrently for non-hybrid writes, reducing end-to-end write latency.


nfnt (commented):
Coderabbit has some good comments here, that should be addressed.
I think that if any of these tasks raise an exception, the other tasks should stop to run. I recommend to use a `asyncio.TaskGroup` for that instead of the suggestion to set `return_exceptions=True` on `gather`.
Furthermore, the comment on shared data is on point. Let's create deep copies of the arguments, otherwise there could be unexpected behavior.
Alternatively, one could use `model_copy(deep=True)` in the mentioned functions to avoid mutating data, but that won't guarantee that there won't be changes in the future that mutate data again. I.e. it's best to create a deep copy here.

siillee:
@nfnt Fair points. The issue with potential partial states of different dbs will be addressed in a separate task I started working on today. The idea of adding transaction-like behaviour to cognify has been there for some time, we just never got around to implementing it. The idea being that we should rollback any changes made if something fails mid-cognify, to avoid partial states. This PR does add to this problem, so there is the option of just postponing the merge of this PR until we have something that solves the issue.

As for the shallow/deep copies, that also makes sense, so I'll look into it.

PR #2691, opened 2026-04-21
feat: clear return types for 'recall'

nfnt:
## Description

Provides a clear OpenAPI schema for the various message types returned by the '/api/v1/recall' API. This has been done by refactoring the 'SessionManager' to return types instead of dictionaries and using these types to describe a union type as return value. The return value is then presented as a 'OneOf' type with the 'source' field as discriminator.

This is using functionality from #2694 to normalize search results.

## Acceptance Criteria

- [x] OpenAPI schema for the `/api/v1/recall` endpoint provides a discriminate union of return types
- [ ] All existing tests pass, besides changing the return types, no functionality has been changed

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

- [x] **I have tested my changes thoroughly before submitting this PR** (See `CONTRIBUTING.md`)
- [x] **This PR contains minimal changes necessary to address the issue/feature**
- [x] My code follows the project's coding standards and style guidelines
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if applicable)
- [x] All new and existing tests pass
- [x] I have searched existing PRs to ensure this change hasn't been submitted already
- [x] I have linked any relevant issues in the description
- [x] My commits have clear and descriptive messages

## DCO Affirmation
I affirm that all code in every commit of this pull request conforms to the terms of the Topoteretes Developer Certificate of Origin.

## Release Notes

* **New Features**
  * Standardized recall endpoint responses with typed result models including source discrimination
  * Normalized search result formatting across all search types for consistent output

* **Improvements**
  * Enhanced API type safety across recall and search endpoints
  * Enforced dataset authorization with UUID mapping for dataset names
  * Improved error handling when authorized datasets are not found

PR #2694, opened 2026-04-21
feat(search): normalize results into SearchResponse envelope

Vasilije1990:
## Summary
- Replace polymorphic \`cognee.search\` return shape (list of dicts, lists, or bare strings depending on access-control mode, \`verbose\` flag, and dataset count) with a validated \`SearchResponse\` envelope.
- Each hit is a \`SearchResultItem\` with \`kind\`, \`text\` (always renderable), \`raw\`, \`metadata\`, \`dataset_id/name\`, optional \`score\`, optional \`structured\`.
- Internal consumers updated: \`recall.py\`, MCP search tool, HTTP routers. Legacy \`_backwards_compatible_search_results\` deleted; obsolete \`SearchResult\` (with \`search_result: Any\`) removed.

## Breaking changes (intentional — no backwards-compat)
- \`cognee.search\` returns \`SearchResponse\`, not \`list\`. Iterate \`response.results\`.
- \`verbose\` parameter removed from \`search()\`, search HTTP DTO, and recall HTTP DTO. Its data is always present on each item via \`raw\` / \`metadata\`.
- Silent single-dataset list unwrap \`[[...]] -> [...]\` removed.
- \`cognee.modules.search.types.SearchResult\` deleted.

## Files
- New: \`cognee/modules/search/types/SearchResponse.py\`, \`cognee/modules/search/methods/normalize_search_result.py\`
- Updated: inner search, public search, search + recall routers, recall, MCP server
- Tests rewritten: \`test_search.py\`, \`test_search_prepare_search_result_contract.py\`, \`test_search_api.py\`, \`test_api_error_responses.py\` (SearchEndpoint), \`test_sync_graph_to_session.py\` (TestRecallSessionMode x2)

## Known downstream breakage not addressed in this PR
- \`cognee/tests/backwards_compatibility/phase{1,2}_*.py\` — explicitly assert the old shape
- \`examples/**/*.py\`, \`notebooks/*.ipynb\` — rely on the old polymorphic shape
- \`cognee-frontend/src/app/(app)/search/SearchPage.tsx\` — TS client

## Test plan
- [x] \`uv run pytest cognee/tests/unit/ --ignore=cognee/tests/unit/infrastructure --ignore=cognee/tests/unit/eval_framework\` — 1202 passed locally
- [ ] CI full suite
- [ ] Manual: \`cognee.search(...)\` via CLI against a seeded dataset, verify \`.results[0].text\`
- [ ] Manual: POST \`/api/v1/search\` — confirm JSON matches \`SearchResponse\` schema
- [ ] Manual: MCP \`search\` tool end-to-end against a live session

## Summary by CodeRabbit

* **New Features**
  * Introduced a unified SearchResponse envelope with per-item kind, text, metadata, raw payload, dataset info, and total count.

* **Bug Fixes**
  * Normalized outputs across retrievers and search types for consistent rendering (including completion and insights results).

* **Refactor**
  * Removed legacy verbose flag and aligned API endpoints and handlers to the new normalized response shape; replaced legacy result types with the new envelope.

* **Tests**
  * Updated unit, integration, and example tests/scripts to assert and consume the new SearchResponse structure.


cognee/api/v1/search/search.py
```diff
@@ -207,7 +196,8 @@ async def search(
 
     client = get_remote_client()
     if client is not None:
-        return await client.search(query_text, search_type=query_type, datasets=datasets)
+        raw = await client.search(query_text, search_type=query_type, datasets=datasets)
+        return SearchResponse.model_validate(raw)
```
nfnt, line 200: We could update `CloudClient` to return a `SearchResponse` as well and save us the `model_validate` here. On the other hand, this would be inconsistent with how it currently works.
But in my opinion, as we have Pydantic types instead, we should aim to use them instead of dictionaries whenever possible.

nfnt (dismissed):
This looks good to me and will be very useful for #2691.
The coderabbit comment in `cognee-mcp/src/server.py` about `FEELING_LUCKY` search types is quite relevant though, we should add the suggested fix.

nfnt:
I added a few commits to fix tests and formatting and merged latest `dev`. With this, all CI check should run successfully.

nfnt:
The backwards-compatibility tests are failing because this is a breaking change. @Vasilije1990, how do we handle these cases? Other than that, all tests and examples have been updated to use `SearchResponse`.

PR #2695, opened 2026-04-21
Feat: Dataset queue in set context async manager

dexters1:
## Description
- Refactor set database context global variables system to work as an async context manager
- Added database queue system management directly to be handled by the setting of the database context global variables 

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

## Summary by CodeRabbit

* **New Features**
  * Introduced a Dataset Queue to limit concurrent operations per dataset and a scoped database context manager to ensure per-dataset execution lifetimes.

* **Bug Fixes**
  * Fixed async transaction rollback behavior in permission workflows for more reliable DB error handling.

* **Tests**
  * Added extensive unit, integration, and end-to-end tests validating dataset-queue concurrency, serialization, slot accounting, and error/release behavior; added CI e2e jobs to run these tests.


dexters1:
Failing CI for Windows unit tests, backwards compatibility and convesation sessions are not related to changes in this PR

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,306 @@
...
+        Idempotent past depth=0: further calls are no-ops because the entry
+        was popped once depth reached zero.
+        """
+        if not self._enabled:
+            return
+        try:
+            task = asyncio.current_task()
+        except RuntimeError:
+            # Called outside a running event loop — nothing to release.
+            return
```
nfnt: This is inconsistent with similar code in `ensure_slot`: According to Python docs, this code is not expected to raise an exception.

```suggestion
        task = asyncio.current_task()
```

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,306 @@
...
+        if task is None:
+            return
```
nfnt: Shouldn't we raise an exception like in `ensure_slot`?

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,306 @@
...
+
+        task_id = id(task)
+        ds_key = f"ds:{dataset_id}" if dataset_id is not None else "ds:<none>"
+
+        entry = self._task_slots.get(task_id, {}).get(ds_key)
+        if entry is None:
+            return
```
nfnt: Raise an exception here? IMO, trying to release the slot for a dataset that isn't tracked here means that the caller is doing something very wrong... didn't they acquire a slot before? There should be an error for this case.

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,306 @@
...
+            slots = self._task_slots.pop(_tid, {})
+            self._registered_tasks.discard(_tid)
+            for entry in slots.values():
+                # Backstop: release whatever's left regardless of depth.
+                entry.release()
+
+        task.add_done_callback(_release_all_on_done)
+
+    # ------------------------------------------------------------ ensure_slot
+    async def ensure_slot(self, dataset_id: Any = None) -> None:
```
nfnt: Do we need to support the case of `dataset_id == None` here? Isn't that already covered by `acquire` below?

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,306 @@
...
+
+        try:
+            yield
+        finally:
+            # Release on scope exit — we don't wait for task end for these.
+            if task_id is not None and slot_key is not None:
+                slots = self._task_slots.get(task_id)
+                if slots is not None:
+                    slots.pop(slot_key, None)
+            release()
```
nfnt, line 280: We could also use `async with` here:


```suggestion
        async with self._semaphore:
            slot_key = None
            if task_id is not None:
                self._ensure_task_cleanup_registered(task, task_id)
                slot_key = f"acquire:{id(release)}"
                # Store a SlotEntry for type uniformity; acquire() is scoped
                # and doesn't use the depth counter.
                self._task_slots[task_id][slot_key] = SlotEntry(release, depth=1)

            try:
                yield
            finally:
                # Release on scope exit — we don't wait for task end for these.
                if task_id is not None and slot_key is not None:
                    slots = self._task_slots.get(task_id)
                    if slots is not None:
                        slots.pop(slot_key, None)
```

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,306 @@
...
+        Unlike ``ensure_slot``/``release_slot_for``, ``acquire`` is always
+        strictly scoped (enter/exit pair in a single ``async with``), so no
+        depth counter is needed — the entry is popped and released on exit.
+        """
+        if not self._enabled:
+            yield
+            return
+
+        task = asyncio.current_task()
+        task_id = id(task) if task is not None else None
```
nfnt, line 251: Similar to the functions above: Do we expect that `task` could be `None`? If not, we should raise an exception here. This would simplify the function.

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,306 @@
...
+        await self._semaphore.acquire()
+        release = _make_release(self._semaphore)
+
+        slot_key = None
+        if task_id is not None:
+            self._ensure_task_cleanup_registered(task, task_id)
+            slot_key = f"acquire:{id(release)}"
+            # Store a SlotEntry for type uniformity; acquire() is scoped
+            # and doesn't use the depth counter.
+            self._task_slots[task_id][slot_key] = SlotEntry(release, depth=1)
```
nfnt, line 270: Do we actually need to track this here? I'm not too deep into Python tasks, so I may be missing something, but here's how I understand it: This function is called as part of a task, acquires a semaphore and later releases that semaphore. As we don't have to track datasets here, this can all be done as part of a single call that won't change the tasks ID. I.e. we don't need to keep track of slots and don't need to clean them up once a task is done.

cognee/context_global_variables.py
```diff
@@ -104,80 +105,154 @@ def is_multi_user_support_possible():
...
+        # which scopes the dataset queue slot to a block and releases it on exit.
+        warnings.warn(
+            "`await set_database_global_context_variables(...)` is deprecated. "
+            "Use `async with set_database_global_context_variables(...):` instead "
+            "for scoped dataset-queue slot management; the `async with` form "
+            "releases the database queue slot automatically on block exit, the deprecated method does not.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        return self._apply().__await__()
```
nfnt, line 211: Is this part of Cognee's public API? Couldn't we simply remove it instead of deprecating it, if it's only used internally? Asking because I only see it being used internally, but I might be missing some context here.

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,306 @@
...
+        logger.debug("Task %d acquiring dataset queue slot for dataset_id=%s", task_id, dataset_id)
+        await self._semaphore.acquire()
+        release = _make_release(self._semaphore)
+
+        self._ensure_task_cleanup_registered(task, task_id)
+        # After registration, the task entry exists in ``_task_slots``.
+        self._task_slots[task_id][ds_key] = SlotEntry(release, depth=1)
+
+    # -------------------------------------------------------- release_slot_for
+    def release_slot_for(self, dataset_id: Any = None) -> None:
```
nfnt, line 195: If this is supposed to only be used as part of a `ContextManager`, we could simplify this function similar to my comments on `acquire` below: With that assumption, the calls to `ensure_slot` and `release_slot_for` should come from the same task and we could remove the whole cleanup functionality.
This is assuming that `release_slot_for` is indeed called (for that we have a `ContextManager`) and that I'm not wrong about the task not changing between calls.

nfnt (commented):
Thanks for improving dataset handling! I had a deep look into the queue implementation and I think there can be some simplifications. But that's if my assumptions about Python task handling are right.. you might be more knowledgeable there, so please take a look at my suggestions and if they make sense.

siillee (commented):
Jan's comments make sense to me as well.
I would ask a question related to CodeRabbit's comments, i.e. what happens when the `run_in_background` flag is set to `true` in operations like cognify and memify?

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,302 @@
...
+            slots = self._task_slots.pop(_tid, {})
+            self._registered_tasks.discard(_tid)
+            for entry in slots.values():
+                # Backstop: release whatever's left regardless of depth.
+                entry.release()
+
+        task.add_done_callback(_release_all_on_done)
+
+    # ------------------------------------------------------------ ensure_slot
+    async def ensure_slot(self, dataset_id: Any = None) -> None:
```
dexters1:
```suggestion
    async def ensure_slot(self, dataset_id) -> None:
```

cognee/infrastructure/databases/dataset_queue/queue.py
dexters1: No we don't, removing it now

cognee/infrastructure/databases/dataset_queue/queue.py
dexters1: Yes, but overall it should not be possible to happen that task is None

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,302 @@
...
+          * the current task doesn't hold a slot for ``dataset_id``.
+
+        Idempotent past depth=0: further calls are no-ops because the entry
+        was popped once depth reached zero.
+        """
+        if not self._enabled:
+            return
+        task = asyncio.current_task()
+        if task is None:
+            return
```
dexters1:
```suggestion
```

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,301 @@
...
+          * there is no running task,
+          * the current task doesn't hold a slot for ``dataset_id``.
+
+        Idempotent past depth=0: further calls are no-ops because the entry
+        was popped once depth reached zero.
+        """
+        if not self._enabled:
+            return
+        task = asyncio.current_task()
+        if task is None:
```
dexters1:
```suggestion
```

cognee/infrastructure/databases/dataset_queue/queue.py
dexters1: Now it is, initially the Queue system I wrote was not written with ContextManager in mind and was handling Queue cleanup on Task exit and relying on the whole system to only be used inside of Tasks (as the Context Variables can only work properly in Tasks)

For a lot of reasons I decided to rework it as a ContextManager to make it much easier to use with no possbility of having dead elements in the Queue. Not everyone will use Cognee as a server.

This piece of code is a remnant of the initial design, did not have time to test the implementation if we removed it so kept it in as a backup, but most likely is not needed anymore

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,300 @@
...
+        """
+        if not self._enabled:
+            return
+        task = asyncio.current_task()
+
+        task_id = id(task)
+        ds_key = f"ds:{dataset_id}" if dataset_id is not None else "ds:<none>"
+
+        entry = self._task_slots.get(task_id, {}).get(ds_key)
+        if entry is None:
```
dexters1:
```suggestion
```

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,299 @@
...
+        """
+        if not self._enabled:
+            return
+        task = asyncio.current_task()
+
+        task_id = id(task)
+        ds_key = f"ds:{dataset_id}" if dataset_id is not None else "ds:<none>"
+
+        entry = self._task_slots.get(task_id, {}).get(ds_key)
+            return
```
dexters1:
```suggestion
```

cognee/modules/memify/memify.py
```diff
@@ -98,14 +98,12 @@ async def memify(
...
             authorized_dataset.id, authorized_dataset.owner_id
-        )
-
-        memory_fragment = await get_memory_fragment(node_type=node_type, node_name=node_name)
-        # Subgraphs should be a single element in the list to represent one data item
-        data = [memory_fragment]
+        ):
+            memory_fragment = await get_memory_fragment(node_type=node_type, node_name=node_name)
+            # Subgraphs should be a single element in the list to represent one data item
+            data = [memory_fragment]
```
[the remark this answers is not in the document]
dexters1: Need to check this

cognee/infrastructure/databases/dataset_queue/queue.py
dexters1: We probably don't, this is a remnant of the initial design. Did not have time to test if it will still work properly if I removed it as there was a rush to get it done.

Now it's more of a fail safe that is most likely not needed

cognee/context_global_variables.py
dexters1: Some of our custom pipelines and code in the community repo still use this old syntax, we don't have to update it all if we keep it backwards compatible

Also some of our users might have built their own code utilizing it so decided to add a warning and deprecation to keep it backwards compatible

While the new Queue system is nice to have, it's not mandatory and old code can be used for regular use

cognee/infrastructure/databases/dataset_queue/queue.py
```diff
@@ -0,0 +1,298 @@
...
+Re-entrancy rules:
+
+* Same task + same dataset → depth counter increments; no new acquire.
+* Same task + different dataset → acquire an additional slot (may block).
+* Different task (e.g. a child task that inherited the ContextVar) → treated
+  as a fresh task; acquires its own independent slot.
+
+Configuration:
+
+* ``DATASET_QUEUE_ENABLED`` — env var. Truthy values enable the queue.
```
hajdul88, line 30: Maybe would be good idea to add these to the env template. Now its disabled by deafult

cognee/memify_pipelines/apply_feedback_weights.py
```diff
@@ -56,23 +56,22 @@ async def apply_feedback_weights_pipeline(
...
-        user=user,
-        run_in_background=run_in_background,
-    )
+        result = await memify(
+            extraction_tasks=extraction_tasks,
+            enrichment_tasks=enrichment_tasks,
+            dataset=dataset_to_write[0].id,
+            data=[{}],
+            user=user,
+            run_in_background=run_in_background,
```
hajdul88, line 73: Before merge we agreed in the call that run_in_background should be tested.

cognee/infrastructure/databases/dataset_queue/queue.py
dexters1: Probably would have been better to do it as an async with, it looks much nicer. I would avoid making the change only because I'd have to test it again and have limited time now

hajdul88 (approved):
LGTM, very nice job especially given the last minute changes

PR #2705, opened 2026-04-23
Graphskills on agentic

Vasilije1990:
Builds on top of #2676 (feat/agentic-search-skills-tools) and adds the
self-improvement loop from `demo/graphskill_COG-4178`. Since #2676
hasn't merged yet, this PR's diff against `dev` includes both changesets.
Merging order can be either: merge #2676 first, then this rebases down
to only the graphskills delta; or merge this as-is and #2676 closes
automatically.

## Summary

* Merges `origin/demo/graphskill_COG-4178` into the agentic retriever
  branch.
* Unifies the two `Skill` DataPoint classes into one at
  `cognee.modules.engine.models.Skill` — single schema, single
  storage. Self-improvement runtime + agentic retriever share the
  same graph nodes.
* Collapses ingest onto a single entry point:
  `cognee.remember("skills/", enrich=...)`. Dropped
  `cognee.skills.ingest` / `.upsert` / `.ingest_meta_skill` — the
  `skills` client is runtime-only now (routing / execution /
  inspect / amendify).
* Content-hash-based upsert inside `add_skills` so
  `cognee.remember` is idempotent and incremental; edits emit
  `SkillChangeEvent`; stale graph + vector rows get cleaned up.

## New user-facing flow

```python
import cognee
from cognee import skills

# Ingest (rich parse + LLM enrichment by default)
await cognee.remember("./my_skills", enrich=True)
await cognee.remember(skills.META_SKILL_PATH, enrich=True)

# Route through the agentic retriever (PR #2676's surface)
answer = await cognee.search(
    "Why are customers leaving?",
    skills_auto_retrieve=True, skills_top_k=3,
)

# Self-improve a skill that's scoring low
diag  = await skills.inspect("churn-analyst")
amdt  = await skills.preview_amendify("churn-analyst")
await skills.amendify(amdt["amendment_id"])
# or one call:
await skills.auto_amendify("churn-analyst")
```

## What changed

**Schema**
* `Skill` extended with the self-improvement fields: `content_hash`,
  `triggers` / `tags`, `instruction_summary`, `resources`,
  `enrichment_model` / `enrichment_confidence`, `task_pattern_candidates`.
* `TaskPattern` and `SkillResource` moved into the same file as
  `Skill` to break a circular `model_rebuild` cascade.
* Canonical field names only (`name` / `procedure` /
  `declared_tools`) — no `@property` aliases, which Pydantic silently
  drops from `model_dump()` and thus from graph nodes.

**Ingest**
* `cognee.modules.tools.ingest_skills.add_skills(source, enrich=True)`
  is the single entry point. Always uses the rich parser (content
  hash, resources, extra_metadata); `enrich=True` adds the LLM
  enrichment pipeline + TaskPattern materialization.
* `cognee.remember("skills/", enrich=...)` threads through.
* Default flipped to `enrich=True` — un-enriched skills have empty
  `triggers` / `tags` / `instruction_summary` and routing drops to
  random.
* `cognee.skills.{ingest,upsert,ingest_meta_skill}` removed. Added
  `cognee.skills.META_SKILL_PATH` so the bundled meta-skill is still
  discoverable (pass to `remember`).
* MCP `ingest_skills` tool now delegates to `cognee.remember`;
  `upsert_skills` tool dropped.

**Rename pass (canonical field names)**
* Graph-node reads in `client.py` / `pipeline.py` / `inspect.py` /
  `amendify.py` / `retrieve.py` / `observe.py` now read
  `.get("name")` / `.get("procedure")` / `.get("declared_tools")`.
* `Skill(...)` constructors in `amendify.py` use canonical kwargs.
* Public dict APIs (`load` / `list` / `recommend_skills`) keep their
  historical `"skill_id"` / `"instructions"` keys for backwards compat
  — the source translation happens inside.
* `SkillRun` / `SkillInspection` / `SkillAmendment` keep their own
  `skill_id` / `instructions` fields (real fields pointing at Skills).

**Deleted dead code**
* `cognee.cognee_skills.pipeline.ingest_skills` / `upsert_skills` —
  unreachable after the client surface change.
* `cognee.modules.tools.loaders.skill_md_loader` — minimal
  frontmatter parser superseded by the rich parser.

## Test plan

* `cognee/tests/unit/modules/search/`: 25/25 ✓
* `cognee/tests/unit/modules/retrieval/`: 226/226 ✓
* `cognee/cognee_skills/tests/`: 99 pass, 9 skipped (6 e2e tests
  that mocked the removed pipeline surface; superseded by the new
  smoke test below)
* **New:** `cognee/cognee_skills/tests/test_smoke_remember_to_amendify.py`
  — end-to-end wiring test of the unified flow:
    * `cognee.remember` auto-dispatches SKILL.md sources
    * canonical `Skill` persisted with rich-parser fields
    * re-ingest is idempotent (content-hash diff)
    * edit emits "updated" `SkillChangeEvent`
    * `skills.*` runtime surface callable, ingest methods absent
    * `SearchType.AGENTIC_COMPLETION` coexists
  * No LLM, no real graph — all mocked at the DB boundary.

## Known follow-ups (not blocking)

- [ ] Rewrite the 6 skipped e2e tests against the new
  `cognee.remember` surface instead of the old `pipeline.ingest_skills`
  mocks.
- [ ] Relocate `SkillRun` / `SkillInspection` / `SkillAmendment` /
  `SkillChangeEvent` into `cognee.modules.engine.models` alongside
  `Skill`. Consistency win, no runtime change.

## Acceptance criteria

- [x] All existing PR #2676 unit tests still pass (251/251).
- [x] Skills ingestion has a single entry point (`cognee.remember`).
- [x] `Skill` is one class, shared between both subsystems.
- [x] Re-running `cognee.remember` on the same folder is idempotent;
  edits produce `SkillChangeEvent`.
- [x] `ruff check` and `ruff format` clean on the branch.


lxobr:
A few high-level points after spending a bit of time with this, happy to discuss any of these:

- Split this into two PRs: AgenticRetriever into a separate PR. It's a good approach that should be slightly generalizad and replaceme the CoT/extension retrievers we can consider deprecating. Skills should be opt-in behavior, not baked into the retriever itself.

- Avoid writing a `SkillRun` node for every opened skill on every call. At current volume this creates a large amount of graph churn, and `inspect_skill` already does a full nodeset scan. Explore session memory first, then persist/promote only on session save or low-score feedback.

- The example in the PR description doesn’t run on this branch. `cognee.skills`, `META_SKILL_PATH`, `skills.inspect`, and `skills.amendify` don’t exist, while `test_skill_ingest.py` asserts `cognee.skills` was removed. Either rewrite the example against the current APIs (`cognee.remember` + `cognee.modules.memify.*`) or ship the missing namespace.

- Skill ingestion currently duplicates large parts of `cognify` (`parsing`, content-hash diffing, ACL backfill, change events, vector cleanup). We should either fold skills into `cognify` as a document type, or expose a dedicated `cognee.ingest_skills(...)` entry point.

About the demo:

- The demo doesn’t yet show a measurable improvement. There’s no baseline or before/after comparison, and `score_review` currently checks review shape via regex rather than review quality. I’d frame this as a smoke test for now, or add a comparison run.

- Daytona’s S3-only storage introduces a lot of workaround complexity (`snapshot tarballs`, layered storage, `sitecustomize.py`, `cognee-cli` wrappers). A simpler example/environment (Modal, VM, local containers) would keep the demo focused on cognee itself.

- Most of the cognee-specific logic is buried under ~800 lines of infra plumbing. Extracting the inline Python scripts into files and splitting `deploy_claude_code_agent` into named stages would make the flow much easier to follow without changing the architecture.

PR #2716, opened 2026-04-24
test: add cognify pipeline status check before search in server test

### LINKED ISSUE #2350: Add wait condition in test_cognee_server_start to ensure cognify pipeline completes before search
*(Created as part of the codebase TODO/FIXME audit #2316)*

In `cognee/tests/test_cognee_server_start.py`, there is an end-to-end integration test (`test_server_is_running`) which triggers a `cognify` pipeline via a POST request and then immediately proceeds to test the `/search` endpoint.

Currently, there is a `TODO: Add test to verify cognify pipeline is complete before testing search`. Because `cognify` can be asynchronous or take varying amounts of time depending on load, this test may be prone to flakiness if the search runs before the pipeline has finished persisting the graph nodes.

**Acceptance Criteria:**
- Implement a polling mechanism or status check in `test_server_is_running` to verify the `cognify` pipeline for the test dataset is `COMPLETED` before proceeding to the search request.
- Ensure the test suite continues to pass reliably in CI.

octo-patch:
Fixes #2350

## Problem

`test_server_is_running` in `cognee/tests/test_cognee_server_start.py` sends a `cognify` request and immediately proceeds to a `search` request, with only a `TODO` comment noting that the pipeline completion should be verified first. This is a source of test flakiness: the synchronous `cognify` API call returns after the pipeline is submitted, but the search may run before graph nodes are fully committed to the database.

## Solution

Add a `_wait_for_cognify_pipeline` helper method that polls
`GET /api/v1/datasets/status?dataset=<id>` until the pipeline
reports `DATASET_PROCESSING_COMPLETED` (success) or `DATASET_PROCESSING_ERRORED`
(failure), or until a 120-second timeout is reached. The helper is called after
the cognify response and before the search request, replacing the old `TODO` comment.

```python
def _wait_for_cognify_pipeline(self, dataset_id, auth_var, timeout=120, poll_interval=5) -> bool:
    # polls GET /api/v1/datasets/status?dataset=<id>
    # returns True on COMPLETED, False on ERRORED or timeout
```

## Testing

The change adds an explicit gate that replaces an implicit race. If the pipeline
hasn't completed within 120 seconds the test fails with a descriptive message
rather than a misleading search error.

Signed-off-by: octo-patch <octo-patch@github.com>

## Summary by CodeRabbit

* **Tests**
  * Enhanced dataset processing test to synchronize with pipeline completion before validation, improving test reliability and preventing premature assertions.

### DIFF
--- a/cognee/tests/test_cognee_server_start.py
+++ b/cognee/tests/test_cognee_server_start.py
@@ -48,6 +48,40 @@ class TestCogneeServerStart(unittest.TestCase):
                 cls.server_process.terminate()
             cls.server_process.wait()
 
+    def _wait_for_cognify_pipeline(
+        self, dataset_id: str, auth_var: str, timeout: int = 120, poll_interval: int = 5
+    ) -> bool:
+        """Poll the dataset status endpoint until the cognify pipeline completes or times out.
+
+        Args:
+            dataset_id: UUID of the dataset to check.
+            auth_var: Authorization header value (e.g. "Bearer <token>").
+            timeout: Maximum seconds to wait before giving up.
+            poll_interval: Seconds between status checks.
+
+        Returns:
+            True if the pipeline completed successfully, False on timeout or error.
+        """
+        headers = {"Authorization": auth_var}
+        status_url = f"http://127.0.0.1:8000/api/v1/datasets/status?dataset={dataset_id}"
+        deadline = time.time() + timeout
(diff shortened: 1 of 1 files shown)


dexters1:
Closing PR as default cognify call does not run in background and is not needed to be polled

PR #2735, opened 2026-04-29
fix: handle markdown-wrapped JSON in infer-schema endpoint

LStromann:
## Summary
- The `/v1/llm/infer-schema` endpoint uses `response_model=str` with `LLMGateway.acreate_structured_output`, so the LLM output is not structurally constrained to raw JSON
- LLMs frequently wrap JSON output in markdown code fences (` ```json ... ``` `) or add surrounding text, causing `json.loads()` to fail with a 422 error
- Add `_extract_json()` helper that strips markdown fences and extracts the JSON object before parsing

## Root cause
```python
llm_output = await LLMGateway.acreate_structured_output(
    ...,
    response_model=str,  # Not a Pydantic model → no structured JSON enforcement
)
schema_dict = json.loads(llm_output)  # Fails on markdown-wrapped output
```

## Fix
`_extract_json()` handles three cases:
1. Clean JSON → parse directly
2. Markdown-fenced JSON (` ```json { ... } ``` `) → strip fences, then parse
3. JSON embedded in text → extract first `{ ... }` block

## Test plan
- [ ] Call `/v1/llm/infer-schema` with a text sample → should return valid schema
- [ ] Call with uploaded files → should return valid schema
- [ ] Verify markdown-wrapped LLM responses are handled correctly

## Summary by CodeRabbit

* **New Features**
  * Schema inference endpoint now returns a structured schema object (preserving $defs) instead of raw text.

* **Bug Fixes**
  * Improved handling and clearer 422 responses when inference output is invalid or unparsable.
  * Extraction remains resilient to JSON in various formats (e.g., fenced blocks, embedded text).

* **Tests**
  * Added unit tests validating structured schema parsing, $defs preservation, and validation error paths.

### DIFF
--- a/cognee/api/v1/llm/routers/get_llm_router.py
+++ b/cognee/api/v1/llm/routers/get_llm_router.py
@@ -4,6 +4,7 @@ from typing import Any, Dict, List, Annotated
 from fastapi import APIRouter, Depends, File, UploadFile as UF, Form
 from fastapi.responses import JSONResponse
 from pydantic import Field
+from pydantic import ConfigDict, ValidationError
 
 from cognee import __version__ as cognee_version
 from cognee.api.DTO import InDTO, OutDTO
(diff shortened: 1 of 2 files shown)


siillee (changes requested):
This can be handled differently imo.

We can just change the response_model to be a dict, or to be a shape we define, and then extract the model from it. It doesn't have to be `response_model=str`.

siillee (approved):

PR #2745, opened 2026-05-01
fix(temporal): extend TemporalRetriever to filter all ontology types, not just Events

### LINKED ISSUE #2429: TemporalRetriever is event-only, blocks ontology-wide temporal filtering
I found a limitation in temporal retrieval that blocks a real use case.

Current behavior:
Temporal retrieval only returns/filters `Event` objects.

Why this is a problem:
- Temporal filtering can’t be applied to other ontology objects.
- If a query asks for time-bounded entities/relations that are not Events, the retriever path cannot express that.
- This makes temporal awareness much narrower than the ontology model.

Suggested direction:
Use query-time extraction logic similar to:
`time_from, time_to = await self.extract_time_from_query(query)`
and apply that temporal window at object level (not just Event type), so filtering can work across ontology entities.

Expected behavior:
Temporal constraints from user query should be reusable for any retrievable ontology object, not only Events.

badalraj9:
Fixes #2429

  ## Problem

  `TemporalRetriever` extracted time windows correctly via `extract_time_from_query()`,
  but then only retrieved `Event`-type nodes via `collect_events()`. Ontology nodes
  such as `Person`, `Organization`, or `Concept` connected to matching timestamps
  were invisible to temporal retrieval.

  ## Solution

  After `collect_time_ids()` returns timestamp-bounded IDs, `TemporalRetriever` now
  also calls `get_id_filtered_graph_data()` to retrieve all ontology node types
  connected to those timestamps.

  `collect_time_ids()` returns a Cypher-formatted string like `'id1', 'id2'`, while
  `get_id_filtered_graph_data()` expects a Python list. The IDs are now parsed into
  a list before calling the generic graph retrieval path.

  Retrieval/context priority:
  1. Event path, preserved as the primary path
  2. Generic ontology nodes via `get_id_filtered_graph_data()` when no Events are found
  3. Existing triplet fallback when no timestamp IDs are found

  ## Changes

  Only `cognee/modules/retrieval/temporal_retriever.py` was modified.

  ## Testing

  - `uv run ruff check cognee/modules/retrieval/temporal_retriever.py`

## Summary by CodeRabbit

* **Bug Fixes**
  * Vector search now runs only when event groups with events exist; search failures are caught and return empty results instead of errors.
  * Temporal retrieval includes time-filtered graph data and consistently returns graph-derived nodes alongside vector results.
  * Event-group inputs are validated to prevent malformed data from breaking retrieval.

* **Refactor**
  * Context assembly and temporal query flow simplified to prefer available vector results, then graph-derived node/edge descriptions, with clear fallbacks.

### DIFF
--- a/cognee/modules/retrieval/temporal_retriever.py
+++ b/cognee/modules/retrieval/temporal_retriever.py
@@ -187,32 +187,46 @@ class TemporalRetriever(GraphCompletionRetriever):
                 has_relevant_events = any(event_groups)
 
         vector_search_results = retrieved_objects.get("vector_search_results")
-        if has_relevant_events and vector_search_results:
-            top_k_events = await self.filter_top_k_events(relevant_events, vector_search_results)
-            return self.descriptions_to_string(top_k_events)
-
-        if retrieved_objects.get("graph_nodes", None):
-            nodes, edges = retrieved_objects.get("graph_nodes")
-            nodes_by_id = {
-                str(node_id): Node(str(node_id), properties) for node_id, properties in nodes
-            }
-            graph_edges = []
-            for source_id, target_id, relationship_type, properties in edges:
-                source_node = nodes_by_id.get(str(source_id))
-                target_node = nodes_by_id.get(str(target_id))
-                if source_node and target_node:
-                    edge_attributes = dict(properties or {})
-                    edge_attributes["relationship_type"] = relationship_type
-                    graph_edges.append(Edge(source_node, target_node, attributes=edge_attributes))
-
-            if not graph_edges and nodes_by_id:
-                return self.descriptions_to_string(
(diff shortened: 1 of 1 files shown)


lxobr:
Thanks for the contribution, @badalraj9!

After reviewing this, I do not think this change fits the intended scope of `TemporalRetriever`. The retriever is meant to work with graphs built using `temporal_cognify=True`, where the temporal pipeline creates the Event/Timestamp structure used for filtering.

The Event-focused logic is intentional. And for graphs built with normal cognify, there is usually no temporal Event/Timestamp layer to filter on.

I’m going to close this PR for now, but thank you again for taking the time to contribute. We really appreciate search-related improvements, and we’d be happy to review a follow-up with a narrower scope and a clear test case showing the behavior it improves.

PR #2748, opened 2026-05-01
fix(cognee-mcp): use content-addressed filename for text uploads (closes silent-write bug)

### LINKED ISSUE #2747: [Bug]: cognee-mcp cognify(data=str) silently drops all writes after the first one due to hardcoded upload filename
### Summary

When the `cognify` MCP tool is called with a raw text string (`cognify(data="...")`), every call after the very first one is silently dropped at Cognee's storage layer. The MCP tool returns `200` with the message *"Background process launched due to MCP timeout limitations."*, the API returns `200` on `/api/v1/add` and `/api/v1/cognify`, no error appears in any container log, and `cognify_pipeline` runs `INITIATED → STARTED → COMPLETED` in roughly 200 ms — but no new row is ever added to the `data` table or the dataset, the graph is unchanged, and the new content is never searchable.

This is a serious bug for the agent-memory use case (an agent calling MCP `cognify` repeatedly to record notes), where the user has every reason to believe writes are succeeding.

### Reproduction

1. Run a Cognee stack in API mode with `cognee-mcp` connected to a `cognee` API server (any recent `cognee/cognee-mcp:main` image; reproduced with image revision `01810275317f20714cbdf82f11a8b6bfc19377be`, repo digest `sha256:fb004981dbc0773dcaf538dd438f096aac686057e8c619de0a7fab9249e7921c`).
2. From any MCP client, call `cognify` twice with two different text strings, e.g. `cognify(data="first memory")` then `cognify(data="second memory")`.
3. Inspect the dataset:
   ```
   GET /api/v1/datasets/<dataset_id>/data
   ```
4. Observe: only **one** new data row exists (from the first call). The second call, and every subsequent call, returned `200` but produced no new row.
5. Inspect storage:
   ```
   ls /app/data/.data_storage/
   ```
   `data.txt` is present from the first call and is never updated, even though calls 2..N uploaded different content.

### Root cause

`cognee-mcp/src/server.py` (the `cognify` tool background path) calls `cognee_client.add(data)`, which posts a multipart upload to `/api/v1/add` with a hardcoded filename:

```python
files = {"data": ("data.txt", str(data), "text/plain")}
```

On the API side, `save_data_item_to_storage()` → `save_data_to_file()` calls `LocalFileStorage.store(..., overwrite=False)`. `LocalFileStorage.store` does:

```python
if overwrite or not os.path.exists(full_file_path):
    ... write file ...
else:
    return self.storage.is_file(file_path)
```

Because `/app/data/.data_storage/data.txt` already exists from the first call, **the new upload's content is never written to disk**. The ingestion pipeline then opens the existing (old) `data.txt`, computes the original `content_hash`, derives the original `data_id` (via `uuid5(NAMESPACE_OID, content_hash + user.id + tenant.id)`), and the dedup branch in `cognee/tasks/ingestion/ingest_data.py` resolves the upload to the existing data row. No new row is created.

The cognify pipeline that follows then sees all data items already marked `DATA_ITEM_PROCESSING_COMPLETED` and exits in ~200 ms with no graph work, which is why the symptom looks like "cognify is silently doing nothing."

This conflicts with Cognee's documented design that ingested files are content-addressed (renamed to `text_<md5>.txt`); the file-based loader path does perform that rename, but the MCP raw-text path does not, leaving a literal `data.txt` collision.

### Why the failure is invisible

- API and MCP both return `200`.
- MCP logs `Cognify process starting` / `Cognify process finished`.
- API returns `200` on `/api/v1/add` and `/api/v1/cognify`.
- No exception is logged anywhere — the storage layer's "skip" is a normal-return path, not an error.
- The `_task_errors` design pattern in MCP would mask any background-task exception even if one were raised.
- `cognify_status` in MCP cannot be used to verify, because in API mode it tries to open a local sqlite path that does not exist in the MCP container (separate bug, see "Related" below).

So nothing in normal operation surfaces the silent drop; users only catch it by directly listing the dataset's data rows and noticing the count never increases.

### Proposed fix

Change the MCP upload to use a content-addressed filename (matching Cognee's documented `text_<hash>.txt` convention), so that distinct payloads never collide on disk:

```python
import hashlib

content = str(data)
digest = hashlib.md5(content.encode("utf-8")).hexdigest()
files = {"data": (f"mcp_{digest}.txt", content, "text/plain")}
```

(SHA-256 would be preferable security-wise, but MD5 matches Cognee's existing content-hash convention used elsewhere in the codebase.)

An alternative server-side fix would be to make `LocalFileStorage` content-address uploads regardless of supplied filename, or to pass `overwrite=True` from the API ingestion path. The MCP-side filename change is the most contained fix and doesn't require any server changes.

### Related

While investigating this, we also confirmed that `cognify_status` in `cognee-mcp` is broken in API mode: it attempts direct local sqlite access and fails with `sqlite3.OperationalError: unable to open database file` because the MCP container has no `/app/data` mount. In API mode, `cognify_status` should call API endpoints (`/api/v1/datasets/status`, `/api/v1/sync/status`, etc.) instead. Happy to file this as a separate issue if useful.

### Environment

- `cognee/cognee-mcp:main` (revision `01810275`, image created 2026-02-09)
- `cognee/cognee:latest` (API on port 8000)
- `cognee-nginx` fronting MCP on port 8001
- `cognee-neo4j` 5.26-community
- Local Docker volume `cognee_cognee_data` mounted at `/app/data` on the API container; no S3 / cloud storage configured.

ai-nuke:
## Summary

Fixes the silent-write bug reported in #2747.

`cognee-mcp`'s `add()` and `remember()` helpers in `cognee-mcp/src/cognee_client.py` posted every text payload to the Cognee API as multipart with a hardcoded filename of `data.txt`. Cognee's `LocalFileStorage.store()` is called with `overwrite=False`, so once `/app/data/.data_storage/data.txt` exists from the first call, every subsequent upload's content is silently discarded at the storage layer — `LocalFileStorage` simply returns the existing file path without writing the new bytes. The ingestion pipeline then opens the *old* `data.txt`, computes the *old* `content_hash`, derives the *old* `data_id` (via `uuid5(NAMESPACE_OID, content_hash + user.id + tenant.id)`), and the dedup branch in `cognee/tasks/ingestion/ingest_data.py` resolves the upload to the existing data row. No new row is created. The cognify pipeline that follows sees all data items already marked `DATA_ITEM_PROCESSING_COMPLETED` and exits in ~200ms with no graph work.

Both `/api/v1/add` and `/api/v1/cognify` return 200, MCP logs show `Cognify process starting` / `Cognify process finished`, no exception is raised anywhere — so the failure is completely invisible to callers. This is especially severe for the agent-memory use case (agents calling `cognify(data=str)` repeatedly to record notes) where the user has every reason to believe writes are succeeding.

## Fix

Derive the upload filename from an MD5 of the payload's content, matching Cognee's documented `text_<hash>.txt` convention used by the file-based loader path. Distinct payloads now land at distinct on-disk paths, so `LocalFileStorage` writes them, ingestion hashes the actual new content, and a new `data` row is created as expected.

The change is contained to `cognee-mcp/src/cognee_client.py`:
- Added `import hashlib`.
- Added a small helper `_content_addressed_upload()` that returns the multipart files dict with a `mcp_<md5>.txt` filename.
- Replaced both occurrences of `files = {"data": ("data.txt", str(data), "text/plain")}` (in `add()` and `remember()`) with a call to the helper.

`hashlib.md5(..., usedforsecurity=False)` is used to make the security-context explicit (this is a content-addressing convention, not a cryptographic guarantee). MD5 was chosen to match the existing `text_<md5>.txt` convention; SHA-256 would be a reasonable alternative if maintainers prefer.

An alternative server-side fix (make `LocalFileStorage` content-address uploads regardless of supplied filename, or pass `overwrite=True` from the API ingestion path) was considered. The MCP-side filename change is the most contained fix, doesn't require any server changes, and brings MCP behavior in line with the existing file-loader convention.

## Test plan

- [x] `ruff check` passes on the modified file.
- [x] `ruff format --check` passes on the modified file.
- [x] `python -m ast` parses the modified file cleanly.
- [ ] Reproducer from #2747 (call `cognify(data="first")` then `cognify(data="second")`, verify two distinct rows in `/api/v1/datasets/<id>/data`) passes against a Docker image built from this branch.

Closes #2747.

## Summary by CodeRabbit

* **Bug Fixes**
  * Uploads now use content-addressed filenames so distinct payloads are uniquely identified, avoiding conflicts.

* **New Features**
  * Recall API now exposes dataset_ids and several explicit retrieval options as first-class parameters.
  * Remote call payloads now include dataset_ids and optional flags (e.g., node name, only_context, verbose) when present.

* **Tests**
  * Added tests (with telemetry disabled) covering recall forwarding behavior and session+dataset_id search scenarios.

### DIFF
--- a/cognee-mcp/src/cognee_client.py
+++ b/cognee-mcp/src/cognee_client.py
@@ -6,6 +6,7 @@ This module provides a unified interface for interacting with Cognee, supporting
 - API mode: Makes HTTP requests to a running Cognee FastAPI server
 """
 
+import hashlib
 import sys
 from typing import Optional, Any, List, Dict
 from uuid import UUID
@@ -13,6 +14,23 @@ from contextlib import redirect_stdout
 import httpx
 from cognee.shared.logging_utils import get_logger
 import json
+
+
+def _content_addressed_upload(data: Any) -> Dict[str, tuple]:
+    """Build a multipart file payload with a content-addressed filename.
+
+    Cognee's local file storage refuses to overwrite an existing file. Posting
+    every text payload under a fixed filename (e.g. ``data.txt``) therefore
+    causes the second and subsequent uploads to be silently dropped and the
+    original file's content to be re-ingested instead. Using a filename derived
+    from the payload's content hash ensures distinct payloads land at distinct
+    paths, matching Cognee's documented ``text_<hash>.txt`` convention for
+    file-based ingestion.
+    """
+    content = str(data)
+    digest = hashlib.md5(content.encode("utf-8"), usedforsecurity=False).hexdigest()
+    return {"data": (f"mcp_{digest}.txt", content, "text/plain")}
+
 
 logger = get_logger()
 
@@ -96,7 +114,7 @@ class CogneeClient:
         if self.use_api:
             endpoint = f"{self.api_url}/api/v1/add"
 
-            files = {"data": ("data.txt", str(data), "text/plain")}
+            files = _content_addressed_upload(data)
             form_data = {
                 "datasetName": dataset_name,
             }
@@ -374,7 +392,7 @@ class CogneeClient:
         """
         if self.use_api:
             endpoint = f"{self.api_url}/api/v1/remember"
-            files = {"data": ("data.txt", str(data), "text/plain")}
+            files = _content_addressed_upload(data)
             form_data = {"datasetName": dataset_name}
             if custom_prompt:
                 form_data["custom_prompt"] = custom_prompt
--- a/cognee/api/v1/recall/recall.py
+++ b/cognee/api/v1/recall/recall.py
@@ -4,7 +4,7 @@ from uuid import UUID
 
 from debugpy.adapter.sessions import Session
 from pydantic import BaseModel, Field
-from typing_extensions import TypedDict, Unpack
+from typing_extensions import TypedDict
 
 from cognee.infrastructure.databases.cache import SessionAgentTraceEntry, SessionQAEntry
 from cognee.memory.entries import normalize_scope
@@ -315,10 +315,24 @@ async def recall(
     query_type: SearchType | None = None,
     *,
     datasets: list[str] | None = None,
+    dataset_ids: list[UUID] | None = None,
     top_k: int = 10,
     auto_route: bool = True,
     scope: str | list[str] | None = None,
-    **kwargs: Unpack[RecallKwargs],
+    system_prompt: str | None = None,
+    system_prompt_path: str = "answer_simple_question.txt",
+    node_name: list[str] | None = None,
+    node_name_filter_operator: str = "OR",
+    only_context: bool = False,
+    session_id: str | None = None,
+    wide_search_top_k: int | None = 100,
+    triplet_distance_penalty: float | None = 6.5,
+    feedback_influence: float = 0.0,
+    verbose: bool = False,
+    retriever_specific_config: dict | None = None,
+    neighborhood_depth: int | None = None,
+    neighborhood_seed_top_k: int | None = None,
+    user: object | None = None,
 ) -> list[RecallResponse]:
     """Search the knowledge graph for relevant information.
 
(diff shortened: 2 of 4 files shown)


Vasilije1990:
@ai-nuke please check MCP again. I fixed most of the issues and this PR is probably outdated by now.

ai-nuke:
Thanks, I rechecked against `v1.0.5`. The original MCP text-upload filename issue from #2747 does appear fixed now: raw text uploads use content-addressed `text_<md5>.txt` filenames, so this PR is likely outdated for that specific bug.

While verifying the latest release, I found a separate recall/API issue that still exists in `v1.0.5` around `/api/v1/recall` forwarding `dataset_ids`/`user`/other options through loose `**kwargs`. I split that into a clean draft PR here: #2777.

PR #2757, opened 2026-05-03
fix(embeddings): auto-derive embedding_dimensions from provider+model

### LINKED ISSUE #2751: embedding_dimensions defaults to 3072 regardless of embedding_model — breaks every non-3072-dim embedder
## Summary

Cognee's embedding config hard-defaults `embedding_dimensions: int = 3072`. That's correct for OpenAI's `text-embedding-3-large` but wrong for every other model, including all common fastembed models. LanceDB then creates its table with `Vector(3072)` and the first write blows up with a pydantic validation error.

## Repro

- cognee 1.0.3 / cognee-mcp 0.5.4
- `EMBEDDING_PROVIDER=fastembed`
- `EMBEDDING_MODEL=BAAI/bge-small-en-v1.5` (returns 384-dim vectors)
- `VECTOR_DB_PROVIDER=lancedb`
- Run any `cognify` call

## Source

`cognee/infrastructure/databases/vector/embeddings/config.py`, ~line 17:

```python
embedding_dimensions: Optional[int] = 3072
```

## Error

```
1 validation error for LanceDataPoint[str, IndexSchema]
vector
  List should have at least 3072 items after validation, not 384
  [type=too_short, input_value=array([...], dtype=float32)]
```

This kills `extract_chunks_from_documents`, `add_data_points`, and `extract_graph_and_summarize` together — they all hit the same persistence path.

Worth highlighting: **graph extraction itself succeeded** before this — the LLM call returned `nodes_extracted: 23, edges_extracted: 32` cleanly. So it's purely a downstream schema-mismatch on the persistence side, not an LLM issue.

## Suggested fix

Auto-derive `embedding_dimensions` from `embedding_model` when the user hasn't explicitly set it:

- fastembed: `TextEmbedding(model_name).model_description.dim`
- LiteLLM-supported models: lookup in `litellm.model_cost[model]['output_vector_size']` (or equivalent)
- If `EMBEDDING_DIMENSIONS` is explicitly set, honor that.

Alternatively, validate the configured `embedding_dimensions` against the actual embedder at startup and fail fast with a helpful error message instead of silently breaking the pipeline at first write.

## Workaround for affected users

Set env: `EMBEDDING_DIMENSIONS=384` (or whatever your embedder produces).

Happy to PR if useful.

Vasilije1990:
## Summary

Closes #2751.

`EmbeddingConfig.embedding_dimensions` was hard-defaulted to `3072`, which is only correct for OpenAI's `text-embedding-3-large`. Every other embedder (fastembed `BAAI/bge-small-en-v1.5` → 384, OpenAI `text-embedding-3-small` → 1536, ollama/nomic-embed → 768, etc.) hit a `Vector(3072)` / actual-dim mismatch on first write into LanceDB:

```
pydantic.ValidationError: 1 validation error for LanceDataPoint[...]
vector
  List should have at least 3072 items after validation, not 384
```

This kills `extract_chunks_from_documents`, `add_data_points`, and `extract_graph_and_summarize` — they all share the same persistence path. Graph extraction itself succeeds (the LLM call returns nodes/edges fine); it's purely a downstream schema-mismatch on persistence.

## Approach

- Default `embedding_dimensions` to `None` (was `3072`).
- In `model_post_init`, when not explicitly set, look up:
  - **fastembed**: `TextEmbedding.list_supported_models()` (handles both `dim` and `embed_dim` registry schema variants).
  - **everything else (litellm-routed)**: `litellm.model_cost[model]['output_vector_size']`, trying full model key, the prefix-stripped key (`text-embedding-3-large`), and `provider/bare` (`openai/text-embedding-3-large`).
- If still unresolvable, fall back to `3072` with a loud warning telling the user to set `EMBEDDING_DIMENSIONS` explicitly. This preserves the pre-fix behavior for users who already had working configs — the default `openai/text-embedding-3-large` still resolves to 3072 via litellm (verified in tests).
- Explicit `EMBEDDING_DIMENSIONS` env always wins.

## Why auto-derive instead of fail-fast

Failing fast on missing dimensions would be a regression for everyone using the OpenAI defaults today. The auto-derive path covers that case silently and the warning only fires when we genuinely don't recognize the model — which is also exactly when users were already broken before this PR.

## Test plan

8 new unit tests in `cognee/tests/unit/infrastructure/databases/vector/test_embedding_config.py`:

- [x] `test_resolve_openai_text_embedding_3_large` — back-compat: default OpenAI model still maps to 3072
- [x] `test_resolve_openai_text_embedding_3_small` — bare model name lookup (1536)
- [x] `test_resolve_unknown_model_returns_none` — unknown model is "unknown" not "guess"
- [x] `test_resolve_returns_none_for_missing_inputs` — null safety
- [x] `test_resolve_fastembed_uses_registry` — fastembed path covered with mocked import (since fastembed is an optional dep)
- [x] `test_config_auto_resolves_when_dimensions_unset` — full config object end-to-end
- [x] `test_config_honors_explicit_dimensions` — explicit override always wins
- [x] `test_config_falls_back_when_unresolvable` — fallback path doesn't crash

All 8 passing locally:
```
8 passed, 8 warnings in 7.53s
```

## Summary by CodeRabbit

* **New Features**
  * Embedding dimensions are now auto-detected from the active embedding engine and applied to the app instead of a hard default.
  * First-run environment checks detect embedding vector size and synchronize it with configuration.

* **Bug Fixes**
  * If auto-detection fails, a warning is logged and the previous default (3072) is preserved as a fallback.

* **Behavioral Changes**
  * Indexing batching now derives from the active engine and concurrent indexing is slightly reduced (4 workers).
  * Retry logging for embedding engines uses warning-level messages for clearer visibility.
  * Embedding outputs are normalized to native Python lists before further processing.

* **Tests**
  * Added unit and async tests covering detection, fallback, registry resolution, and environment overrides.

[Review Change Stack](https://app.coderabbit.ai/change-stack/topoteretes/cognee/pull/2757?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


NMZivkovic (dismissed):

Vasilije1990:
@siillee unit tests are failing + community tests

dexters1 (changes requested):
Needs to be changed, changes are not compatible with all embedding engines (FastEmbed test is failing). Will fix it tomorrow

cognee/infrastructure/llm/utils.py
```diff
@@ -141,3 +149,34 @@ async def test_embedding_connection() -> None:
...
+
+def determine_embedding_dimensions(detected_dimensions: int) -> None:
+    """
+    Apply embedding-dimension policy using a single already-produced test vector size.
+
+    Rules:
+    - If EMBEDDING_DIMENSIONS is explicitly provided by the user, keep it as source of truth.
+    - Otherwise, sync config and active embedding engine dimensions to detected size.
+    """
+    configured_dimensions_raw = os.getenv("EMBEDDING_DIMENSIONS")
```
dexters1, line 162: We shouldn't use EMBEDDING_DIMENSIONS directly from os env variables and get it from config instead.

This will allow us in the future to support multiple different embedding/llm models in one long running Cognee instance

cognee/infrastructure/databases/vector/embeddings/config.py
```diff
@@ -2,6 +2,62 @@
...
+    Returns the dimension count if we can confidently determine it, or None
+    if the model is unknown to litellm and fastembed. Defensive against
+    optional-dependency and registry-schema variations — never raises.
+    """
+    if not provider or not model:
+        return None
+
+    provider_lower = provider.lower()
+    # Strip "openai/" / "azure/" / etc. prefix from "openai/text-embedding-3-large"
+    bare_model = model.split("/", 1)[1] if "/" in model else model
```
dexters1: Format can also be something like:

baseten/openai/gpt-oss-120b

for LLM/Embedding model names

dexters1: fixed this

cognee/infrastructure/llm/utils.py
dexters1: Will leave it in for now as we only return here and not use the value

dexters1 (approved):

PR #2778, opened 2026-05-04
test: add Docker remember recall e2e

Vasilije1990:
Adds an opt-in Docker E2E pytest that starts a built Cognee image, posts data to /api/v1/remember, queries /api/v1/recall with graph scope, and validates graph results contain the expected fact. Updates the reusable backend Docker workflow to run in two explicit phases: Part 1 builds the Docker image from the checked-out commit, and Part 2 runs this test against that exact image tag.

## Summary by CodeRabbit

* **Tests**
  * Added an opt-in Docker end-to-end smoke test that launches a real container, verifies service health, and validates the remember → graph recall flows (chunk and graph completion modes) and aggregated result content.

* **Chores**
  * CI workflow updated to build, validate Docker images and run targeted Docker E2E tests as part of automated pipelines.


siillee (dismissed):

.github/workflows/backend_docker_build_test.yml
```diff
@@ -2,25 +2,69 @@ name: build test | Docker image
 
 on:
   workflow_call:
```
dexters1, line 4: Should we have the test run on PR towards dev/main? Or as part of the test suite for Docker tests

PR #2792, opened 2026-05-07
fix(pghybrid): guard None metadata in add_nodes_with_vectors and clean up collection-key construction

adamhunter:
## Summary

Two follow-ups on coderabbitai feedback against PR #2587 that were marked "addressed" but never fixed. Same function (`add_nodes_with_vectors`), same review thread, both small — bundling.

### Bug 1 (live) — `dp.metadata` not guarded against `None`

```python
for field_name in dp.metadata.get("index_fields", []):
    # AttributeError: 'NoneType' object has no attribute 'get'
```

`metadata` is `Optional` on subclasses of `DataPoint`. Any DataPoint with `metadata=None` reaching `add_nodes_with_vectors` crashes the call. The sibling code at [`cognee/tasks/storage/index_data_points.py:35`](https://github.com/topoteretes/cognee/blob/dev/cognee/tasks/storage/index_data_points.py#L35) already has the canonical guard:

```python
if not hasattr(data_point, "metadata") or not data_point.metadata:
    continue
```

**Fix:** mirror that guard at the top of the per-DataPoint loop. Missing metadata silently skips vector indexing (matching `index_data_points` semantics); the `graph_node` insert still runs, so the node row is preserved.

### Bug 2 (latent code smell) — lossy collection-key round-trip

The vector grouping built keys as `f"{type_name}_{field_name}"` and recovered them via `rsplit("_", 1)`:

```python
collection = f"{type(dp).__name__}_{field_name}"   # "CodeGraphEntity_source_code"
...
await self._vector.create_vector_index(
    collection.rsplit("_", 1)[0],   # "CodeGraphEntity_source"
    collection.rsplit("_", 1)[1],   # "code"
)
```

For multi-underscore field names like `source_code`, the recovered split looks wrong on its face. **However, no observable corruption occurs in practice.** `PGVectorAdapter.create_vector_index` is:

```python
async def create_vector_index(self, index_name, index_property_name):
    await self.create_collection(f"{index_name}_{index_property_name}")
```

The args are immediately re-joined with the same separator. F-string concatenation across an underscore is associative — `(a + "_" + b) + "_" + c == a + "_" + (b + "_" + c)` — so the final table name `"CodeGraphEntity_source_code"` is identical regardless of where the rsplit boundary fell.

But the round-trip is a footgun. Any future change to `_validate_table_name`, `create_collection`, or how `create_vector_index` joins its args would silently break search routing because the rsplit-then-rejoin assumes both sides use exactly the same join. **This PR removes that fragile assumption.**

**Fix:** restructure `vector_groups` from `Dict[str, ...]` (joined-string keys) to `Dict[Tuple[str, str], List[DataPoint]]` (tuple keys). Reconstruct the joined string only at the SQL identifier step in `_validate_table_name`, where it's immediately validated and quoted.

## Backwards compatibility

- **API:** unchanged. `vector_groups` is a private local variable; no method signatures touched.
- **Behavior:** unchanged for any DataPoint that worked before. The `metadata=None` case used to crash and now skips silently — strict improvement; no caller can have depended on the crash.
- **Data layout:** **unchanged.** Same pgvector tables are created and written to. No migration required. The new `test_collection_key_round_trip_is_backwards_compatible` test pins this invariant — if anyone ever changes `PGVectorAdapter.create_vector_index` in a way that breaks the f-string associativity, that test fails and surfaces the impact.

## Dedupe search

Before opening:

| Query | Hits |
| --- | --- |
| `gh pr list --search "rsplit"` | only #2587 (the original feature) |
| `gh pr list --search "metadata index_fields"` | nothing touching the pghybrid adapter |
| `gh pr list --search "2587"` | #2791 (my prior batching fix), #2587, no others on these lines |
| `gh pr list --search "pghybrid collection"` | only #2587, #2755 |

No competing PR.

## Test plan

New unit tests at `cognee/tests/unit/infrastructure/databases/test_postgres_hybrid_collection_key.py`. Same fake-adapter pattern as the batching tests already in this directory; both use `pytest.importorskip("asyncpg")` / `pytest.importorskip("pgvector")` so they're skipped without the postgres extra.

Four cases:

1. **Single-underscore field sanity** — `metadata={"index_fields": ["name"]}` → `create_vector_index("_SingleNode", "name")`. Sanity check, passes before and after the fix.
2. **Multi-underscore field preserves full name** — `metadata={"index_fields": ["source_code"]}` → asserts `create_vector_index` called with `("_CodeGraphEntity", "source_code")`, NOT `("_CodeGraphEntity_source", "code")`. **Failed before fix** with the corrupted args; passes after. (Final table name was identical pre- and post-fix due to f-string associativity; this test verifies the args themselves are now correct, eliminating the latent footgun.)
3. **`metadata=None` does not crash** — DataPoint with `metadata: Optional[dict] = None`. **Failed before fix** with `AttributeError: 'NoneType' object has no attribute 'get'`; passes after, with `create_vector_index` not called and the `graph_node` INSERT still executed.
4. **Backwards-compat invariant** — calls `PGVectorAdapter.create_vector_index` directly with both the pre-fix and post-fix arg shapes; asserts both produce the same final collection name. Pins the f-string associativity property the rest of the backwards-compat claim rests on.

TDD red→green confirmed: ran tests #1–#3 against the unmodified adapter, observed exactly the expected failures, applied the fix, then added test #4 to pin the backwards-compat invariant.

### Test evidence

```
$ .venv/bin/pytest cognee/tests/unit/infrastructure/databases/test_postgres_hybrid_collection_key.py cognee/tests/unit/infrastructure/databases/test_postgres_hybrid_batching.py -v
======================== 10 passed, 8 warnings in 1.7s =========================

$ .venv/bin/ruff check cognee/infrastructure/databases/hybrid/postgres/adapter.py cognee/tests/unit/infrastructure/databases/test_postgres_hybrid_collection_key.py
All checks passed!

$ .venv/bin/pre-commit run --files cognee/infrastructure/databases/hybrid/postgres/adapter.py cognee/tests/unit/infrastructure/databases/test_postgres_hybrid_collection_key.py
Trim Trailing Whitespace.................................................Passed
Fix End of Files.........................................................Passed
ruff (legacy alias)......................................................Passed
ruff format..............................................................Passed
```

`ty check` on the touched files reports the same 7 pre-existing diagnostics on this branch as on `dev` (verified by stash + check + unstash) — no new type errors introduced.

## Not in scope

- Same-string `(type, field)` collisions (e.g. `("Foo_bar", "baz")` vs `("Foo", "bar_baz")` would produce the same joined string). Pre-existing, separate from coderabbit's flag.
- `add_edges_with_vectors` (already fixed in #2791).

### DIFF
--- a/cognee/infrastructure/databases/hybrid/postgres/adapter.py
+++ b/cognee/infrastructure/databases/hybrid/postgres/adapter.py
@@ -278,20 +278,27 @@ class PostgresHybridAdapter(GraphDBInterface, VectorDBInterface):
         await self._graph.initialize()
         now = datetime.now(timezone.utc)
 
-        # Group data points by (type_name, field_name) for vector indexing
-        vector_groups: Dict[str, List[Tuple[DataPoint, str]]] = {}
+        # Group data points by (type_name, field_name) for vector indexing.
+        # Keying by tuple (not by joined string) avoids the lossy rsplit("_", 1)
+        # round-trip that would corrupt multi-underscore field names like
+        # ``source_code`` into ``("..._source", "code")``.
+        vector_groups: Dict[Tuple[str, str], List[DataPoint]] = {}
         for dp in data_points:
+            # ``metadata`` is Optional on DataPoint; mirror the guard at
+            # ``index_data_points.py:35`` so points lacking metadata are
+            # silently skipped instead of crashing on ``None.get``.
+            if not hasattr(dp, "metadata") or not dp.metadata:
+                continue
+            type_name = type(dp).__name__
             for field_name in dp.metadata.get("index_fields", []):
                 if getattr(dp, field_name, None) is None:
                     continue
-                collection = f"{type(dp).__name__}_{field_name}"
-                vector_groups.setdefault(collection, [])
-                vector_groups[collection].append((dp, field_name))
+                vector_groups.setdefault((type_name, field_name), []).append(dp)
 
-        # Embed all texts grouped by collection
-        embeddings_by_collection: Dict[str, List[Tuple[DataPoint, List[float], str]]] = {}
-        for collection, items in vector_groups.items():
-            valid_items = [(dp, getattr(dp, field_name, None)) for dp, field_name in items]
+        # Embed all texts grouped by (type_name, field_name)
+        embeddings_by_group: Dict[Tuple[str, str], List[Tuple[DataPoint, List[float], str]]] = {}
+        for (type_name, field_name), dps in vector_groups.items():
+            valid_items = [(dp, getattr(dp, field_name, None)) for dp in dps]
             valid_items = [(dp, t.strip() if isinstance(t, str) else t) for dp, t in valid_items]
             valid_items = [(dp, t) for dp, t in valid_items if t is not None]
             if not valid_items:
@@ -301,15 +308,13 @@ class PostgresHybridAdapter(GraphDBInterface, VectorDBInterface):
             vectors = []
             for i in range(0, len(texts), batch_size):
                 vectors.extend(await self._vector.embed_data(texts[i : i + batch_size]))
-            embeddings_by_collection[collection] = [
+            embeddings_by_group[(type_name, field_name)] = [
                 (dp, vec, t) for (dp, t), vec in zip(valid_items, vectors)
             ]
 
         # Ensure vector collection tables exist
-        for collection in vector_groups:
-            await self._vector.create_vector_index(
-                collection.rsplit("_", 1)[0], collection.rsplit("_", 1)[1]
-            )
+        for type_name, field_name in vector_groups:
+            await self._vector.create_vector_index(type_name, field_name)
 
         # Build all rows in Python, then one INSERT per table inside the transaction.
         core_keys = {"id", "name", "type"}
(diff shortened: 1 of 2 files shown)


adamhunter:
Closing — investigated further and concluded both fixes are speculative cleanup, not real bug fixes against current production paths.

**Bug 2 (latent footgun, the rsplit cleanup):** Backwards-compatible by accident. `PGVectorAdapter.create_vector_index` immediately re-joins its two args with `f"{a}_{b}"`, and f-string concatenation across the underscore is associative, so the lossy `rsplit("_", 1)` round-trip produces the same final collection name as the post-fix tuple-keyed code. No observable corruption today; only a footgun if anyone ever changes the join.

**Bug 1 (the metadata=None guard):** Not actually reachable. Every DataPoint subclass in the codebase declares `metadata: dict = {...}` with a non-None default; pydantic on the base class rejects `metadata=None` construction; nothing in `convert_node_to_data_point` or any extraction path can produce a DataPoint with `metadata=None`. Triggering the crash requires writing a custom subclass with `metadata: Optional[dict] = None`, which is what the test had to do. Likely why the original PR #2587 author marked the coderabbit comment "addressed" without code changes.

Closing as draft because this is the case the draft state is for. Will revive if either becomes reachable in practice.

Thanks for the time.

PR #2811, opened 2026-05-11
fix: agent display name leaks user ID

LStromann:
## Summary
- Agent display names in `/v1/activity/agents` included the user ID suffix (e.g. "my open claw bot+85c3f9ec 3edf 446a 80ba" instead of "my open claw bot")
- Root cause: the old code used `rsplit("-", 1)` to parse the agent name from the internal email, which split on hyphens inside the UUID
- Changed the internal email separator from `+` to `++` (users can't include `++` in names) and updated parsing to split on `++`
- Added backwards compatibility: parses `++` first, falls back to legacy `+` for existing OSS agents

## Test plan
- [ ] Create a new agent with spaces in the name (e.g. "My Open Claw Bot") and verify the display name shows correctly without any UUID fragment
- [ ] Verify existing agents created with old `+` separator still display correctly (backwards compat)

## Release Notes

* **Bug Fixes**
  * Improved agent identity parsing for email-based agent accounts to recognize new local-part formats and reliably extract agent type and identifier, while preserving sensible fallbacks for legacy formats. This reduces misidentified agents and improves agent lookup accuracy.

[Review Change Stack](https://app.coderabbit.ai/change-stack/topoteretes/cognee/pull/2811)


nfnt (commented):
It looks like the PR description doesn't match the implementation. This is splitting at `+`, not at `++`, let's update the description.

Furthermore, Coderabbit mentions backwards-compatibility here. Is this something we need to consider, or can we ignore it, as this is mostly a nuisance when using existing agent identifiers which still use `-` as separator?

Unrelated to your changes, be we should add unit tests for these routers soon-ish, as some of them contain quite some complexity.

LStromann:
Updated — thanks for the review.

**Description**: Fixed to match the implementation (uses `+`, not `++`).

**Backwards compatibility**: Added a fallback for legacy agents that use `-` throughout. The parsing now:
1. Checks for `+{user_id}` (current format) → splits on last `+`
2. Falls back to `rsplit("-", 1)` and validates the suffix looks like a hex UUID fragment (≥8 hex chars) — this handles old agents without incorrectly chopping hyphenated names
3. If neither matches, uses the full local part as the display name

nfnt (approved):
LGTM!
I'm not a fan of us using email addresses to distinguish agents in this way, but that refactoring is something to be done at a later point.

PR #2813, opened 2026-05-11
feat: add possibility of custom rollback handlers to run_tasks. add c…

siillee:
…ognify rollback handler

## Description

This PR adds rollback mechanisms to pipelines. In case of custom pipelines, one can define a custom rollback mechanism as well, and pass it to run_tasks, if they want to have it. It is optional.
A rollback mechanism here entails deleting entries in various databases, in case of a crash in the middle of execution. Since we use multiple different databases, partial states are possible, and we want to avoid them.

This PR introduces a rollback mechanism for the cognify pipeline only (at the time of writing this description; add pipeline might also get a rollback here). This rollback is called at startup also, in case of hard crashes.

This PR adds the `pipeline_run_id` column to nodes and edges in the relational database. This PR also changes the way `pipeline_run_id` is generated, making it unique for every run of a pipeline, instead of it being derived from `pipeline_id` and `dataset_id`.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

## Summary by CodeRabbit

* **New Features**
  * Per-run identifiers added across pipelines and ingestion so run-scoped artifacts can be tracked and cleaned.
  * Pluggable async rollback handlers with startup invocation to recover stale STARTED runs after a configurable age.
  * Durable push to storage occurs before marking runs complete.

* **Bug Fixes**
  * Rollback isolates deletions to the failed run/dataset and preserves legacy/other-run artifacts.

* **Chores**
  * Migration and model/indexing support for a nullable pipeline-run ID field.

* **Tests**
  * New integration and unit tests covering rollback, recovery, idempotency, and isolation.


Vasilije1990:
## Automated test + rollback-logic review

Ran the full new test suite locally on this branch and did an adversarially-verified review of the rollback/recovery logic. Sharing results in case they're useful before merge.

### ✅ Tests: all green (34 passed)

Each file run in an isolated data root; the integration test self-mocks the LLM and uses local ladybug/lancedb, so no external deps.

| Test file | Result |
|---|---|
| `unit/modules/cognify/test_rollback.py` | 2 passed |
| `unit/modules/cognify/test_recovery.py` | 2 passed |
| `unit/modules/pipelines/test_run_tasks_rollback.py` | 1 passed |
| `unit/tasks/storage/test_add_data_points.py` | 24 passed |
| `integration/tasks/test_cognify_rollback_recovery.py` | 5 passed (41.8s) |

### 🔎 Logic review: 6 correctness gaps the tests don't cover

The happy path is well covered, but the rollback/recovery **design** has gaps. Ordered by severity (each verified against source):

#### HIGH

**1. Recovery never resets pipeline status → cognify-only retry silently skipped** — `recovery.py:57-86`
After startup recovery deletes a stale run's nodes/edges, it never writes a new `PipelineRun` status row, so the latest status stays `DATASET_PROCESSING_STARTED`. A subsequent `cognify()` hits `check_pipeline_run_qualification` → "already being processed" and skips — but the graph was just emptied. Recoverable only via a full `add()→cognify()` (which calls `reset_dataset_pipeline_run_status`), not the obvious cognify-only retry. Suggested fix: call `reset_pipeline_run_status` from recovery (and/or the rollback handler).

**2. No liveness check → a live run on another worker gets rolled back** — `recovery.py:38-55`
"Stale" is defined purely as status `STARTED`, with no age threshold / heartbeat / owner / lock. On a multi-replica or rolling deploy sharing a DB (supported with Postgres backends), a booting replica rolls back nodes a *live* replica is still writing → silently incomplete graph. The default container runs gunicorn `-w 1`, so single-pod deployments are safe; the risk is specific to shared-DB multi-process topologies. Suggested fix: add an age threshold or a heartbeat/lease before treating a `STARTED` run as abandoned.

**3. `push_to_s3` failure after completion destroys completed data** — `run_tasks.py:157-200`
On the success path, `yield PipelineRunCompleted(...)` happens *before* `graph_engine.push_to_s3()` / `relational_engine.push_to_s3()`, all inside the same `try`. With `STORAGE_BACKEND=s3`, a transient S3 error after a fully successful cognify falls into the new `except` block → the rollback handler deletes valid data, **and** the consumer receives both `PipelineRunCompleted` and `PipelineRunErrored` for one run. Suggested fix: move the S3 push before yielding completion, or run it in its own try that doesn't trigger rollback.

#### MEDIUM

**4. `on_conflict_do_nothing` keeps the first run's `pipeline_run_id`** — `upsert_nodes.py` / `upsert_edges.py`
The relational row id (`uuid5` of tenant+user+dataset+data_id+node.id) deliberately excludes `pipeline_run_id`, and the upsert is `on_conflict_do_nothing(index_elements=["id"])`. So a re-cognify leaves existing rows tagged with run1's id while `add_data_points` re-writes the graph under run2 — run2's rollback then filters by `pipeline_run_id == run2` and misses them (under-deletes). Default `incremental_loading=True` short-circuits before `add_data_points`, so this manifests with `incremental_loading=False` or after a swallowed run1 rollback failure.

**5. `extract_dlt_fk_edges` writes untagged graph nodes/edges** — `extract_dlt_fk_edges.py:223-234`
This task (Task #5 in the default pipeline, after `add_data_points`) calls `graph_engine.add_nodes/add_edges` directly with no `pipeline_run_id` and no relational ledger row, so the DLT schema nodes + FK edges are invisible to the rollback handler and orphaned permanently on failure/recovery. Bounded by deterministic `uuid5` ids (re-runs reuse ids), but the rollback contract genuinely fails for these artifacts.

#### LOW

**6. ERRORED runs re-rolled-back on every restart** — `recovery.py:43-49`
Recovery includes `DATASET_PROCESSING_ERRORED` runs, which were already rolled back inline at error time (`run_tasks.py:172-190`); combined with #1 (status never advances), the same run is re-selected and the handler re-runs on every startup, forever. Idempotent (no data damage — the deletes short-circuit on empty result sets), just repeated startup work and misleading "recovery completed" logs.

### Suggested test additions
- "re-cognify after startup recovery succeeds" (covers #1)
- "a live/in-progress run is not rolled back by a concurrent startup" (covers #2)
- S3-backend completion where the push fails after `PipelineRunCompleted` (covers #3)

---
🤖 Generated with [Claude Code](https://claude.com/claude-code) — automated test run + adversarially-verified static review. Findings are pointers; please sanity-check against intended behavior.

Vasilije1990 (approved):

PR #2826, opened 2026-05-13
perf(subprocess): id-based concurrent RPC for LanceDB harness

GraDKh:
## Summary

- Replaces the session-wide `threading.Lock` in `cognee_db_workers/harness.py` with per-request ids + a reader thread routing responses to per-call `concurrent.futures.Future`s — sync and async callers share the same primitive.
- Worker dispatches async handlers via `asyncio.create_task` under a bounded semaphore (`SUBPROCESS_WORKER_MAX_INFLIGHT`, default 16); sync handlers stay inline so **Kuzu is unaffected** (sync `.call()` path + non-thread-safe `Connection`).
- Core invariant: the reader's `finally` always runs `_fail_all_pending` — every path that ends the session (crash, shutdown, respawn, reader-internal exception) drains the pending dict and propagates `SubprocessTransportError` to every waiter; caller `try/finally` pops on cancel/timeout, so no futures leak.

## Why

The single `_rpc_lock` made concurrent `await session.call_async(...)` calls queue at the lock even when the underlying `lancedb.AsyncConnection` could overlap. Under 10 concurrent users, `/api/v1/add` was 6.4× slower in subprocess mode than embedded (1371 ms vs 213 ms avg), with a 9.3 s p95 — pure lock-queue wait.

## Measurements

Same locust workload (`cognee/tests/performance/locust_performance_analysis.py`, 10 users / 1/s spawn / 5 min), `*_SUBPROCESS_ENABLED=true`:

| Endpoint | Baseline subprocess | Embedded (intra-process) | **After** | Δ | After ÷ embedded |
|---|---:|---:|---:|---:|---:|
| `/api/v1/add` avg | 1371 ms | 213 ms | **246 ms** | **5.57×** | 1.15× |
| `/api/v1/add` p95 | 9300 ms | 370 ms | **520 ms** | **17.9×** | 1.41× |
| `/api/v1/cognify` avg | 33 089 ms | 36 263 ms | **27 291 ms** | **1.21×** | 0.75× (beats embedded) |
| `/api/v1/search` avg | 527 ms | 425 ms | 550 ms | ~flat | 1.29× |
| Throughput | 0.57 req/s | 0.56 req/s | **0.69 req/s** | **+21 %** | — |

The 33 ms residual `/add` gap to embedded is IPC marshalling (pickle on `multiprocessing.Queue` + reader hop), irreducible without changing the transport. The locust patch in this PR (`TimeoutError` added to `wait_for_server`'s caught exception set) makes subprocess mode actually startable under that harness — without it, the gevent-wrapped `urlopen` timeout escaped the 240 s deadline loop.

## Should we do the same for Kuzu next?

Probably not, based on the table above. Kuzu already inherits the main-process wins (no more `_rpc_lock`, reader routing) from this PR. Worker-side concurrency for Kuzu would need a `kuzu.Connection` pool + a read/write protocol split + a write semaphore (Kuzu serializes writes at the DB level anyway) — for an estimated **~50–100 ms upside on `/search`** and basically nothing on `/add` or `/cognify`. The full math is in `cognee/tests/performance/results/after_lancedb_concurrent/SUMMARY.md` locally (gitignored).

## Test plan

- [x] `pytest cognee/tests/unit/infrastructure/databases/test_subprocess_session_failures.py` — 20/20 pass (one assertion adjusted: per-call timeout no longer marks the whole session closed — semantic change inherent to concurrent RPC)
- [x] `pytest cognee/tests/unit/infrastructure/databases/test_subprocess_session_concurrent.py` — 8/8 new tests pass: concurrent completion, slow-doesn't-block-fast, crash-with-N-pending fails every waiter + clears `_pending`, sync/async interleaving, caller cancellation pops registry, per-call timeout cleans registry + session stays open, shutdown drains in-flight, late response after timeout silently dropped
- [x] Pre-commit clean (ruff + ruff-format)
- [x] LanceDB integration script `python cognee/tests/test_lancedb.py` failure on `len(result) > 15` reproduces with `VECTOR_DB_SUBPROCESS_ENABLED=false` → pre-existing LLM-extraction count flakiness, not introduced here
- [ ] CI

## Summary by CodeRabbit

* **New Features**
  * Concurrent request handling with per-request identifiers for accurate routing
  * Improved worker protocol and session lifecycle for safe respawn, shutdown, and restart

* **Bug Fixes**
  * Per-call timeouts no longer close the entire session
  * Liveness checks made more robust; broader error handling when probing servers
  * Late or unsolicited responses dropped without affecting in-flight calls

* **Performance**
  * Slow operations no longer block fast ones

* **Tests**
  * Extensive concurrency and failure-mode test suite added; failure-path tests updated

---


GraDKh:
@copilot review

GraDKh:
@coderabbitai full review

PR #2830, opened 2026-05-13
fix: stop subprocess watchdog from killing workers when parent is PID 1

GraDKh:
## Summary

Fixes a startup failure where cognee SDK calls in Docker (Kuzu + LanceDB subprocess mode) fail with `SubprocessTransportError: Subprocess init timed out after 60.0s`. Root cause is a logic bug in the worker's parent-liveness watchdog, not a Docker resource limit or permissions issue.

### What was happening

Every spawned worker started a daemon thread that polled `os.getppid()` and self-`os._exit(0)`'d when `current_ppid != original_ppid or current_ppid == 1`. The `== 1` clause was intended to detect "reparented to init after parent death" — but in container deployments where cognee runs as the container entrypoint, the worker's **legitimate** parent is PID 1. So the watchdog killed every worker on its very first poll, the parent never received the READY sentinel, and `wait_for_ready` sat blocked for 60s on a queue read.

Signature in the failing log: `proc=<SpawnProcess pid=42 parent=1 stopped exitcode=0 daemon>` — clean self-exit, not a crash.

### Changes

- **`set_pdeathsig()` now returns `bool`.** `run_worker_loop` skips the polling watchdog whenever the kernel-side pdeathsig is armed (always the case on Linux). `pdeathsig` fires on actual parent death regardless of pid, so it is immune to the PID-1 case.
- **`start_parent_liveness_watchdog` drops the `current_ppid == 1` clause.** Belt-and-suspenders for macOS/Windows fallbacks where pdeathsig isn't available. The `current_ppid != original_ppid` check alone covers reparenting correctly.
- **`wait_for_ready` polls for early child death** instead of a single 60s blocking queue read. Every init-failure error now embeds `pid=… exitcode=… alive=…` so future regressions of this shape are diagnosable from one log line.
- **Post-death drain race fixed.** When a worker `put`s an error Response and immediately dies, the producer-side feeder thread may not have flushed bytes onto the pipe before death. `_drain_response_after_death` polls for a bounded window (500 ms) and also catches `EOFError`/`OSError` from pipe corruption, so legitimate worker-side error text isn't dropped.
- **libc resolved portably.** `set_pdeathsig` now uses `ctypes.util.find_library('c')` with a `CDLL(None)` fallback instead of hard-coding `libc.so.6`. Without this, musl-based distros (Alpine) and other non-glibc Linuxes would silently fall through to the polling watchdog and re-expose the original PID-1 bug.
- **Signal-decoded exitcodes in diagnostics.** Negative exitcodes now render as `exitcode=-9 (SIGKILL — likely OOM kill or \`docker kill\`)`, `exitcode=-11 (SIGSEGV — native crash — check faulthandler dump …)`, etc. Hints attached for SIGKILL/SIGSEGV/SIGABRT/SIGTERM/SIGBUS/SIGFPE; other signals get the bare name.
- **`faulthandler` enabled in the worker.** A native crash inside Kuzu/LanceDB now dumps the Python traceback that led there to stderr before the process dies, so segfaults stop being a bare `exitcode=-11` mystery. Opt-out via `SUBPROCESS_FAULTHANDLER_DISABLED=1` for debugger workflows.

Same harness is shared by Kuzu and LanceDB workers, so a single fix covers both.

## Test plan

- [x] Unit tests for every new path (timeout, clean-exit-no-ready, drain-race, signal-decoding, faulthandler gating, pdeathsig-skips-watchdog) — 7 init-path tests pass locally
- [x] `ruff format` + `ruff check` clean
- [x] `Code Quality`, `check-dco`, `CodeQL` CI jobs green
- [x] Existing unit-test regex assertions (`"init timed out"`, `"init failed"`) remain substrings of the new messages
- [ ] Re-run the failing recall in the cloud Docker container — should now succeed
- [ ] Optional follow-up (deploy hygiene, not in this PR): adopt `tini`/`--init` so cognee never runs as PID 1
- [ ] Optional follow-up (separate PR): capture worker stderr into a ring buffer the harness embeds in init-failure errors, so the faulthandler dump appears in the parent's exception rather than only in container logs

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable parent-process monitoring (Linux-only behavior where applicable) and reduced false watchdog exits
  * Faster detection and clearer reporting when workers exit before signaling readiness
  * Improved diagnostics for startup/termination failures, including clearer exit-code and signal information

* **Tests**
  * New regression tests covering early exits, queued init errors, signal-terminated workers, and parent-monitoring behavior

[Review Change Stack](https://app.coderabbit.ai/change-stack/topoteretes/cognee/pull/2830)

---

### DIFF
--- a/cognee/tests/unit/infrastructure/databases/test_subprocess_session_failures.py
+++ b/cognee/tests/unit/infrastructure/databases/test_subprocess_session_failures.py
@@ -22,6 +22,7 @@ from cognee_db_workers.harness import (
     Response,
     SubprocessSession,
     SubprocessTransportError,
+    _describe_exitcode,
     run_worker_loop,
     spawn_without_main,
 )
@@ -124,6 +125,17 @@ def _put_error_then_die_worker(req_q, resp_q):
     # ``Queue`` feeder thread's flush against process death. Exercises
     # the post-death drain path in ``wait_for_ready``.
     resp_q.put(Response(error="init-side boom"))
+
+
+def _signal_killed_worker(req_q, resp_q):
+    # Send ourselves SIGTERM before READY to exercise the signal-decoding
+    # path in ``_describe_exitcode``. We use SIGTERM rather than SIGKILL so
+    # the negative exitcode is portable (SIGKILL=-9 on Linux but the test
+    # body inspects whichever signal name appears — see assertions below).
+    import os as _os
+    import signal as _signal
+
+    _os.kill(_os.getpid(), _signal.SIGTERM)
 
 
 def _wait_for_worker_exit(session, timeout: float = 5.0) -> None:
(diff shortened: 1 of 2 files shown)


GraDKh:
@coderabbitai review

dexters1 (approved):

PR #2857, opened 2026-05-18
fix: guard against None message in LLM adapter transcription paths

qizwiz:
## What

Fix three LLM adapter files where `choices[0].message` could be `None`:

- `cognee/infrastructure/llm/structured_output_framework/litellm_instructor/llm/generic_llm_api/adapter.py` — add guard before `TranscriptionReturnType` construction
- `cognee/infrastructure/llm/structured_output_framework/litellm_instructor/llm/ollama/adapter.py` — extend existing guard to also check `message is None`
- `cognee/infrastructure/llm/structured_output_framework/litellm_instructor/llm/mistral/adapter.py` — extend existing truthy check to also guard against `message is None`

## Why

Some providers (e.g. Gemini via OpenAI-compatible endpoint on PROHIBITED_CONTENT) return HTTP 200 with `choices[0].message = null` rather than an HTTP error code, causing `AttributeError: 'NoneType' object has no attribute 'content'`. The existing guards only check that `choices` is non-empty, not that `choices[0].message` is also non-None.

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened validation of AI responses to reject missing or empty choice arrays.
  * Added checks to detect and reject null or absent message objects before reading content, preventing runtime errors.
  * Improved error handling so transcription and structured-output flows fail gracefully when responses lack usable message data.

[Review Change Stack](https://app.coderabbit.ai/change-stack/topoteretes/cognee/pull/2857?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


cognee/infrastructure/llm/structured_output_framework/litellm_instructor/llm/generic_llm_api/adapter.py
```diff
@@ -287,6 +287,8 @@ async def create_transcript(self, input: str) -> TranscriptionReturnType:
             max_retries=self.MAX_RETRIES,
         )
 
+        if not response.choices or response.choices[0].message is None:
+            raise ValueError("LLM returned empty or filtered response")
```
dexters1: Can we specify the ValueError to be something like:

"Transcription failed. No response received."

So it is more understandable what the issue was and where it happened

```suggestion
            raise ValueError("Transcription failed. No response received.")
```

dexters1 (approved):

PR #2876, opened 2026-05-21
COG-5963 perf(llm): cache openai-python type-introspection helpers

GraDKh:
## Summary

openai-python rebuilds its `ChatCompletion` response tree on every HTTP response, repeatedly calling `get_origin` / `get_args` / `is_annotated_type` / `is_literal_type` against the same set of static types. At cognify scale (200 docs ≈ 600+ structured-output calls), this introspection accounts for **~20% of CPU**.

This PR wraps the four helpers in `functools.lru_cache(maxsize=4096)` and rebinds them at every known openai import site — both the source modules and the names already imported into `_models`, `_response`, `_legacy_response`, and `_base_client`. Patching only the source module is insufficient because most call sites use `from ._compat import get_origin`, which captures the name at import time.

The cache is installed once from `cognee.infrastructure.llm.__init__` and is idempotent. Unhashable arguments fall back to the uncached function so the cache can never block a real LLM response.

## Measurement

Cognify on 200 generated documents (`time.process_time()` + `resource.getrusage`, no sampling profiler, same workload across runs):

| | baseline | patched | Δ |
|---|---:|---:|---:|
| cognify CPU (user+sys) | **93.85s** (avg of 2 runs) | **67.70s** | **-26.15s (-28%)** |

Two baseline runs measured 93.32s and 94.38s (within 1.1%), so the saving is reliable.

A pyinstrument-style profile shows the openai helpers (`get_origin` / `get_args` / `is_*_type`) disappearing entirely from the top-30 self-time table; the remaining work is unavoidable Pydantic value validation.

## Trade-offs / limitations

- Module-side-effect at import time. Justified: the patch must run before any openai SDK call, and `cognee.infrastructure.llm` is the natural anchor. Cost is bounded — one-time, idempotent.
- The list of rebind targets is hard-coded against openai-python's current module layout. A future openai-python refactor could introduce new call sites this misses; the failure mode is "cache miss → uncached call", i.e. today's behaviour. Cache hits return `cache_info()` for observability.

## Test plan
- [x] `ruff check` + `ruff format --check`
- [x] Smoke test: importing `cognee.infrastructure.llm` installs the cache; `install()` is idempotent; `cache_info()` is exposed on the rebound helpers
- [x] End-to-end cognify of 200 docs completes successfully and shows the predicted CPU reduction
- [ ] CI

## Summary by CodeRabbit

* **Chores**
  * Adds an idempotent runtime caching layer for OpenAI type-introspection helpers that applies automatically.
  * Improves latency by reducing redundant introspection work.
  * Increases stability with graceful fallback/no-op behavior when caching cannot be applied.


Vasilije1990:
Updated the branch: merged current `dev` (was ~1,240 commits behind; the PR's two files merged cleanly, no conflicts).

**Why CI was red (June 3 run):** none of the failures trace to this change —
- `test_search_db.py::test_e2e_graph_vector_consistency` (`assert 160 == 20`) ran through repeated `RateLimitError: 429` retries that duplicated edges — flake in the old base
- `Integration - Infrastructure` failed in `test_kuzu_adapter.py` (`_drop_native_resources` monkeypatch) — kuzu adapter area since reworked in the ladybug migration
- Windows 3.13 unit failed on the `WinError 32` file-lock flake, fixed on `dev` since

**Re-verified the patch against the current stack** (openai has since jumped to 2.43.0, and this module touches openai's private internals):
- all 20 hard-coded rebind sites still exist in openai 2.43.0
- `install()` succeeds, is idempotent, and cached `get_origin`/`get_args` return correct results
- `import cognee.infrastructure.llm` on the merged branch installs the cache as intended
- `pre-commit` (ruff lint + format) passes on both files

**Heads-up on the fresh CI run:** the CI LLM key is currently out of credits, so every LLM-touching job fails with `LLMPaymentRequiredError` (402) repo-wide right now — those failures will appear on this PR too and are unrelated. Re-run once the key is topped up.

PR #2881, opened 2026-05-21
feat: accumulate concurrent embedding API calls

GraDKh:
## Summary

Adds `AccumulatingEmbeddingEngine`, an opt-in coalescing wrapper around any `EmbeddingEngine`. When `ACCUMULATE_EMBEDDING_CALLS=true`, concurrent `embed_text` calls share a queue and are flushed together once either the inner engine's batch size is reached or the oldest queued item has waited `ACCUMULATE_EMBEDDING_TIMEOUT_MS` (default 100 ms).

## Motivation

Telemetry on a 100-document cognify run showed the pipeline was issuing many small embedding requests:

- **48% of OpenAI embedding API calls had `batch_size ≤ 3`**, but those calls carried only **~6% of the strings being embedded** — they burned ~36% of total embedding wall time on per-request overhead.
- Root cause: `add_data_points` runs once per chunk group in the cognify pipeline, and `index_data_points` only batches within that group. With small groups (e.g. 1 `TextDocument` → 1 string), every type/field becomes a single-string call.

## Impact

Same 100-document workload, single sequential run with the accumulator off vs on:

| | Off | **On** | Δ |
|---|---|---|---|
| Embedding API calls | 711 | **366** | −49% |
| `bs ≤ 3` calls | 343 | **24** | −93% |
| **Cognify wall time** | **334.9s** | **273.2s** | **−18.4%** |

## What changed

- `cognee/infrastructure/databases/vector/embeddings/AccumulatingEmbeddingEngine.py` (new) — transparent wrapper. Uses `asyncio.Lock` for queue mutation, dispatches outside the lock, retains in-flight task references, short-circuits when all waiters are cancelled, and fails loudly on vector-count mismatch from the inner engine.
- `cognee/tests/unit/infrastructure/test_accumulating_embedding_engine.py` (new) — 11 unit tests covering coalescing, timeout/batch-size flush conditions, oversized bypass, error propagation, cancellation, attribute forwarding, and constructor validation.
- `config.py`, `get_embedding_engine.py` — two new config fields and factory wiring. Wrap happens after engine construction; singleton preserved via the existing `@lru_cache`.
- `.env.template` — documents `ACCUMULATE_EMBEDDING_CALLS` and `ACCUMULATE_EMBEDDING_TIMEOUT_MS`.

Defaults to **off**; existing behavior is unchanged.

## Design notes

- **Scoped to a single event loop.** Internal state is protected by `asyncio.Lock`. cognee runs FastAPI on one loop, so this is sufficient; the docstring is explicit about the limitation.
- **Recursive context-window splits bypass the queue.** `LiteLLMEmbeddingEngine` recurses on `ContextWindowExceededError` via `self.embed_text` — that `self` is the inner engine, not the wrapper, so recursion goes straight to the API. This is intentional (a merged batch that exceeded the context window should be re-split, not re-queued) and documented in the module docstring.
- **Cancellation safety.** `_dispatch` catches `BaseException` (deliberately broad) so that a cancelled dispatch task propagates `CancelledError` to every waiter via `set_exception`. Using `except Exception` instead would leave waiters hanging forever on their futures.

## Test plan

- [x] Unit tests pass (`pytest cognee/tests/unit/infrastructure/test_accumulating_embedding_engine.py` — 11/11, ~5s)
- [x] End-to-end benchmark with `ACCUMULATE_EMBEDDING_CALLS=false` (baseline cognify time)
- [x] End-to-end benchmark with `ACCUMULATE_EMBEDDING_CALLS=true` (−18% cognify time)
- [ ] CI green

## Summary by CodeRabbit

* **New Features**
  * Optional embedding request coalescing: concurrent embed requests can be batched and flushed by timeout or when batch-size is reached; oversized requests bypass batching. New commented env template variables added for ACCUMULATE_EMBEDDING_CALLS and ACCUMULATE_EMBEDDING_TIMEOUT_MS.

* **Tests**
  * Added comprehensive unit tests covering coalescing behavior, timeout and immediate flush, oversized bypass, error propagation, response validation, cancellations, and config serialization/validation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/topoteretes/cognee/pull/2881?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


GraDKh:
@coderabbitai review

siillee (commented):
LGTM, but maybe one more pair of eyes with more experience with asynchronous python should take a look

.env.template
```diff
@@ -77,6 +77,12 @@ STRUCTURED_OUTPUT_FRAMEWORK="instructor"
 #EMBEDDING_BATCH_SIZE=36
 # If not provided, LLM_API_KEY is used for embeddings too.
 #EMBEDDING_API_KEY="your_api_key"
+# Coalesce concurrent embed_text calls into one batched API request.
+# Helps pipelines (e.g. cognify) that emit many small embedding requests.
+# Flushes when EMBEDDING_BATCH_SIZE strings are queued or the oldest one
+# has waited ACCUMULATE_EMBEDDING_TIMEOUT_MS milliseconds.
+#ACCUMULATE_EMBEDDING_CALLS=false
+#ACCUMULATE_EMBEDDING_TIMEOUT_MS=100
```
dexters1: Can we also define size of embeddings that will be accumulated in a single request before being sent? Some providers don't support a big number of embeddings to be part of one batch call.

GraDKh: At the moment we are reading the size of the batch from the inner embedding provider. Do you think we should still have a separate setting?

GraDKh:
## Summary of Fixes

All Copilot and CodeRabbit feedback has been addressed in commit 5abca17e1:

✅ **Type Annotations**
- Added return type annotation to `create_embedding_engine()`
- Added return type hint to `FakeEngine.embed_text()`

✅ **Documentation**
- Updated `create_embedding_engine()` docstring with new parameters
- Added docstrings to 9 test functions for clarity

✅ **Test Coverage**
- Added `test_config_rejects_non_positive_accumulate_timeout_ms()` to validate config validation

✅ **Code Quality**
- All code formatted with ruff
- All 22 tests passing ✓
- Type checking passing with ty ✓

The PR is now ready for CI validation.

.env.template
dexters1: No, you are right we already have an env variable for it.

at the moment it is defined by the EMBEDDING_BATCH_SIZE environment variable

The batch_size there is arbitrarily set and not a real limit, it can be increased and decreased depending on the embedding model and embedding provider. We need to be careful with how big the batch can be:

Image

I remember adding it as 36 as the default because Gemini could not handle a bigger batch of embeddable points, OpenAI could handle at least 100 as far as I remember (probably even more)

These limits have probably changed overtime I added this a long time ago. If there could be a way to check what this limit is and auto set the variable based on that (if no env setting was defined) it could go a long way optimizing this

GraDKh: If we need auto-adjusting the limit I would propose to implement it as a separate PR and merge this one as is. WDYT?

cognee/infrastructure/databases/vector/embeddings/config.py
```diff
@@ -81,6 +81,13 @@ class EmbeddingConfig(BaseSettings):
     embedding_max_completion_tokens: Optional[int] = 8191
     embedding_batch_size: Optional[int] = None
     huggingface_tokenizer: Optional[str] = None
+    # When True, embed_text calls are coalesced via AccumulatingEmbeddingEngine:
+    # concurrent callers share a queue and are flushed together once either
+    # embedding_batch_size strings are pending or the oldest item has waited
+    # accumulate_embedding_timeout_ms milliseconds. Helps when the cognify
+    # pipeline emits many small (bs=1..3) requests.
+    accumulate_embedding_calls: bool = False
+    accumulate_embedding_timeout_ms: int = 100
```
dexters1: @GraDKh can we reduce the default flush timeout? It will reduce search speed by 100ms 

We are trying to get it under 1s (when no LLMs are used)

cognee/infrastructure/databases/vector/embeddings/config.py
```diff
@@ -81,6 +81,13 @@ class EmbeddingConfig(BaseSettings):
     embedding_max_completion_tokens: Optional[int] = 8191
     embedding_batch_size: Optional[int] = None
     huggingface_tokenizer: Optional[str] = None
+    # When True, embed_text calls are coalesced via AccumulatingEmbeddingEngine:
+    # concurrent callers share a queue and are flushed together once either
+    # embedding_batch_size strings are pending or the oldest item has waited
+    # accumulate_embedding_timeout_ms milliseconds. Helps when the cognify
+    # pipeline emits many small (bs=1..3) requests.
+    accumulate_embedding_calls: bool = False
```
dexters1: Overall seems like a nice idea might be good to have it turned on by default if we can have a lower default timeout

cognee/infrastructure/databases/vector/embeddings/config.py
GraDKh: Reduced to 20 ms

cognee/infrastructure/databases/vector/embeddings/config.py
GraDKh: Set to true by default

.env.template
dexters1: Yes, not sure if we need an auto-adjusting limit necessarily, but wanted to mention that it is arbitrary at this moment.

I also am not sure how much does it affect speed if too many data points are grouped in one embedding call, something that should be checked before we implement an auto-adjusting limit (if we implement it)

cognee/infrastructure/databases/vector/embeddings/get_embedding_engine.py
```diff
@@ -73,6 +77,10 @@ def create_embedding_engine(
           for specific providers.
         - llm_api_key: API key for the LLM service, to be used if embedding_api_key is not
           provided.
+        - accumulate_embedding_calls: Whether to enable coalescing of concurrent embed_text
+          calls into batched API requests (default: False).
+        - accumulate_embedding_timeout_ms: Maximum wait time in milliseconds before flushing
+          accumulated calls (default: 100).
```
dexters1, line 83:
```suggestion
          accumulated calls (default: 20).
```

dexters1:
A pretty good idea, definitely reduces RPM for embedding requests I just don't know if it is worth it as it does reduce search speed by 0.2s always which could make it not be worth it. Closing this one for now, but we will revisit it if we notice issues with RPMs

For larger documents processing bypasses the accumulator entirely as the batching of the embedding calls in the pipeline is already larger or the same as the accumulator size

PR #2911, opened 2026-05-28
fix: send string data as plain form field in CloudClient.remember to enable session cache (closes #2888)

### LINKED ISSUE #2888: [Bug]: remember(session_id=...) via HTTP/MCP silently no-ops despite documented "session cache" behavior
### Bug Description

`POST /api/v1/remember` (and therefore the MCP `remember` tool in API mode) accepts a `session_id` parameter per the documented [Tools Reference](<https://docs.cognee.ai/cognee-mcp/mcp-tools>) and returns `status="session_stored"`, but writes nothing to the session cache. The cognee-mcp client always wraps `data` as a multipart file upload, and the backend's `_add_to_session` explicitly skips file-shaped data — so the documented `remember → recall → improve` session-aware workflow is functionally broken via every HTTP path. Direct Python library calls (`cognee.remember(session_id=...)`) work correctly, isolating the defect to the HTTP endpoint.

### Steps to Reproduce

1. Start cognee backend with `CACHE_BACKEND=redis` (or `fs` — same outcome). Confirm Redis adapter loads:

   ```
   docker logs cognee 2>&1 | grep "Successfully connected to Redis"
   # cognee.shared.logging_utils: Successfully connected to Redis at redis:6379 [RedisAdapter]
   ```
2. Confirm Redis is empty:

   ```
   docker exec redis redis-cli DBSIZE
   # (integer) 0
   ```
3. Hit the remember endpoint directly (no MCP, no client library), matching the multipart shape that `cognee-mcp/src/cognee_client.py:466-480` sends:

   ```
   echo "session test content" > /tmp/test.txt
   curl -X POST http://localhost:8001/api/v1/remember \
     -F "data=@/tmp/test.txt" \
     -F "datasetName=test_dataset" \
     -F "session_id=repro_session"
   ```
4. Inspect response and storage state:

   ```
   # Response:
   # {"status":"session_stored","dataset_name":"test_dataset","dataset_id":null,
   #  "pipeline_run_id":null,"items_processed":0,"elapsed_seconds":2.07e-7,
   #  "session_ids":["repro_session"]}
   
   docker exec redis redis-cli KEYS '*'
   # (empty)
   
   docker exec postgres psql -U cognee -d cognee_db \
     -c "SELECT session_id FROM session_records WHERE session_id='repro_session';"
   # (0 rows)
   ```
5. Compare to direct Python library call inside the same container (works correctly):

   ```
   docker exec cognee /app/.venv/bin/python -c "
   import asyncio, cognee
   async def main():
       await cognee.remember('text', session_id='direct_python_test',
                             dataset_name='test_dataset')
   asyncio.run(main())
   "
   # Log: "Successfully connected to Redis at redis:6379"
   # Log: "remember: added entry to session 'direct_python_test'"
   
   docker exec redis redis-cli KEYS '*'
   # 1) "agent_sessions:<user_uuid>:direct_python_test"
   ```

### Expected Behavior

Per [`docs.cognee.ai/cognee-mcp/mcp-tools` > Tools Reference > `remember`](<https://docs.cognee.ai/cognee-mcp/mcp-tools>):

> | `session_id` | `str` | `None` | **When set, stores in session cache instead of permanent graph memory.** |

Per [Sessions and Caching](<https://docs.cognee.ai/core-concepts/sessions-and-caching>):

> Sessions are keyed by `agent_sessions:{user_id}:{session_id}`.

So step 4 should produce a Redis key `agent_sessions:<user_uuid>:repro_session`, a corresponding row in the `session_records` table, and an actually-stored session Q&A entry retrievable via `cognee.session.get_session(session_id="repro_session")`.

### Actual Behavior

* HTTP response says `status="session_stored"` with `items_processed=0` and `elapsed_seconds: ~2e-7` (200 nanoseconds — the handler returns immediately without doing real work).
* No Redis key is created.
* No `session_records` row is inserted.
* `cognee.session.get_session(session_id="repro_session")` returns `[]`.
* `recall(query, session_id="repro_session")` finds nothing in session memory and falls straight through to the permanent graph (where the data was never stored either).

The handler returns success despite skipping the write.

### Environment

* OS: macOS 25.4.0 (Darwin) on Apple M4 Pro, Docker Desktop
* Python version: 3.12.13 (inside container)
* Cognee version: `1.1.0-local` (built from `cognee/cognee:main` at commit on or after upstream/main `0187fd8a4`, rebuilt 2026-05-22)
* cognee-mcp: `cognee/cognee-mcp:main` from Docker Hub (current as of 2026-05-22)
* LLM Provider: Ollama (`qwen2.5:32b-instruct-q5_K_M`) reached via `host.docker.internal:11434/v1`
* Databases: PostgreSQL 17 + pgvector (relational + vector), Neo4j 5.26 (graph), Redis 7-alpine (session cache)
* `REQUIRE_AUTHENTICATION=false`, default user UUID confirmed identical between direct Python and HTTP request paths
* Deployment: docker-compose, `cognee` and `cognee-mcp` on shared `cognee-network`
* Structured output framework: BAML (backend), instructor (mcp) — defect reproduces under both

### Logs/Error Messages

```shell
# Backend access log shows the POST arriving with HTTP 200:
192.168.65.1:39453 - "POST /api/v1/remember HTTP/1.1" 200

# But NO log line from cognee.shared.logging_utils about session write,
# unlike the direct-Python path which always logs:
# "remember: added entry to session '<session_id>'"

# Enable DEBUG to see the skip:
# LOG_LEVEL=DEBUG produces:
# "remember: skipping session write for placeholder-only payload ([UploadFile]...)"


No errors or exceptions raised — the silent-no-op pattern means there's nothing visible at INFO level.
```

### Additional Context

Two upstream files disagree on the data shape expected by the session-cache write path.

`cognee-mcp/src/cognee_client.py:466-480` — wraps every `data` argument as a multipart file upload, regardless of whether `session_id` is set:

```python
async def remember(self, data, dataset_name="main_dataset", session_id=None, custom_prompt=None):
    if self.use_api:
        endpoint = f"{self.api_url}/api/v1/remember"
        files = self._text_upload(data)          # ← always wraps as file
        form_data = {"datasetName": dataset_name}
        if session_id:
            form_data["session_id"] = session_id
        response = await self.client.post(endpoint, files=files, data=form_data, ...)
```

`cognee/api/v1/remember/remember.py:125-156` — backend explicitly skips session writes when data arrives as a file upload:

```python
async def _add_to_session(session_id: str, data, user):
    """Add a Q&A entry to the session cache.

    Sessions store chat-shaped content (prompts, assistant answers,
    Q&A turns). File-upload data coerces to placeholder strings like
    ``[UploadFile]`` / ``[file: name]`` — those are useless in the
    session cache and pollute recall results, so they're skipped.
    """
    ...
    text = _data_to_text(data)
    stripped = text.strip()
    if any(stripped.startswith(prefix) for prefix in _SESSION_PLACEHOLDER_PREFIXES):
        logger.debug(
            "remember: skipping session write for placeholder-only payload"
        )
        return                                    # ← silent skip, status flag already set
```

The status response is set BEFORE the skip check, so the API caller is told `session_stored` regardless.

### Workarounds attempted (none address the defect)

* Switching `CACHE_BACKEND` between `fs` and `redis`: no difference — the cache adapter is never invoked because the library returns before reaching it.
* Migrating cognee-mcp from a host venv install to the canonical Dockerized `cognee/cognee-mcp:main`: same client code path, same backend, same defect.
* Setting `STRUCTURED_OUTPUT_FRAMEWORK=instructor` vs `baml`: unrelated.
* Using `cognee/cognee:main` (rolling) vs pinned `:1.1.0`: same source code as of upstream/main `0187fd8a4`.

The only working workaround is bypassing the HTTP endpoint — calling `cognee.remember(session_id=...)` directly via Python in-process.

### Impact

Disables the documented session-aware workflow for every MCP client. The introductory example from [`cognee-mcp/README.md`](<https://github.com/topoteretes/cognee/blob/main/cognee-mcp/README.md#-basic-usage>) does not work:

```python
remember(data="Temporary working note", session_id="agent-session-1")
recall(query="What changed?", session_id="agent-session-1")
```

Because `remember(session_id=...)` silently drops, `recall(session_id=...)` always falls through to graph (which is empty for session content), and `improve(session_ids=...)` has nothing to bridge into permanent memory.

### Related prior work in this area

* topoteretes/cognee#2855 (closed) — `MCP recall in default config fails with "'NoneType' object has no attribute 'id'" — wrapper does not pass user to cognee.recall`. Different defect (recall, missing user) but in the same `cognee-mcp/src/cognee_client.py` wrapper file. Pattern is the same: MCP wrapper doesn't deliver behavior promised by the docs.
* topoteretes/cognee#2809 (open) — `cognee-cli --api-url doesn't support remember/recall/improve/forget`. CLI dispatcher misses v1.0 commands. Different surface, related theme of HTTP-endpoint paths not matching documented behavior.
* topoteretes/cognee#2796 (open) — Unrelated tokenizer issue; mentioned only because it appears in `session_id remember` keyword searches.

### Suggested fixes (pick one — all three resolve the issue)

1. **Client-side**: In `cognee-mcp/src/cognee_client.py`, branch on `session_id`. When set, POST a JSON body to `/api/v1/remember/entry` (the typed-entry endpoint that already exists at `cognee/api/v1/remember/routers/get_remember_router.py:123-176`) with a `QAEntry` constructed from the data string. When unset, keep the current multipart-file behavior. This matches the apparent intent of the dual endpoint design.
2. **Library-side**: In `cognee/api/v1/remember/remember.py`, when `session_id` is set AND data arrives as `UploadFile`-shaped, read the file content into a string before applying the placeholder filter. Only skip if the resolved text is actually a placeholder (e.g., `[file: <empty>]`), not just because the wire shape is multipart.
3. **API-side**: Have `POST /api/v1/remember` route internally to `/api/v1/remember/entry` whenever `session_id` is present and `data` is a single text file.

Whichever is chosen, the response from the HTTP endpoint should not return `status="session_stored"` when the write was skipped — `items_processed: 0` plus an explicit `status="skipped_placeholder"` (or similar) would prevent the silent-failure pattern.

### Pre-submission Checklist

- [X] I have searched existing issues to ensure this bug hasn't been reported already
- [X] I have provided a clear and detailed description of the bug
- [X] I have included steps to reproduce the issue
- [X] I have included my environment details

botbikamordehai2-sketch:
## What

When `CloudClient.remember()` receives a string `data` argument, it wraps it as a multipart file upload (`io.BytesIO` with `filename="data.txt"`). The backend's `_add_to_session` explicitly skips file-shaped data, so the session cache is never populated — causing the documented `remember → recall → improve` session-aware workflow to silently no-op for all HTTP/MCP paths.

## Fix

Remove the `io.BytesIO` wrapping and `filename` parameter when `data` (or list items) is a plain string. Plain strings are now sent as regular form fields with `content_type="text/plain"`, which the backend recognises as text content eligible for session cache storage.

File uploads (file-like objects with a `.read()` method) are unaffected and continue to be sent as multipart file fields.

Closes #2888

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of text data submissions to use more efficient transmission format.
  * Ensured file uploads continue to work as expected while optimizing string data processing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/topoteretes/cognee/pull/2911?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cognee/api/v1/serve/cloud_client.py
+++ b/cognee/api/v1/serve/cloud_client.py
@@ -68,10 +68,12 @@ class CloudClient:
 
         # Handle data — string or file-like objects
         if isinstance(data, str):
+            # Send plain strings as a regular form field so the backend
+            # treats them as text (not file uploads) and can write them
+            # to the session cache via _add_to_session.
             form.add_field(
                 "data",
-                io.BytesIO(data.encode("utf-8")),
-                filename="data.txt",
+                data,
                 content_type="text/plain",
             )
         elif isinstance(data, list):
@@ -79,8 +81,7 @@ class CloudClient:
                 if isinstance(item, str):
                     form.add_field(
                         "data",
-                        io.BytesIO(item.encode("utf-8")),
-                        filename="data.txt",
+                        item,
                         content_type="text/plain",
                     )
                 elif hasattr(item, "read"):


dexters1:
Thank you for your contribution! But the issue was already resolved in a different PR

botbikamordehai2-sketch:
Thanks for the note — closing since the issue was already resolved in another PR.

PR #2914, opened 2026-05-28
feat(remember): add graph_model, ontology_key params; fix UploadFile background task; remove DatasetConfiguration

LStromann:
### `POST /v1/remember` — new params + background task fix
- Added `graph_model` (JSON-serialised schema) and `ontology_key` Form fields to the remember endpoint, matching the cognify endpoint.
- **Bug fix**: with `run_in_background=True`, Starlette closes `UploadFile` (SpooledTemporaryFile) handles when the HTTP response is sent. The asyncio background task then calls `add()` on closed files — nothing was ever stored. Pre-read all file bytes into `BytesIO` during the request handler so data survives into the background task.

### `POST /v1/cognify` — remove DatasetConfiguration
- Removed DB lookup fallback: the router was reading `DatasetConfiguration` from the DB when `graph_model`/`custom_prompt` were not in the payload. The UI stores config client-side and always sends params explicitly — a DB fallback means clearing a graph model in the UI has no effect (old DB value restored on next run).
- Removed DB write-back after cognify.
- Removed the multi-dataset bug: config was read from `datasets[0]` and applied uniformly to all datasets.
- Both params are now truly optional and default to `KnowledgeGraph` / `None` when not supplied.

## Test plan
- [ ] `POST /v1/remember` with files + `run_in_background=true` — files should be stored and cognify should run
- [ ] `POST /v1/remember` with `graph_model` JSON — custom model applied to cognify
- [ ] `POST /v1/remember` with `ontology_key` — ontology applied to cognify
- [ ] `POST /v1/cognify` without `graph_model` — defaults to KnowledgeGraph (no DB lookup)
- [ ] `POST /v1/cognify` with `graph_model` — custom model used, not persisted to DB
- [ ] Clearing graph model in UI then cognifying — no stale config restored from DB

### DIFF
--- a/cognee/api/v1/cognify/routers/get_cognify_router.py
+++ b/cognee/api/v1/cognify/routers/get_cognify_router.py
@@ -176,39 +176,8 @@ def get_cognify_router() -> APIRouter:
                     }
                 }
 
-            # Resolve graph model and custom prompt: use payload values,
-            # fall back to stored DatasetConfiguration, then defaults.
             graph_model_schema = payload.graph_model
             custom_prompt = payload.custom_prompt
-
-            if datasets and (not graph_model_schema or not custom_prompt):
-                try:
-                    from uuid import UUID as _UUID
-                    from cognee.modules.data.models import DatasetConfiguration
-                    from cognee.infrastructure.databases.relational import get_relational_engine
-                    from sqlalchemy import select
-
-                    first_ds = datasets[0]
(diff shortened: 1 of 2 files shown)


LStromann:
Superseded by two focused PRs: #2915 (background UploadFile fix) and #2916 (graph_model param + DatasetConfiguration removal).

PR #2915, opened 2026-05-28
fix(remember): materialize UploadFile bytes before background task

LStromann:
## Summary

With `run_in_background=True`, `POST /v1/remember` returns an HTTP response immediately. Starlette then closes the request scope, which invalidates the `UploadFile` objects (backed by `SpooledTemporaryFile`). The asyncio background task calls `add()` on already-closed file handles — so uploaded files were never stored.

**Fix**: pre-read all file bytes into `BytesIO` during the request handler (while the scope is still open) before calling `cognee_remember()`, so data is available when the background task runs.

## Test plan
- [ ] `POST /v1/remember` with files and `run_in_background=true` — files should be stored and cognify should run
- [ ] `POST /v1/remember` with files and `run_in_background=false` (blocking) — should work as before

## Summary by CodeRabbit

**Bug Fixes**
- Fixed an issue where uploaded files became inaccessible when using background processing mode. Files are now properly retained in memory and remain fully accessible throughout the entire background processing workflow, improving reliability of asynchronous file operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/topoteretes/cognee/pull/2915?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cognee/api/v1/remember/routers/get_remember_router.py
+++ b/cognee/api/v1/remember/routers/get_remember_router.py
@@ -1,3 +1,4 @@
+from io import BytesIO
 from uuid import UUID
 
 from fastapi import APIRouter, HTTPException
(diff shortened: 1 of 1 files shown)


dexters1:
Closing this one as we handled in a different PR

PR #2938, opened 2026-05-30
fix: bind module-level asyncio locks to the running event loop

Dmheath1:
## Description

I hit this running cognee from a process that drives work through more than one event loop. After the first loop finished, later calls started failing with:

```
RuntimeError: <asyncio.locks.Lock object at 0x...> is bound to a different event loop
```

The cause is that a handful of locks are created at import time as module-level `asyncio.Lock()` objects. On CPython 3.10+ an `asyncio.Lock` binds itself to the running loop the first time it is awaited on the contended path (`asyncio/mixins.py` `_get_loop`). The uncontended `acquire()` fast path never calls `_get_loop()`, so the binding only happens once the lock is actually contended, which is why the failure looked intermittent to me at first. Once a module-level lock has bound to one loop, contending it from any other loop raises, so in practice a module-level lock can only be used from a single event loop for the lifetime of the process.

This affects anything that runs cognee from more than one loop in the same process:

- calling `asyncio.run()` more than once,
- a web framework that creates a fresh event loop per request,
- `pytest-asyncio` with function-scoped event loops,
- multi-worker or multi-loop ingestion.

The module-level locks this change covers are `_registry_lock`, `_improve_registry_lock` and the cached per-`(session_id, op)` locks in `cognee/infrastructure/locks/session_lock.py`, plus `_create_db_lock` in `cognee/infrastructure/databases/relational/create_db_and_tables.py`.

The fix is a small `LoopBoundLock` (`cognee/infrastructure/locks/loop_bound_lock.py`) that keeps one real `asyncio.Lock` per running loop, created on first use and stored in a `WeakKeyDictionary` keyed by the loop so locks for finished loops are garbage-collected. An `asyncio.Lock` can only synchronize tasks inside a single loop anyway, so a per-loop lock keeps the exact same single-loop guarantee while removing the cross-loop crash. I swapped `LoopBoundLock` into those sites.

On scope: I kept this to the sites where a per-loop lock is the correct primitive, which is also the highest-severity case (these locks bind at import time and are shared for the whole life of the process). I deliberately left `_first_run_lock` in `cognee/modules/pipelines/layers/setup_and_check_environment.py` alone: it guards a once-per-process flag (`_first_run_done`), so it wants a loop-agnostic single-flight guard rather than a per-loop lock, which is a separate change. Two other module-level `asyncio.Lock()` objects (`update_status_lock` in `cognee/api/v1/cognify/cognify.py` and `cognee/modules/pipelines/operations/pipeline.py`) are dead code (only referenced by a commented-out line), so they never bind to a loop and I left them untouched. Locks created per instance inside `__init__` on adapters that are then cached across loops (for example through `closing_lru_cache`) are the same class of problem but a larger, separate change. Happy to follow up on any of these.

## Acceptance Criteria

- cognee can be used from more than one event loop in the same process without raising "bound to a different event loop".
- Single-loop behavior is unchanged: same-key session operations still serialize, different-key operations still run concurrently, and database setup is still guarded by a lock.
- New unit tests cover the cross-loop case and pin the preserved semantics, and they fail without the change.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots

I added unit tests in `cognee/tests/unit/infrastructure/locks/`. Local run:

    $ python -m pytest cognee/tests/unit/infrastructure/locks/ -q
    11 passed

    $ ruff check cognee/infrastructure/locks/
    All checks passed!

    $ ty check cognee/infrastructure/locks/loop_bound_lock.py cognee/infrastructure/locks/session_lock.py
    All checks passed!

The cross-loop tests force contention on purpose, so they fail against the current code: without the fix, `test_session_lock_works_across_event_loops` raises `RuntimeError: ... is bound to a different event loop`. I also checked that the bug reproduces and the fix holds on Python 3.11, 3.13 and 3.14.

## Pre-submission Checklist

## DCO Affirmation

## Summary by CodeRabbit

* **Bug Fixes**
  * Resolved event loop binding errors in async locking that could cause runtime failures when handling multiple concurrent event loops.

* **Tests**
  * Added comprehensive unit tests for lock behavior across multiple event loops and contention scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/topoteretes/cognee/pull/2938?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cognee/infrastructure/databases/relational/create_db_and_tables.py
+++ b/cognee/infrastructure/databases/relational/create_db_and_tables.py
@@ -1,8 +1,8 @@
-import asyncio
+from cognee.infrastructure.locks.loop_bound_lock import LoopBoundLock
 
 from .get_relational_engine import get_relational_engine
 
-_create_db_lock = asyncio.Lock()
+_create_db_lock = LoopBoundLock()
 
 
 async def create_db_and_tables():
--- a/cognee/infrastructure/locks/__init__.py
+++ b/cognee/infrastructure/locks/__init__.py
@@ -1,3 +1,9 @@
+from .loop_bound_lock import LoopBoundLock
 from .session_lock import release_improve_lock, session_lock, try_acquire_improve_lock
 
-__all__ = ["release_improve_lock", "session_lock", "try_acquire_improve_lock"]
+__all__ = [
+    "LoopBoundLock",
+    "release_improve_lock",
+    "session_lock",
+    "try_acquire_improve_lock",
+]
--- a/cognee/infrastructure/locks/loop_bound_lock.py
+++ b/cognee/infrastructure/locks/loop_bound_lock.py
@@ -1,0 +1,74 @@
+"""A module-safe async lock that binds to the running event loop on use.
+
+A plain ``asyncio.Lock()`` created at import time lazily binds to the first
+event loop that contends it (CPython 3.10+, via
+``asyncio.mixins._LoopBoundMixin``) and afterwards raises
+``RuntimeError: ... is bound to a different event loop`` if it is contended again
+from a different running loop. Any process that drives cognee from more than one
+event loop -- repeated ``asyncio.run(...)`` calls, a fresh loop per request,
+``pytest-asyncio`` with function-scoped loops, or a multi-worker ingest -- can
+therefore hit that error on cognee's module-level locks.
+
+``LoopBoundLock`` keeps one real ``asyncio.Lock`` per running loop, created on
+first use and stored in a ``WeakKeyDictionary`` keyed by the loop so locks for
+closed loops are garbage-collected. Each loop gets its own lock, which is the
+only thing an ``asyncio.Lock`` can do anyway -- it cannot synchronize across
+loops -- so a module-level singleton was already incorrect for multi-loop
+callers. Single-loop behavior is unchanged; the cross-loop failure is removed.
+"""
+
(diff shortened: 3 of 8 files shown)


Dmheath1:
Thanks for the review. Addressed both points in fbdb21c8e:

1. `setup_and_check_environment` (`_first_run_done`): you're right that a per-loop lock doesn't preserve the once-per-process contract across loops. I reverted that file out of this PR and scoped the change to the sites where per-loop locking is the correct semantic: the session locks (keyed per `(session_id, op)`) and the idempotent `create_db_and_tables`. The first-run guard needs a loop-agnostic single-flight primitive, which is a separate change I'd rather not bundle here.

2. `LoopBoundLock`: added docstrings to `acquire`, `release`, and `_lock_for_running_loop`. On the `WeakKeyDictionary` thread-safety, I left the `threading.Lock` out as you suggested, since the same instance isn't accessed from multiple OS threads here (a running loop is single-threaded, so two threads never key on the same loop). Glad to add it as a safeguard if you'd prefer.

I also added a small guard test (`test_module_level_locks.py`) asserting the converted module-level locks stay loop-bound, so a raw `asyncio.Lock` can't be reintroduced at those sites unnoticed.

Vasilije1990:
@Dmheath1 looks like a good catch. @dexters1 can you review? Seems like an improvement

PR #2939, opened 2026-05-31
feat(visualization): schema inventory SDK + interactive memory schema explorer

Vasilije1990:
## Why

The graph visualization shows *what* the system understood, but verifying coverage meant chat probing (e.g. "list the names" returning a partial set). This adds a way to **see what was understood** — a per-type inventory with counts and samples — and turns the Schema tab into an interactive, multi-level explorer of how data, actors, and operations shape the memory.

## What

**SDK**
- `get_schema_inventory(dataset, samples_per_type)` → `[{type, count, samples, relationships}]`. Resolves `Entity → EntityType` via the `is_a` edge and answers "are all N captured?" with a **graph aggregation** (not top-k retrieval, which structurally under-reports list queries).
- `get_memory_provenance_graph()` / `visualize_memory_provenance()` — project the relational actor/ownership layer (Tenant → User → Agent → Dataset/"brain" → Data/file → memory, plus agent read/write and agent-written Sessions) into a graph **purely from the relational DB** (works without the graph backend / LLM).

**Schema view (standalone HTML)**
- Clean, airy Segment-style reskin of the schema view + entity inspector panel; default light theme so Graph and Schema match.
- Actor/ownership **levels** with column headers and **instance mini-cards** (top-5, expand/collapse). Semantic entity types now land in the Entities column.
- **Instance inspector** with breadcrumb navigation + close button.
- **Click-to-spotlight**: highlight an instance's connections *in place* (no re-layout), with **d3 pan/zoom** and panel-aware auto-fit. Edges arc below the cards to avoid overlap.
- **Transformations impact-layer**: a toggleable rail of operations (cognify, memify, improve/self-improvement, feedback weighting, forget, …) with color-coded **produces / enriches / modifies / removes** edges to the schema types they affect — driven by a curated operations catalog corroborated by live node provenance.

## Test plan
- `pytest cognee/tests/unit/modules/visualization/ cognee/tests/unit/api/test_get_schema_inventory.py cognee/tests/unit/api/test_memory_provenance.py` → **48 passing** (SDK, provenance builder, preprocessor operation layer, rendered schema markup).
- Demos: `examples/python/schema_inventory_demo.py`, `examples/python/memory_provenance_demo.py` render `~/cognee_*_demo.html`.

## Notes
- Scoped strictly to the visualization/SDK work; unrelated working-tree changes were intentionally excluded.
- Follow-ups: node_set-region targeting, and a `PipelineRun`/`TaskRun` transformation timeline (the catalog is structured to support it).

## Summary by CodeRabbit

* **New Features**
  * Schema inventory HTTP endpoint; memory-provenance projection and HTML export; operation-impact layer; interactive schema inspector with type/instance drill-down and highlighting

* **Improvements**
  * Redesigned Schema UI (side inspector, expandable instance cards, spotlights) and default light mode; deterministic bounded samples and richer relationship metadata

* **Tests**
  * Unit tests for inventory, provenance projection, and preprocessing

* **Examples**
  * End-to-end demos for schema inventory and memory provenance visualization


dexters1:
image

There are issues with the visualization:
1) Two duplicate TextDocument nodes have been created for one document
2) DocumentChunk is not handled properly at all it is missing all its connections and description
3) Ruff formatting for the PR
4) Will this feature work when we have multiple large datasets? Regular visualization barely works on one large graph/dataset

Would adding information to regular and provenance visualization about which user/agent added which document to the dataset be an interesting addition? We have this information stored in the relational DB

dexters1 (dismissed):
Added comment on found issues

dexters1:
> image
> There are issues with the visualization:
> 
> 1. Two duplicate TextDocument nodes have been created for one document
> 2. DocumentChunk is not handled properly at all it is missing all its connections and description

Regular visualization on the same data for comparison

image

cognee/api/v1/visualize/get_schema_inventory.py
```diff
@@ -0,0 +1,179 @@
...
+            When set, scoping mirrors the visualize router via
+            ``set_database_global_context_variables``.
+        samples_per_type: maximum number of sample instance names per type.
+        sort: ``"count"`` orders types by descending count, then type name.
+
+    Returns:
+        A list of dicts with keys ``type``, ``count``, ``samples``,
+        ``sample_size``, and ``relationships``. Each ``relationships`` entry is
+        ``{"to_type", "relation", "count"}`` aggregated over edges.
+    """
```
nfnt: A docstring needs to be the first statement in a function body to be usable.

```suggestion
    """Summarize the knowledge graph by semantic type.

    Parameters:
        dataset: optional dataset id/name to scope the graph databases to.
            When set, scoping mirrors the visualize router via
            ``set_database_global_context_variables``.
        samples_per_type: maximum number of sample instance names per type.
        sort: ``"count"`` orders types by descending count, then type name.

    Returns:
        A list of dicts with keys ``type``, ``count``, ``samples``,
        ``sample_size``, and ``relationships``. Each ``relationships`` entry is
        ``{"to_type", "relation", "count"}`` aggregated over edges.
    """
    if samples_per_type < 0:
        raise ValueError("samples_per_type must be non-negative")
```

cognee/api/v1/visualize/memory_provenance.py
```diff
@@ -0,0 +1,395 @@
...
+"""
+
+from typing import Any, Dict, List, Optional, Tuple
+
+from cognee.shared.logging_utils import get_logger
+
+logger = get_logger()
+
+Node = Tuple[str, Dict[str, Any]]
+EdgeData = Tuple[str, str, str, Dict[str, Any]]
```
nfnt: How about using a Dataclass instead of tuples here? This makes it clearer what the tuple entries are meant to be. Especially for `EdgeData` it's not immediately clear what the various string entries are used for.

cognee/api/v1/visualize/get_schema_inventory.py
```diff
@@ -0,0 +1,179 @@
...
+            degree[source_id] += 1
+        if target_id in degree:
+            degree[target_id] += 1
+    return degree
+
+
+async def get_schema_inventory(
+    dataset: Optional[Union[str, UUID]] = None,
+    samples_per_type: int = 5,
+    sort: str = "count",
```
nfnt, line 80: Are there other values of `sort` possible? It seems that only "count" has an effect. If "do not sort" is meant as "use any other value than 'count'", then this isn't good UX. Let's, e.g., use a bool instead, or restrict which input values are possible.

cognee/api/v1/visualize/get_schema_router.py
```diff
@@ -0,0 +1,57 @@
...
+from cognee.modules.users.models import User
+from cognee.shared.logging_utils import get_logger
+
+logger = get_logger()
+
+
+def get_schema_router() -> APIRouter:
+    router = APIRouter()
+
+    @router.get("/inventory", response_model=List[Any])
```
nfnt: The response model isn't very OpenAPI friendly. This is fine if the results are assumed to only get consumed by our visualization. If we want additional use-cases in SDKs that consume this data, a more defined response model should be used.

cognee/api/v1/visualize/routers/get_schema_router.py
nfnt, line 1: Shouldn't this file be in a `routers` folder?

cognee/api/v1/visualize/memory_provenance.py
```diff
@@ -0,0 +1,395 @@
...
+
+
+def build_provenance_graph(
```
nfnt, line 105: What's the use-case of having this function as part of cognee's public API? It is used in `get_memory_provenance_graph` after that function has gathered all data necessary to call this function from internal state. But why call this function directly? Where would a user gather the state necessary to call this function and prepare it in the necessary shapes?

cognee/api/v1/visualize/memory_provenance.py
```diff
@@ -0,0 +1,395 @@
...
+    *,
+    tenants: Optional[List[Dict[str, Any]]] = None,
+    users: Optional[List[Dict[str, Any]]] = None,
+    datasets: Optional[List[Dict[str, Any]]] = None,
+    files: Optional[List[Dict[str, Any]]] = None,
+    agents: Optional[List[Dict[str, Any]]] = None,
+    sessions: Optional[List[Dict[str, Any]]] = None,
+    memory: Optional[Dict[str, Any]] = None,
```
nfnt: We could use `TypedDict`s to describe the required shape of the input values.

nfnt (dismissed):
Looks good to me!

nfnt (approved):

PR #2940, opened 2026-05-31
feat(api): HTTP router for schema inventory + memory provenance (tenant-scoped permissions)

Vasilije1990:
## Summary

Stacked on #2939. That PR adds `get_schema_inventory`, `get_memory_provenance_graph`, and `visualize_memory_provenance` as **SDK functions only**. This PR exposes them over HTTP so FastAPI clients (e.g. the cognee SaaS frontend's Schema tab) can consume them, **with per-caller permission scoping**.

Targets `feature/schema-inventory-sdk` because it depends on those functions; rebase onto `dev` once #2939 merges.

## Endpoints
- `GET /api/v1/schema/inventory?dataset_id=<uuid>&samples_per_type=5&sort=count`
  → JSON `[{type, count, samples, sample_size, relationships:[{to_type, relation, count}]}]`
- `GET /api/v1/schema/provenance?include_memory=false`
  → HTML memory-provenance visualization (Tenant → User → Agent → Brain → File → memory)

## Permissions (important)
- **`/inventory`** enforces **read** permission on the dataset via `get_authorized_existing_datasets([dataset_id], "read", user)`; `get_schema_inventory` then self-scopes the graph DBs to that dataset's owner. Returns **403** on `PermissionDeniedError`; internal errors return a generic **409** (no raw exception text leaked to the client — logged server-side).
- **`/provenance` is tenant-scoped.** The underlying `get_memory_provenance_graph()` reads the relational actor/ownership layer, which spans **all** tenants/users/datasets/files. Exposing it unscoped would let any authenticated user see every other tenant's users (incl. emails), brains, files and agents. So:
  - `get_memory_provenance_graph` / `visualize_memory_provenance` gained `scope_tenant_ids` / `scope_user_ids` filters (applied to tenants, users, datasets, and cascaded to agents/sessions; memory is filtered to in-scope datasets and edges to in-scope nodes; out-of-scope registered/in-process agents are dropped). **No scope ⇒ global**, preserving the OSS local single-user default.
  - The router **always** passes a scope: the caller's `tenant_id` (SaaS) or, if none, the caller's own `user_id` (OSS) — the HTTP path is never fail-open.

Implementation mirrors the sibling `get_visualize_router` conventions (`get_authenticated_user` + `get_authorized_existing_datasets`; `/visualize/multi` is `is_superuser`-gated). `get_schema_inventory` and the provenance helpers are exposed over HTTP **only** by this router.

## Verification
- `/inventory`: `401` unauthenticated; `403` on an unauthorized/missing dataset; per-type rows for an authorized, cognified dataset.
- `/provenance`: `200` returning HTML scoped to the caller's tenant/user (a second tenant's user gets none of the first tenant's actors/data).


nfnt (commented):
This is looking great but depending on the expected use-case, we might need a clearer defined schema for the return values, see my comments on #2939.
While `/provenance` returns HTML, `/inventory` returns JSON without a defined schema. If we want to have maintainable consumers of that data, an OpenAPI schema for that endpoint is very helpful.

nfnt (approved):
LGTM!

PR #2966, opened 2026-06-03
fix(chunking): align LangchainChunker with unified max_chunk_size API

Mubashirrrr:
## Description

`LangchainChunker` was never updated when #626 ("Eliminate the use of `max_chunk_tokens` and use a unified `max_chunk_size`") changed the chunker API. It still targets the **old** `Chunker.__init__(document, get_text, max_chunk_tokens, chunk_size=1024)` signature and references attributes/fields that no longer exist, so it is **completely broken** — selecting it crashes `cognify`.

Two independent failures:

1. **Cannot be instantiated via the standard call path.** Every `Document.read()` does `chunker_cls(self, max_chunk_size=..., get_text=...)`, but `LangchainChunker.__init__` has no `max_chunk_size` parameter → `TypeError: __init__() got an unexpected keyword argument 'max_chunk_size'`. And even bypassing that, `super().__init__(document, get_text, max_chunk_tokens, chunk_size)` passes 4 positional args to the current 3-arg base `Chunker.__init__` → `TypeError: __init__() takes 4 positional arguments but 5 were given`.
2. **`read()` builds an invalid `DocumentChunk`.** It passes `word_count=` and `token_count=` (fields `DocumentChunk` does not define — silently dropped by pydantic's `extra="ignore"`) and never sets the **required** `chunk_size`, so construction raises a `ValidationError`.

`LangchainChunker` is user-selectable via `--chunker LangchainChunker` in both the `cognify` and `remember` CLI commands, so any user choosing it hits this immediately.

### Reproduction

```python
# Standard call path (matches every Document.read):
LangchainChunker(document, max_chunk_size=128, get_text=get_text)
# TypeError: __init__() got an unexpected keyword argument 'max_chunk_size'
```

## Fix

Align the class with the unified `max_chunk_size` convention (same pattern as `TextChunker` / `TextChunkerWithOverlap`):
- Rename the `max_chunk_tokens` parameter to `max_chunk_size`.
- Call the current 3-arg base `super().__init__(document, get_text, max_chunk_size)`.
- Emit `chunk_size=token_count` instead of the non-existent `word_count` / `token_count`.

The `RecursiveCharacterTextSplitter` config params (`chunk_size`, `chunk_overlap`) are unchanged.

## Tests

Added `cognee/tests/unit/modules/chunking/test_langchain_chunker.py` with two tests: one instantiates the chunker via the `Document` call path, one runs `read()` end-to-end (with a mocked embedding engine) and asserts each yielded object is a `DocumentChunk` with a populated `chunk_size` and monotonic indices. Both **fail before** this change (`TypeError`) and **pass after**.

## DCO Affirmation

## Summary by CodeRabbit

* **Refactor**
  * Renamed and clarified chunking parameters; chunk size is now bounded by a new max limit and overlap is clamped accordingly. Error messaging updated to reference the max size.

* **Tests**
  * Added regression tests ensuring the chunker respects small max sizes, yields properly sized chunks, and preserves chunk indexing and chunk_size reporting.

### DIFF
--- a/cognee/modules/chunking/LangchainChunker.py
+++ b/cognee/modules/chunking/LangchainChunker.py
@@ -21,11 +21,11 @@ class LangchainChunker(Chunker):
         self,
         document,
         get_text: callable,
-        max_chunk_tokens: int,
+        max_chunk_size: int,
         chunk_size: int = 1024,
         chunk_overlap=10,
     ):
-        super().__init__(document, get_text, max_chunk_tokens, chunk_size)
+        super().__init__(document, get_text, max_chunk_size)
 
         self.splitter = RecursiveCharacterTextSplitter(
             chunk_size=chunk_size,
@@ -38,12 +38,11 @@ class LangchainChunker(Chunker):
             for chunk in self.splitter.split_text(content_text):
                 embedding_engine = get_vector_engine().embedding_engine
                 token_count = embedding_engine.tokenizer.count_tokens(chunk)
-                if token_count <= self.max_chunk_tokens:
+                if token_count <= self.max_chunk_size:
                     yield DocumentChunk(
                         id=uuid5(NAMESPACE_OID, chunk),
                         text=chunk,
-                        word_count=len(chunk.split()),
-                        token_count=token_count,
+                        chunk_size=token_count,
                         is_part_of=self.document,
                         chunk_index=self.chunk_index,
                         cut_type="missing",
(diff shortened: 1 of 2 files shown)


Mubashirrrr:
Good catch, thanks. Pushed `386f7a9` which defaults `chunk_size` to `max_chunk_size` so the splitter actually honors the unified budget (every `Document.read` instantiates the chunker with only `max_chunk_size`). I also clamped `chunk_overlap` to `< effective_chunk_size` so a very small budget can't trip `RecursiveCharacterTextSplitter`'s "overlap larger than chunk size" error. Added a regression test (`test_langchain_chunker_honors_small_max_chunk_size`) that raises before the change and passes after.

Issue #2902, opened 2026-05-27
[Bug]: Cognee hangs forever following the github quickstart with locally hosted models


cheribral:
### Bug Description

When trying cognee as described in the quickstart section of the github README.md,  the script hangs forever.  Trying the equivalent with the command line does the same.

My config is basically this, running through omlx on an m4 mac.
```
LLM_API_KEY=***
LLM_PROVIDER=custom
LLM_MODEL='openai/Qwen3.6-35B-A3B-oQ6'
LLM_ENDPOINT=http://inference_host:8000/v1
COGNEE_SKIP_CONNECTION_TEST=false

EMBEDDING_PROVIDER="openai_compatible"
EMBEDDING_MODEL="openai/embeddinggemma-300m-6bit"
EMBEDDING_ENDPOINT="http://inference_host:8000/v1"
EMBEDDING_API_KEY=***
LOG_LEVEL=DEBUG
```

When I run the script or the CLI, it just hangs forever.  At that point it just hangs and I have ctrl-c several times to kill it.

### Steps to Reproduce

1. Setup a local inference server
2. Follow the quickstart to install
3. Configure the .env file in a similar manner to above and then copy the script from the README
4. Run the script

### Expected Behavior

It would return to the shell after running the script or the command line tool.

### Actual Behavior

It hangs forever.

### Environment

- OS: Debian 13.4
- Python: 3.11.15
- Cognee version: 1.1.0
- LLM provider: omlx (openai compatible) running on an M4 Max.

### Logs/Error Messages

```shell
These are the last lines in the output.

executing functools.partial(<built-in method execute of sqlite3.Cursor object at 0xffff18e207c0>, 'SELECT data_1.id AS data_1_id, datasets.id AS datasets_id, datasets.name AS datasets_name, datasets.created_at AS datasets_created_at, datasets.updated_at AS datasets_updated_at, datasets.owner_id AS datasets_owner_id, datasets.tenant_id AS datasets_tenant_id \nFROM data AS data_1 JOIN dataset_data AS dataset_data_1 ON data_1.id = dataset_data_1.data_id JOIN datasets ON datasets.id = dataset_data_1.dataset_id \nWHERE data_1.id IN (?)', ('0b6fda8f3c3c5169994d3160e38c2706',))

[...]

Issue #2796, opened 2026-05-08
[Bug]:  cognee.remember调用失败


iamyishan:
### Bug Description

代码执行到：result1 = await cognee.remember("Cognee turns documents into AI memory.")
然后卡住一会儿，打印出日志
Switching to TikToken default tokenizer. [OpenAICompatibleEmbeddingEngine]
然后就终止了
代码如下：

import cognee
import asyncio
async def main():
    print("1. 存储记忆到永久知识图谱...")
    result1 = await cognee.remember("Cognee turns documents into AI memory.")
    print(f"   存储结果: {result1}")
    
    print("\n2. 存储记忆到会话缓存...")
    result2 = await cognee.remember("User prefers detailed explanations.", session_id="chat_1")
    print(f"   存储结果: {result2}")

    print("\n3. 检索记忆 (自动路由)...")
    results = await cognee.recall("What does Cognee do?")
    print(f"   找到 {len(results)} 条结果:")
    for i, result in enumerate(results):
        print(f"   [{i+1}] {result}")

    print("\n4. 检索会话记忆...")
    results = await cognee.recall("What does the user prefer?", session_id="chat_1")
    print(f"   找到 {len(results)} 条结果:")
    for i, result in enumerate(results):
        print(f"   [{i+1}] {result}")

    print("\n5. 清理记忆...")
    await cognee.forget(dataset="main_dataset")
    print("   清理完成")

if __name__ == '__main__':
    asyncio.run(main())

### Steps to Reproduce

对OpenAICompatibleEmbeddingEngine进行测试，代码如下：
import sys
import traceback

print('Step 1: Import OpenAICompatibleEmbeddingEngine')
try:
    from cognee.infrastructure.databases.vector.embeddings.OpenAICompatibleEmbeddingEngine import OpenAICompatibleEmbeddingEngine
    print('   Success')
except Exception as e:
    print(f'   Failed: {e}')
    traceback.print_exc()
    sys.exit(1)

print('Step 2: Create OpenAICompatibleEmbeddingEngine instance')
try:
    engine = OpenAICompatibleEmbeddingEngine(
        model='embed',
        dimensions=1024,
        endpoint='http://10.35.208.2:4001/v1',
        api_key='.',
        batch_size=36,
    )
    print('   Success')
except Exception as e:
    print(f'   Failed: {e}')
    traceback.print_exc()
    sys.exit(1)

print('Step 3: Test tokenizer')
try:
[...]

Issue #2752, opened 2026-05-02
Preflight LLM connection test times out for non-OpenAI providers (verified endpoint healthy)


jason-edstrom:
## Summary

The "Testing connection to LLM endpoint…" preflight check at the start of `cognify` / `remember` hangs for 30 seconds and times out when `LLM_PROVIDER` is set to anything other than `openai`, even when the endpoint is verified healthy via raw `curl` from the same machine.

## Repro

- cognee 1.0.3 / cognee-mcp 0.5.4
- `LLM_PROVIDER=anthropic`
- `LLM_MODEL=claude-haiku-4-5-20251001`
- Valid Anthropic API key with credit on the account
- Run any `remember` (permanent mode) or `cognify` call

## Error

```
LLM connection test timed out after 30s.
Check that your LLM endpoint is reachable and responding.
Set COGNEE_SKIP_CONNECTION_TEST=true to bypass this check.
```

The error message itself recommends bypassing — that's a hint that this check has a known reliability problem.

## Confirmation that the endpoint is fine

Same key, same model, hit directly from the same machine in <1s:

```bash
curl -sS https://api.anthropic.com/v1/messages \
  -H "x-api-key: $KEY" \
  -H "anthropic-version: 2023-06-01" \
  -H "content-type: application/json" \
  -d '{"model":"claude-haiku-4-5-20251001","max_tokens":10,"messages":[{"role":"user","content":"hi"}]}'
# returns proper content + usage in ~600ms
```

So the endpoint is reachable. The preflight is misrouting, sending OpenAI-shaped payload to Anthropic, or otherwise misconfigured for non-OpenAI providers.

## Suggested fix

Route the preflight through the same provider adapter the actual call will use, so it's correct-by-construction. Or, for a generic ping that handles all providers, use:

```python
await litellm.acompletion(
    model=f"{llm_provider}/{llm_model}",
    messages=[{"role": "user", "content": "hi"}],
    max_tokens=1,
)
```

Note that this will also expose the `max_tokens` issue for the Anthropic path (filed separately) — fixing that bug first will make this preflight work as a side effect.

## Workaround for affected users

`COGNEE_SKIP_CONNECTION_TEST=true` in env. The check then exits early and the actual call surfaces real errors with proper diagnostics.

Happy to PR if useful.

Vasilije1990:
Merged, did a new release to handle it

Issue #2456, opened 2026-03-21
[Bug]: Search endpoint hangs after successful data upload and cognify (vLLM + Ollama)


Drexster:
## Issue Summary
The `/api/v1/search` endpoint hangs and times out after data upload and cognify processing complete. Server operational, auth works, datasets API works, but search queries never return results.

### Steps to Reproduce

Steps to Reproduce

1. Deploy Cognee via Docker with vLLM + Ollama config (see above)
2. Authenticate via `/api/v1/auth/login` (works ✅)
3. Create dataset via `/api/v1/datasets` (works ✅)
4. Upload 175 text chunks via `/api/v1/add` using multipart form-data (works ✅)
5. Run cognify via `/api/v1/cognify` (appears to process successfully ✅)
6. Search via `/api/v1/search`:

 bash
curl -X POST http://localhost:8002/api/v1/search \
     -H "Authorization: Bearer <token>" \
     -H "Content-Type: application/json" \
     -d '{"query": "test"}'

## Expected Behavior
Search endpoint should return results with scores and matched text.

### With `LLM_PROVIDER=openai` config:
- Search hangs for 30+ seconds
- Logs show error:

  [error] Error during brute force search for query: ['test']. Error: Constructor parameter should be str
- Request times out, returns empty response

### With `LLM_PROVIDER=custom` config (per docs):
- Search hangs for 30+ seconds
- **Zero logs** — request accepted, then complete silence
- Request times out, returns empty response

## Environment
- **Cognee Version:** `0.5.5-local` (Docker image `cognee/cognee:latest`)
- **Deployment:** Docker on Windows server 2022/WSL2
- **LLM Provider:** vLLM (local) via OpenAI-compatible endpoint
- **Embedding Provider:** Ollama (local) - nomic-embed-text
- **Vector Store:** LanceDB (file-based)
- **Graph Store:** NetworkX (in-memory)
- **Data:** 175 text chunks uploaded successfully
- **Cognify Status:** Appears to have processed (entity extraction logs visible)

### Working Config (current) env # 
#LLM Config
LLM_PROVIDER=custom
LLM_MODEL=hosted_vllm/Qwen3-8B
LLM_ENDPOINT=http://host.docker.internal:8000/v1
LLM_API_KEY=.
LLM_INSTRUCTOR_MODE=json_schema_mode

# Embedding Config
EMBEDDING_PROVIDER=ollama
EMBEDDING_MODEL=nomic-embed-text
[...]

PR #2969, opened 2026-06-04
fix(llm): bound structured-output calls + tolerate preflight parse failures (#2456, #2752, #2796, #2902)

Vasilije1990:
## Why
Recurring root cause from a 6-month issue triage: LLM/embedding calls could hang indefinitely (#2119, #2362, #2456, #2752, #2796, #2902).

## What
Assessment found the **embedding paths** (LiteLLM/OpenAICompatible/Ollama engines) and the **tenacity retry decorators** were already fixed (timeouts + `retry_if_not_exception_type` for auth/validation). Two real gaps remained and are fixed here:

- **#2456 / #2796 / #2902** — structured-output adapters (Generic/CUSTOM/vLLM, OpenAI, Ollama) had no per-call timeout, so `cognify`/`search`/`remember` could hang forever. Bounded centrally at `LLMGateway.acreate_structured_output` via a new `LLM_CALL_TIMEOUT_SECONDS` (default 300; `<=0` disables). One change covers all providers + instructor/BAML.
- **#2752** — the preflight `test_llm_connection` re-raised any error from a `str`-response round-trip, so healthy non-OpenAI/local providers failed the check. Now only timeouts / genuine connection / auth errors are fatal; parse failures are treated as reachable.

## Tests
4 new tests in `test_llm_call_timeouts.py` (verified the gateway test fails at 8s without the wrapper and raises in ~1s with it; preflight test fails when tolerant clauses reverted). `ty` clean; ruff clean.

## Risk
The 300s default could cut off very-long completions — operators raise/disable via `LLM_CALL_TIMEOUT_SECONDS`. Centralized at the gateway, so a future adapter bypassing it wouldn't inherit the bound.

## Summary by CodeRabbit

* **New Features**
  * Configurable per-call LLM timeout (default 300s); timed-out calls raise a clear TimeoutError and prevent post-call usage recording.

* **Bug Fixes**
  * Connection preflight now treats structured-output/parse failures as reachable (unless a timeout/auth/connection error occurs), reducing false negatives.

* **Tests**
  * Added unit tests covering timeout behavior, disabling timeouts, and preflight connection semantics.

### DIFF
--- a/cognee/infrastructure/llm/LLMGateway.py
+++ b/cognee/infrastructure/llm/LLMGateway.py
@@ -1,3 +1,4 @@
+import asyncio
 from collections.abc import Coroutine
 from typing import Any, TypeVar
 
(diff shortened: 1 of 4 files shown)


dexters1 (changes requested):
Even with Base10 we have multiple times per day where the LLM response takes longer than 300s (sometimes 600s) but it does reply in the end.

The amount of traffic and requests the model needs to handle makes the response time slower, for non corporate users this will happen even more often.

Instead of killing the request on timeout can we give the users a warning that their requests are taking a long time to complete with additional info there? 

Or we can make it an optional feature that requests will be killed, but turned off by default

Adding this and killing requests will break a lot if not most heavy load cognify runs

PR #3006, opened 2026-06-07
fix(retrieval): add distance-based filtering to extract_relevant_node_ids (fixes #2720)

### LINKED ISSUE #2720: Graph-completion retrieval returns identical subgraph regardless of query (LanceDB filter not propagating)
## Summary

Search queries through `/api/v1/search` return an identical graph subgraph regardless of query content. Distinct queries — semantically relevant and semantically orthogonal — produce the same LLM context, because the graph expansion step appears to project the full dataset subgraph instead of a subgraph filtered around LanceDB's top-K hits.

## Environment

- Cognee 1.0.2
- Storage: LanceDB (vector), Kuzu (graph), SQLite (metadata)
- Embedding: `snowflake-arctic-embed2:latest` via Ollama (1024-dim)
- LLM: `phi4:14b` via Ollama
- Self-hosted, single-node deployment

## Symptom

Example: a dataset with 44 DocumentChunks, 903 graph nodes, 2,219 edges total.

Two orthogonal queries via `/api/v1/search` (any `query_type` — INSIGHTS, CHUNKS, GRAPH_COMPLETION, SUMMARIES all reproduce):

| Query | Response |
|---|---|
| "feedback weight poisoning LanceDB" | `"The document contains information about a code change suggestion for the MIN_CONVICTION value."` |
| "penguins in Antarctica eat fish" | `"The document contains information about a code change suggestion for the MIN_CONVICTION value."` |

The response text is essentially identical (sometimes paraphrased but always pointing to the same document).

### Write layer is healthy

Direct inspection of LanceDB via `lancedb` Python client (bypassing cognee's retriever):

```python
tbl = db.open_table("DocumentChunk_text")
# Row count: 44
# Schema: id string, vector fixed_size_list<float>[1024], payload struct<...>

# First 5 rows:
# Row 0: dim=1024 norm=1.0000 mean=0.000268 std=0.031249
# Row 1: dim=1024 norm=1.0000 mean=0.000294 std=0.031249
# Row 2: dim=1024 norm=1.0000 mean=0.000273 std=0.031249
# Row 3: dim=1024 norm=1.0000 mean=0.000321 std=0.031248
# Row 4: dim=1024 norm=1.0000 mean=0.000303 std=0.031249

# Pairwise cosine similarity (excluding diagonal):
# 0.341, 0.405, 0.258, 0.435, 0.381, 0.398, 0.379, 0.266, 0.495, 0.313
# → healthy diversity range
```

### Direct LanceDB search works correctly

```python
# Query 1: "feedback weight poisoning LanceDB"
tbl.search(qv).limit(5)
# Top result: "Cognee Operational Gotchas" doc (distance 1.2625)
# Distances: 1.26 - 1.38 (tight, matching topic)

# Query 2: "penguins Antarctica fish"
tbl.search(qv).limit(5)
# Top result: same "Cognee Operational Gotchas" doc BUT distance 1.6705
# Distances: 1.67 - 1.75 (looser, no real match)
# Different ordering + different set in top-5 from Query 1
```

So LanceDB returns query-dependent top-K with meaningful distance differences. The write layer and vector index are both sound.

### Cognee's retriever is the broken layer

Cognee container logs during the above two queries show:
```
Vector collection retrieval completed: Retrieved distances from 6 collections in 0.30s
Retrieving ID-filtered graph from database.
ID-filtered retrieval: 503 nodes and 1496 edges in 0.03s
Graph projection completed: 503 nodes, 1496 edges in 0.02s
Completed resolving edges to text — node_count: 15, connection_count: 10
```

**Identical `503 nodes / 1496 edges` for every query**, across all four search types tested. The 503 is also consistent with the dataset's total node count range (903 total nodes, so ~55% of the graph — consistent with a broad dataset-scoped projection rather than a query-targeted one).

This strongly suggests the filter path from LanceDB results → `relevant_node_ids` → `project_graph_from_db` is not being applied. The graph projection step is pulling a dataset-wide subgraph instead of a subgraph filtered around the LanceDB hits.

## Suspected code path

Skimmed the retrieval pipeline:

- `cognee/api/v1/search/search.py` → retriever class
- `cognee/modules/retrieval/graph_completion_retriever.py:165` → `brute_force_triplet_search(...)` call with `wide_search_top_k=self.wide_search_top_k` (default 100)
- `cognee/modules/retrieval/utils/brute_force_triplet_search.py` → `_get_top_triplet_importances` → `get_memory_fragment` → `project_graph_from_db` (or `project_neighborhood_from_db` if `neighborhood_depth` set)

In `get_memory_fragment`, when `neighborhood_depth is None` (the default path), it calls `project_graph_from_db` with `relevant_ids_to_filter=relevant_node_ids`. If `relevant_node_ids` is `None` — which happens when `wide_search_limit is None` in `_get_top_triplet_importances` — the projection covers the full dataset subgraph instead of a filtered one.

Whether `wide_search_limit` gets a value depends on the caller chain. The observed symptom (constant 503-node projection across queries) is consistent with `wide_search_limit` being `None` at the projection call site.

Worth checking: is there a path where `wide_search_top_k=100` on the retriever doesn't translate to `wide_search_limit=100` inside `_get_top_triplet_importances`? A parameter name mismatch or pass-through bug along that chain would produce exactly this symptom.

## Expected behavior

Graph projection should be filtered to a subgraph constructed around the LanceDB top-K node IDs (seed + N-hop neighborhood), not the full dataset subgraph. Different queries → different seed IDs → different projected subgraphs → different LLM context → different responses.

## Reproducer sketch

1. Cognify a dataset with ~50 heterogeneous documents (mixed topics).
2. Verify via direct `lancedb` Python client that embeddings are diverse (pairwise cosine similarity has meaningful spread).
3. Query `/api/v1/search` with two topically-orthogonal strings.
4. Observe identical or near-identical responses.
5. Check container logs for the `ID-filtered retrieval: N nodes and M edges` line — if N and M are identical across queries (and on the order of the full dataset size rather than a small seed-neighborhood), the filter isn't propagating.

## Workaround

Direct LanceDB similarity search + external graph walk returns the expected query-dependent results. Bypassing `/api/v1/search` with a custom thin retriever restores correct behavior while keeping cognee's ingestion pipeline (which works correctly and is the valuable part of the system).

## Why this is important

This is a **silent** failure mode. The `/api/v1/search` endpoint returns 200 OK with plausible-looking text. There is no error, no warning, no status change. Graph + LanceDB both appear healthy. Only a side-by-side comparison of responses to orthogonal queries surfaces the bug. Users unaware of this bug will operate on degraded retrieval for extended periods without realizing it.

Willing to help diagnose further if maintainers would find it useful.

yubingz:
## Summary

Fixes #2720 — Graph-completion retrieval returns identical subgraph regardless of query.

### Root Cause

When `wide_search_top_k` (default 100) exceeds the number of entries in a vector collection (common on small datasets), the vector search returns **all** entries regardless of query relevance. `extract_relevant_node_ids()` then collects every node ID from every collection, making the graph filter ineffective — the ID-filtered projection returns the same subgraph for every query.

Example: a dataset with 44 DocumentChunks and ~50 entities. With `limit=100`, ALL entries are returned from each collection regardless of query. The `extract_relevant_node_ids()` method collects 500+ IDs (more than half the graph), and the `WHERE n.id IN $target_ids OR m.id IN $target_ids` Cypher clause returns most of the graph (503/903 nodes in the issue reporter's case), making results query-independent.

### Fix

Add an optional `max_distance` parameter (default **1.5** cosine distance) to `NodeEdgeVectorSearch.extract_relevant_node_ids()`. Results whose cosine distance exceeds the threshold are excluded from `relevant_node_ids`, so the graph projection is actually filtered to query-relevant nodes.

Cosine distance ranges from 0 (identical) to 2 (opposite). A threshold of 1.5 means "closer than random" — orthogonal vectors have cosine distance ~1.0, so 1.5 keeps relevant results while filtering out clearly unrelated hits.

### Changes

| File | Change |
|------|--------|
| `node_edge_vector_search.py` | Add `max_distance` param to `extract_relevant_node_ids()`, filter by score, add diagnostic logging |
| `brute_force_triplet_search.py` | Add `wide_search_max_distance` param, pass through to `_get_top_triplet_importances` |
| `graph_completion_retriever.py` | Add `wide_search_max_distance` param, pass through to `brute_force_triplet_search` |
| `search.py` (API + methods) | Thread `wide_search_max_distance` through the search pipeline |
| `get_search_type_retriever_instance.py` | Pass `wide_search_max_distance` to all graph-based retrievers |
| `test_extract_relevant_node_ids.py` | 9 unit tests covering filter logic, edge cases, and backward compatibility |

### Backward Compatibility

Setting `wide_search_max_distance=None` restores the old behaviour (return all IDs regardless of score). The default is 1.5, which is a conservative threshold that filters clearly irrelevant results while preserving all potentially relevant ones.

### Testing

```python
# Demonstration of the fix
vs = NodeEdgeVectorSearch.__new__(NodeEdgeVectorSearch)
vs.query_list_length = None
vs.node_distances = {
    "Entity_name": [
        _make_scored_result("id1", 0.5),   # relevant
        _make_scored_result("id2", 1.8),   # irrelevant
        _make_scored_result("id3", 1.5),   # boundary
    ]
}

# Old behaviour (max_distance=None): returns all 3 IDs
# New behaviour (max_distance=1.5): returns only id1 and id3
```

### Relationship to PR #2998

PR #2998 fixed the PostgresAdapter by adding `get_id_filtered_graph_data()`. This PR addresses a different layer of the same problem: even when the adapter correctly implements ID-filtered retrieval, the IDs passed to it can be too many (covering most of the graph) if the vector search returns everything. Both fixes are complementary.

## Release Notes

* **New Features**
  * Added a configurable `wide_search_max_distance` parameter to search operations, enabling fine-tuned control over vector distance filtering with a default threshold of 1.5 cosine distance.

* **Tests**
  * Added comprehensive unit tests for distance-based filtering in vector search, validating backward compatibility, threshold behavior, and edge cases across multiple scenarios.


e-hosseini:
The `extract_relevant_node_ids` fix itself seems right. 

However `get_search_type_retriever_instance` always puts `wide_search_max_distance` into the kwargs for the decomposition/cot/context-extension/summary/temporal retrievers now, but only `GraphCompletionRetriever` got updated to actually accept that param. The other five have closed `__init__` signatures, no `**kwargs`, so `retriever_cls(**retriever_args)` will raise a TypeError as soon as any of those search types get used.

There's an existing test in `test_get_search_type_retriever_instance.py` (`test_graph_search_retrievers_receive_feedback_influence`, parametrized across all six types, unmocked) that should fail on 5/6 cases once CI runs. The "Test Suites" workflow is currently sitting in "awaiting approval," which is probably why this hasn't shown up as a red yet.

Just thread the param through the same way `wide_search_top_k` already is, on each of the five subclasses. Example for the decomposition one:

```diff
         node_name_filter_operator: str = "OR",
         wide_search_top_k: Optional[int] = 100,
+        wide_search_max_distance: Optional[float] = 1.5,
         triplet_distance_penalty: Optional[float] = 6.5,
...
             node_name_filter_operator=node_name_filter_operator,
             wide_search_top_k=wide_search_top_k,
+            wide_search_max_distance=wide_search_max_distance,
             triplet_distance_penalty=triplet_distance_penalty,
```

Same two-liner needed in `GraphCompletionCotRetriever`, `GraphCompletionContextExtensionRetriever`, `GraphSummaryCompletionRetriever`, and `TemporalRetriever` , same shape in all of them as far as I checked.

This test would also help, since the existing ones fail with a TypeError pointing at `feedback_influence` rather than the actual cause:

```python
@pytest.mark.asyncio
@pytest.mark.parametrize(
    ("search_type", "expected_class"),
    [
        (SearchType.GRAPH_COMPLETION, GraphCompletionRetriever),
        (SearchType.GRAPH_COMPLETION_DECOMPOSITION, GraphCompletionDecompositionRetriever),
        (SearchType.GRAPH_COMPLETION_COT, GraphCompletionCotRetriever),
        (SearchType.GRAPH_COMPLETION_CONTEXT_EXTENSION, GraphCompletionContextExtensionRetriever),
        (SearchType.GRAPH_SUMMARY_COMPLETION, GraphSummaryCompletionRetriever),
        (SearchType.TEMPORAL, TemporalRetriever),
    ],
)
async def test_graph_search_retrievers_accept_wide_search_max_distance(search_type, expected_class):
    import cognee.modules.search.methods.get_search_type_retriever_instance as mod

    retriever_instance = await mod.get_search_type_retriever_instance(
        search_type, query_text="q", wide_search_max_distance=0.8,
    )
    assert isinstance(retriever_instance, expected_class)
    assert retriever_instance.wide_search_max_distance == 0.8
```

Ran this against current `main` and it fails with `TypeError: __init__() got an unexpected keyword argument 'wide_search_max_distance'` on 5/6 of the parametrized cases.

Vasilije1990:
Conflicts with `main` are resolved and review comments addressed:

- **@e-hosseini's TypeError** — `wide_search_max_distance` is now threaded through all five remaining subclasses (`GraphCompletionDecompositionRetriever`, `GraphCompletionCotRetriever`, `GraphCompletionContextExtensionRetriever`, `GraphSummaryCompletionRetriever`, `TemporalRetriever`), same shape as `wide_search_top_k`. Also added the suggested parametrized test (`test_graph_search_retrievers_accept_wide_search_max_distance`) covering all six graph search types.
- **Silent drop / API TypeError** — while fixing the above I found the module-level `search()` in `cognee/modules/search/methods/search.py` never accepted the param (the API layer passes it, which would have raised `TypeError` on every `cognee.search()` call), and `search_in_datasets_context` accepted it but never forwarded it to `get_retriever_output`, so user-supplied values were silently replaced by the default. Both threading gaps are fixed.
- **Test placement** — `test_extract_relevant_node_ids.py` was added under repo-root `tests/unit/`, which CI never runs (`pytest cognee/tests/unit/`). Moved it to `cognee/tests/unit/modules/retrieval/` and removed the `sys.modules` stubbing, which would have overwritten attributes on real modules (e.g. `get_vector_engine`) for the rest of a shared pytest session.
- **CodeRabbit docstrings** — `wide_search_max_distance` documented in the api/v1 `search`, `authorized_search`, `search_in_datasets_context`, and `GraphCompletionRetriever.__init__` docstrings.

Locally: `pytest cognee/tests/unit/modules/retrieval cognee/tests/unit/modules/search` → 435 passed, 1 skipped; ruff format/check clean.

PR #3011, opened 2026-06-08
fix(graph): key edges by relationship type, not field name

rshkarin:
### What this fixes

`get_graph_from_model` flattens a `DataPoint` tree into `(nodes, edges)` and dedups edges with a key built from the **field name**:

    edge_key = f"{data_point_id}_{target_datapoint.id}_{field_name}"

but stores and later dedups them by the resolved **relationship name** (`Edge.relationship_type`, falling back to the field name).

The two keys disagree, with two consequences:

- **Silent edge loss.** When one field holds several edges to the *same* target with *different* relationship types, e.g.

      employees = [
          (Edge(relationship_type="manages"), bob),
          (Edge(relationship_type="mentors"), bob),
      ]

  both produce the key `…_bob_employees`, so only the first edge is emitted and `mentors` is dropped before it ever reaches `deduplicate_nodes_and_edges` or the graph DB.

- **Spurious duplicates.** The same `(source, target, relationship)` reached through two different fields gets two different keys, so a duplicate is emitted that only the later dedup pass / `ON CONFLICT DO NOTHING` removes.

### The fix

Key the dedup set on `relationship_name` instead of `field_name`:

    edge_key = f"{data_point_id}_{target_datapoint.id}_{relationship_name}"

This is the same edge identity already used by `deduplicate_nodes_and_edges` (`source + relationship + target`) and the graph adapters. Distinct relationships between a node pair now survive, identical ones collapse at the source, and `added_edges` becomes consistent with the `visited_properties` cycle key (which already uses `relationship_name`). Behaviour is unchanged when `field_name == relationship_name`, the common case.

### Tests

Added to `cognee/tests/unit/interfaces/graph/get_graph_from_model_unit_test.py`:

- `test_same_field_same_target_distinct_relationships_both_survive` — two relationship types to the same target via one field are both emitted (fails before the fix: `['manages'] != ['manages', 'mentors']`).
- `test_same_triple_via_two_fields_collapses_to_single_edge` — same triple via two fields collapses to one edge.
- `test_distinct_relationships_survive_deduplication` — both survive the `get_graph_from_model → deduplicate_nodes_and_edges` hand-off used by `add_data_points`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Edge deduplication now collapses duplicates by (source, target, relationship) so distinct relationship types between the same nodes are preserved and redundant edges are removed.

* **Tests**
  * Added regression tests covering multiple relationship types to the same target and deduplication across multiple fields to prevent edge loss.

### DIFF
--- a/cognee/modules/graph/utils/get_graph_from_model.py
+++ b/cognee/modules/graph/utils/get_graph_from_model.py
@@ -277,8 +277,13 @@ async def get_graph_from_model(
     ):
         relationship_name = _get_relationship_key(field_name, edge_metadata)
 
-        # Create edge if not already added
-        edge_key = f"{data_point_id}_{target_datapoint.id}_{field_name}"
+        # Create edge if not already added.
+        # Key on relationship_name (not field_name) so that distinct relationship
+        # types between the same pair of nodes are preserved, and so the same
+        # (source, target, relationship) triple reached via different fields
+        # collapses to a single edge — matching the edge identity used by
+        # deduplicate_nodes_and_edges and the graph DB.
+        edge_key = f"{data_point_id}_{target_datapoint.id}_{relationship_name}"
         if edge_key not in added_edges:
             edge_properties = _create_edge_properties(
                 data_point.id, target_datapoint.id, relationship_name, edge_metadata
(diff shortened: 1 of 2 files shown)


dexters1:
@rshkarin just to note changing ID of any graph/vector or relational element requires running migrations for the change for existing Graph/Vector and Relational stores.

I am currently working on an effort to create a migration system for our graph/vector stores, if we merge changes proposed here we need to create migrations for them as well to update existing stores

dexters1 (changes requested):
Checked this with @lxobr we might want to avoid creating multiple edges between nodes, the choice to deduplicate edges was intentional originally. Would be great to hear more examples of where we would need multiple edges between two nodes though. 

Marking this as request changes for now to avoid accidental merge and moved it to draft

PR #3020, opened 2026-06-09
fix: clearer local/MCP setup errors (json extension, unsupported files, client errors)

Vasilije1990:
Addresses three rough edges reported during local + Cognee MCP setup.

## 1. `recall` fails with `Extension: json ... has not been installed`

Kuzu/Ladybug's `json` extension was installed best-effort against a throwaway DB and **silently swallowed** at two layers, so a network-restricted MCP container never actually got it — and the failure only surfaced later as a cryptic Binder error when `recall`/temporal queries run `json_extract(...)`.

- **`cognee-mcp/Dockerfile`**: pre-install the extension at **build time** (network available) so it's baked into the image. Runs the same install path the app uses, as the same non-root user with a stable `HOME=/app`, so the cached extension lands exactly where the running server looks for it.
- **`ladybug/adapter.py`**: on the real connection, fall back to `INSTALL JSON; LOAD JSON;` when `LOAD` alone fails (idempotent, no-op when cached), and emit a **loud, actionable warning** instead of a silent `info` log when it genuinely can't load — pointing at the network/pre-install/`INSTALL json; LOAD json;` fixes.

## 2. Uploading an unsupported file (e.g. `.pptx`) returned `ERROR: {}`

Two compounding bugs:
- The loader raised `No loader found for file: <path>` with no hint that `.pptx` needs an optional loader.
- `PipelineRunErrored` carries the error on `payload` (`repr(error)`), but the add router read a non-existent `error` attribute and fell back to the model repr — so the client got an unhelpful body.

Fixes:
- **`LoaderEngine`**: new error names the extension, lists supported extensions, and tells the user which extra to install for office/doc formats — e.g. ``'.pptx' files need an optional document loader... Install it with `pip install cognee[docling]` (or `cognee[unstructured]`)``.
- **add router** (`get_add_router.py`): surface `add_run.payload` as the error detail.

## 3. Frontend logged `[LOCAL-API] ✗ ... ERROR: {}`

`handleServerErrors` rejected with a **plain object**, which serializes to `{}` in `console.error`.
- **`handleServerErrors.ts`**: reject with a real `Error` carrying the server's `detail`/`error`/`message`, while still copying parsed fields onto it so callers reading `error.detail` keep working.

## Testing
- `ruff` / `pre-commit` clean; Python files compile; `tsc` clean on the frontend file.
- Loader message verified locally: `.pptx` →
  `No loader found for file '...deck.pptx' (extension '.pptx'). '.pptx' files need an optional document loader that is not installed. Install it with \`pip install cognee[docling]\` (or \`cognee[unstructured]\`) and retry. Supported extensions: ...`

## Notes
- The Dockerfile pre-install is best-effort (`|| echo …`): a failed build-time download won't break the image, and the adapter fallback installs on first run if the container has network.
- The `json`-extension fix targets the default Ladybug/Kuzu backend; other graph backends are unaffected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Server and authentication errors now display more detailed, actionable messages
  * Pipeline failures provide clearer error feedback to users
  * File loader errors include available formats and installation instructions when applicable

* **Improvements**
  * JSON extension automatically installs as fallback when initial load fails, improving database compatibility

### DIFF
--- a/cognee-frontend/src/utils/handleServerErrors.ts
+++ b/cognee-frontend/src/utils/handleServerErrors.ts
@@ -1,5 +1,32 @@
 import { redirect } from "next/navigation";
 
+// Build a real Error (not a bare object) from a server error response.
+//
+// Rejecting with a plain object is why callers logged "ERROR: {}" — a plain
+// object has no useful default serialization in console.error. An Error
+// serializes with its message (and stack). The server's parsed JSON fields
+// (detail/error/message/...) are copied onto the Error so existing callers that
+// read `error.detail`/`error.error`/`error.message` keep working.
+function buildServerError(body: string, status: number, statusText: string): Error {
+  let parsed: Record<string, unknown> = {};
+  if (body) {
+    try {
+      parsed = JSON.parse(body);
+    } catch {
+      parsed = { message: body };
+    }
+  }
+  const detail = parsed.detail ?? parsed.error ?? parsed.message;
+  const message =
+    typeof detail === "string" && detail.trim()
+      ? detail
+      : (body && body.trim()) || statusText || `Request failed with status ${status}`;
+  const error = Object.assign(new Error(message), parsed, { status, statusText });
+  // Ensure the resolved message wins even if `parsed` carried its own `message`.
+  error.message = message;
+  return error;
+}
+
 export default function handleServerErrors(
   response: Response,
   retry: ((response: Response) => Promise<Response>) | null = null,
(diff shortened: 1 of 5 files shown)


zain-sajid:
Hey @Vasilije1990,

I still ran into the `Extension: json ... has not been installed` on my machine (macOS) when [setting up the MCP locally using docker](https://docs.cognee.ai/cognee-mcp/mcp-quickstart):

Things I tried:

- Deleted the old image and pulled the latest — JSON error persisted (image doesn't actually bundle the extension).
- Added --dns 8.8.8.8 — got past DNS but hit a cache-dir creation failure.

What finally fixed it was setting HOME=/tmp (even without --dns).

`docker run -d --rm -p 8000:8000 \
  -e TRANSPORT_MODE=http -e HOME=/tmp \
  --env-file .env --name cognee-mcp cognee/cognee-mcp:main`

PR #3024, opened 2026-06-10
feat(migration): import memory from other providers, export to portable COGX format

Vasilije1990:
## What

Bidirectional memory migration, built hub-and-spoke around **COGX** (Cognee eXchange format, a versioned JSONL archive) instead of N×M pairwise converters.

### Import — `cognee.remember()` accepts provider sources

```python
from cognee.migration import Mem0Source, GraphitiSource, LettaSource, COGXArchiveSource

await cognee.remember(Mem0Source("mem0_export.json"))
await cognee.remember(GraphitiSource("graphiti_dump.json", mode="hybrid"))
await cognee.remember(LettaSource("agent.af", mode="preserve"))
```

A `MemorySource` is any object that async-yields COGX records — `remember()` routes it via the same type-dispatch already used for `MemoryEntry`. Three fidelity modes per source:

- **re-derive** (default): ingest raw content (episodes, documents, memories), run cognify — Cognee's own extraction builds the graph
- **preserve**: map the source's already-extracted entities/facts directly to native `Entity` DataPoints + custom edges via `add_data_points` — zero LLM calls
- **hybrid**: both (default for Graphiti, which keeps verbatim episodes *and* a derived graph)

### Export — `cognee.export()` in the SDK

```python
await cognee.export("main_dataset", format="graphml")   # also: cogx / json / cypher
await cognee.remember(COGXArchiveSource("backup_cogx"))  # restore / instance migration
```

### Push — local instance → Cognee Cloud

```python
await cognee.push("main_dataset")                # or: cognee-cli push main_dataset
await cognee.push("main_dataset", target_dataset="prod", mode="hybrid")
result = await cognee.push("big_dataset", run_in_background=True)  # poll result.pipeline_run_id
```

`push` composes the two halves above into a local→cloud migration: export the dataset as a COGX archive → tar → upload through the existing `CloudClient.remember()` multipart endpoint with `content_type="cogx-archive"` → the remember endpoint unpacks it (path-traversal safe) and routes it through `import_memory_source(COGXArchiveSource(...))`. Default `mode="preserve"` lands the locally extracted graph with zero LLM calls; auth reuses the `serve` stack (live connection → `~/.cognee/cloud_credentials.json` → `COGNEE_SERVICE_URL`/`COGNEE_API_KEY`), so the UX is `cognee-cli serve` once, then `cognee-cli push` any time. Unlike `sync` (raw files + remote re-cognify), push preserves the existing graph and is additive/idempotent on re-runs.

## Design notes

- **Idempotent by construction**: deterministic `data_id`s (uuid5 of `external_system:external_id`) and `generate_node_id(name)` entity ids — re-running an interrupted import is safe, and preserved facts merge into the existing graph vocabulary instead of forming a parallel graph
- **Temporal truth preserved**: Zep/Graphiti `valid_at`/`invalid_at` are carried onto edge properties, so the data is in place when bi-temporal search lands
- Imports run as ordinary pipelines (`migration_import_pipeline`) — batching, OTel spans, run status, ledger provenance all inherited
- Provider parsers are file/dict based and tolerant of key-name variants; live-API connectors are follow-ups

## Testing

- 108 unit tests (`cognee/tests/unit/migration/` + CLI), pure — no DB, no LLM, no network: archive transport round-trip, tar path-traversal/symlink/decompression-bomb rejection, a full export→import round-trip regression suite (raw-node fidelity, no fabricated UUID entities, bounded batching, version gating), the remember-router archive branch via TestClient, import orchestration dispatch per mode, client-resolution precedence, and mocked end-to-end push
- End-to-end smoke (mock embeddings, zero LLM): preserve-mode Graphiti import → idempotent re-run → JSON export with `invalid_at` intact → COGX export → restore round-trip → GraphML/Cypher emits

## Follow-ups (not in this PR)

Live-API connectors, `VectorDBInterface.scroll()` + embeddings/blobs/relational rows in COGX archives (push currently re-embeds on the receiving side), REST export endpoint, external-ID map table, reverse emitters (Cognee → Mem0/Zep/.af), delta/tombstone protocol for push (currently additive-only)


cognee/modules/migration/cmif.py
```diff
@@ -0,0 +1,231 @@
...
+
+
+def read_manifest(directory: Union[str, Path]) -> Optional[CMIFManifest]:
+    manifest_path = Path(directory) / MANIFEST_FILE
+    if not manifest_path.exists():
+        return None
+    return CMIFManifest.model_validate_json(manifest_path.read_text(encoding="utf-8"))
+
+
+def read_archive(directory: Union[str, Path]) -> Iterator[CMIFRecord]:
```
NMZivkovic: Do we need to read nodes.jsonl here for re-import to work?
If I understand it correctly, if we do export and then remmember it will scramble imported nodes.

cognee/api/v1/export/__init__.py
NMZivkovic, line 1: Add at least one e2e test for export->archive->re-import so we are ready for the future fixes with backwards compatibility tests.

NMZivkovic (approved):

NMZivkovic (changes requested):
1. Add at least one happy path regression test.
2. Potential data loss

PR #3034, opened 2026-06-11
fix: background remember fails on uploads with 'I/O operation on closed file'

LStromann:
## Summary
- With `run_in_background=true`, `POST /v1/remember` returns the response before ingestion runs; Starlette then closes each `UploadFile`'s temp file on request teardown, so the background pipeline fails every item with `I/O operation on closed file` → `PipelineRunFailedError (422)`. The endpoint still reports 200/"running", so uploads silently never appear in the dataset.
- Fix: before spawning the background task, copy each upload into a pipeline-owned `SpooledTemporaryFile` (`_detach_request_files`), preserving the filename via a small `UploadFile` duck-type. Blocking mode is untouched — the request is still open there.
- Affects any client uploading files with background mode, including the cloud UI which always sends `run_in_background=true`.

## Test plan
- [x] New unit tests: detached uploads survive closing the original handle, read position independence, non-file items pass through
- [x] `pytest cognee/tests/unit/api/` passes locally
- [ ] End-to-end: `curl -F 'data=@file.pdf' -F 'datasetName=x' -F 'run_in_background=true' /api/v1/remember` then confirm the document appears in the dataset and the pipeline run completes

### DIFF
--- a/cognee/api/v1/remember/remember.py
+++ b/cognee/api/v1/remember/remember.py
@@ -138,6 +138,39 @@ def _data_to_text(data) -> str:
     return f"[{type(data).__name__}]"
 
 
+class _DetachedUpload:
+    """Duck-types an UploadFile (.file / .filename) with a pipeline-owned buffer."""
+
+    def __init__(self, file, filename):
+        self.file = file
+        self.filename = filename
+
+
+def _detach_request_files(data):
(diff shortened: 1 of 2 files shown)


LStromann:
Closing — investigating the regression introduced by the dev merge instead of adding a new detach mechanism.

PR #3057, opened 2026-06-12
fix(ladybug): retry LOAD EXTENSION with INSTALL on the live connection

Vasilije1990:
## What

Fixes the CI failure where `cognee.add()` dies during graph-engine init with:

```
RuntimeError: Binder exception: Extension: json is an official extension and has not been installed.
You can install it by: install json.
```

## Why it happens

`LadybugAdapter.create_subprocess` warms up the JSON extension via `install_json_extension_local`, which runs `INSTALL JSON;` on a throwaway database and **swallows every failure** (by design — already-installed and offline cases both surface as raises there). When the install fails for a real reason — e.g. a transient network error downloading the extension on a fresh CI runner with an empty extension cache — the warm-up silently does nothing, and the subsequent `LOAD EXTENSION JSON;` in the worker hard-fails with the binder error above. Because the swallow is silent, the root cause never appears in logs.

## Fix

- `cognee_db_workers/kuzu_worker.py::_load_extension`: catch exactly the "has not been installed" binder error, run `INSTALL <ext>;` on the **live** connection, and retry the LOAD once. Unrelated load errors propagate unchanged. If the install genuinely can't succeed (e.g. no network at all), the INSTALL now raises with the real cause instead of the confusing downstream error.
- `cognee_db_workers/_kuzu_helpers.py`: the warm-up remains best-effort, but the swallowed exception is now printed to stderr so CI logs show *why* the warm-up failed.

## Testing

- 3 new pure unit tests (`test_kuzu_worker_load_extension.py`): direct load when installed, install-and-retry when not, unrelated errors propagate without an install attempt — no ladybug, no subprocess.
- Full `tests/unit/infrastructure/databases/` suite passes, including the worker import-hygiene allowlist test (no new imports in the worker).
- ruff check + format clean.

### DIFF
--- a/cognee/tests/unit/infrastructure/databases/test_kuzu_worker_load_extension.py
+++ b/cognee/tests/unit/infrastructure/databases/test_kuzu_worker_load_extension.py
@@ -1,0 +1,75 @@
+"""Unit tests for the Kuzu worker's LOAD EXTENSION retry-with-install path.
+
+The warm-up ``INSTALL JSON`` on a throwaway database is best-effort and can
+fail silently (e.g. a transient network error downloading the extension on a
+fresh CI runner). ``_load_extension`` must then install on the live
+connection and retry, instead of surfacing Ladybug's
+"Extension: json ... has not been installed" binder error.
+
+Pure tests: fake connection/registry, no ladybug, no subprocess.
+"""
+
+import pytest
+
+from cognee_db_workers.harness import Request
+from cognee_db_workers.kuzu_protocol import OP_LOAD_EXTENSION
+from cognee_db_workers.kuzu_worker import _load_extension
+
+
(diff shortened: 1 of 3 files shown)


Vasilije1990:
Added retry-with-backoff for the INSTALL itself (3 attempts, 2s apart): analysis of dev run [27425923988](https://github.com/topoteretes/cognee/actions/runs/27425923988) shows the failure is intermittent (same S3 job green at 12:22 and 12:55, failed at 15:35 on identical code — no infra changes in the commit range, ladybug 0.17.1 stable since Jun 2), i.e. a transient network failure downloading the extension. A single install attempt at LOAD time could hit the same blip; the retry rides it out, and each failed attempt is logged to stderr.

PR #3072, opened 2026-06-14
feat(cache): implement session-context CRUD on SqlCacheAdapter (sqlite/postgres)

Vasilije1990:
## Why

The session-context guidance layer on `feature/session-context-mvp` (PR #2989) added four abstract methods to `CacheDBInterface` — `create_session_context_entry`, `get_session_context_entries`, `update_session_context_entry`, `delete_session_context` — and implemented them on the **Redis, FS, and Tapes** adapters.

That branch was cut **before** #3028 made `SqlCacheAdapter` the **default (`sqlite`) cache backend**. As a result, the default backend does not implement the new interface, and once #2989 merges into `dev` the abstract methods would leave `SqlCacheAdapter` unimplementable — breaking session memory on the out-of-the-box backend (and Postgres).

This branch merges current `dev` into the feature branch and adds the missing SQL implementation so the guidance layer works on the default backend.

## What

- **New `cache_session_context` table** (`sql/tables.py`): append-only, holds kind-discriminated `"context"`/`"feedback"` dicts. The payload's `id` is promoted to an `entry_id` column for direct `UPDATE`, mirroring how `cache_qa_entries` promotes `qa_id`. Session + partial-expiry indexes match the existing tables.
- **Four methods on `SqlCacheAdapter`**, reusing the shared TTL helpers (sliding expiry on write, expired rows filtered on read) and the `FOR UPDATE` + deadlock-retry pattern for shallow-merge updates.
- **Lifecycle wiring**: session context is dropped by `delete_session` and `prune`, and swept by the throttled global TTL purge.
- **10 new unit tests** in `test_sql_adapter_crud.py`: insertion ordering, both entry kinds, shallow-merge update, missing-`entry_id` → `False`, missing-`id` rejection, per-session isolation, TTL set/refresh, and expiry exclusion.

## Testing

```
pytest cognee/tests/unit/infrastructure/databases/cache/   # 129 passed
pytest cognee/tests/unit/infrastructure/session/           # 106 passed
```

Behaviour matches the Redis/FS adapters: `update_*` returns `True`/`False` on hit/miss, `delete_session_context` reports whether live rows existed, reads exclude expired rows.


Vasilije1990:
Follow-up after an adversarial review against the Redis/FS reference impls (2nd commit):

- **Critical fix:** `SqlCacheAdapter.create_qa_entry` / `update_qa_entry` didn't accept `used_session_context_ids`, but the merged `CacheDBInterface` declares it and `SessionManager.add_qa` forwards it **unconditionally** — so *every* `add_qa` on the default sqlite backend would have raised `TypeError`. Now threaded through the full QA path. This was a second instance of the same "feature branch predates #3028" gap, on the QA side.
- **Minor parity fix:** `create_session_context_entry` no longer rejects an id-less payload (Redis/FS accept it); it falls back to a synthetic `entry_id` for the NOT NULL column without mutating the stored payload.

Verified `SqlCacheAdapter.__abstractmethods__ == []` (fully concrete). Tests: `cache/` + `session/` unit suites = **236 passed**.

Vasilije1990:
Pushed `0f2baffdb`: **default `auto_feedback=True`** so the session-context guidance layer is always on out of the box (the gate is `caching and auto_feedback`, and `caching` already defaults to `True`). No env vars required; `AUTO_FEEDBACK=false` opts out.

- **Trade-off:** adds one structured-output LLM call per answered turn (turn analysis / feedback detection). The path is fail-open — an LLM error degrades to an empty analysis rather than blocking the answer.
- Updated the config docstring + the cache-config default/`to_dict` tests.
- Verified end-to-end on the sqlite adapter: with zero env vars, a turn's durable guidance is extracted, persisted in `cache_session_context`, and rendered into the next turn's active-context block.

Heads-up unrelated to this change: two pre-existing integration tests in `test_session_manager_sql.py` (`..._feedback_only_no_new_qa`, `..._feedback_and_followup_adds_qa`) fail in my environment because they exercise the real feedback-detection LLM call; they fail identically with these commits stashed, so they're not introduced here.

PR #3075, opened 2026-06-15
fix(pipelines): bound run_info data to prevent unbounded pipeline_runs growth

### LINKED ISSUE #3074: pipeline_runs table grows without bound: run_info stores full input payload on every run
### What happened

I hit a local out-of-memory while running cognee and traced it to the `pipeline_runs` table in the SQLite relational store. It had grown to 6.4 GB across ~23k rows (averaging ~280 KB each). Everything else in the database (nodes, edges) was under 100 MB combined, so this single table was essentially the entire file.

### Cause

`log_pipeline_run_start`, `log_pipeline_run_complete` and `log_pipeline_run_error` all build `run_info` the same way:

```python
if not data:
    data_info = "None"
elif isinstance(data, list) and all(isinstance(item, Data) for item in data):
    data_info = [str(item.id) for item in data]
else:
    data_info = str(data)   # full payload stored verbatim
```

When `data` is not a list of `Data` records (e.g. raw text passed to `add`/`remember`), the entire payload gets stringified into `run_info["data"]` on every run. There is no size cap and no pruning, so the table grows without bound.

As far as I can tell the column is never read back from the database anywhere in the codebase. The places that look like readers operate on the in-memory `PipelineRunInfo` object returned from `cognify()`, and only touch `.status` / `.pipeline_run_id`. So `run_info["data"]` is write-only audit data.

### Impact

Opening the store pulls this large table into process memory, which is what caused the OOM on a memory-constrained machine.

This is the same failure mode that #2549 fixed for the `queries` / `results` search-history tables. `pipeline_runs` was not covered by that change, and the code is unchanged on `dev`.

### Suggested fix

Cap the stringified payload (truncate to a few hundred chars with a marker recording the original length), leaving the `Data`-list and empty-input branches as they are. I have a PR ready that does this behind a shared helper plus a unit test.

### Environment

cognee 1.1.2 (also confirmed present on `dev`), SQLite relational backend.

P7AC1D:
## Description

While debugging a local out-of-memory I traced it to the `pipeline_runs` table in the SQLite store growing to several GB (~23k rows, ~280 KB each). Every pipeline run logs a `run_info` row, and `log_pipeline_run_start` / `complete` / `error` store the input payload via the `str(data)` fallback whenever `data` is not a list of `Data` records. In my case raw text passed to `add` / `remember` was being stored verbatim on every run.

That column is never read back from the database anywhere (the apparent readers operate on the in-memory `PipelineRunInfo` returned by `cognify()` and only use `.status` / `.pipeline_run_id`), so it is write-only audit data that grows without limit, and opening the store pulls it into process memory.

This is the same failure mode #2549 fixed for the `queries` / `results` search-history tables; `pipeline_runs` was not covered there.

The fix extracts the shared branch into `summarize_run_info_data()` and caps the stringified payload at 512 chars with a truncation marker that records the original length. Empty input still maps to `"None"`, and lists of `Data` records still reduce to their ids, so existing behaviour is unchanged for those cases.

Closes #3074.

## Acceptance Criteria

* `run_info["data"]` no longer stores unbounded payloads: large inputs are truncated to 512 chars followed by a `[truncated, N chars total]` marker.
* Existing behaviour preserved for empty input (`"None"`) and lists of `Data` (list of ids).
* New unit test covers empty / Data-list / small / large payloads.

Local test run:

```
cognee/tests/unit/modules/pipelines/test_summarize_run_info_data.py ....   [4 passed]
```

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/modules/pipelines/operations/log_pipeline_run_complete.py
+++ b/cognee/modules/pipelines/operations/log_pipeline_run_complete.py
@@ -1,19 +1,15 @@
 from uuid import UUID
 from cognee.infrastructure.databases.relational import get_relational_engine
-from cognee.modules.data.models import Data
 from cognee.modules.pipelines.models import PipelineRun, PipelineRunStatus
 from typing import Any
+
+from cognee.modules.pipelines.utils import summarize_run_info_data
 
 
 async def log_pipeline_run_complete(
     pipeline_run_id: UUID, pipeline_id: UUID, pipeline_name: str, dataset_id: UUID, data: Any
 ):
-    if not data:
-        data_info = "None"
-    elif isinstance(data, list) and all(isinstance(item, Data) for item in data):
-        data_info = [str(item.id) for item in data]
-    else:
-        data_info = str(data)
+    data_info = summarize_run_info_data(data)
 
     pipeline_run = PipelineRun(
         pipeline_run_id=pipeline_run_id,
--- a/cognee/modules/pipelines/operations/log_pipeline_run_error.py
+++ b/cognee/modules/pipelines/operations/log_pipeline_run_error.py
@@ -1,8 +1,9 @@
 from uuid import UUID
 from cognee.infrastructure.databases.relational import get_relational_engine
-from cognee.modules.data.models import Data
 from cognee.modules.pipelines.models import PipelineRun, PipelineRunStatus
 from typing import Any
+
+from cognee.modules.pipelines.utils import summarize_run_info_data
 
 
 async def log_pipeline_run_error(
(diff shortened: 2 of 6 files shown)


Vasilije1990:
@P7AC1D does it make sense to truncate? This is useful for async runs when you want to get a status of the pipeline and understand how and where things are in the flow

We'll check internally and see to simplify, but in general, running sqlite is not recommended for prod workloads

dexters1:
@Vasilije1990 will merge it for now to prevent unbounded pipeline growth, we should see to refactor this in the future instead of truncating str input to provide data id properly to store in the DB so the full data can be reconstructed later based on ID

dexters1:
@P7AC1D Thank you for bringing this to our attention and contributing to the project!

PR #3093, opened 2026-06-18
fix(remember): materialize uploads before background ingestion (closed-file)

LStromann:
## Problem

`remember(..., run_in_background=True)` with a file / `UploadFile` input **silently ingests nothing**. The HTTP request closes the upload stream as soon as it returns, but the background task reads it afterwards:

```
Item 0 failed: I/O operation on closed file.
PipelineRunFailedError: Pipeline run failed. Data item could not be processed. (422)
```

No data is added and no graph is built — and because the caller already received its "running" result, the failure is only visible in logs. This is what breaks "upload a file" in a hosted/API setting where `run_in_background=True` is the norm.

## Fix

Before scheduling the background task (i.e. **still inside the request**, while the streams are open), materialize any file-like / `UploadFile` items to on-disk temp copies, preserving the original filename, and hand those stable paths to `add`/`cognify`. The temp dir is cleaned up when the task finishes. This mirrors the existing `content_type="skills"` materialization in this same module.

Blocking mode (`run_in_background=False`) is untouched — the streams are still open there.

## Repro

`POST /api/v1/remember` with a file upload + `run_in_background=true`:
- **Before:** background log shows `I/O operation on closed file` → `422`; dataset `data` stays empty.
- **After:** the upload is materialized in-request, so the background `add`/`cognify` ingests the file normally.

Independent of (and complementary to) the detached-user background fix in #3092.

### DIFF
--- a/cognee/api/v1/remember/remember.py
+++ b/cognee/api/v1/remember/remember.py
@@ -1063,6 +1063,35 @@ async def _remember_inner(
             await improve(**improve_kwargs)
 
     if run_in_background:
+        import os
+        import tempfile
+        from pathlib import Path as _Path
+
+        # The request closes UploadFile / file-like inputs as soon as it returns,
+        # but the background task reads them afterwards -> "I/O operation on closed
+        # file" and the upload silently ingests nothing. Materialize file-like
+        # uploads to on-disk temp copies NOW (still inside the request) so the
+        # background task ingests stable files; clean them up once it finishes.
+        _bg_tmp_dir: Optional[tempfile.TemporaryDirectory] = None
+        _items = data if isinstance(data, list) else [data]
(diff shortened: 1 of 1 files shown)


LStromann:
Closing in favor of #3095 (dexter), which fixes the same background closed-file bug with a cleaner, more general approach — a reusable `_materialize_stream_for_background` helper that buffers streams in-memory, correctly awaits async `UploadFile.read()`, and handles `DataItem`/list inputs. 

Note for whoever merges: #3095 addresses **Bug B (closed file)** only. The related **detached-user** background bug is separate and still needs #3092 — in a multi-tenant / `ENABLE_BACKEND_ACCESS_CONTROL` deployment, both must land together for background `remember` uploads to work.

Issue #2681, opened 2026-04-20
[Feature]: Enforce type hints and gradually fix/add types


nfnt:
### Problem Statement

Currently, the code in cognee has well defined type hints in almost all of its Python files. Not all of them are correct though and unfortunately these types aren't enforced by CI.

There are mentions of type checking with `mypy` and various files as well as an (probably outdated) `mypy.ini`. Therefore it seems that adding type checks to CI shouldn't be too hard.

### Proposed Solution

It's unrealistic to fix all types in the cognee code base at once, as it'll result in a very large diff that's hard to review. Instead, we can roll out these changes gradually until everything is covered.

I.e. a first step is to add type checking to CI, but don't have it check any files yet. In follow-up PRs more and more files are fixed for types and added to CI until most of the code is covered.

For type checking, let's use [ty](https://github.com/astral-sh/ty) instead of [mypy](https://www.mypy-lang.org/). While `ty` is currently in beta, it already covers almost everything the `mypy` does, while at the same time being very fast and providing a LSP. This means it can be used directly in an editor to give immediate feedback on wrong types. 

### Use Case

Having proper types and type checking improves the developer experience, as it eliminates type related errors when refactoring, provides a clearer function interface, and improves editor auto-completion.

Furthermore, agentic coding can benefit from the additional guard provided by type checking.

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this feature hasn't been requested already
- [x] I have provided a clear problem statement and proposed solution
- [x] I have described my specific use case

darshan-codes-ai:
Hi! I'd like to help with this. I'm comfortable with Python 
type hints and would like to start adding them to some functions. 
@Vasilije1990 @nfnt could you assign this to me or point me 
to a good starting file? @maintainers

priyanshihihi:
Hi! I'd love to contribute to this. I can start by setting up ty in the CI config (the first step you described), and then fix types in a few Python files in follow-up PRs. I'm comfortable with Python type hints and this seems like a great way to make a meaningful, incremental contribution. Tagging @Vasilije1990  and @nfnt t — could you assign this to me so I can get started? @maintainers

aryan-65548:
Hi! I'd love to work on the first step here — adding ty to CI with an empty include list so the infrastructure is in place without requiring all types to be fixed upfront. I can also fix types in a small initial set of files as a starting point. Could a maintainer assign this to me?@Vasilije1990

PR #3105, opened 2026-06-20
fix(types): make run_custom_pipeline type-clean and enforce ty checks (#2681)

pooja-bhavani:
### Summary

Continues the gradual `ty` type-checking rollout tracked in #2681 by making `cognee/modules/run_custom_pipeline` type-clean and adding it to the enforced `[tool.ty.src].include` set.

### Changes

- **Fix `invalid-parameter-default` errors**: `tasks` and `user` were annotated as non-optional but defaulted to `None`. Wrapped both in `Optional[...]` to match the actual default.
- **Fix a latent runtime bug**: with the default `tasks=None`, `[*tasks]` would raise `TypeError: argument of type 'NoneType' is not iterable`. Guarding with `*(tasks or [])` preserves behavior for valid inputs and makes the documented `None` default safe.
- **Enable type checking**: added `cognee/modules/run_custom_pipeline` to `[tool.ty.src].include`.

### Verification

- `uv run ty check .` → passes (exit 0)
- `uv run ruff check .` → passes
- `uv run ruff format --check .` → passes

Relates to #2681.

### DIFF
--- a/cognee/modules/run_custom_pipeline/run_custom_pipeline.py
+++ b/cognee/modules/run_custom_pipeline/run_custom_pipeline.py
@@ -12,10 +12,10 @@ logger = get_logger()
 
 
 async def run_custom_pipeline(
-    tasks: Union[List[Task], List[str]] = None,
+    tasks: Optional[Union[List[Task], List[str]]] = None,
     data: Any = None,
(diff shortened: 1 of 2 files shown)


dexters1:
These are not optional parameters

PR #3110, opened 2026-06-20
fix(ontology): try multiple RDF formats for file-object ontology input

### LINKED ISSUE #2907: RDFLibOntologyResolver hardcodes format="xml" for file-object input, silently dropping Turtle/N3 ontologies
## Summary

When an ontology is supplied as a **file-like object**
(e.g. uploaded via the `/api/v1/ontologies` endpoint),
`RDFLibOntologyResolver` parses it with a hardcoded
`format="xml"`. Any non-RDF/XML serialization (Turtle,
N3, JSON-LD, etc.) fails to parse, the exception is
swallowed, and the resolver ends up with an **empty
graph**. Silent failure: `cognify` "succeeds" with the
given `ontologyKey`, but nothing anchors — every node
ends up `ontology_valid: false`.

## Location

`cognee/modules/ontology/rdf_xml/RDFLibOntologyResolver.py` (current `main`), file-object branch:

    content = file_obj.read()
    self.graph.parse(data=content, format="xml")   # hardcoded

The path-based branch a few lines down does NOT force a
format (`self.graph.parse(file_path)`), so the same
ontology loads as a path but fails as a file object.

## Symptom

Log storm of `OntologyAdapter: No close match found for '<x>' in category 'classes'`, `Lookup built: 0 classes, 0 individuals`, and all nodes `ontology_valid: false`.

## Fix options

- Derive format from filename via `rdflib.util.guess_format(filename)`, or
- Pass an explicit format/content-type from the upload endpoint, or
- Try a small set of formats and use the first that parses, or
- At minimum, hard-error when no file object parses instead of proceeding with an empty graph.

## Workaround

Convert to RDF/XML before upload:
`rdflib.Graph().parse(data=ttl, format="turtle").serialize(format="xml")`.

garimapahwa:
## Summary
- `RDFLibOntologyResolver` hardcoded `format="xml"` when parsing file-like ontology objects. Any non-RDF/XML serialization (Turtle, N3, JSON-LD) failed to parse; the exception was swallowed and the resolver silently ended up with an empty graph (cognify "succeeds" but every node is `ontology_valid: false`).
- Now tries `xml`, `turtle`, `n3`, `json-ld` in order for each file object and keeps the first format that parses successfully. Each attempt parses into a scratch graph first, so a failed attempt never leaves partial triples in `self.graph`.

## Test plan
- [x] Added `test_ontology_adapter_loads_turtle_file_object`, which fails on `main` (graph stays empty) and passes with this fix.
- [x] `pytest cognee/tests/unit/modules/ontology/test_ontology_adapter.py`

Fixes #2907

### DIFF
--- a/cognee/modules/ontology/rdf_xml/RDFLibOntologyResolver.py
+++ b/cognee/modules/ontology/rdf_xml/RDFLibOntologyResolver.py
@@ -57,11 +57,31 @@ class RDFLibOntologyResolver(BaseOntologyResolver):
                     for file_obj in file_objects:
                         try:
                             content = file_obj.read()
-                            self.graph.parse(data=content, format="xml")
-                            loaded_objects.append(file_obj)
-                            logger.info("Ontology loaded successfully from file object")
                         except Exception as e:
-                            logger.warning("Failed to parse ontology file object: %s", str(e))
+                            logger.warning("Failed to read ontology file object: %s", str(e))
+                            continue
+
+                        parsed = False
+                        for rdf_format in ("xml", "turtle", "n3", "json-ld"):
+                            try:
+                                candidate_graph = Graph()
+                                candidate_graph.parse(data=content, format=rdf_format)
+                            except Exception:
+                                continue
+                            self.graph += candidate_graph
+                            parsed = True
+                            logger.info(
+                                "Ontology loaded successfully from file object (format=%s)",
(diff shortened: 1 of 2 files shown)


garimapahwa:
@mpotter @adamhunter @aenawi Please review this PR.

PR #3113, opened 2026-06-20
fix(pipeline): truncate pipeline run_info payload to prevent unbounded DB growth

### LINKED ISSUE #3074: pipeline_runs table grows without bound: run_info stores full input payload on every run

aniruddhaadak80:
Closes #3074

When input data to pipelines isn't a list of Data items, the raw stringified payload could cause unbounded growth in the SQLite pipeline_runs table. This PR extracts the serialization logic into a reusable format_data_info helper that caps stringified payloads at 500 characters.

### DIFF
--- a/cognee/modules/pipelines/operations/log_pipeline_run_complete.py
+++ b/cognee/modules/pipelines/operations/log_pipeline_run_complete.py
@@ -8,12 +8,9 @@ from typing import Any
 async def log_pipeline_run_complete(
     pipeline_run_id: UUID, pipeline_id: UUID, pipeline_name: str, dataset_id: UUID, data: Any
 ):
-    if not data:
-        data_info = "None"
-    elif isinstance(data, list) and all(isinstance(item, Data) for item in data):
-        data_info = [str(item.id) for item in data]
-    else:
-        data_info = str(data)
+    from cognee.modules.pipelines.utils.format_data_info import format_data_info
+
+    data_info = format_data_info(data)
 
     pipeline_run = PipelineRun(
         pipeline_run_id=pipeline_run_id,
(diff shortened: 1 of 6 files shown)


dexters1:
Closing PR, solution to cap payload to 500 chars does not seem like the right solution for this

PR #3133, opened 2026-06-20
fix(ui): make MCP Docker launch work on Colima

### LINKED ISSUE #2744: Support Colima as a Docker Desktop alternative for cognee UI / MCP launch
## Problem

When running the `cognee` CLI / UI, the MCP server is launched by shelling out to the `docker` binary in [`cognee/api/v1/ui/ui.py`](https://github.com/topoteretes/cognee/blob/main/cognee/api/v1/ui/ui.py):

```python
subprocess.run(["docker", "pull", image], check=True)
...
docker_cmd = ["docker", "run", "--name", container_name, "-p", f"{mcp_port}:8000", "--rm", "-e", "TRANSPORT_MODE=sse", ...]
mcp_process = subprocess.Popen(docker_cmd, ...)
```

This implicitly assumes Docker Desktop. [Colima](https://github.com/abiosoft/colima) is a popular open-source, free-for-commercial-use alternative to Docker Desktop on macOS and Linux. Many users cannot use Docker Desktop because of its licensing terms, and switching the daemon to Colima is the standard workaround.

## What works today

Because Colima ships a `docker` CLI shim that talks to the Colima VM socket, `cognee` UI does mostly run on Colima — `docker pull` and `docker run` succeed:

```
Starting Cognee MCP server with Docker...
main: Pulling from cognee/cognee-mcp
c831750efa1d: Pull complete
383e294faf4c: Download complete
...
```

## What breaks / friction points on Colima

1. **`host.docker.internal` is not populated by default.** [`cognee-mcp/entrypoint.sh`](https://github.com/topoteretes/cognee/blob/main/cognee-mcp/entrypoint.sh) rewrites `localhost` → `host.docker.internal` so the container can reach a host-side cognee API. On Colima this hostname does not resolve unless the user starts Colima with `--network-address`, so any flow that uses the host API silently breaks inside the container.
2. **No detection or guidance.** If `docker` is missing or the daemon is unreachable, the launch fails with a generic subprocess error rather than a hint pointing the user at Docker Desktop / `colima start`.
3. **Docs assume Docker Desktop.** README and quickstart do not mention Colima as a supported runtime.

## Proposal

Three small changes, in increasing order of effort:

1. **Docs:** add a short "Using Colima" section to `cognee-mcp/README.md` covering `brew install colima docker`, `colima start`, and the `host.docker.internal` caveat.
2. **Preflight in `ui.py`:** before `docker pull`, run `docker info` (or `docker version`) and, on failure, surface a clear message — e.g. "Docker daemon not reachable — start Docker Desktop or run `colima start`". Cheap and runtime-agnostic.
3. **`host.docker.internal` fallback in `entrypoint.sh`:** if `host.docker.internal` does not resolve, fall back to `host.lima.internal`, then to the default gateway IP. This makes the host-API flow work on Colima out of the box.

Happy to send a PR for (1) and (2) if maintainers are open to it.

## Environment

- macOS (Darwin 25.3.0, arm64)
- `cognee` 1.0.3
- Colima as the container runtime in place of Docker Desktop

guglxni:
Closes #2744

## Summary

- Extract `build_mcp_docker_command()` so MCP container launch is testable
- Add `--add-host host.docker.internal:host-gateway` for Docker Desktop and Colima
- Set `API_URL` to `http://host.docker.internal:<port>` instead of `localhost` from inside the container

## Test plan

- [x] `pytest cognee/tests/unit/api/v1/ui/test_mcp_docker_command.py`

### DIFF
--- a/cognee/api/v1/ui/ui.py
+++ b/cognee/api/v1/ui/ui.py
@@ -21,6 +21,48 @@ from .npm_utils import run_npm_command
 logger = get_logger()
 
 
+def _mcp_docker_api_url(backend_port: int) -> str:
+    """Backend URL reachable from inside the MCP container (Docker Desktop / Colima)."""
+    return f"http://host.docker.internal:{backend_port}"
+
+
+def build_mcp_docker_command(
+    *,
+    container_name: str,
+    mcp_port: int,
+    image: str,
+    backend_port: Optional[int] = None,
+    env_file: Optional[str] = None,
+) -> list[str]:
+    """Build the docker run argv for launching the MCP server container."""
+    docker_cmd = [
+        "docker",
+        "run",
+        "--name",
+        container_name,
+        "-p",
+        f"{mcp_port}:8000",
+        "--rm",
+        "--add-host",
+        "host.docker.internal:host-gateway",
+        "-e",
+        "TRANSPORT_MODE=sse",
+    ]
+
+    if backend_port is not None:
+        docker_cmd.extend(
+            [
+                "-e",
+                f"API_URL={_mcp_docker_api_url(backend_port)}",
+            ]
+        )
+    elif env_file:
+        docker_cmd.extend(["--env-file", env_file])
+
+    docker_cmd.append(image)
+    return docker_cmd
+
+
 def _check_docker_available() -> Tuple[bool, str]:
     """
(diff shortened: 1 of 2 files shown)


guglxni:
Reopening to refresh issue linking

PR #3159, opened 2026-06-20
fix(build): remove invalid uv exclude-newer that dropped constraints

abhay-codes07:
## Description
While installing the project with uv, I noticed it printed a warning and then
ignored the dependency constraints:

    warning: Failed to parse `pyproject.toml` during settings discovery:
      TOML parse error at line 232, column 17
          |
      232 | exclude-newer = "2 days"
          |                 ^^^^^^^^
      failed to parse year in date "2 days"

Both pyproject.toml and cognee-mcp/pyproject.toml set exclude-newer = "2 days"
under [tool.uv]. uv's exclude-newer expects an absolute date (like 2025-01-01),
not a duration like "2 days", so uv can't parse it and throws away the whole
[tool.uv] table — with just a warning, then it carries on. The problem is that
the table also holds constraint-dependencies (the protobuf / pyasn1 / wheel /
nltk / spacy / scipy pins, and onnxruntime in cognee-mcp), so all of those get
silently dropped and the resolver stops honoring them.

This flag was added in #2828 to skip package releases newer than 2 days, but
since "2 days" isn't a valid value it never actually did that — it only broke
parsing. There's no way to express a rolling "2 days" window as a static value
in uv, so I removed the invalid line (and its comment) from both files. That
makes the [tool.uv] table parse again and brings the constraints back. If the
team still wants the 2-day window, it would need a real mechanism like passing
a computed date to `uv lock --exclude-newer` in CI — happy to do that in a
separate PR.

## Acceptance Criteria
- uv no longer fails to parse [tool.uv].
- The constraint-dependencies are applied again.
- Verified locally: `uv pip install -e . --dry-run` resolves all 132 packages
  with 0 warnings (before this change it warned on every run). See screenshot.

## Type of Change

## Screenshots
image

## Pre-submission Checklist
- [ ] I have added tests that prove my fix is effective (n/a — build config)
- [x] All new and existing tests pass (uv resolves cleanly)

## DCO Affirmation

### DIFF
--- a/cognee-mcp/pyproject.toml
+++ b/cognee-mcp/pyproject.toml
@@ -64,9 +64,6 @@ constraint-dependencies = [
     "onnxruntime>=1.24.1 ; python_version >= '3.14'",
 ]
 
-# Note: exclude newer packages that are less than 2 days old
-exclude-newer = "2 days"
-
 [project.scripts]
 cognee = "src:main"
 cognee-mcp = "src:main_mcp"
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -228,8 +228,6 @@ exclude = [
 packages = ["cognee", "cognee_db_workers", "distributed", "kuzu"]
 
 [tool.uv]
-# Note: exclude newer packages that are less than 2 days old
-exclude-newer = "2 days"
 constraint-dependencies = [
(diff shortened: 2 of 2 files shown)


dexters1:
@abhay-codes07 update your UV version

PR #3161, opened 2026-06-20
fix(ingestion): persist data_size on re-ingest update

### LINKED ISSUE #3160: [Bug]: Re-ingesting a changed file never updates the stored data_size (wrong attribute name in update path)
### Bug Description

In `store_data_to_dataset` (`cognee/tasks/ingestion/ingest_data.py`), the branch that updates an already-existing `Data` record sets the file size with:

```python
data_point.file_size = original_file_metadata["file_size"]
```

The `Data` model (`cognee/modules/data/models/Data.py`) has no `file_size` column - the column is `data_size`. The create branch in the same function uses it correctly (`data_size=original_file_metadata["file_size"]`); only the update branch has the wrong name.

Because `file_size` is not a mapped column, that assignment just creates a plain instance attribute. SQLAlchemy's unit-of-work only flushes mapped column attributes, so the new size is never written and the persisted `data_size` silently keeps its old value. No error is raised, so re-ingesting a file whose size changed leaves stale metadata in the relational store.

### Steps to Reproduce

The defect is at the ORM level, so it can be shown deterministically against the real `Data` model without a full pipeline run:

```python
from sqlalchemy import inspect
from cognee.modules.data.models import Data

cols = set(inspect(Data).columns.keys())
print("data_size is a column:", "data_size" in cols)  # True
print("file_size is a column:", "file_size" in cols)  # False

dp = Data(data_size=100)
dp.file_size = 200          # exactly what the update branch does
print(dp.data_size)         # 100 - the real column was never touched
```

End-to-end trigger: `add()` a file, then `add()` it again after its size has changed (so the update branch runs). The `data` row's `data_size` stays at the original value.

### Expected Behavior

Re-ingesting an existing record with a changed size updates the persisted `data_size` column to the new value.

### Actual Behavior

`data_size` keeps its original value. The new size is assigned to an unmapped `file_size` attribute and dropped on flush. Output from the snippet above:

Image

### Environment

- OS: Linux (WSL2) - but the bug is source-level and OS-independent
- Python version: 3.12
- Cognee version: dev branch (reproduced at current HEAD)
- Database: default SQLite (relational store)

### Logs/Error Messages

```shell

```

### Additional Context

_No response_

### Pre-submission Checklist

parveshsaini:
Closes #3160

## Description
While looking through store_data_to_dataset in cognee/tasks/ingestion/ingest_data.py, I noticed a small typo in the update path that's used when an existing file is ingested again.

Right now it updates the file size like this:

```python
data_point.file_size = original_file_metadata["file_size"]
```

The problem is that the Data model doesn't actually have a file_size column. The mapped column is called data_size (cognee/modules/data/models/Data.py). Interestingly, the create path just a few lines below already uses the correct field:

```python
data_size=original_file_metadata["file_size"]
```

Since file_size isn't a mapped SQLAlchemy column, assigning to it only creates a regular Python attribute on the object. SQLAlchemy doesn't track or persist that attribute, so when the session is flushed, the updated file size is never written to the database. As a result, if a file is re-ingested after its size changes, the stored data_size remains outdated without raising any error.

The fix is simply to change the attribute from file_size to data_size so the update is applied to the actual database column.

## Acceptance Criteria
- Re-ingesting an existing `Data` record with a new size updates the persisted `data_size` column.
- Added a regression test that drives the real `ingest_data` update path against an in-memory SQLite relational engine and asserts the stored `data_size` reflects the new value. It fails before the change and passes after it.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots
`uv run pytest cognee/tests/unit/tasks/ingestion/ -q`
image

Without the fix the same test fails:
image

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/tasks/ingestion/ingest_data.py
+++ b/cognee/tasks/ingestion/ingest_data.py
@@ -183,7 +183,7 @@ async def ingest_data(
                 data_point.owner_id = user.id
                 data_point.content_hash = new_content_hash
                 data_point.raw_content_hash = storage_file_metadata["content_hash"]
-                data_point.file_size = original_file_metadata["file_size"]
+                data_point.data_size = original_file_metadata["file_size"]
                 data_point.external_metadata = ext_metadata
                 data_point.node_set = json.dumps(node_set) if node_set else None
                 data_point.tenant_id = user.tenant_id if user.tenant_id else None
--- a/cognee/tests/unit/tasks/ingestion/test_ingest_update_data_size.py
+++ b/cognee/tests/unit/tasks/ingestion/test_ingest_update_data_size.py
@@ -1,0 +1,150 @@
+"""Regression test for the re-ingestion `data_size` update path.
+
+When an already-ingested file is added again with a different size, the update
+branch of ``store_data_to_dataset`` (in ``cognee/tasks/ingestion/ingest_data.py``)
+must write the new size to the ``Data.data_size`` column. A prior version wrote
+to ``data_point.file_size`` — an attribute the ``Data`` model does not declare —
+so SQLAlchemy's unit-of-work never flushed it and the persisted ``data_size``
+stayed stale.
+
+This test drives the real ``ingest_data`` update branch against an in-memory
+SQLite relational engine and asserts the persisted ``data_size`` reflects the
+new value. It fails before the fix and passes after it.
+"""
+
+import importlib
+import os
+import tempfile
+from contextlib import asynccontextmanager
+from types import SimpleNamespace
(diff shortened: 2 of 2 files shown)


parveshsaini:
Hey team! Gentle bump on this. Would appreciate a quick review whenever you get a chance.

@dexters1 @Vasilije1990 @siillee @LStromann @lxobr 

Thanks! 🚀

Vasilije1990 (approved):

PR #3162, opened 2026-06-20
fix(ingestion): correct data_size attribute when updating existing data points (fixes #3160)

### LINKED ISSUE #3160: [Bug]: Re-ingesting a changed file never updates the stored data_size (wrong attribute name in update path)

aniruddhaadak80:
Fixes #3160 by replacing the incorrect file_size attribute with the correct data_size attribute when updating existing data points in the ingestion pipeline.

### DIFF
--- a/cognee/tasks/ingestion/ingest_data.py
+++ b/cognee/tasks/ingestion/ingest_data.py
@@ -183,7 +183,7 @@ async def ingest_data(
                 data_point.owner_id = user.id
                 data_point.content_hash = new_content_hash
                 data_point.raw_content_hash = storage_file_metadata["content_hash"]
-                data_point.file_size = original_file_metadata["file_size"]
+                data_point.data_size = original_file_metadata["file_size"]
                 data_point.external_metadata = ext_metadata
                 data_point.node_set = json.dumps(node_set) if node_set else None
                 data_point.tenant_id = user.tenant_id if user.tenant_id else None


dexters1:
This PR was initially made by a different contributor who raised the issue

PR #3164, opened 2026-06-20
fix(storage): await file_exists in LocalFileStorage.get_size

abhay-codes07:
## Description
While reading through the local file storage adapter I noticed `get_size`
calls `file_exists` without awaiting it:

```python
async def get_size(self, file_path: str) -> int:
    parsed_storage_path = get_parsed_path(self.storage_path)
    return (
        os.path.getsize(os.path.join(parsed_storage_path, file_path))
        if self.file_exists(file_path)   # <-- file_exists is async, not awaited
        else 0
    )
```

`file_exists` is an `async def`, so `self.file_exists(file_path)` returns a
coroutine object. A coroutine is always truthy, so the `if` is always true:
the `else 0` branch is dead and `os.path.getsize(...)` runs unconditionally.
For a path that doesn't exist this raises `FileNotFoundError` instead of
returning `0`, and Python also emits a "coroutine was never awaited" warning.

Everywhere else in the codebase `file_exists` is awaited (e.g. in
`S3FileStorage.store`, `SqlAlchemyAdapter`, `StorageManager`), so this is just
a missing `await`. The fix is one word.

## Acceptance Criteria
- `get_size` returns `0` for a missing file (instead of raising).
- `get_size` still returns the real size for an existing file.
- No "coroutine was never awaited" warning.

Verified locally (before the fix `get_size("missing")` raised
`FileNotFoundError`; after it returns `0`), and added a unit test that covers
both the missing-file and existing-file cases. See screenshot.

## Type of Change

## Screenshots
image

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/files/storage/LocalFileStorage.py
+++ b/cognee/infrastructure/files/storage/LocalFileStorage.py
@@ -204,7 +204,7 @@ class LocalFileStorage(Storage):
 
         return (
             os.path.getsize(os.path.join(parsed_storage_path, file_path))
-            if self.file_exists(file_path)
+            if await self.file_exists(file_path)
             else 0
         )
 
--- a/cognee/tests/unit/infrastructure/test_local_file_storage_get_size.py
+++ b/cognee/tests/unit/infrastructure/test_local_file_storage_get_size.py
@@ -1,0 +1,28 @@
+import os
+import tempfile
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


abhay-codes07:
@dexters1 @Vasilije1990 get_size does not await file_exists, so the coroutine is always truthy and missing files raise instead of returning 0. Small fix with a regression test, ready when you are.

PR #3169, opened 2026-06-20
fix(graph): await async query in Neo4jAdapter.has_node

abhay-codes07:
## Description
`Neo4jAdapter.has_node` calls the async `query` method without awaiting it:

```python
async def has_node(self, node_id: str) -> bool:
    results = self.query(            # <-- query is async, not awaited
        f"""
            MATCH (n:`{BASE_LABEL}`)
            WHERE n.id = $node_id
            RETURN COUNT(n) > 0 AS node_exists
        """,
        {"node_id": node_id},
    )
    return results[0]["node_exists"] if len(results) > 0 else False
```

`query` is an `async def`, so `self.query(...)` returns a coroutine. The next
line then does `len(results)` / `results[0]` on that coroutine, which raises
`TypeError: object of type 'coroutine' has no len()` — `has_node` can never
return a value, and the coroutine is left un-awaited.

Every other call site in this adapter awaits `query`; `has_node` was the only
one that didn't. The fix is a missing `await`.

## Acceptance Criteria
- `has_node` returns a bool instead of raising `TypeError`.
- Returns `True`/`False` based on the query result.

Added a unit test (with a mocked async `query`) that fails on the old code
(`TypeError`: coroutine has no `len()`) and passes with the fix. See screenshot.

## Type of Change

## Screenshots
image

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
+++ b/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
@@ -176,7 +176,7 @@ class Neo4jAdapter(GraphDBInterface):
 
             - bool: True if the node exists, otherwise False.
         """
-        results = self.query(
+        results = await self.query(
             f"""
                 MATCH (n:`{BASE_LABEL}`)
                 WHERE n.id = $node_id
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@dexters1 @Vasilije1990 has_node calls query without awaiting it, so len() runs on a coroutine and raises. One await plus a mocked test, no live Neo4j needed. Should be a trivial review when you get a chance.

Vasilije1990 (approved):

PR #3172, opened 2026-06-20
fix(graph): implement is_empty in NeptuneGraphDB

abhay-codes07:
## Description
`NeptuneGraphDB` does not implement `is_empty`, which `GraphDBInterface`
declares as an `@abstractmethod`:

```python
# graph_db_interface.py
@abstractmethod
async def is_empty(self) -> bool:
    """Return True when the graph contains no nodes."""
    logger.warning("is_empty() is not implemented")
    return True
```

`ladybug`, `neo4j` and `postgres` all implement it; Neptune does not. Because
the abstract method is left unimplemented, `NeptuneGraphDB.__abstractmethods__`
is `{'is_empty'}`, so the class can't be instantiated at all:

```
TypeError: Can't instantiate abstract class NeptuneGraphDB without an
implementation for abstract method 'is_empty'
```

i.e. the Neptune graph backend is unusable. (And even ignoring the abstract
check, the interface's fallback body just logs a warning and returns `True`, so
it would report every graph as empty.)

This PR implements `is_empty` with a simple node-count query scoped to the
graph's node label, returning `True` only when there are no nodes — consistent
with the neo4j/ladybug/postgres implementations.

## Acceptance Criteria
- `NeptuneGraphDB` is no longer abstract (can be instantiated).
- `is_empty()` returns `True` for an empty graph and `False` otherwise.

Added a regression test that asserts `is_empty` is no longer in
`__abstractmethods__` and that the method returns the correct value (query
mocked; no live Neptune needed). It fails on the old code and passes with the
fix. See screenshot.

## Type of Change

## Screenshots
image

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/databases/graph/neptune_driver/adapter.py
+++ b/cognee/infrastructure/databases/graph/neptune_driver/adapter.py
@@ -751,6 +751,15 @@ class NeptuneGraphDB(GraphDBInterface):
             logger.error(f"Failed to get neighborhood: {error_msg}")
             raise Exception(f"Failed to get neighborhood: {error_msg}") from e
 
+    async def is_empty(self) -> bool:
+        """Return True if the graph contains no nodes."""
+        query = f"""
+            MATCH (n:{self._GRAPH_NODE_LABEL})
+            RETURN count(n) AS node_count
+        """
+        result = await self.query(query)
+        return result[0]["node_count"] == 0
+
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@Vasilije1990 @dexters1 NeptuneGraphDB cannot be instantiated at all right now because is_empty was left abstract. Implemented it with a regression test. Small and self-contained, would appreciate a merge.

PR #3174, opened 2026-06-20
fix(data): serialize null tenant_id as null in Dataset.to_json (#3173)

### LINKED ISSUE #3173: [Bug]: Dataset.to_json() serializes a null tenant_id as the string "None" instead of null
### Bug Description

`Dataset.to_json()` serializes the nullable `tenant_id` column with `str(self.tenant_id)`. When a dataset has no tenant (the normal case in single-tenant / non-multi-tenant deployments, where `tenant_id` is `None`), `str(None)` produces the literal string `"None"`, so the JSON payload contains `"tenantId": "None"` instead of `null`.

The same method already handles the nullable `updated_at` correctly (`self.updated_at.isoformat() if self.updated_at else None`), so this looks like an oversight on the `tenant_id` line. `owner_id` is not nullable, so it is unaffected.

`cognee/modules/data/models/Dataset.py`

```python
def to_json(self) -> dict:
    return {
        ...
        "updatedAt": self.updated_at.isoformat() if self.updated_at else None,
        "ownerId": str(self.owner_id),
        "tenantId": str(self.tenant_id),   # <-- str(None) -> "None" when tenant_id is None
        ...
    }
```

### Steps to Reproduce

No database or LLM needed — it is pure serialization:

```python
from cognee.modules.data.models.Dataset import Dataset

d = Dataset()
d.id = "00000000-0000-0000-0000-000000000001"
d.name = "ds"
d.created_at = datetime(2024, 1, 1, tzinfo=timezone.utc)
d.updated_at = None
d.owner_id = "00000000-0000-0000-0000-000000000002"
d.tenant_id = None
d.data = []

print(d.to_json()["tenantId"])   # -> "None"  (a string)
```

### Expected Behavior

`"tenantId"` is JSON `null` when the dataset has no tenant.

### Actual Behavior

`"tenantId"` is the string `"None"`.

### Environment

- cognee: dev branch (reproduced at current HEAD)
- Python: 3.x
- Not dependent on OS / DB / LLM (source-level).

pooja-bhavani:
## Description

While reading the dataset models I noticed `Dataset.to_json()` serializes the nullable `tenant_id` column with `str(self.tenant_id)`. For any dataset without a tenant — the normal case in single-tenant / non-multi-tenant setups where `tenant_id` is `None` — `str(None)` returns the literal string `"None"`, so the payload ends up with `"tenantId": "None"` instead of `null`.

The same method already handles the nullable `updated_at` the right way (`... if ... else None`), so I applied that same pattern to `tenant_id`. `owner_id` is non-nullable, so I left it as `str(self.owner_id)`.

## Acceptance Criteria

- A dataset with `tenant_id = None` serializes `"tenantId": null`.
- A dataset with a tenant still serializes `"tenantId"` as the stringified UUID.

```python
# before:  d.to_json()["tenantId"] == "None"
# after:   d.to_json()["tenantId"] is None   (when tenant_id is None)
```

Added a unit test (`cognee/tests/unit/modules/data/test_dataset_to_json_tenant.py`) covering both cases. It fails before the change and passes after. No DB/LLM needed.

`uv run ruff check .` passes; the change adds no `ty` diagnostics.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

Fixes #3173.

### DIFF
--- a/cognee/modules/data/models/Dataset.py
+++ b/cognee/modules/data/models/Dataset.py
@@ -43,6 +43,6 @@ class Dataset(Base):
             "createdAt": self.created_at.isoformat(),
             "updatedAt": self.updated_at.isoformat() if self.updated_at else None,
             "ownerId": str(self.owner_id),
-            "tenantId": str(self.tenant_id),
+            "tenantId": str(self.tenant_id) if self.tenant_id else None,
             "data": [data.to_json() for data in self.data],
         }
--- a/cognee/tests/unit/modules/data/test_dataset_to_json_tenant.py
+++ b/cognee/tests/unit/modules/data/test_dataset_to_json_tenant.py
@@ -1,0 +1,35 @@
+"""Regression test: Dataset.to_json() must serialize a null tenant_id as JSON
+null, not the string "None".
+
+`tenant_id` is a nullable column, so in single-tenant deployments it is None.
+`str(None)` yields the literal "None", leaking a bogus string into the payload.
+The same method already handles the nullable `updated_at` correctly with
+`... if ... else None`; this pins tenant_id to that same convention.
+"""
+
+from datetime import datetime, timezone
+
+from cognee.modules.data.models.Dataset import Dataset
+
+
+def _make_dataset(tenant_id):
+    d = Dataset()
+    d.id = "00000000-0000-0000-0000-000000000001"
+    d.name = "ds"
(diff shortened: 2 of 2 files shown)


dexters1:
This should not be changed

PR #3242, opened 2026-06-21
fix(telemetry): anchor fire-and-forget telemetry task against gc

### LINKED ISSUE #3231: [Bug]: Telemetry task is not strongly referenced and can be GC'd mid-execution
### Bug Description
In [cognee/shared/utils.py](https://github.com/topoteretes/cognee/blob/dev/cognee/shared/utils.py#L272-L273), the telemetry helper `send_telemetry` fires a fire-and-forget background task to send the request via:
```python
    loop = asyncio.get_running_loop()
    loop.create_task(_send_telemetry_request(payload))
```
According to Python `asyncio` documentation, the event loop only holds *weak* references to tasks. If a task created by `create_task()` (or `loop.create_task()`) is not strongly referenced elsewhere, it can be garbage-collected at any time during an `await` point, silently aborting the background task mid-run.

Since the returned `Task` from `loop.create_task` is discarded, under memory pressure or GC cycles during processing, the telemetry HTTP request can be aborted, resulting in lost or incomplete telemetry logs.

This is the exact same hazard already guarded against in `closing_lru_cache.py` (via `_PENDING_CLOSE_TASKS`) and recently fixed in pipelines/sync background tasks.

### Source Code Reference
- File: [cognee/shared/utils.py](https://github.com/topoteretes/cognee/blob/dev/cognee/shared/utils.py#L272-L273)

### Suggested Fix
Anchor the task in a module-level set and discard it on completion:
```python
_TELEMETRY_TASKS: set[asyncio.Task] = set()

# inside send_telemetry:
    loop = asyncio.get_running_loop()
    task = loop.create_task(_send_telemetry_request(payload))
    _TELEMETRY_TASKS.add(task)
    task.add_done_callback(_TELEMETRY_TASKS.discard)
```

garimapahwa:
## Summary
- `send_telemetry` in `cognee/shared/utils.py` fired a background task via `loop.create_task(_send_telemetry_request(payload))` and discarded the result. asyncio's event loop only holds a *weak* reference to tasks — without a strong reference elsewhere, the gc can collect the task mid-run, silently dropping the telemetry request.
- Anchor the task in a module-level `_TELEMETRY_TASKS` set, removing it via `add_done_callback` once it completes — the same pattern already used for eviction tasks in `closing_lru_cache.py`.

## Test plan
- [x] Added `test_send_telemetry_anchors_task_against_gc`: spies on the real event loop's `create_task` to confirm the produced task is anchored in `_TELEMETRY_TASKS` while in-flight and removed once done. Fails on `main` (`_TELEMETRY_TASKS` doesn't exist), passes with this fix.
- [x] Updated `test_send_telemetry_includes_api_key_tracking_id`'s mock loop to return a task-like stub (production code now calls `.add_done_callback()` on the result of `create_task`).
- [x] `pytest cognee/tests/unit/shared/test_telemetry_tracking.py cognee/tests/test_telemetry.py`

Fixes #3231


garimapahwa:
@mpotter @adamhunter @aenawi Please review and merge the PR's

PR #3257, opened 2026-06-21
fix(retrieval): align triplet context results with their entities

abhay-codes07:
## Description

`TripletSearchContextProvider` runs a brute-force triplet search per entity and then labels each set of results with the entity it came from. `_get_search_tasks` builds the tasks with a *filtering* comprehension — it skips any entity whose concatenated text (`name`/`description`/`text`) is empty:

```python
tasks = [
    brute_force_triplet_search(query=f"{entity_text} {query}", ...)
    for entity in entities
    if (entity_text := self._get_entity_text(entity)) is not None
]
```

So when any entity is text-less, the gathered `results` list is shorter than `entities`. But `_results_to_context` re-zipped against the **full** `entities` list:

```python
for entity, entity_triplets in zip(entities, results):
```

`zip` then pairs each result with the wrong entity from the first skipped one onward, and silently drops the trailing entities. The triplet context handed to the LLM ends up attributed to the wrong entity (and some entities disappear), degrading retrieval quality whenever at least one entity lacks text.

The fix makes `_get_search_tasks` return the surviving entities alongside their tasks, and `get_context` zips those — so every result is paired with the entity it was actually searched for.

## Acceptance Criteria

- Each entity's triplet results are labeled with that same entity, even when an earlier entity was skipped for having no text.
- Entities with searchable text are never dropped.

Regression test (text-less entity between two valid ones; fails on `dev`, passes with the fix):

```
--- BEFORE: dev code (bug present) ---
FAILED cognee/tests/unit/modules/retrieval/test_triplet_context_provider_alignment.py::test_context_pairs_each_entity_with_its_own_result
  - AssertionError: 'Context for Gamma:\nRESULT::Gamma Q' not found (Gamma was dropped / misaligned)
1 failed in 0.47s

--- AFTER: fix applied ---
1 passed in 0.06s
```

## Type of Change

## Screenshots
image

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/modules/retrieval/context_providers/TripletSearchContextProvider.py
+++ b/cognee/modules/retrieval/context_providers/TripletSearchContextProvider.py
@@ -43,20 +43,31 @@ class TripletSearchContextProvider(BaseContextProvider):
         entities: List[DataPoint],
         query: str,
         memory_fragment: CogneeGraph,
-    ) -> List:
-        """Creates search tasks for valid entities."""
-        tasks = [
-            brute_force_triplet_search(
-                query=f"{entity_text} {query}",
-                top_k=self.top_k,
-                collections=self.collections,
-                properties_to_project=self.properties_to_project,
-                memory_fragment=memory_fragment,
+    ) -> tuple[List[DataPoint], List]:
+        """Creates search tasks for the entities that have searchable text.
+
+        Returns the surviving entities alongside their tasks so callers can keep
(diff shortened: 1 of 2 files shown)


abhay-codes07:
@Vasilije1990 @dexters1 Retrieval correctness: the triplet context provider zips results against the full entity list after filtering text-less entities, so labels shift and trailing entities drop. Fix keeps the lists aligned, with a test.

PR #3275, opened 2026-06-22
fix(ci): nightly llama-cpp — swap to Phi-3.5-mini, pull from HuggingFace

NMZivkovic:
## What
- Swap the nightly Llama-cpp test model **Phi-3-mini-4k q4 → Phi-3.5-mini-instruct Q4_K_M** (MIT).
- Pull it **directly from HuggingFace** (public) instead of the private S3 mirror, removing the AWS-credential dependency.

## Why — not an instructor/code regression
The job fails on `InstructorRetryException` → `nodes.N.description Field required`: the 4-bit Phi-3-mini sometimes omits the required per-node `description`. Investigation of the last-green (Jun 15) → first-red (Jun 16) window shows nothing in the code path changed:
- instructor was **`1.15.1` at both** the green and red commits — no version change.
- The forward-compat fix (#3080) only wraps bare *primitives*; `KnowledgeGraph` (a `BaseModel`) passes through untouched.
- The GGUF was **SHA-pinned** — no model drift.

So it's a structurally fragile test: a weak quantized model against a strictly-required field. Jun 15 was the last lucky sampling draw.

## Fix — better model, no schema change
Phi-3.5-mini is the same-vendor successor (MIT), newer/stronger, ~2.4 GB so still within the runner's RAM budget. **`description` stays required** — the product/data contract is unchanged; only the test model improves.

## Validation
Ran cognee's real extraction path (`extract_content_graph` → `KnowledgeGraph`, `description` **required**) on the exact CI input text, 5× per model:

| Model | Result |
|---|---|
| Phi-3-mini-q4 (old) | drops node `description` → `InstructorRetryException` |
| **Phi-3.5-mini Q4_K_M (this PR)** | **5/5 produced all node descriptions** ✅ |
| Qwen2.5-3B Q4_K_M (alt) | 5/5 ✅ (not chosen — "other" license vs MIT) |

Download SHA-verified against HF (`e4165e3a…`); the pinned SHA matches the file the workflow fetches.

## HuggingFace vs S3
The old step mirrored to S3 "to avoid HuggingFace 429 rate limits in CI." For a single file pulled once nightly, 429 risk is low and `curl --retry 5 --retry-all-errors` covers it — and it removes the dependency on `AWS_S3_DEV_USER` credentials. If the team prefers keeping the S3 mirror, the model swap alone (uploading the GGUF to the bucket) is the alternative.

### DIFF
--- a/.github/workflows/test_llamacpp.yml
+++ b/.github/workflows/test_llamacpp.yml
@@ -10,7 +10,7 @@ jobs:
 
   run_llama-cpp_test:
 
-    # needs 4 Gb RAM for phi3-mini in a container which the smallest runner has
+    # needs ~4 Gb RAM for the GGUF model in a container which the smallest runner has
     runs-on: ubuntu-22.04
     steps:
       - name: Checkout repository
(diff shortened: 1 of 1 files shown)


dexters1:
@NMZivkovic HuggingFace requests have a rate limit, it is the reason I created and moved the files to our own S3 bucket, removing S3 use might cause flakiness again

PR #3280, opened 2026-06-22
Fix Cognify WebSocket auth

fallintoplace:
## Summary
- replace the Cognify subscription direct dependency call with a WebSocket-specific JWT auth helper
- accept bearer headers or auth cookies, hydrate active users, and preserve default-user fallback when auth is disabled
- verify read access to the pipeline run dataset before queue subscription and graph streaming

## Root cause
The WebSocket path manually called `get_authenticated_user(cookie=..., strategy_cookie=..., user_manager=..., bearer=None)`, but the current helper only accepts dependency-injected users. That raises during subscription auth and closes clients as unauthorized.

## Validation
- `UV_CACHE_DIR=/private/tmp/uv-cache uv run pytest cognee/tests/unit/api/test_cognify_websocket_auth.py -q`
- `UV_CACHE_DIR=/private/tmp/uv-cache uv run pytest cognee/tests/unit/api/test_api_error_responses.py -q`
- `UV_CACHE_DIR=/private/tmp/uv-cache uv run ruff check cognee/api/v1/cognify/routers/get_cognify_router.py cognee/tests/unit/api/test_cognify_websocket_auth.py`
- `UV_CACHE_DIR=/private/tmp/uv-cache uv run ruff format --check cognee/api/v1/cognify/routers/get_cognify_router.py cognee/tests/unit/api/test_cognify_websocket_auth.py`

### DIFF
--- a/cognee/api/v1/cognify/routers/get_cognify_router.py
+++ b/cognee/api/v1/cognify/routers/get_cognify_router.py
@@ -9,9 +9,15 @@ from fastapi import APIRouter, WebSocket, Depends, WebSocketDisconnect, status
 from starlette.status import WS_1000_NORMAL_CLOSURE, WS_1008_POLICY_VIOLATION
 
 from cognee.api.DTO import InDTO
+from cognee.modules.data.methods import get_authorized_dataset
 from cognee.modules.pipelines.methods import get_pipeline_run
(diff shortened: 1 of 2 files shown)


Vasilije1990:
Won't touch this at this point

PR #3291, opened 2026-06-23
fix(permissions): keep shared-dataset documents when filtering by name

### LINKED ISSUE #3290: [Bug]: get_document_ids_for_user drops shared-dataset documents when filtered by name
### Bug Description

`get_document_ids_for_user(user_id, datasets=...)` is documented to return "documents ids for which the user has read permission." Without the `datasets` filter it does exactly that: it joins the `ACL` table on `principal_id == user_id` and `Permission.name == "read"`, so it correctly returns documents from every dataset the user can read, including datasets owned by someone else and shared via an ACL.

But when you pass `datasets=[name]` to narrow the result, the function re-resolves each name to a dataset id with an extra `Dataset.owner_id == user_id` condition. For a dataset that was shared to the user (read ACL) but is owned by another principal, that lookup returns `None`, the follow-up membership query becomes `DatasetData.dataset_id == None` (matches nothing), and all of the shared dataset's documents are silently dropped.

So filtering by a dataset's *own name* removes documents that the unfiltered call returns, which also contradicts the docstring and the function's own ACL-based first query. Because cognee returns an empty list on permission denial as well, this looks indistinguishable from "you have no access," making it easy to miss.

Affected symbol: `cognee.modules.users.permissions.methods.get_document_ids_for_user` (the `Dataset.owner_id == user_id` predicate in the name→id lookup).

### Steps to Reproduce

Minimal, no LLM/network needed ,  drives the real function against a throwaway SQLite relational engine seeded with a dataset owned by user B and shared to user A with a read ACL, containing one document:

```python
import asyncio
from uuid import uuid4

from cognee.modules.data.models import Data, Dataset, DatasetData
from cognee.modules.users.models import ACL, Permission
from cognee.modules.users.models.Principal import Principal
from cognee.modules.users.permissions.methods import get_document_ids_for_user


async def main():
    engine = get_relational_engine()
    await engine.create_database()

    owner_id, reader_id = uuid4(), uuid4()      # B owns, A only has a read ACL
    dataset_id, data_id = uuid4(), uuid4()
    name = "shared_dataset"

    async with engine.get_async_session() as session:
        async with session.begin():
            session.add(Principal(id=owner_id, type="principal"))
            session.add(Principal(id=reader_id, type="principal"))
            session.add(Dataset(id=dataset_id, name=name, owner_id=owner_id))
            session.add(Data(id=data_id, name="doc1", data_size=10))
            session.add(DatasetData(dataset_id=dataset_id, data_id=data_id))
            perm = Permission(name="read")
            session.add(perm)
            session.add(ACL(principal_id=reader_id, dataset_id=dataset_id, permission=perm))

    print("unfiltered :", [str(i) for i in await get_document_ids_for_user(reader_id)])
    print("by name    :", [str(i) for i in await get_document_ids_for_user(reader_id, datasets=[name])])


asyncio.run(main())
```

### Expected Behavior

The reader has read permission on `shared_dataset`, so filtering by that dataset's name should still return its document:

Image

### Actual Behavior

The name filter drops the shared document:

Image

The emitted SQL for the name→id lookup shows the cause:

Image

It filters on `owner_id = reader_id`, but the reader does not own the dataset, so it resolves to NULL and the next query runs `... WHERE dataset_data.dataset_id IS NULL ...`, matching nothing :(

### Environment

- OS: Ubuntu (Docker, `ghcr.io/astral-sh/uv:python3.12-bookworm-slim`)
- Python version: 3.12
- Cognee version: `dev` branch (reproduced at current HEAD)
- LLM Provider: n/a (not exercised)
- Database: default SQLite (relational)

### Pre-submission Checklist

parveshsaini:
Closes #3290

## Description

`get_document_ids_for_user(user_id, datasets=...)` is meant to return documents the user has read access to, and its initial ACL-based query correctly includes datasets that have been shared with the user.

The problem occurs when `datasets=[...]` is used. The dataset name lookup adds an ownership check:

```python
select(Dataset.id).where(
    Dataset.name == dataset,
    Dataset.owner_id == user_id,   # <- only matches datasets the user OWNS
)
```

Because of this, datasets shared through ACLs but owned by someone else are not found. The function then silently excludes documents from those datasets. As a result, an unfiltered call returns the documents, but adding the dataset's own name as a filter makes them disappear.

This is a fail-closed bug that doesn't leak data, but it incorrectly hides documents the user is allowed to read.

## Acceptance Criteria

- A dataset owned by user B and shared to user A through a `read` ACL, containing one document, must yield that document for user A both with and without a `datasets=[name]` filter.
- The unfiltered path is unchanged.
- Added regression test `cognee/tests/unit/users/permissions/test_get_document_ids_for_user_shared_dataset.py` drives the real `get_document_ids_for_user` against a throwaway SQLite engine seeded with the real ORM models. It fails before the change (name-filtered call returns `[]`) and passes after.

Verified in a clean `ghcr.io/astral-sh/uv:python3.12-bookworm-slim` container:

```
===== FAIL-BEFORE (unfixed source) =====
FAILED cognee/tests/unit/users/permissions/test_get_document_ids_for_user_shared_dataset.py::test_name_filter_keeps_acl_shared_dataset_documents
EXIT_BEFORE=1
===== PASS-AFTER (fixed source) =====
cognee/tests/unit/users/permissions/test_get_document_ids_for_user_shared_dataset.py::test_name_filter_keeps_acl_shared_dataset_documents PASSED [100%]
EXIT_AFTER=0
```

`ruff format --check` and `ruff check` both pass on the two changed files.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots

image

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/modules/users/permissions/methods/get_document_ids_for_user.py
+++ b/cognee/modules/users/permissions/methods/get_document_ids_for_user.py
@@ -49,7 +49,7 @@ async def get_document_ids_for_user(user_id: UUID, datasets: list[str] = None) -
                         await session.scalars(
                             select(Dataset.id).where(
                                 Dataset.name == dataset,
-                                Dataset.owner_id == user_id,
+                                Dataset.id.in_(dataset_ids),
                             )
                         )
                     ).one_or_none()
--- a/cognee/tests/unit/users/permissions/test_get_document_ids_for_user_shared_dataset.py
+++ b/cognee/tests/unit/users/permissions/test_get_document_ids_for_user_shared_dataset.py
@@ -1,0 +1,68 @@
+import importlib
+import os
+import tempfile
+from uuid import uuid4
+
+import pytest
+
+from cognee.infrastructure.databases.relational.sqlalchemy.SqlAlchemyAdapter import (
+    SQLAlchemyAdapter,
+)
+from cognee.modules.data.models import Data, Dataset, DatasetData
+from cognee.modules.users.models import ACL, Permission
+from cognee.modules.users.models.Principal import Principal
+
+gdifu = importlib.import_module(
+    "cognee.modules.users.permissions.methods.get_document_ids_for_user"
+)
+gdd = importlib.import_module("cognee.modules.data.methods.get_dataset_data")
+get_document_ids_for_user = gdifu.get_document_ids_for_user
+
+OWNER_ID = uuid4()  # user B -- owns the dataset
+READER_ID = uuid4()  # user A -- only has a read ACL on it
+DATASET_ID = uuid4()
+DATA_ID = uuid4()
+DATASET_NAME = "shared_dataset"
+
+
+async def _make_engine():
+    """Create a throwaway SQLite-backed relational engine seeded with a dataset
+    owned by user B and shared to user A with a read ACL, containing one doc."""
+    tmp = tempfile.NamedTemporaryFile(suffix=".db", delete=False)
+    tmp.close()
+    engine = SQLAlchemyAdapter(f"sqlite+aiosqlite:///{tmp.name}")
+    await engine.create_database()
+
+    async with engine.get_async_session() as session:
+        async with session.begin():
+            session.add(Principal(id=OWNER_ID, type="principal"))
+            session.add(Principal(id=READER_ID, type="principal"))
+            session.add(Dataset(id=DATASET_ID, name=DATASET_NAME, owner_id=OWNER_ID))
+            session.add(Data(id=DATA_ID, name="doc1", data_size=10))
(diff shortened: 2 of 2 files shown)


dexters1:
@parveshsaini thank you for flagging this, made a PR based on your changes here:

https://github.com/topoteretes/cognee/pull/3307/changes

Your commits will be stored there so you will be part of the git history, we wanted to resolve the issue in a different way. Datasets should never be resolved by name when not necessary as multiple users can own/have datasets with the same name

parveshsaini:
Aha - makes sense! Thanks @dexters1 🚀

PR #3301, opened 2026-06-23
fix: Cache lock context manager can release an invalid lock (#3294)

### LINKED ISSUE #3294: [Bug]: Cache lock context manager can release an invalid lock / release wrong lock
### Bug Description

CacheDBInterface.hold_lock() does lock = self.acquire_lock() then finally: self.release_lock(lock). If acquire_lock() fails partially, returns None, or throws after acquiring, release_lock(lock) may attempt releasing None or releasing a lock not actually held.

### Steps to Reproduce

1. Use a cache adapter where acquire_lock() can return None or throw after acquiring.
2. Enter with cache_db.hold_lock(): ....

### Expected Behavior

 Either:
acquire_lock() reliably returns a valid lock handle, or
hold_lock() should not call release_lock unless the lock was successfully acquired.

### Actual Behavior

release_lock(lock) runs unconditionally in finally, potentially releasing an invalid lock handle.

### Environment

cognee/infrastructure/databases/cache/cache_db_interface.py

### Logs/Error Messages

```shell

```

### Additional Context

_No response_

### Pre-submission Checklist

AmanM006:
## Summary

`CacheDBInterface.hold_lock()` called `self.release_lock(lock)` unconditionally in the `finally` block, even when `acquire_lock()` returned `None` (e.g. the lock was not available) or raised an exception after a partial acquisition. This caused `release_lock` to be invoked with an invalid / `None` lock handle, which concrete adapters (Redis, FS, SQL, Tapes) do not expect and can mishandle.

## Fix

Two minimal changes inside `hold_lock()`:

1. `yield` changed to `yield lock` so the caller can inspect or use the lock handle inside the `with` block if needed.
2. `self.release_lock(lock)` wrapped in `if lock:` so the call is skipped entirely when `acquire_lock()` returned a falsy value or raised before returning.

Nothing outside `hold_lock()` was touched.

## Tests

Added `cognee/tests/unit/infrastructure/databases/cache/test_cache_db_interface.py` with 11 regression tests across three scenarios:

- **(a)** `acquire_lock` returns `None` ? `release_lock` is never called
- **(b)** `acquire_lock` returns a valid handle ? `release_lock` is called with that exact handle, even when the `with` body raises
- **(c)** `acquire_lock` raises ? `release_lock` is never called

All tests use a lightweight `_MinimalAdapter` stub; no real infrastructure is required.

Closes #3294

## Type of change
- [x] Bug fix

## Testing
- `python -m py_compile` passes on both modified files
- 11 unit tests cover all three failure scenarios described in the issue

### DIFF
--- a/cognee/infrastructure/databases/cache/cache_db_interface.py
+++ b/cognee/infrastructure/databases/cache/cache_db_interface.py
@@ -41,12 +41,18 @@ class CacheDBInterface(ABC):
     def hold_lock(self):
         """
         Context manager for safely acquiring and releasing the lock.
+
+        ``release_lock`` is only invoked when ``acquire_lock`` actually returned a
+        valid (truthy) lock handle, so a failed/partial acquisition (e.g. ``None``
+        returned, or an exception raised after partially acquiring) can never
+        result in releasing an invalid or wrong lock.
         """
         lock = self.acquire_lock()
         try:
-            yield
+            yield lock
         finally:
-            self.release_lock(lock)
+            if lock:
+                self.release_lock(lock)
 
     async def add_qa(
         self,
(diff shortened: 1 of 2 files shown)


dexters1:
If the lock was not acquired it should fail and cause a traceback

PR #3303, opened 2026-06-23
fix(cache): safeguard hold_lock context manager against releasing invalid locks

Chandan11232:
The CacheDBInterface.hold_lock() context manager was executing self.release_lock(lock) unconditionally within its finally block. If an issue occurred during the initial acquire_lock() step (such as returning None or failing partially), the context manager would pass an invalid handle to the cleanup method. Depending on the concrete database adapter implementation, this could trigger silent errors, unexpected crashes during resource cleanup, or inadvertently release an active lock held by another worker.

Solution:

Wrapped the cleanup step inside the finally block with a strict if lock is not None: guard clause.
This prevents passing unacquired or null lock states down to the cache adapters while fully preserving the underlying generator protocols required by Python's @contextmanager.
Created a comprehensive suite of unit tests tracking both successful execution pathways and handling scenarios where a lock acquisition evaluates to None.

DCO Affirmation


dexters1:
If there is no lock or it is invalid it should fail so we can fix the lock mechanism

PR #3306, opened 2026-06-23
feat: propagate LLM budget exhaustion through Cognee API as HTTP 402

nfnt:
## Description

When an LLM provider (or LiteLLM proxy enforcing a token budget)
have no budget, cognee now raises LLMPaymentRequiredError instead of
retrying or surfacing a generic 500. All nine LLM adapters detect budget
exhaustion on any exception and convert it to LLMPaymentRequiredError,
which is excluded from tenacity retries. The cognify, search, recall,
remember, and llm routers catch LLMPaymentRequiredError before
CogneeValidationError and return HTTP 402
with an informative body so callers know to top up their token budget.

Adds a central is_budget_exhausted_error() helper in llm/exceptions.py
that consolidates three detection cases:
- HTTP 402 from any LLM provider directly
- litellm.BudgetExceededError raised by litellm's own budget manager
- LiteLLM proxy HTTP 429 with response body type="budget_exceeded"
  (proxy-specific: distinguishes budget exhaustion from ordinary rate
  limiting; depends on LiteLLM proxy response shape, see comment)

All nine LLM adapters call this helper instead and raise
LLMPaymentRequiredError on match.

## Acceptance Criteria

* [HTTP 402](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/402) responses in general and [HTTP 429](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/429) reponses by LiteLLM proxy raise `LLMPaymentRequiredError`
* Existing and added tests pass

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation


cognee/api/v1/recall/routers/get_recall_router.py
```diff
@@ -184,6 +185,14 @@ async def recall(payload: RecallPayloadDTO, user: User = Depends(get_authenticat
                 include_references=payload.include_references,
             )
             return jsonable_encoder(results)
+        except LLMPaymentRequiredError as error:
```
dexters1, line 196: These are not needed if a Cognee based exception is inherited properly, they already handle JSONResponse on their own, would be good to confirm before we remove these though that the exception is handled properly and desired info is propagated

This piece of code catches Cognee based exceptions and handles the JSONResponse return:

Image

If all desired info is propagated with it

dexters1 (changes requested):
Remove catching of the LLMPaymentRequiredError in routers and verify the existing JSONResponse mechanism for Cognee exceptions provides desired JSONResponse return

cognee/api/v1/recall/routers/get_recall_router.py
nfnt: The routers where I'm adding these have a catch-all that would swallow these errors. The exception handlers won't be used because of this. E.g. in `get_recall_router.py:217`:

```python
        except Exception as error:
            logger = get_logger()
            logger.error("Recall endpoint error: %s", error, exc_info=True)
            return JSONResponse(
                status_code=409,
                content={"error": "An error occurred during recall."},
            )
```

means that this function won't be able to raise exceptions that could be caught be an exception handler. Every exception not handled before this will be handled here.

dexters1 (approved):

Issue #3292, opened 2026-06-23
[Bug]: index_graphiti_objects mutates a shared metadata dict - GraphitiNode/EdgeType indexes embed the wrong field


yashigupta4623:
### Bug Description

**Describe the bug :**
In cognee/tasks/temporal_awareness/index_graphiti_objects.py, indexed copies share a single metadata dict, so for any DataPoint with multiple index_fields they all collapse to the last field — and that field's value is what actually gets embedded for every index.

**Where**
Lines 65-66 (for GraphitiNode) and 92-93 (for EdgeType):

indexed_data_point = graphiti_node.model_copy()           # shallow copy
indexed_data_point.metadata["index_fields"] = [field_name]  # mutates the shared dict
model_copy() is shallow (Pydantic v2), so indexed_data_point.metadata is the same dict object as the source's. Each loop iteration overwrites metadata["index_fields"] on that shared dict, so every copy ends up with [last_field].

**Impact**
GraphitiNode declares index_fields = ["name", "summary", "content"]. Because LanceDBAdapter.index_data_points selects the value to embed via getattr(data_point, data_point.metadata["index_fields"][0]), the name and summary indexes embed the content value instead of their own. Temporal-awareness retrieval on name/summary is silently wrong. EdgeType indexing has the same issue.

Note: This is the same class of bug as [#2529](https://github.com/topoteretes/cognee/issues/2529), but a different code path — the fix for #2529 (index_data_points.py) does not touch index_graphiti_objects.py, so this needs its own fix.

_Suggested fix_
Give each copy its own metadata dict instead of mutating the shared one:

```python
indexed_data_point = graphiti_node.model_copy()           # shallow copy
indexed_data_point.metadata["index_fields"] = [field_name]  # mutates the shared dict
```
(same for the EdgeType branch). I'm happy to open a PR with a regression test.

### Steps to Reproduce

Minimal reproduction of the shared-dict mutation:

    from cognee.tasks.temporal_awareness.graphiti_model import GraphitiNode

    node = GraphitiNode(name="Alice", summary="short bio", content="full content")
    # default metadata: {"index_fields": ["name", "summary", "content"]}

    a = node.model_copy()
    a.metadata["index_fields"] = ["name"]
    b = node.model_copy()
    b.metadata["index_fields"] = ["summary"]

    print(a.metadata["index_fields"])     # expected ['name']
    print(b.metadata["index_fields"])      # ['summary']
[...]

PR #3309, opened 2026-06-23
fix(temporal_awareness): prevent metadata mutation in index_graphiti_objects (#3292)

Chandan11232:
This PR resolves a silent data-correctness bug where Pydantic v2's `.model_copy()` creates a shallow copy of the `metadata` dictionary field in `index_graphiti_objects.py`. 

Because the dictionary reference was shared across loop iterations, assigning `indexed_data_point.metadata["index_fields"] = [field_name]` mutated the source object. This resulted in all vector collections collapsing to embed only the final index field (e.g., `content` overwriting `name` and `summary`), making temporal-awareness search over those fields incorrect.

### Changes
* **State Isolation:** Replaced shallow mutation with explicit dictionary unpacking (`{**...}`) for the `GraphitiNode` loop.
* **Edge Type Safety:** Applied the same isolated dictionary pattern to the `EdgeType` loop.
* **Regression Testing:** Added `tests/test_index_graphiti_objects.py` containing 6 unit tests to validate dictionary isolation, memory pointers, and loop length stability under mock conditions.

### Checklist
- [x] My code follows the code style of this project.
- [x] I have performed a self-review of my own code.
- [x] I have added tests that prove my fix is effective.
- [x] All new and existing tests passed.


Vasilije1990:
We don't want to maintain the graffiti piece since we built our own temporal resolution.

PR #3333, opened 2026-06-24
fix: initialize database before resolving CLI user

### LINKED ISSUE #3267: [Bug]: Postgres database is not initialised when running the cli
### Bug Description

After setting up a new cognee install with a fresh postgres database, all commands will fail as if it cannot connects.  Watching the logs on the database server, I can see that it tries to query the database before setting up any tables, which obviously fails. 

The workaround that I found was to run `uvicorn cognee.api.client:app` which _does_ seem to initialise the database properly.  (I am not entirely sure that it is correct yet, since cognee now consumes documents instantly and doesn't seem to cognify.)

### Steps to Reproduce

Configure cognee to use a newly created postgres database.
Run cognee-cli
It will throw an error.

### Expected Behavior

It should create the proper schema in the new database.

### Actual Behavior

It tries to query a blank database, and then throws the error:
Error: DatabaseNotCreatedError: The database has not been created yet. Please call `await setup()` first. (Status code: 422)

### Environment

Mac OS
cognee 1.2.1-local
Postgres 18

### Logs/Error Messages

```shell
[DEBUG] Real error in get_default_user: ProgrammingError('(sqlalchemy.dialects.postgresql.asyncpg.ProgrammingError) <class \'asyncpg.exceptions.UndefinedTableError\'>: relation "principals" does not exist')

[error    ] DatabaseNotCreatedError: The database has not been created yet. Please call `await setup()` first. (Status code: 422) [cognee.shared.logging_utils]
Error: DatabaseNotCreatedError: The database has not been created yet. Please call `await setup()` first. (Status code: 422)
```

### Additional Context

_No response_

### Pre-submission Checklist

IamNishant51:
Running cognee-cli against a fresh Postgres database fails with DatabaseNotCreatedError because the CLI calls get_default_user() before the database schema has been created. The API server handles this in its lifespan hook (create_database + run_migrations), but the CLI has no equivalent initialization.

Added create_database() and run_migrations() calls at the top of resolve_cli_user(), which is the common entry point for all CLI commands that need a user. Both calls are idempotent — they're no-ops on subsequent invocations.

Fixes #3267

### DIFF
--- a/cognee/cli/user_resolution.py
+++ b/cognee/cli/user_resolution.py
@@ -16,6 +16,13 @@ async def resolve_cli_user(user_id: Optional[str] = None, strict: bool = False):
     error instead — used by ownership-sensitive commands (e.g. ``agents``) where a
     silent fallback to the default user would break the isolation the flag promises.
     """
+    from cognee.infrastructure.databases.relational import get_relational_engine
+    from cognee.run_migrations import run_migrations
+
+    db_engine = get_relational_engine()
+    await db_engine.create_database()
+    await run_migrations()
+
     from cognee.modules.users.methods import get_default_user
 
     if not user_id:


Vasilije1990:
Should be resolved with the latest changes.

PR #3340, opened 2026-06-24
fix(ingestion): reopen S3 temp file by name on Windows (NamedTemporaryFile delete=False)

### LINKED ISSUE #3339: [Bug]: S3 ingestion fails on Windows — NamedTemporaryFile(delete=True) reopened by name raises PermissionError
# [Bug]: S3 ingestion fails on Windows — `NamedTemporaryFile(delete=True)` reopened by name raises `PermissionError`

## Bug Description

When ingesting data from an `s3://` path, `data_item_to_text_file()` writes the
downloaded bytes to a `tempfile.NamedTemporaryFile` and then passes
`temp_file.name` to the loader, which **reopens the file by name while the
original handle is still open**.

On Windows, `NamedTemporaryFile` defaults to `delete=True`, which opens the file
with an exclusive lock (`O_TEMPORARY`). Reopening that path by name while the
handle is open is not permitted on Windows, so ingestion of any S3 data item
fails with `PermissionError: [WinError 32]` / `[Errno 13] Permission denied`.

This is platform-specific: it works on Linux/macOS (where a `NamedTemporaryFile`
can be reopened by name) but always fails on Windows.

**Location:** [`cognee/tasks/ingestion/data_item_to_text_file.py`](https://github.com/topoteretes/cognee/blob/dev/cognee/tasks/ingestion/data_item_to_text_file.py) (the `s3` branch)

```python
with tempfile.NamedTemporaryFile(mode="wb", suffix=path_info.suffix) as temp_file:
    await pull_from_s3(data_item_path, temp_file)
    temp_file.flush()
    loader = get_loader_engine()
    return await loader.load_file(temp_file.name, preferred_loaders), loader.get_loader(
        temp_file.name, preferred_loaders   # <-- reopens by name while handle is open
    )
```

Note: the other three `NamedTemporaryFile` usages in the codebase
(`SqlAlchemyAdapter`, the ladybug adapter, `node_setup`) already use
`delete=False` and close the handle before reopening by name — this S3 path is
the one place that doesn't, so it looks like an oversight rather than an
intentional difference.

## Steps to Reproduce

Real trigger (Windows):

1. On Windows, configure S3 access and call `cognee.add("s3://your-bucket/file.txt")`
   (or any pipeline that routes an `s3://` item through `data_item_to_text_file`).
2. Ingestion fails when the loader reopens the temp file by name.

Minimal, dependency-free reproduction of the underlying mechanism (run on Windows):

```python
import tempfile

# Mirrors the current cognee S3 branch: open with default delete=True,
# write+flush, then reopen BY NAME while the handle is still open.
with tempfile.NamedTemporaryFile(mode="wb", suffix=".txt") as temp_file:
    temp_file.write(b"hello world")
    temp_file.flush()
    with open(temp_file.name, "r", encoding="utf-8") as f:  # what the loader does
        print(f.read())
```

## Expected Behavior

S3 data items are ingested successfully on Windows (same as on Linux/macOS).

## Actual Behavior

On Windows, the reopen-by-name fails:

```
PermissionError: [Errno 13] Permission denied: 'C:\\Users\\...\\AppData\\Local\\Temp\\tmpXXXXXXXX.txt'
```

(Confirmed reproducing on Windows 11, Python 3.13.7.)

## Proposed Fix

Mirror the `delete=False` pattern already used elsewhere in the codebase: create
the temp file with `delete=False`, close the handle before the loader reopens it
by name, and unlink it in a `finally` block. (Happy to open a PR — I have the fix
and a regression test ready.)

## Environment

- OS: Windows 11
- Python version: 3.13.7
- Cognee version: 1.2.1
- LLM Provider: N/A (filesystem/ingestion bug, provider-independent)

YashasviThakur:
# fix(ingestion): reopen S3 temp file by name on Windows (NamedTemporaryFile delete=False)

## What

`data_item_to_text_file()` (S3 branch) wrote the downloaded object to a
`tempfile.NamedTemporaryFile` created with the default `delete=True`, then passed
`temp_file.name` to the loader, which reopens the file **by name while the handle
is still open**. On Windows that raises `PermissionError [WinError 32]`, so every
S3 ingestion fails on Windows.

This changes the temp file to `delete=False`, closes the handle before the loader
reopens it, and unlinks the file in a `finally` block (so there is no temp-file
leak). This mirrors the `delete=False` pattern already used by the SQLAlchemy and
ladybug S3 temp-file paths.

Fixes #3339.

## Why this is correct

- The loader reads the file *during* `load_file()` / `get_loader()` (both of which
  open the path by name), and the downstream consumer in `ingest_data.py` uses the
  loader's returned storage path — not the temp file — so the temp file can safely
  be unlinked once both calls return.
- Behavior on Linux/macOS is unchanged.

## Testing

Added `cognee/tests/unit/tasks/test_data_item_to_text_file.py`:

- Drives the S3 branch with a fake loader that **reopens the temp file by name**
  (exactly what triggered the Windows failure) and asserts the written bytes are
  read back. This is strict on Windows and benign on POSIX.
- Asserts the temp file is created with `delete=False` and is removed afterwards —
  a cross-platform guard so a revert to `delete=True` fails on any OS.

Verified on Windows 11 / Python 3.13.7: the test passes on the fix and fails
(`PermissionError`) on the pre-fix code.

### DIFF
--- a/cognee/tasks/ingestion/data_item_to_text_file.py
+++ b/cognee/tasks/ingestion/data_item_to_text_file.py
@@ -45,13 +45,29 @@ async def data_item_to_text_file(
             # TODO: Rework this to work with file streams and not saving data to temp storage
             # Note: proper suffix information is needed for OpenAI to handle mp3 files
             path_info = Path(parsed_url.path)
-            with tempfile.NamedTemporaryFile(mode="wb", suffix=path_info.suffix) as temp_file:
+            # NamedTemporaryFile defaults to delete=True, which on Windows keeps an
+            # exclusive lock on the file and forbids reopening it by name while the
+            # handle is still open. The loader below reopens the file by name, so we
+            # create it with delete=False, close our handle first, and clean it up
+            # ourselves. (Mirrors the delete=False pattern used by the SQLAlchemy and
+            # ladybug S3 temp-file paths.)
+            temp_file = tempfile.NamedTemporaryFile(
+                mode="wb", suffix=path_info.suffix, delete=False
+            )
+            try:
                 await pull_from_s3(data_item_path, temp_file)
                 temp_file.flush()  # Data needs to be saved to local storage
+                temp_file.close()  # release the handle so the loader can reopen by name
                 loader = get_loader_engine()
                 return await loader.load_file(temp_file.name, preferred_loaders), loader.get_loader(
                     temp_file.name, preferred_loaders
                 )
+            finally:
+                temp_file.close()  # idempotent; covers the early-exception path
+                try:
+                    os.unlink(temp_file.name)
+                except OSError:
+                    pass
 
         # data is local file path
         elif parsed_url.scheme == "file":
--- a/cognee/tests/unit/tasks/test_data_item_to_text_file.py
+++ b/cognee/tests/unit/tasks/test_data_item_to_text_file.py
@@ -1,0 +1,89 @@
+"""Regression test for the Windows temp-file reopen bug in
+``data_item_to_text_file`` (S3 ingestion path).
+
+``tempfile.NamedTemporaryFile`` defaults to ``delete=True``, which on Windows
+keeps an exclusive lock on the file and forbids reopening it *by name* while the
+handle is still open. The S3 branch writes the downloaded bytes to such a temp
+file and then hands ``temp_file.name`` to the loader, which reopens it by name --
+raising ``PermissionError`` on Windows.
+
+The fix creates the temp file with ``delete=False``, closes the handle before the
+loader reopens it, and unlinks it afterwards. These tests lock that in:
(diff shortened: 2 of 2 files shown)


rshkarin:
@YashasviThakur pease add DCO sign-off: "I affirm that all code in every commit of this pull request conforms to the terms of the Topoteretes Developer Certificate of Origin" into your PR description.

YashasviThakur:
@rshkarin Added the DCO affirmation to the PR description — should be good now.

YashasviThakur:
@rshkarin DCO is resolved and CI is green aside from the secrets-gated test suites (expected for external PRs, per the earlier discussion). Ready for review whenever you get a chance — thanks!

PR #3342, opened 2026-06-24
fix(ingestion): derive basename from Windows paths in classify()

### LINKED ISSUE #3341: [Bug]: On Windows, classify() stores the full file path as the document name (splits on '/' only)
# [Bug]: On Windows, `classify()` stores the full file path as the document name (splits on `/` only)

## Bug Description

`cognee.modules.ingestion.classify()` derives the name of a binary stream like this:

```python
# cognee/modules/ingestion/classify.py
if isinstance(data, BufferedReader) or isinstance(data, SpooledTemporaryFile):
    return BinaryData(data, filename if filename else str(data.name).split("/")[-1])
```

`data.name` for a file-backed `BufferedReader` is the path the file was opened
with. On Windows that path uses backslashes (e.g. `C:\Users\me\report.pdf`) and
contains **no forward slashes**, so `str(data.name).split("/")[-1]` returns the
**entire path** instead of the basename.

That value becomes `BinaryData.name`, which flows through `get_file_metadata(...)`
into the stored document's `name` / metadata (and `data_point.name`). So on
Windows, ingested local files are recorded with names like
`C:\Users\me\report.pdf` instead of `report.pdf`.

This is platform-specific: it works on Linux/macOS (forward-slash paths) and is
wrong on Windows.

The codebase already has the correct idiom elsewhere —
`_normalize_filename` in `cognee/tasks/ingestion/utils.py` does
`str(filename).replace("\\", "/").split("/")[-1]` — so `classify()` looks like an
oversight.

## Steps to Reproduce

Minimal, dependency-free reproduction (run on Windows):

```python
import tempfile, os

d = tempfile.mkdtemp()
fpath = os.path.join(d, "annual_report.pdf")
open(fpath, "wb").write(b"%PDF-1.4 test")

reader = open(fpath, "rb")            # a real io.BufferedReader, like ingestion uses
print("reader.name:", reader.name)    # C:\Users\...\annual_report.pdf

# what classify() currently does:
print("derived name:", str(reader.name).split("/")[-1])
```

Or, end to end: on Windows, ingest a local file and inspect the stored document
name — it will be the full `C:\...` path rather than the filename.

## Expected Behavior

The document name is the file's basename (`annual_report.pdf`) on all platforms.

## Actual Behavior

On Windows the document name is the full path, e.g.:

```
C:\Users\lenovo\AppData\Local\Temp\tmpkfq79s6i\annual_report.pdf
```

(Confirmed on Windows 11, Python 3.13.7.)

## Proposed Fix

Normalize both separators before taking the basename (mirrors
`_normalize_filename`):

```python
derived_name = str(data.name).replace("\\", "/").split("/")[-1]
```

I have the fix and cross-platform regression tests ready — happy to open a PR.

## Environment

- OS: Windows 11
- Python version: 3.13.7
- Cognee version: 1.2.1
- LLM Provider: N/A (ingestion/metadata bug, provider-independent)

YashasviThakur:
# fix(ingestion): derive basename from Windows paths in classify()

## What

`classify()` derived a binary stream's name with `str(data.name).split("/")[-1]`.
On Windows, `data.name` is a backslash path (e.g. `C:\dir\file.pdf`) with no
forward slashes, so the split returned the **entire path** as the file's name.
That value propagates into the stored document name and metadata
(`BinaryData.name` → `get_file_metadata` → `data_point.name`), so on Windows
ingested local files were recorded with their full `C:\...` path instead of the
basename.

This normalizes both `\` and `/` separators before taking the basename, matching
the convention already used by `_normalize_filename` in
`cognee/tasks/ingestion/utils.py`.

```diff
- return BinaryData(data, filename if filename else str(data.name).split("/")[-1])
+ derived_name = str(data.name).replace("\\", "/").split("/")[-1]
+ return BinaryData(data, filename if filename else derived_name)
```

Fixes #3341.

## Testing

Added `cognee/tests/unit/modules/ingestion/test_classify.py` with cross-platform
regression tests:

- Windows-style input (`C:\Users\foo\report.pdf`) → `report.pdf`
- POSIX-style input (`/home/user/report.pdf`) → `report.pdf`
- explicit `filename` argument takes precedence over the stream name

Because the Windows-style input contains backslashes regardless of host, a revert
to splitting on `/` only fails the test on **any** OS (Linux CI included).
Verified on Windows 11 / Python 3.13.7: the fix returns the basename where the
pre-fix code returned the full path.

### DIFF
--- a/cognee/modules/ingestion/classify.py
+++ b/cognee/modules/ingestion/classify.py
@@ -14,7 +14,11 @@ def classify(
         return TextData(data)
 
     if isinstance(data, BufferedReader) or isinstance(data, SpooledTemporaryFile):
-        return BinaryData(data, filename if filename else str(data.name).split("/")[-1])
+        # Normalize both POSIX ("/") and Windows ("\") separators before taking the
+        # basename. On Windows, data.name is a backslash path (e.g. C:\dir\file.pdf),
+        # so splitting on "/" alone would keep the whole path as the file's name.
+        derived_name = str(data.name).replace("\\", "/").split("/")[-1]
+        return BinaryData(data, filename if filename else derived_name)
 
     try:
         from s3fs import S3File
--- a/cognee/tests/unit/modules/ingestion/test_classify.py
+++ b/cognee/tests/unit/modules/ingestion/test_classify.py
@@ -1,0 +1,40 @@
+"""Regression tests for ``classify`` filename derivation.
+
+``classify`` derived a binary stream's name with ``str(data.name).split("/")[-1]``.
+On Windows, ``data.name`` is a backslash path (e.g. ``C:\\dir\\file.pdf``) with no
+forward slashes, so the split returned the *entire path* as the file's name --
+which then propagated into the stored document name / metadata.
+
+The fix normalizes both ``\\`` and ``/`` before taking the basename. These tests
+are cross-platform strict: a revert to splitting on ``/`` only fails on any OS,
+because the Windows-style input below contains backslashes regardless of host.
+"""
+
+import io
+from unittest.mock import MagicMock
+
+from cognee.modules.ingestion.classify import classify
+from cognee.modules.ingestion.data_types import BinaryData
+
+
+def _fake_binary_stream(name: str) -> MagicMock:
+    # spec=io.BufferedReader makes isinstance(stream, BufferedReader) pass in classify.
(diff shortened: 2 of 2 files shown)


rshkarin:
@YashasviThakur thanks for checking out! Superseded by https://github.com/topoteretes/cognee/pull/4157.

PR #3344, opened 2026-06-24
fix(config): validate cache_root_directory as an absolute path

### LINKED ISSUE #3343: [Bug]: cache_root_directory skips the absolute-path validation applied to every other root directory
# [Bug]: `cache_root_directory` skips the absolute-path validation applied to every other root directory

## Bug Description

`BaseConfig.validate_paths` enforces absolute paths for the root directories:

```python
# cognee/base_config.py
# Require absolute paths for root directories
self.data_root_directory = ensure_absolute_path(self.data_root_directory)
self.system_root_directory = ensure_absolute_path(self.system_root_directory)
self.logs_root_directory = ensure_absolute_path(self.logs_root_directory)
```

`ensure_absolute_path` raises `ValueError` on a relative path (and passes `s3://`
URLs through unchanged). But `cache_root_directory` — also a root directory — is
**the only one of the four not validated**.

As a result, a relative `CACHE_ROOT_DIRECTORY` is silently accepted and later
resolved against the current working directory (see `cognee/shared/cache.py`,
which feeds it into `get_file_storage(...)` and `os.path.abspath(...)`), so the
cache lands in different places depending on where the process is started.
The same relative value for `data_root_directory` / `system_root_directory` /
`logs_root_directory` raises a clear error — an inconsistent, surprising
behavior.

## Steps to Reproduce

```python
from cognee.base_config import BaseConfig

# A relative value for the documented siblings is rejected:
BaseConfig(data_root_directory="some_relative_dir")   # -> ValueError

# But the same relative value for cache_root_directory is silently accepted:
cfg = BaseConfig(cache_root_directory="some_relative_dir")
print(cfg.cache_root_directory)   # 'some_relative_dir'  (resolved vs cwd later)
```

## Expected Behavior

`cache_root_directory` is validated like the other root directories: relative
paths are rejected with the same `ValueError`, and `s3://` URLs are preserved.

## Actual Behavior

A relative `cache_root_directory` is accepted and silently resolved against the
current working directory, inconsistent with the other three root directories.

## Proposed Fix

Validate `cache_root_directory` alongside its siblings (safe for `s3://`, which
`ensure_absolute_path` returns unchanged):

```python
self.cache_root_directory = ensure_absolute_path(self.cache_root_directory)
```

I have the fix and regression tests ready — happy to open a PR.

## Environment

- Cognee version: 1.2.1 (`dev`)
- Python version: 3.13.7
- OS: cross-platform (config-validation bug)
- LLM Provider: N/A

YashasviThakur:
# fix(config): validate cache_root_directory as an absolute path

## What

`BaseConfig.validate_paths` enforces absolute paths for `data_root_directory`,
`system_root_directory` and `logs_root_directory` via `ensure_absolute_path`
(which raises on a relative path and passes `s3://` URLs through unchanged).
`cache_root_directory` is a root directory too, but was the only one of the four
left unvalidated.

So a relative `CACHE_ROOT_DIRECTORY` was silently accepted and later resolved
against the current working directory (`cognee/shared/cache.py` →
`get_file_storage(...)` / `os.path.abspath(...)`), making the cache location
depend on the process's cwd — while the same value for its siblings raises a
clear error.

```diff
 self.data_root_directory = ensure_absolute_path(self.data_root_directory)
 self.system_root_directory = ensure_absolute_path(self.system_root_directory)
 self.logs_root_directory = ensure_absolute_path(self.logs_root_directory)
+self.cache_root_directory = ensure_absolute_path(self.cache_root_directory)
```

`ensure_absolute_path` returns `s3://` URLs unchanged, so the S3 cache
auto-configuration above it is unaffected.

Fixes #3343.

## Testing

Added `cognee/tests/unit/test_base_config.py`:

- relative `cache_root_directory` is rejected (regression guard; fails on the
  pre-fix code)
- a relative sibling (`data_root_directory`) is rejected too (shows the intended,
  consistent behavior)
- `s3://` cache path is preserved unchanged (the fix doesn't break S3)
- an absolute cache path is accepted

Cross-platform (no OS-specific behavior).

### DIFF
--- a/cognee/base_config.py
+++ b/cognee/base_config.py
@@ -33,6 +33,9 @@ class BaseConfig(BaseSettings):
         self.data_root_directory = ensure_absolute_path(self.data_root_directory)
         self.system_root_directory = ensure_absolute_path(self.system_root_directory)
         self.logs_root_directory = ensure_absolute_path(self.logs_root_directory)
+        # cache_root_directory is a root directory too; validate it like the others
+        # (ensure_absolute_path passes s3:// URLs through unchanged).
+        self.cache_root_directory = ensure_absolute_path(self.cache_root_directory)
 
         return self
 
--- a/cognee/tests/unit/test_base_config.py
+++ b/cognee/tests/unit/test_base_config.py
@@ -1,0 +1,48 @@
+"""Regression tests for BaseConfig root-directory validation.
+
+`validate_paths` enforces absolute paths for the root directories via
+`ensure_absolute_path` (which raises on a relative path and passes ``s3://`` URLs
+through unchanged). ``cache_root_directory`` is a root directory too, but was the
+only one of the four not validated, so a relative ``CACHE_ROOT_DIRECTORY`` was
+silently accepted and later resolved against the current working directory --
+unlike its siblings, which reject the same value with a clear error.
+
+These tests lock in that ``cache_root_directory`` is validated identically.
+"""
+
+from pathlib import Path
+
+import pytest
+
+from cognee.base_config import BaseConfig
+
+
+def _clear_env(monkeypatch):
+    monkeypatch.delenv("CACHE_ROOT_DIRECTORY", raising=False)
+    monkeypatch.delenv("STORAGE_BACKEND", raising=False)
+
+
+def test_relative_cache_root_directory_is_rejected(monkeypatch):
+    _clear_env(monkeypatch)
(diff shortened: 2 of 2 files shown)


dexters1:
Thanks for the PR @YashasviThakur! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #3354, opened 2026-06-24
fix(serve): health-check first when reusing saved cloud credentials

### LINKED ISSUE #3249: [Feature]: Add “Connection Check + Fallback” to cloud serve()
### Problem Statement

In cloud mode, serve() can be blocked by Management/Auth token expiry even when the cached creds.api_key and creds.service_url would still successfully connect. This makes SDK startup fragile when Management/Auth endpoints are temporarily unavailable.

### Proposed Solution

- When loading saved credentials, attempt client._health_check() immediately using creds.service_url + creds.api_key.
- If health check succeeds, reuse the connection without requiring is_token_expired(creds) to be false.
- Only perform Auth0 refresh / reauth as a fallback when the health check fails.

### Alternatives Considered

_No response_

### Use Case

Users who are using the SDK with previously saved credentials want cognee.serve() to work reliably even if Auth0/Management services are slow or intermittently failing.

### Implementation Ideas

In cognee/api/v1/serve/serve.py inside _serve_cloud(), reorder logic:
Build CloudClient(creds.service_url, creds.api_key).
Try _health_check() first.
If it passes: set remote client and return.
If it fails: then run the existing token-expiry refresh/reauth flow (including refresh_access_token).
Add a short timeout/retry strategy for _health_check() to reduce flakiness.
Keep existing behavior as a fallback to preserve security expectations (do not assume token validity; rely on health check success).

### Additional Context

_No response_

### Pre-submission Checklist

syf2211:
Fixes #3249.

## Problem

`cognee.serve()` in cloud mode currently gates on `is_token_expired(creds)` *before* trying to reach the cognee service. When the saved `(service_url, api_key)` is perfectly able to talk to the service, a slow or temporarily-down Auth0 / Management backend still blocks SDK startup — because we never even try the cheap local API-key path.

Compounding the issue: the existing `CloudClient._health_check()` inherits the session's 5-minute total / 30-second connect timeout. A single health probe at startup could stall for half a minute on a degraded network before falling through to token refresh.

## Fix

Two minimal, additive changes:

1. **`CloudClient._health_check()` accepts an optional `timeout`**, defaulting to `None` (preserves prior behaviour for every existing caller). A new class attribute `HEALTH_CHECK_TIMEOUT = ClientTimeout(total=5, sock_connect=2)` is defined for the fast path.

2. **`_serve_cloud` adds a fast path**: before any token check, attempt a short-timeout health probe against the cached `(service_url, api_key)`. On success, set the remote client and return — no Auth0 round-trip. On failure or exception, fall through to the existing token-expired / refresh / device-code path *unchanged*.

This is strictly a fast-path addition. Every existing behaviour is preserved as the fallback.

## Tests

- `test_cloud_client_health_check_timeout.py` — locks in the default behaviour of `_health_check` (no explicit timeout → no `timeout` kwarg) and the new fast-path timeout (5s / 2s), plus non-2xx and connection-failure handling.
- `test_serve_cloud_health_check_first.py` — drives the actual `_serve_cloud` with mocked external calls. Asserts:
  - **token expired + health OK → uses saved creds, refresh NOT called** (the new behaviour).
  - **token expired + health fails → still falls through to refresh** (regression guard for the original path).

All 9 tests pass (3 pre-existing + 6 new). `ruff check` and `ruff format --check` both clean.

## Out-of-scope

The issue mentions a follow-up "Add short timeout/retry strategy for _health_check() to reduce flakiness" — I've added the short timeout, but not a retry strategy. A single 5s attempt should be plenty for the common case; if maintainers want retries, that should be a follow-up so we can discuss policy (retry budget, exponential vs fixed backoff, etc.) on its own.

## Notes on test patching

`cognee/api/v1/serve/serve.py:111-129` lazy-imports its dependencies (`from X import Y`) inside `_serve_cloud`. I verified empirically that `patch.object(creds_mod, "load_credentials", ...)` works correctly because each call re-evaluates the import — patching the *source module* is the right target, not the enclosing `serve_mod`. There's a long docstring in the test explaining this for future contributors.

### DIFF
--- a/cognee/api/v1/serve/cloud_client.py
+++ b/cognee/api/v1/serve/cloud_client.py
@@ -31,6 +31,11 @@ class CloudClient:
     # can legitimately exceed any fixed total; per-read inactivity stays
     # bounded instead. Applied per-request, only to archive uploads.
     UPLOAD_TIMEOUT = aiohttp.ClientTimeout(total=None, sock_connect=30, sock_read=300)
+    # Health checks should fail fast so that startup can fall through to
+    # token refresh or device-code login. 5s total is plenty for a healthy
+    # service; any slower means Auth0/management is degraded, which is
+    # exactly the case we want to detect quickly. See issue #3249.
+    HEALTH_CHECK_TIMEOUT = aiohttp.ClientTimeout(total=5, sock_connect=2)
 
     async def _get_session(self) -> aiohttp.ClientSession:
         if self._session is None or self._session.closed:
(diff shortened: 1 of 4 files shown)


dexters1:
Thanks for the PR @syf2211! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #3427, opened 2026-06-24
fix(cli): log full traceback on delegated API command failure (#3335)

### LINKED ISSUE #3335: [Bug]: CLI “delegate to API server” exception loses original stack/context
### Bug Description

In cognee/cli/_cognee.py, when --api-url is used and command delegation fails, the CLI prints only str(ex) and returns 1. The original exception details are only re-raised when debug is enabled. This makes failures hard to diagnose in production/CI and can mask the real root cause.

### Steps to Reproduce

- Start CLI with --api-url pointing to an API server.
- Trigger any exception inside cognee.cli.api_dispatch.dispatch(args) (e.g., misconfigured server URL, bad auth, or an internal handler error).

### Expected Behavior

CLI should print the full error context (stack trace) for non-debug runs at least in a log file, and/or include the exception class and stack trace in output.

### Actual Behavior

The CLI catches Exception as ex and does:
fmt.error(str(ex))
returns 1
The exception is only re-raised when debug.is_debug_enabled() is true.

### Environment

- OS: Windows 11
- Code path: cognee/cli/_cognee.py (main() function)

### Logs/Error Messages

```shell
No full stack trace is printed by default; only the message:
fmt.error(str(ex))
When debug is enabled, it would re-raise:
raise ex
```

### Additional Context

_No response_

### Pre-submission Checklist

nexiouscaliver:
Hi, this addresses #3335.

When a command is delegated to the API server through `--api-url` and that delegation fails, the CLI only prints the error message and returns exit code 1. The exception type and full traceback are thrown away unless `--debug` is on, which makes failures in CI and production hard to trace back to the real cause.

The change is small and stays in the delegated-command path. In the except block in `cognee/cli/_cognee.py` I now log the full exception with `exc_info=True` before printing the friendly message. cognee uses structlog, so that renders the complete traceback into the log regardless of whether debug mode is enabled. The console output and exit code are unchanged, so nothing about the user-facing behavior changes.

I scoped this to the delegated-command path that the issue describes. The same pattern appears in two sibling except blocks in `main()`, the UI-start path and the local-command path, but I left those alone to keep this change focused. Happy to follow up on those separately if you want them handled the same way.

I added a regression test to the CLI unit suite that makes a delegated command fail with `--debug` off and asserts the full exception is logged with `exc_info`, while the friendly message is still printed and the exit code stays 1.

Closes #3335

---


dexters1:
Thanks for the PR @nexiouscaliver! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #3435, opened 2026-06-24
feat(migration): add LangChain and LlamaIndex memory sources

Ashutoshx7:
## Summary

A lot of people already have memory living inside LangChain or LlamaIndex.
Chat histories, docstores, knowledge graph indices, entity stores, and
there wasn't a clean way to pull that into Cognee. This PR adds two
`MemorySource` adapters so you can hand any of those objects to
`cognee.remember()` and have them land as the right Cognee types.

For LangChain, chat history becomes episodes, docstores and vector-store
docstores become documents, `ConversationKGMemory` triples become entities
and facts, and `ConversationEntityMemory` entity stores become entities.
LlamaIndex works the same way. Docstores and node lists become documents,
and `KnowledgeGraphIndex` / `PropertyGraphIndex` triples become entities
and facts. If you're on LlamaIndex's newer property graph stuff,
`BaseNode.relationships` is preserved too.

Both adapters are duck-typed on purpose, so they don't import LangChain
or LlamaIndex at all. You can pass live framework objects, plain Python
dicts/lists shaped the same way, or a JSON export from either library
and it all goes through the same path. That keeps the optional extras
truly optional and makes the tests trivial to run.

Speaking of tests, they're pure. No library installs, no DB, no LLM
calls. The existing `import_memory_source` `FakeSource` tests already
cover the `remember()` dispatch wiring, so these new tests just exercise
the shape-mapping logic.

Default is re-derive mode, since the stuff coming out of these stores is
mostly lossy chunks and messages. You usually want Cognee to re-extract
rather than trust whatever the upstream framework decided was a "memory."

## Test plan
- [x] `pytest cognee/tests/unit/migration/test_langchain_source.py`
- [x] `pytest cognee/tests/unit/migration/test_llama_index_source.py`
- [x] Existing `FakeSource` tests still pass
- [x] `pre-commit run --all-files`

## AI disclosure

I used Claude Code (Anthropic) to help with this PR, mostly to map out
the LangChain and LlamaIndex memory APIs (there's a lot of surface area
across versions), sanity-check which objects make sense to migrate to
which Cognee type, and draft the duck-typed adapters and tests. I read
through everything, ran the tests locally, and made the design calls
(e.g. duck-typing instead of hard imports, re-derive as the default).
Flagging it here for transparency.


Ashutoshx7:
@dexters1 lmk your thoughts on this

PR #3355, opened 2026-06-24
fix: make HUGGINGFACE_TOKENIZER optional for Ollama embeddings

darshan-codes-ai:
## Summary
Fixes #3353

HUGGINGFACE_TOKENIZER was incorrectly treated as a required variable
alongside EMBEDDING_PROVIDER, EMBEDDING_MODEL, and EMBEDDING_DIMENSIONS.
This caused a ValidationError on import for every new user setting up
Cognee locally with Ollama — before any code even runs.

## Changes
- Removed HUGGINGFACE_TOKENIZER from the required embedding env vars check
- Updated the error message to reflect the actual required variables
- HUGGINGFACE_TOKENIZER remains usable as an optional variable

## Testing
Verified that importing cognee with Ollama embedding config and WITHOUT
HUGGINGFACE_TOKENIZER now works correctly:
- OS: Windows 11
- Python: 3.13.7
- Cognee: 1.2.1-local
- Embedding: nomic-embed-text via Ollama

Issue #3353, opened 2026-06-24
[Bug]: [Bug]: HUGGINGFACE_TOKENIZER incorrectly required when using Ollama embeddings


darshan-codes-ai:
### Bug Description

When setting up Cognee with Ollama as both LLM and embedding provider, 
Cognee throws a validation error requiring `HUGGINGFACE_TOKENIZER` even 
though it should be optional for Ollama users.

- OS: Windows 11
- Python: 3.13.7
- Cognee version: 1.2.1
- Embedding Provider: Ollama (nomic-embed-text)

### Steps to Reproduce

Set these environment variables and import cognee:

```python
os.environ["EMBEDDING_PROVIDER"] = "ollama"
os.environ["EMBEDDING_MODEL"] = "nomic-embed-text:latest"
os.environ["EMBEDDING_DIMENSIONS"] = "768"
```

### Steps to Reproduce

1. Install Cognee with Ollama support:
   pip install "cognee[ollama]"

2. Set these environment variables before importing cognee:

   os.environ["LLM_PROVIDER"] = "ollama"
   os.environ["LLM_MODEL"] = "phi3:latest"
   os.environ["LLM_ENDPOINT"] = "http://localhost:11434/v1"
   os.environ["LLM_API_KEY"] = "ollama"
   os.environ["EMBEDDING_PROVIDER"] = "ollama"
   os.environ["EMBEDDING_MODEL"] = "nomic-embed-text:latest"
   os.environ["EMBEDDING_ENDPOINT"] = "http://localhost:11434/api/embed"
   os.environ["EMBEDDING_DIMENSIONS"] = "768"

3. Then import cognee:
   import cognee

4. The error appears immediately on import — before any 
   add() or cognify() is even called.

### Expected Behavior

`HUGGINGFACE_TOKENIZER` should be optional when using Ollama embeddings. 
It should only be required when the user explicitly needs HuggingFace 
token counting. New users setting up local Ollama will hit this confusing 
error with no clear fix in the docs.

### Actual Fix Applied

Adding this variable resolved it:

### Actual Behavior

Cognee throws a ValidationError immediately on import and 
crashes before any code runs:

ValidationError: 1 validation error for LLMConfig
  Value error, You have set some but not all of the required 
  environment variables for embeddings (EMBEDDING_PROVIDER, 
  EMBEDDING_MODEL, EMBEDDING_DIMENSIONS, HUGGINGFACE_TOKENIZER). 
[...]

PR #3441, opened 2026-06-25
fix: make huggingface_tokenizer optional in OllamaEmbeddingEngine

darshan-codes-ai:
## Summary
Related to #3353

When using Ollama as the embedding provider, OllamaEmbeddingEngine 
was defaulting to "Salesforce/SFR-Embedding-Mistral" as the 
huggingface_tokenizer even when the user never set HUGGINGFACE_TOKENIZER.
This caused silent wrong tokenizer usage as noted by @Vasilije1990 
in PR #3355.

## Changes
- Changed huggingface_tokenizer parameter default from 
  "Salesforce/SFR-Embedding-Mistral" to None in OllamaEmbeddingEngine
- Updated get_tokenizer() to skip loading when tokenizer is None
- Tokenizer loading is now fully optional for Ollama users

## Testing
Verified that importing cognee with Ollama embedding config and 
WITHOUT setting HUGGINGFACE_TOKENIZER works correctly with no 
wrong defaults:
- OS: Windows 11
- Python: 3.13.7
- Embedding: nomic-embed-text via Ollama

### DIFF
--- a/cognee/infrastructure/databases/vector/embeddings/OllamaEmbeddingEngine.py
+++ b/cognee/infrastructure/databases/vector/embeddings/OllamaEmbeddingEngine.py
@@ -65,7 +65,7 @@ class OllamaEmbeddingEngine(EmbeddingEngine):
         dimensions: Optional[int] = 1024,
         max_completion_tokens: int = 512,
         endpoint: Optional[str] = "http://localhost:11434/api/embed",
-        huggingface_tokenizer: str = "Salesforce/SFR-Embedding-Mistral",
+        huggingface_tokenizer: Optional[str] = None,
         batch_size: int = 100,
     ):
         self.model = model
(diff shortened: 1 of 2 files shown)


darshan-codes-ai:
Hi @Vasilije1990 @dexters1 @siillee 👋

I wanted to follow up on this PR. This fix makes huggingface_tokenizer optional in OllamaEmbeddingEngine 
so that users don't silently get the wrong default tokenizer when using Ollama embeddings.

This is a follow-up to my merged PR #3355 which made HUGGINGFACE_TOKENIZER optional at the config level. This PR completes the fix at the engine level as suggested by @Vasilije1990 in the review of #3355.

I have tested this on:
- OS: Windows 11
- Python: 3.13.7
- Cognee: 1.2.1-local
- Embedding: nomic-embed-text via Ollama

Would really appreciate a review when you get a chance! 🙏
Thank you for your time and for maintaining this amazing project!

PR #3448, opened 2026-06-25
fix: prevent unbounded Redis session list growth and OOM on retrievalFix redis memory leak

Rishiii2:
This PR addresses an unbounded memory growth issue in the Redis fast-cache and a corresponding memory exhaustion risk in Python.

While auditing the session memory pipeline, I noticed that rpush is used without any list truncation (ltrim) in the RedisAdapter. Since every interaction also refreshes the session TTL, active long-running sessions will grow indefinitely in Redis. Furthermore, retrieving these sessions via _load_entries fetches the entire list into the Python heap simultaneously via lrange(0, -1), which could cause massive latency spikes or OOM errors for large sessions.

To fix this safely, this PR:

Introduces max_session_history_items to CacheConfig (defaulting to 1000). Setting it to None acts as an escape hatch to keep the legacy unbounded behavior.
Implements ltrim as a sliding window in the Redis adapter immediately after appending new turns.
Updates _load_entries to perform chunked reads when fetching full lists, providing defense-in-depth against OOMs for any existing massive lists already in production.
Note on Safety / Data Loss: Since index_session_qa is called synchronously alongside create_qa_entry, every turn is still durably indexed into the Vector DB (SESSION_QA_VECTOR_COLLECTION) before ltrim drops anything. The Redis fast-cache safely retains the most recent 1000 turns for low-latency context, while the semantic graph accurately retains all historical context.

Type of Change
 Bug fix (non-breaking change which fixes an issue)
 Performance Improvement
Testing
Added test_max_session_history_limit in test_redis_adapter_crud.py to verify the ltrim sliding window logic works as expected.
Verified no regressions in existing cache CRUD tests.

### DIFF
--- a/cognee/infrastructure/databases/cache/config.py
+++ b/cognee/infrastructure/databases/cache/config.py
@@ -31,6 +31,8 @@ class CacheConfig(BaseSettings):
     - auto_feedback: When caching is True, run automatic feedback detection and session-context
       guidance on each query (default True). Adds one structured-output LLM call per answered
       turn; set AUTO_FEEDBACK=false to disable.
+    - max_session_history_items: Maximum number of QA turns, agent traces, and session contexts
+      kept in fast-cache per session (default: 1000). Set to None to keep unbounded history.
     """
 
     cache_backend: Literal["redis", "fs", "tapes", "sqlite", "postgres"] = "sqlite"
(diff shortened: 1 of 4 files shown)


dexters1:
Thanks for the PR @Rishiii2! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #3469, opened 2026-06-25
fix(retrieval): align triplet search results with searchable entities

### LINKED ISSUE #3465: [Bug]: TripletSearchContextProvider misaligns entities and results when an entity has empty text
### Bug Description

TripletSearchContextProvider builds search tasks only for entities with non-empty text,
but then pairs the results back against the full entities list using zip(). When any
entity has empty text, len(results) < len(entities), so zip() pairs each result with the
wrong entity and silently drops the trailing ones. The LLM is handed context attributed
to the wrong entities, with no error raised.

_get_search_tasks filters at TripletSearchContextProvider.py:56-57 (tasks built only for
entities with text), while get_context zips against the unfiltered list at line 92.

### Steps to Reproduce

1. Call a retrieval path that uses TripletSearchContextProvider with a set of entities
   where at least one entity has empty/missing text and at least one later entity has text.
2. _get_search_tasks skips the empty-text entity, so results is shorter than entities.
3. get_context does zip(entities, results) over mismatched lengths.

Minimal repro (mocking the search so no live LLM/DB is needed): construct entities
[A(text=""), B(text="..."), C(text="...")], stub the search to return one result per
text-bearing entity, and assert each returned context block maps to the correct entity.
The pairing is off by one and C is dropped.

### Expected Behavior

Each entity's context should map to that same entity's triplets, and no entity should be
silently dropped — regardless of whether some entities have empty text.

### Actual Behavior

Because results is filtered but entities is not, zip() shifts the pairing after the first
empty-text entity and drops the trailing entities entirely. Wrong context is attributed
to the wrong entities and returned to the LLM with no exception. Confirmed by source
tracing; the alignment test reproduces it deterministically with the search mocked.

### Environment

Traced against source at commit 15e48600c (v1.2.1); intended runtime Python 3.10-3.14
per pyproject.toml.

### Logs/Error Messages

```shell
None — the failure is silent (no exception). Manifests as incorrect retrieval context.
```

### Additional Context

Fix: return the filtered entity list alongside the tasks and zip/format against that
filtered list so results and entities stay parallel. I have this on a branch
(fix/triplet-context-zip-alignment, cut from 15e48600c) with a unit test that mocks the
search to assert per-entity alignment, and am happy to open a PR.

### Pre-submission Checklist

syf2211:
Fixes #3465

## Summary

`TripletSearchContextProvider.get_context()` now pairs triplet search results with the same filtered entity list used to build search tasks, instead of the full input entity list.

## Motivation

`_get_search_tasks()` only creates tasks for entities with non-empty searchable text, but `get_context()` previously zipped `asyncio.gather()` results against the unfiltered `entities` list. When any entity had empty text, `zip()` shifted pairings and silently dropped trailing entities, returning context attributed to the wrong entities.

## Changes

- Filter `searchable_entities` with the same `_get_entity_text()` guard used by `_get_search_tasks()`.
- Pass `searchable_entities` into `_results_to_context()` so each result maps to the entity that produced it.
- Add a focused unit test that mocks triplet search and asserts per-entity alignment when a leading entity has empty text.

## Tests

```bash
uv run pytest cognee/tests/unit/modules/retrieval/test_triplet_search_context_provider.py -q
# 1 passed

uv run ruff check cognee/modules/retrieval/context_providers/TripletSearchContextProvider.py cognee/tests/unit/modules/retrieval/test_triplet_search_context_provider.py
# All checks passed
```

## Notes

- `SummarizedTripletSearchContextProvider` inherits `get_context()` and is covered by the same fix.
- No behavior change for the all-empty-entities case; it still returns `"No valid entities found for context search."`

### DIFF
--- a/cognee/modules/retrieval/context_providers/TripletSearchContextProvider.py
+++ b/cognee/modules/retrieval/context_providers/TripletSearchContextProvider.py
@@ -88,5 +88,8 @@ class TripletSearchContextProvider(BaseContextProvider):
         if not search_tasks:
             return "No valid entities found for context search."
 
+        searchable_entities = [
+            entity for entity in entities if self._get_entity_text(entity) is not None
+        ]
         results = await asyncio.gather(*search_tasks)
-        return await self._results_to_context(entities, results)
+        return await self._results_to_context(searchable_entities, results)
--- a/cognee/tests/unit/modules/retrieval/test_triplet_search_context_provider.py
+++ b/cognee/tests/unit/modules/retrieval/test_triplet_search_context_provider.py
@@ -1,0 +1,45 @@
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, patch
+
+import pytest
+
+from cognee.modules.retrieval.context_providers.TripletSearchContextProvider import (
+    TripletSearchContextProvider,
+)
+
+
+@pytest.mark.asyncio
+async def test_get_context_pairs_results_with_matching_entities_when_some_have_empty_text():
+    provider = TripletSearchContextProvider()
+
+    entities = [
+        SimpleNamespace(name="", description="", text=""),
+        SimpleNamespace(name="", description="entity b"),
+        SimpleNamespace(name="", description="entity c"),
+    ]
+
+    async def fake_triplet_search(*, query: str, **_kwargs):
+        entity_text = query[: -len(" user query")]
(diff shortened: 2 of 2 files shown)


dexters1:
@syf2211 closing PR as a duplicate, a different contributor already made a PR for the same issue

PR #3489, opened 2026-06-26
fix(llm): use provider-aware defaults for ollama chunk sizing

### LINKED ISSUE #3436: [Bug]: Chunk size is based on the output token limit, not the model context window (Ollama)
### Bug Description

cognee decides how big each document chunk should be using max_completion_tokens. That value is the model's output token limit, not its context window. 
For Ollama, the model name is never found in LiteLLM's lookup table, so the value returns None and cognee falls back to its default of 16384. Half of that is 8192, and after the embedding cap the chunk size becomes 8191 tokens. Ollama loads these models with a default context of 4096 tokens (older Ollama versions used 2048). So cognee builds chunks about twice as large as the context the model actually has.

### Steps to Reproduce

1. Configure cognee for Ollama using the documented .env settings (LLM_PROVIDER="ollama", LLM_MODEL="llama3.1:8b", and the matching embedding settings).

2. In a Python shell, call cognee's own chunk-sizing functions:

from cognee.infrastructure.llm.utils import (
    get_model_max_completion_tokens,
    get_max_chunk_tokens,
)

for m in ["llama3.1:8b", "llama3.2:3b", "qwen2.5:7b", "mistral:7b"]:
    print(m, "->", get_model_max_completion_tokens(m))

print("chunk size cognee uses:", get_max_chunk_tokens())

3. See that every lookup prints None and the chunk size is 8191, which is larger than the model's 4096 context.

### Expected Behavior

For providers whose context is not known to LiteLLM (Ollama, llama_cpp), the chunk size should come from a configurable context window value, so chunks fit the context the model actually loads.

### Actual Behavior

Every model lookup returns None, cognee falls back to 16384, and get_max_chunk_tokens() returns 8191. That is larger than Ollama's default context of 4096, so chunks overflow the context during cognify.

### Environment

- OS: Windows 11 (10.0.26100)
- Python: 3.14.0
- Cognee version: 1.2.1-local
- LLM Provider: Ollama (llama3.1:8b)
- Embedding Provider: Ollama (nomic-embed-text:latest)
- Vector / Graph DB: LanceDB + Ladybug

### Logs/Error Messages

```shell
llama3.1:8b -> None
llama3.2:3b -> None
qwen2.5:7b -> None
mistral:7b -> None
chunk size cognee uses: 8191
```

### Additional Context

Suggested fix: add a configurable ollama_num_ctx value (the same pattern already used by llama_cpp_n_ctx), use it to size the chunk cutoff, and pass it as num_ctx on the Ollama requests. This keeps the chunk size and the loaded context in agreement.

Image

### Pre-submission Checklist

kadamprashant1:
***

Fixes #3436 

### What this PR does
Resolves an issue where Cognee generates chunks larger than the Ollama context window because LiteLLM's `model_cost` lookup returns `None` for Ollama models. 

Instead of falling back to a hardcoded 16384 token limit, the chunk calculation now safely falls back to a provider-aware context value.

### Key Changes
- **`utils.py`**: Updated `get_model_max_completion_tokens()` to use provider-specific context values when LiteLLM lookup fails.
- **`config.py`**: Introduced `ollama_num_ctx` (default `4096`) to `LLMConfig`, mirroring the existing `llama_cpp_n_ctx` pattern.
- **`.env.template`**: Added `OLLAMA_NUM_CTX` documentation so users can easily override the default context window if necessary.

*(Note: Users who wish to modify the context allocated at runtime inside Ollama can continue to do so securely using the pre-existing `LLM_ARGS` mechanism.)*

***

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -525,6 +525,8 @@ WEB_SCRAPER_MAX_DELAY=10.0
 #LLM_MODEL="llama3.1:8b"
 #LLM_PROVIDER="ollama"
 #LLM_ENDPOINT="http://localhost:11434/v1"
+# Context window for Ollama models (Ollama default is 4096)
+#OLLAMA_NUM_CTX=4096
 #EMBEDDING_PROVIDER="ollama"
 #EMBEDDING_MODEL="nomic-embed-text:latest"
 #EMBEDDING_ENDPOINT="http://localhost:11434/api/embed"
--- a/cognee/infrastructure/llm/config.py
+++ b/cognee/infrastructure/llm/config.py
@@ -74,6 +74,7 @@ class LLMConfig(BaseSettings):
     llama_cpp_n_ctx: int = 2048
     llama_cpp_n_gpu_layers: int = 0
     llama_cpp_chat_format: str = "chatml"
+    ollama_num_ctx: int = 4096
 
     fallback_api_key: str = ""
     fallback_endpoint: str = ""
--- a/cognee/infrastructure/llm/utils.py
+++ b/cognee/infrastructure/llm/utils.py
@@ -50,7 +50,7 @@ def get_model_max_completion_tokens(model_name: str) -> int | None:
 
     Checks if the provided model name is present in the predefined model cost dictionary. If
     found, it logs the maximum token count for that model and returns it. If the model name
-    is not recognized, it logs an informational message and returns None.
+    is not recognized, it falls back to provider-specific context sizes.
 
     Parameters:
     -----------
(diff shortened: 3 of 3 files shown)


dexters1:
Thanks for the PR @kadamprashant1! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

kadamprashant1:
Thanx, @dexters1, I will make changes accordingly.

kadamprashant1:
Added changes in New PR #5170

PR #3528, opened 2026-06-27
fix(migration): declare e_relationship_name key in GraphML export

### LINKED ISSUE #3527: [Bug]: GraphML export emits undeclared e_relationship_name key — output unreadable by networkx.read_graphml
### Bug Description

`write_graphml` in `cognee/modules/migration/formats.py` writes a `<data key="e_relationship_name">` element for every edge, but never emits a matching `<key id="e_relationship_name" for="edge" .../>` declaration.

The edge `<key>` declarations are collected only from each edge's *properties* dict, while `relationship_name` is a separate element of the edge tuple `(source, target, relationship_name, properties)` — so it is never registered.

GraphML requires every `<data key="X">` to reference a declared `<key id="X">`. The emitted file is well-formed XML, so the existing `test_graphml_is_valid_xml` (which only parses with `xml.etree.ElementTree`) passes — but strict GraphML readers reject it. The module docstring lists "Gephi/yEd/NetworkX interop" as the purpose, so essentially every export containing at least one edge is unreadable.

### Steps to Reproduce

```python
import tempfile
from pathlib import Path

import networkx as nx
from cognee.modules.migration.formats import write_graphml

nodes = [("id-1", {"name": "Alice"}), ("id-2", {"name": "Berlin"})]
edges = [("id-1", "id-2", "lives_in", {"edge_text": "Alice lives in Berlin"})]

dest = Path(tempfile.mktemp(suffix=".graphml"))
write_graphml(nodes, edges, dest)
nx.read_graphml(dest)
```

### Actual Behavior

```
networkx.exception.NetworkXError: Bad GraphML data: no key e_relationship_name
```

No `<key id="e_relationship_name" ...>` is declared, although every edge emits `<data key="e_relationship_name">`.

### Expected Behavior

The export is valid GraphML, readable by `networkx.read_graphml` (and other strict GraphML consumers), with `relationship_name` preserved on each edge.

### Environment

Reproducible on the current `dev` tip. `networkx` is a core dependency (`networkx>=3.4.2,<4`). No infrastructure or API keys required.

Vrund4591:
## Problem

`write_graphml` (`cognee/modules/migration/formats.py`) emits a `<data key="e_relationship_name">` element on every edge, but never declares that key. The edge `<key>` declarations are built only from each edge's *properties* dict, whereas `relationship_name` comes from the edge tuple `(source, target, relationship_name, properties)` — so it is never registered.

GraphML requires every `<data key="X">` to have a matching `<key id="X">` declaration. The output is well-formed XML (so the existing `test_graphml_is_valid_xml`, which only checks XML validity via `ElementTree`, passes), but strict GraphML readers reject it:

```
networkx.exception.NetworkXError: Bad GraphML data: no key e_relationship_name
```

Since virtually every edge carries a relationship name, essentially every export containing edges is unreadable by NetworkX — one of the interop targets named in the module docstring ("GraphML for Gephi/yEd/NetworkX interop").

## Fix

Declare the `relationship_name` edge key whenever edges are present, so the `<data key="e_relationship_name">` emitted per edge has a matching `<key>` declaration.

## Tests

Added two regression tests to `TestFormatEmitters` (both fail before this change, pass after):
- `test_graphml_data_keys_are_all_declared` — every `<data key>` references a declared `<key id>`.
- `test_graphml_round_trips_through_networkx` — `networkx.read_graphml` reads the export and preserves `relationship_name`.

Fixes #3527

### DIFF
--- a/cognee/modules/migration/formats.py
+++ b/cognee/modules/migration/formats.py
@@ -49,6 +49,13 @@ def write_graphml(nodes: List[Node], edges: List[Edge], destination: Path) -> No
     for _, properties in nodes:
         for key in properties or {}:
             node_keys.setdefault(key)
+    # Every edge emits a "relationship_name" datum below. It comes from the edge
+    # tuple rather than the per-edge properties dict, so the loop below would never
+    # register it; declare it explicitly whenever edges are present, otherwise the
+    # emitted <data key="e_relationship_name"> has no matching <key> declaration and
+    # strict GraphML readers (e.g. networkx.read_graphml) reject the file.
+    if edges:
+        edge_keys.setdefault("relationship_name")
     for _, _, _, properties in edges:
         for key in properties or {}:
             if key not in _SKIP_EDGE_KEYS:
--- a/cognee/tests/unit/migration/test_migration.py
+++ b/cognee/tests/unit/migration/test_migration.py
@@ -478,6 +478,32 @@ class TestFormatEmitters:
         assert len(graph.findall(f"{namespace}node")) == 2
         assert len(graph.findall(f"{namespace}edge")) == 1
 
+    def test_graphml_data_keys_are_all_declared(self, tmp_path):
+        # Every <data key="..."> must reference a declared <key id="...">. The edge
+        # relationship_name datum was emitted without a matching <key> declaration,
+        # which strict GraphML readers reject even though the XML is well-formed.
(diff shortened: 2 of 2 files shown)


dexters1:
Thanks for the PR @Vrund4591! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #3535, opened 2026-06-27
fix: reject relative paths when local file ingestion is disabled

Chandan11232:
## Description
While reviewing `save_data_item_to_storage`, I noticed an inconsistency in how the `accept_local_file_path` setting handles path branches. While absolute paths and `file://` URIs correctly raise an `IngestionError` when local file ingestion is disabled, the relative-path branch lacked an `else` block entirely. 

Consequently, when `ACCEPT_LOCAL_FILE_PATH=False`, providing an existing relative path (e.g., `data/secret.txt`) would silently fall through to the text-storage handler, storing the literal path string as document content instead of blocking it.

This PR fixes the issue by adding the missing symmetric `else: raise IngestionError(...)` block to ensure all local path types are handled consistently and securely.

## Acceptance Criteria
- With `ACCEPT_LOCAL_FILE_PATH=False`, passing an existing relative path now raises an `IngestionError`.
- With `ACCEPT_LOCAL_FILE_PATH=True`, relative paths continue to successfully resolve to `file://` URIs (no behavior change).

## Type of Change
- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/cognee/api/v1/serve/serve.py
+++ b/cognee/api/v1/serve/serve.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import asyncio
 import os
 import time
 from typing import TYPE_CHECKING, Optional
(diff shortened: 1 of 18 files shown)


dexters1:
@Chandan11232 PR has to many unrelated changes to PR description

Issue #3585, opened 2026-06-27
[Bug]: LadybugAdapter.get_neighborhood(edge_types=...) crashes with a Kuzu internal assertion (UNREACHABLE_CODE) on variable-length path filtering


mohitjeswani01:
### Bug Description

`get_neighborhood(node_ids=..., depth=..., edge_types=[...])` crashes on the Ladybug/Kuzu backend whenever `edge_types` is supplied. Unfiltered neighborhood queries (`edge_types=None`) work correctly — the crash happens only when an edge-type allow-list is passed.

Edge-type filtering works on the Neo4j and Postgres graph backends; only Ladybug/Kuzu is affected. Since Ladybug is the default backend, edge-filtered neighborhood queries are unusable out of the box.

This surfaced while implementing `SearchType.NEIGHBORHOOD` (#3376), which passes `edge_types` straight through to this primitive.

### Steps to Reproduce

On a Ladybug (default) graph, run:

```python
import asyncio, os, tempfile
from cognee.infrastructure.databases.graph.ladybug.adapter import LadybugAdapter


class FakeNode:
    def __init__(self, id, name="", type=""):
        self._d = {"id": str(id), "name": name, "type": type}

    def model_dump(self):
        return dict(self._d)


async def main():
    adapter = LadybugAdapter(db_path=os.path.join(tempfile.mkdtemp(), "graph"))
    await adapter.add_nodes([FakeNode("A"), FakeNode("B"), FakeNode("D")])
    await adapter.add_edges([("A", "B", "KNOWS", {}), ("A", "D", "WORKS_AT", {})])

    # Unfiltered works:
    nodes, _ = await adapter.get_neighborhood(node_ids=["A"], depth=2)
    print(sorted(n for n, _ in nodes))  # -> ['A', 'B', 'D']

    # Filtered crashes:
    await adapter.get_neighborhood(node_ids=["A"], depth=2, edge_types=["KNOWS"])

    await adapter.close()


asyncio.run(main())
```

### Expected Behavior

The filtered call returns a `KNOWS`-only neighborhood: `{A, B}`. `D`, which is reachable only via a `WORKS_AT` edge, is excluded. This is the behavior on Neo4j and Postgres.

### Actual Behavior

The filtered call raises:
RuntimeError: Assertion failed in file ".../parsed_parameter_expression.h" on line 21: UNREACHABLE_CODE

[...]

PR #3586, opened 2026-06-27
feat(search): add SearchType.NEIGHBORHOOD (structural N-hop neighborhood search) (#3376)

### LINKED ISSUE #3376: [hackathon] Search type: NEIGHBORHOOD / ego-graph (N-hop, edge-type filtered)
Part of #3375.

## Idea
A search type that returns the **structural N-hop neighborhood** around the entities a query mentions — no LLM/vector reranking required — with optional **edge-type filtering**. "Give me everything within 2 hops of `Acme`, following only `works_at`/`owns` edges."

## Why
`GRAPH_COMPLETION` defaults to a flat (depth-0) projection; N-hop expansion is only a hidden `neighborhood_depth` tuning knob with no first-class search type. This is the single highest-leverage addition because the primitive **already exists**.

## Implementation (lowest effort)
- New `SearchType.NEIGHBORHOOD`.
- New retriever: resolve seed nodes from query text (reuse the vector seed-resolution in `brute_force_triplet_search`), then call `get_neighborhood(node_ids, depth, edge_types)` (already abstract on every backend), return the subgraph (and an optional LLM summary variant).
- Params: `depth` (default 1-2), `edge_types` (optional allow-list), `seed_top_k`. Plumb via `retriever_specific_config` like existing retrievers.
- Register in `search_core_registry`.

## Acceptance
- `recall(query, query_type=SearchType.NEIGHBORHOOD, retriever_specific_config={"depth":2,"edge_types":["works_at"]})` returns the depth-2, edge-filtered subgraph around the matched entity.
- Works on Ladybug + Neo4j + Postgres graph (uses only `get_neighborhood`).

Files: `SearchType.py`, `modules/retrieval/`, `get_search_type_retriever_instance.py`, `graph_db_interface.get_neighborhood`.

mohitjeswani01:
## Description

Closes #3376.

This adds `SearchType.NEIGHBORHOOD`, a search type that returns the structural N-hop neighborhood around the entities a query mentions, with no LLM step involved.

My reasoning for the approach: while reading the search code I noticed the `get_neighborhood` primitive already exists and is abstract across all backends, but no search type actually exposes it — `GRAPH_COMPLETION` only reaches it through a hidden `neighborhood_depth` knob. So the work here is mostly a thin, honest entry point to a primitive that's already there, rather than new graph logic. I deliberately call `get_neighborhood` directly instead of going through `brute_force_triplet_search`, because that path re-scores the expansion nodes with vectors, and I wanted this retriever to be genuinely structural — the vector engine is touched exactly once, only to resolve the seed nodes from the query. After that it's pure graph traversal.

A few decisions worth explaining:

- **Truncation:** the primitive returns a flat node/edge list with no hop distance, so a hub entity could dump a huge subgraph. I added a `max_nodes` cap and, since I had to derive distance myself, did a small multi-source BFS from the seeds to keep the *closest* hops first when truncating. Seeds are never dropped even if they exceed the cap.
- **Multi-seed:** I checked the primitive and it already unions and dedupes the per-seed ego-graphs, so I didn't reimplement that — I just pass the seeds in and rely on it, and added a test that proves the union/dedupe behaviour.
- **Determinism:** the vector seed-resolution returns ids from a set (non-deterministic order), so I sort candidates by `(score, id)` before slicing to `seed_top_k`, so the same query always picks the same seeds.

**Direction is undirected.** While building this I found that `get_neighborhood` traverses undirected on every backend with no direction parameter, so `edge_types` is a symmetric *type* filter, not a direction control. Directional traversal would mean changing every adapter, so I left it out of scope on purpose.

**One real limitation I hit and want to be upfront about:** passing `edge_types` currently crashes on the default Ladybug/Kuzu backend — it raises a Kuzu internal assertion (`UNREACHABLE_CODE`) on the variable-length path predicate. Unfiltered neighborhood queries work fine on every backend, and edge-type filtering works on Neo4j and Postgres. This is an upstream `get_neighborhood` bug, not something in this retriever (the retriever just forwards `edge_types`, which the unit test confirms). I filed it separately as #3585, and the integration test strict-`xfail`s that one case on Ladybug so a fix in #3585 will flip it red.

The LLM-summarised variant mentioned in the issue is intentionally left as a follow-up — this PR is structural-only.

## Acceptance Criteria

From #3376:

- `recall(query, query_type=SearchType.NEIGHBORHOOD, retriever_specific_config={"depth":2,"edge_types":[...]})` returns the depth-N, edge-filtered subgraph around the matched entities. **Met on Neo4j/Postgres.** On the default Ladybug/Kuzu backend the unfiltered case works; the `edge_types` case is blocked by the upstream defect #3585 (documented above and strict-`xfail`'d).
- Works on Ladybug + Neo4j + Postgres via `get_neighborhood`. **Met** — structural/unfiltered queries verified on Ladybug (default) in the integration test; Neo4j/Postgres exercise the same path under their extras.

Proof it works:

- **Unit tests** (9, mocked, no DB/LLM): deterministic seed ordering + `seed_top_k` slicing, closest-hops-first `max_nodes` truncation, seeds-never-dropped, `edge_types` pass-through, structured no-LLM return with the vector engine used exactly once, and empty-graph / no-seed short-circuits.
- **Integration test** (real Ladybug graph): depth-1 multi-edge-type expansion, multi-seed union/dedupe, and `max_nodes` truncation on real data; the edge-filter case is strict-`xfail`'d per #3585.

Files: `cognee/modules/retrieval/neighborhood_retriever.py` (new), `SearchType.py`, `get_search_type_retriever_instance.py`, `normalize_search_payload.py`, plus unit + integration tests and a `skill.md` doc entry.

## Type of Change

## Screenshots

**Unit tests — 9 passed:**
image

**Integration tests — 3 passed, 1 xfailed (edge-filter, per #3585):**
image

## Pre-submission Checklist

## DCO Affirmation


dexters1:
Thanks for the PR @mohitjeswani01! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #3588, opened 2026-06-27
fix(llm): derive basename from Windows paths in Mistral transcription

### LINKED ISSUE #3587: [Bug]: On Windows, Mistral transcription sends the full file path as file_name (splits on '/' only)
# [Bug]: On Windows, Mistral transcription sends the full file path as `file_name` (splits on `/` only)

## Bug Description

`MistralAdapter.create_transcript` derives the API `file_name` like this:

```python
# cognee/infrastructure/llm/.../mistral/adapter.py
file_name = input.split("/")[-1]
...
self.mistral_client.audio.transcriptions.complete(
    model=transcription_model,
    file={"content": f, "file_name": file_name},
)
```

`input` is the audio file path. On Windows that's a backslash path
(e.g. `C:\Users\me\audio\clip.mp3`) with **no forward slashes**, so
`input.split("/")[-1]` returns the **entire path** instead of `clip.mp3`. That
full path is then sent to the Mistral API as `file_name`.

This is the same separator bug already fixed for `classify()` and handled
correctly by `_normalize_filename` in `cognee/tasks/ingestion/utils.py`.

## Steps to Reproduce

On Windows, with the Mistral provider configured for audio transcription, call
the transcription path with a local Windows audio path. The `file_name` sent to
the API is the full `C:\...` path rather than the basename.

Minimal illustration of the mechanism:

```python
input = r"C:\Users\me\audio\clip.mp3"
print(input.split("/")[-1])  # -> 'C:\\Users\\me\\audio\\clip.mp3'  (whole path)
```

## Expected Behavior

`file_name` is the basename (`clip.mp3`) on all platforms.

## Actual Behavior

On Windows, `file_name` is the full backslash path.

## Proposed Fix

Normalize both separators before taking the basename (matching `classify()`):

```python
file_name = str(input).replace("\\", "/").split("/")[-1]
```

I have the fix and cross-platform regression tests ready — happy to open a PR.

## Environment

- OS: Windows 11
- Python version: 3.13.7
- Cognee version: dev
- LLM Provider: Mistral (audio transcription)

YashasviThakur:
# fix(llm): derive basename from Windows paths in Mistral transcription

## What

`MistralAdapter.create_transcript` derived the API `file_name` with
`input.split("/")[-1]`. On Windows the audio path is a backslash path
(e.g. `C:\audio\clip.mp3`) with no forward slashes, so the split sent the
**entire path** as `file_name` to the Mistral API instead of the basename.

This normalizes both `\` and `/` separators before taking the basename, matching
the fix already used in `classify()` and `_normalize_filename`
(`cognee/tasks/ingestion/utils.py`).

```diff
- file_name = input.split("/")[-1]
+ file_name = str(input).replace("\\", "/").split("/")[-1]
```

Fixes #3587.

## Testing

Added `cognee/tests/unit/infrastructure/llm/test_mistral_transcript_basename.py`:

- Windows-style input (`C:\Users\me\audio\clip.mp3`) → `file_name` == `clip.mp3`
- POSIX-style input (`/home/me/audio/clip.mp3`) → `file_name` == `clip.mp3`

The test builds the adapter via `__new__` (no API keys needed) and mocks
`open_data_file` + the Mistral client, asserting the `file_name` passed to the
API. Because the Windows-style input contains backslashes regardless of host, a
revert to splitting on `/` only fails the test on any OS (Linux CI included).


cognee/infrastructure/llm/structured_output_framework/litellm_instructor/llm/mistral/adapter.py
```diff
@@ -180,7 +180,10 @@ async def create_transcript(self, input: str, **kwargs: Any) -> TranscriptionRet
         transcription_model = self.transcription_model
         if self.transcription_model.startswith("mistral"):
             transcription_model = self.transcription_model.split("/")[-1]
-        file_name = input.split("/")[-1]
+        # Normalize Windows ("\") and POSIX ("/") separators before taking the
+        # basename; on Windows `input` is a backslash path, so splitting on "/"
+        # alone would send the full path as the file name to the API.
+        file_name = str(input).replace("\\", "/").split("/")[-1]
```
dexters1, line 186: Maybe it would be better to handle path based operations like this for example:

from pathlib import Path

file_name = Path(path).name

as the Path library is intended to handle paths for all OS types

YashasviThakur: Good catch on preferring `pathlib` — pushed a4e32d8e with `PureWindowsPath(input).name` instead of the manual replace.

One nuance worth flagging: a plain `pathlib.Path(input).name` wouldn't quite work here. On POSIX hosts, `Path` resolves to `PurePosixPath`, which doesn't treat "\\" as a separator — so for a Windows-style input like `C:\Users\me\audio\clip.mp3`, `Path(input).name` on Linux/macOS CI would return the whole string unchanged, not `clip.mp3`. That's exactly the bug this PR is fixing, just reintroduced.

`PureWindowsPath` always treats both `/` and `\` as separators regardless of host OS, so it gives the correct basename on any platform — which is what the existing tests assert (they run the same Windows-style-path assertion on whatever OS CI happens to be).

PR #3597, opened 2026-06-28
fix(api): support singular dataset fields and enforce dataset isolation in search/recall

### LINKED ISSUE #3520: [Bug]: /api/v1/recall ignores dataset_name and returns cross-tenant data on Cognee Cloud (managed)
## Summary

On the managed Cognee Cloud endpoint, `POST /api/v1/recall` ignores the `dataset_name` (and presumably `dataset_id`) filter and returns one result per dataset on the tenant. The auth principal sees `GRAPH_COMPLETION` syntheses from datasets it never ingested into. When those datasets have no real content, the LLM hallucinates plausible-sounding completions (e.g. "eldaa is a Star Wars Jedi", "eldaa is a data-analysis library", "eldaa is an EHR platform").

## Environment

- **Endpoint:** managed Cognee Cloud (`https://tenant-8eb9b5c9-dee8-4ffd-a0e4-c4385e7be39d.aws.cognee.ai`)
- **Auth:** `X-Api-Key` header
- **Date observed:** 2026-06-27
- **Cognee version:** unknown (managed tenant, no version banner)
- **Datasets on the tenant:** 4 (`default_dataset`, `cloudflare_workers`, `eldaa`, `remember_club_crm_development`) — only `eldaa` has content we ingested

## Steps to reproduce

```bash
curl -X POST "$COGNEE_BASE_URL/api/v1/recall" \
  -H "X-Api-Key: $COGNEE_API_KEY" \
  -H "Content-Type: application/json" \
  -d '{"query":"What is eldaa?","dataset_name":"eldaa","session_id":"any"}'
```

## Expected

Response contains only results for the `eldaa` dataset (i.e. `dataset_name == "eldaa"` for every entry).

## Actual

Response is an array of 4 entries — one per dataset on the tenant. The 3 non-target entries contain hallucinated graph completions. The hallucinated content varies between calls (different `session_id` → different completions), so the source is real per-dataset retrieval against vector stores / graph nodes the principal should not have access to.

The same call with `search_type: "CHUNKS"` or `"SUMMARIES"` likely has the same problem given the same underlying code path; I observed it on the default `GRAPH_COMPLETION`.

## Related upstream issue

[#2867](https://github.com/topoteretes/cognee/issues/2867) — same symptom on **self-hosted** when `ENABLE_BACKEND_ACCESS_CONTROL=false`. Closed without a merged fix; the root-cause analysis in that issue points to `cognee/modules/search/methods/search.py:291-339` and the retrievers (`chunks_retriever.py`, `summaries_retriever.py`, `graph_completion_retriever.py` → `brute_force_triplet_search.py`) never receiving a dataset arg.

## What is different about this report

The Cognee docs ([Multi-User Mode overview](https://docs.cognee.ai/core-concepts/multi-user-mode/multi-user-mode-overview)) say multi-user isolation (`ENABLE_BACKEND_ACCESS_CONTROL`, default ON ≥ v0.5.0) auto-isolates data per user — recall is supposed to be scoped to datasets the auth principal can read. **On the managed Cloud endpoint, this isolation is not happening** even though the docs and the default-on env var should make it the default behaviour. So either:

1. The Cloud-managed storage backend doesn't support the per-dataset routing the multi-user mode requires, and the platform silently degrades to no-isolation (same observable behaviour as the self-hosted AC-disabled path).
2. The proxy in front of the Cloud API is bypassing the permission check.

Either way, the **public API surface returns cross-tenant data**, which is a security/privacy concern — users reasonably trust the per-user auth key to scope what they see.

## Suggested fix areas

- Confirm the Cloud platform either enforces multi-user isolation or rejects the call with a clear error when storage doesn't support it.
- If proxy-bypass is the cause, audit the proxy layer.
- In parallel, plumb `dataset_ids` through the retrievers as #2867 reporter proposed (Phase 1 refactor: accept and ignore; Phase 2: payload-filter on the vector search). That fix benefits self-hosted AC-disabled users too.

## Workaround

Client-side filter on every recall call:

1. Parse the JSON array response.
2. Keep only entries where `dataset_name == <requested>`.
3. If the filtered list is empty, treat the recall as "no information" — do not trust any non-target-dataset result, even if it sounds relevant.
4. Cross-check synthesised answers against the actual codebase / docs before acting on them.

BhaveshBhardwaj:
## Description
In environments where storage collections are shared or physical isolation is disabled, performing search/recall queries could retrieve context chunks from other datasets. Furthermore, if a dataset was empty or irrelevant context was retrieved, the completion engine still called the LLM with empty text, prompting it to hallucinate plausible-sounding completions. Lastly, singular `dataset_name` and `dataset_id` request payloads were ignored by the FastAPI DTO schemas.

To resolve these issues, the following changes were made:
1. **API Schema Payload Validator:** Added `dataset_name` and `dataset_id` to `RecallPayloadDTO` and `SearchPayloadDTO` along with a pre-validator (`map_singular_dataset_fields`) to map them into their list equivalents for backwards compatibility.
2. **Relational Database Fetching:** Created a new helper `get_dataset_node_ids` to query node UUID slugs associated with dataset IDs inside the relational `Node` model mapping table.
3. **Application-level Filtering:** Plumbed allowed node IDs into `get_retriever_output` to discard chunks, summaries, entities, and facts not belonging to the authorized datasets before generating context.
4. **Hallucination Prevention:** Added an early intercept inside `generate_completion` to return a standard fallback response when context is empty/whitespace, rather than invoking the LLM.

## Acceptance Criteria
* Singular payload fields map correctly to datasets list.
* Retriever filters out results not associated with target datasets at the application level.
* LLM returns context fallback message instead of hallucinating on empty context.
* Unit tests verify singular fields parsing, output filtering, and completion fallback.

## Type of Change
- [ ] Code refactoring
- [ ] Other (please specify):

Fixes #3520

### DIFF
--- a/cognee/api/v1/recall/routers/get_recall_router.py
+++ b/cognee/api/v1/recall/routers/get_recall_router.py
@@ -1,11 +1,11 @@
 from datetime import datetime
-from typing import List, Optional, Union
+from typing import Any, List, Optional, Union
 from uuid import UUID
 
 from fastapi import APIRouter, Depends
 from fastapi.encoders import jsonable_encoder
 from fastapi.responses import JSONResponse
-from pydantic import Field
+from pydantic import Field, model_validator
 
 from cognee import __version__ as cognee_version
 from cognee.api.DTO import InDTO, OutDTO
@@ -49,6 +49,36 @@ class RecallPayloadDTO(InDTO):
             "when both are provided. Leave empty to resolve by name."
         ),
     )
+    dataset_name: Optional[str] = Field(
+        default=None,
+        description="Optional singular dataset name to search. Added to datasets list.",
+    )
+    dataset_id: Optional[UUID] = Field(
+        default=None,
+        description="Optional singular dataset UUID to search. Added to dataset_ids list.",
+    )
+
+    @model_validator(mode="before")
+    @classmethod
+    def map_singular_dataset_fields(cls, data: Any) -> Any:
+        if isinstance(data, dict):
+            dataset_name = data.get("dataset_name")
+            if dataset_name is not None:
+                datasets = data.get("datasets")
+                if datasets is None:
+                    data["datasets"] = [dataset_name]
+                elif isinstance(datasets, list) and dataset_name not in datasets:
+                    data["datasets"].append(dataset_name)
+
+            dataset_id = data.get("dataset_id")
+            if dataset_id is not None:
+                dataset_ids = data.get("dataset_ids")
+                if dataset_ids is None:
+                    data["dataset_ids"] = [dataset_id]
+                elif isinstance(dataset_ids, list) and dataset_id not in dataset_ids:
+                    data["dataset_ids"].append(dataset_id)
+        return data
+
     query: str = Field(default="What is in the document?")
     system_prompt: Optional[str] = Field(
         default="Answer the question using the provided context. Be as brief as possible."
--- a/cognee/api/v1/search/routers/get_search_router.py
+++ b/cognee/api/v1/search/routers/get_search_router.py
@@ -5,7 +5,7 @@ from uuid import UUID
 from fastapi import APIRouter, Depends, status
 from fastapi.encoders import jsonable_encoder
 from fastapi.responses import JSONResponse
-from pydantic import Field
+from pydantic import Field, model_validator
 
 from cognee import __version__ as cognee_version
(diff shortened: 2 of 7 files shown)


dexters1:
Thank you for your contribution! But currently we do not want to support single parameter support for dataset_name/dataset_id for the endpoints

PR #3639, opened 2026-06-28
feat: add Turso (libSQL) as a vector database option

### LINKED ISSUE #3583: [hackathon] Add Turso as a vector database option
## Task

Add Turso (libSQL) as a vector database backend, selectable with `VECTOR_DB_PROVIDER=turso`. Implement a new `TursoVectorAdapter(VectorDBInterface)` that stores embeddings using libSQL's native vector type and runs similarity search with libSQL's built-in methods— directly mirroring how the PGVector adapter was added.

## Design

Implement every `VectorDBInterface` abstract method (`cognee/infrastructure/databases/vector/vector_db_interface.py`):
   `has_collection`, `create_collection`, `create_data_points`, `retrieve`,
   `search`, `batch_search`, `delete_data_points`, `prune`, `embed_data`.
Override the useful optional ones too: `create_vector_index`,
   `index_data_points`, `remove_belongs_to_set_tags`, `run_migrations`, `close`.

Make sure it adheres to other systems in Cognee:
1) It will be a part of the create_vector_engine code in Cognee and usable easily
2) A dataset database handler (read about Cognee dataset database handlers from our docs) will be written for it and properly working with Cognee in a multi-user mode with which the permission system already built in Cognee will fully work
3) The cache when getting a vector engine will work (No work here is needed unless new configration options are added)
4) Make sure it can be used by just setting the environment variables (through the VectorConfig class)

## Testing

Make sure you run as much tests as possible from our e2e tests with Turso as the vector database and attach e2e test results that you have ran along with the solution.

therealharshit:
## Description

Fixes #3583.

This adds Turso (libSQL) as a vector database option. Set `VECTOR_DB_PROVIDER=turso` and cognee
stores its embeddings in libSQL. I followed the PGVector adapter, so I kept the new
`TursoVectorAdapter` as close to it as I could: each collection is a libSQL table with an `F32_BLOB`
vector column, search uses `vector_distance_cos`, and the NodeSet (`belongs_to_set`) filtering that
PGVector does with Postgres JSONB I rewrote using SQLite's JSON1 functions.

The part I want to flag is the async handling. cognee's vector interface is async, so my first
instinct was the async `libsql-client`. When I actually tried it, I hit a wall: in embedded
(local-file) mode it falls back to plain SQLite and doesn't have libSQL's vector functions at all.
The only client that gives you vectors on a local file is `libsql-experimental`, which is
synchronous. So I went with that and run the DB calls through `asyncio.to_thread` so they don't block
the event loop. I deliberately kept all the sync-client code in one place (`_get_connection` /
`_run`), so if we move to a native-async client later, that's the only spot that changes and none of
the SQL moves. 
Happy to align this with whatever async approach you'd prefer for the Turso adapter(or propose mine).

## What it does (the four points from the issue)

- **Plugs into `create_vector_engine`** — `VECTOR_DB_PROVIDER=turso` routes to the adapter, with a
  clear `ImportError` if the `cognee[turso]` extra isn't installed.
- **Multi-user** — added a dataset database handler (`TursoVectorDatasetDatabaseHandler`) and
  registered it, so under access control each dataset gets its own libSQL file and the permission
  system keeps them isolated.
- **Cache** — no new config, so the existing vector-engine cache works as-is.
- **Env vars only** — just set `VECTOR_DB_PROVIDER=turso` (plus `VECTOR_DB_URL` / `VECTOR_DB_KEY` for
  cloud) and you're done.

## Testing

I wrote 12 unit tests that hit a real embedded libSQL DB (no LLM needed) covering the whole adapter:
create/search/delete, cosine ordering, NodeSet OR and AND, the belongs_to_set merge, tag removal, and
prune. I also checked the dataset handler end to end (create a dataset, confirm another dataset can't
see its data, delete it) and ran the adapter against a real Turso cloud DB to make sure the remote
path works over the network.

For the bigger end-to-end tests (`test_permissions.py` and `test_search_db.py`), I wired them to run
against Turso in CI. I couldn't get a clean local run of those because I only have a free-tier LLM
key, and cognee fires a lot of `cognify` calls in parallel which blows past the rate limit. Could you
approve the CI workflows so those run with the repo's LLM secrets? I'll fix anything they turn up.

## Type of Change

## Screenshots

Screenshot 2026-06-29 at 3 32 15 AM

## Pre-submission Checklist

- [x] I've tested my changes (12 adapter unit tests + the dataset handler + a live Turso cloud run;
  the heavier e2e runs in CI)
- [x] This PR is scoped to the feature, nothing extra
- [x] Code follows the style guide (`ruff` is clean)
- [x] Added tests that show the feature works
- [x] Added docs (the Turso block in `.env.template`)
- [x] All tests pass — unit tests pass locally; the full e2e is validated by the new CI jobs
- [x] Searched existing PRs
- [x] Linked the issue
- [x] Commits are clear (Conventional Commits, DCO signed off)

## DCO Affirmation

I affirm that all code in every commit of this pull request conforms to the terms of the Topoteretes
Developer Certificate of Origin.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -153,13 +153,22 @@ GRAPH_DATASET_DATABASE_HANDLER="kuzu"
 #  Provider-specific connection details.
 ################################################################################
 
-# Supported (built-in): pgvector | lancedb
+# Supported (built-in): pgvector | lancedb | turso
 # Community adapters (separate packages): qdrant | weaviate | milvus | chromadb
 VECTOR_DB_PROVIDER="lancedb"
 #VECTOR_DB_URL=
 #VECTOR_DB_KEY=
 # Handler for multi-user access control (per-dataset DB creation).
 VECTOR_DATASET_DATABASE_HANDLER="lancedb"
+
+# Turso / libSQL (requires the turso extra: pip install cognee"[turso]")
+# Embedded (local file):
+#VECTOR_DB_PROVIDER="turso"
+#VECTOR_DB_URL="/absolute/path/to/cognee.turso.db"
+# Remote Turso cloud:
+#VECTOR_DB_PROVIDER="turso"
+#VECTOR_DB_URL="libsql://your-db.turso.io"
+#VECTOR_DB_KEY="your_turso_auth_token"
 
 # Connection pool tuning for PGVector per-dataset engines (JSON).
 # When ENABLE_BACKEND_ACCESS_CONTROL=true each dataset gets its own engine; this controls
--- a/.github/workflows/vector_db_tests.yml
+++ b/.github/workflows/vector_db_tests.yml
@@ -189,7 +189,7 @@ jobs:
           EMBEDDING_DIMENSIONS: 300
           EMBEDDING_MODEL: ${{ secrets.EMBEDDING_MODEL }}
           EMBEDDING_API_KEY: ${{ secrets.EMBEDDING_API_KEY }}
-        run: uv run python ./cognee/tests/test_search_db.py
+        run: uv run pytest cognee/tests/test_search_db.py -v --log-level=INFO
 
   run-turso-multi-user-tests:
     name: Turso Multi-User Tests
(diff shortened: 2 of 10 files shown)


therealharshit:
Hey @dexters1, 
I've tested the adapter locally (12 unit tests on a real libSQL DB, the dataset handler, and a live Turso cloud run, all passing). For the `test_permissions.py` and `test_search_db.py` e2e tests, I've wired them to run against Turso in CI but can't run them locally (Rate limit on my API key). "

Could you approve the CI workflows so they run with the repo's secrets? Happy to fix anything they surface.

rshkarin:
@therealharshit thanks for your contribution! Your changes have been merged.

PR #3655, opened 2026-06-29
refactor(config): unify PostgreSQL credential fallback 

### LINKED ISSUE #3384: [hackathon] Unify database configuration (one Postgres credential block, consistent naming)
Part of #3381.

## Problem: the Postgres "everything" combo needs the same credentials three times
There is **no shared DB config**. Three independent field families, three connection-string builders:
- **Relational:** `DB_HOST`, `DB_PORT`, `DB_USERNAME`, `DB_PASSWORD`, `DB_NAME`, `DB_PROVIDER=postgres` → `create_relational_engine.py`.
- **pgvector:** `VECTOR_DB_HOST/PORT/USERNAME/PASSWORD/NAME`, `VECTOR_DB_PROVIDER=pgvector` → `create_vector_engine.py` (falls back to relational config only if these are fully absent, with a warning).
- **postgres-graph:** `GRAPH_DATABASE_HOST/PORT/USERNAME/PASSWORD/NAME`, `GRAPH_DATABASE_PROVIDER=postgres` → `get_graph_engine.py` (same warning-laden fallback).

So a single Postgres deployment requires `DB_*` + `VECTOR_DB_*` + `GRAPH_DATABASE_*` — same host/port/user/pass three times. (A `USE_UNIFIED_PROVIDER=pghybrid` flag is a fourth, separate way to force reuse.)

## Other confirmed inconsistencies
- **Three prefixes:** `DB_` vs `VECTOR_DB_` vs `GRAPH_DATABASE_` (full word "DATABASE" only for graph).
- **Provider strings disagree for the same engine:** relational/graph use `"postgres"`, vector uses `"pgvector"`; internal handler names add a third — `"postgres_graph"`. So Postgres has 3 names depending on layer.
- **Port type/default drift:** relational `db_port: str|None` (None), vector `vector_db_port: int` (1234), graph `graph_database_port: int` (123).
- **Auth naming drift:** `DB_PASSWORD` vs `VECTOR_DB_KEY` vs `GRAPH_DATABASE_KEY`/`GRAPH_DATABASE_PASSWORD`.
- **URL form only for some:** graph/vector accept a single `*_URL`; relational doesn't.
- `extra="allow"` on all → misspelled DB env vars are silently ignored.

## Proposed simplification
- **One shared connection block** (host/port/username/password/name/url) that relational, vector, and graph can each reference, so the all-Postgres combo is configured once. Per-store overrides still allowed but optional.
- **Reconcile provider strings:** accept `postgres` everywhere (keep `pgvector` as an alias), and stop exposing the internal `postgres_graph` handler name as user-facing.
- **Make `*_URL` available for all three** (relational included); when a URL is given, don't also require broken-out fields.
- **Consistent port type + default** across the three configs.
- Make the relational-config fallback the documented default (no warning) rather than a last resort.

## Acceptance
- Setting one Postgres credential block runs relational + pgvector + postgres-graph with no repetition and no fallback warnings.
- `VECTOR_DB_PROVIDER=postgres` works (alias of pgvector).
- A typo'd DB env var is reported, not silently ignored.

Files: `cognee/infrastructure/databases/{relational,vector,graph}/config.py`, `create_relational_engine.py`, `create_vector_engine.py`, `get_graph_engine.py`, `cognee/api/v1/config/config.py`.

armaanjain-byte:
## Description

This PR formalizes the existing PostgreSQL credential fallback so that a single `DB_*` credential block can be used across relational, pgvector and postgres-graph deployments.

### Changes

- Accept `VECTOR_DB_PROVIDER=postgres` as an alias for `pgvector`
- Add `DB_URL` support to the relational configuration and engine
- Propagate `DB_URL` through the vector and graph fallback paths
- Remove fallback warnings now that the shared `DB_*` configuration is the intended behavior
- Preserve `VECTOR_DB_*` and `GRAPH_DATABASE_*` precedence over shared `DB_*`
- Add regression tests covering the new configuration behavior
- Leave `extra="allow"` unchanged in all configuration classes, following the maintainer's guidance

## Acceptance Criteria

- [x] `VECTOR_DB_PROVIDER=postgres` behaves identically to `pgvector`
- [x] Shared `DB_*` credentials work for relational, vector and graph databases
- [x] Store-specific overrides retain precedence
- [x] `DB_URL` works without requiring broken-out credentials
- [x] Regression tests added

## Testing

```bash
uv run python -m pytest cognee/tests/unit/infrastructure/databases/test_shared_pg_fallback.py -v
```

Result:

```
13 passed
```

## Screenshots

Regression test output is attached below.

Screenshot 2026-06-29 121607
Screenshot 2026-06-29 121624
Closes #3384

### DIFF
--- a/cognee/infrastructure/databases/graph/get_graph_engine.py
+++ b/cognee/infrastructure/databases/graph/get_graph_engine.py
@@ -344,27 +344,25 @@ def _create_graph_engine(
             else:
                 from cognee.infrastructure.databases.relational import get_relational_config
 
-                logger.warning(
-                    "Postgres graph credentials are not fully configured; "
-                    "falling back to the relational database configuration. "
-                    "Set GRAPH_DATABASE_HOST/PORT/USERNAME/PASSWORD/NAME explicitly "
-                    "to avoid this fallback."
-                )
-
+                # Fall back to the relational database configuration.
                 relational_config = get_relational_config()
-                db_username = relational_config.db_username
-                db_password = relational_config.db_password
-                db_host = relational_config.db_host
-                db_port = relational_config.db_port
-                db_name = relational_config.db_name
 
-                if not (db_host and db_port and db_name and db_username and db_password):
-                    raise EnvironmentError("Missing required Postgres graph credentials!")
+                if relational_config.db_url:
+                    connection_string = relational_config.db_url
+                else:
+                    db_username = relational_config.db_username
+                    db_password = relational_config.db_password
+                    db_host = relational_config.db_host
+                    db_port = relational_config.db_port
+                    db_name = relational_config.db_name
 
-                connection_string: str = (
-                    f"postgresql+asyncpg://{db_username}:{db_password}"
-                    f"@{db_host}:{db_port}/{db_name}"
-                )
+                    if not (db_host and db_port and db_name and db_username and db_password):
+                        raise EnvironmentError("Missing required Postgres graph credentials!")
+
+                    connection_string: str = (
+                        f"postgresql+asyncpg://{db_username}:{db_password}"
+                        f"@{db_host}:{db_port}/{db_name}"
+                    )
 
         from .postgres.adapter import PostgresAdapter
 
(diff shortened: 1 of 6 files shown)


dexters1:
Thanks for the PR @armaanjain-byte! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #3674, opened 2026-06-29
fix: prevent cascading cancellations in add_data_points asyncio.gather calls

HEETMEHTA18:
## Summary

All four `asyncio.gather` calls in `add_data_points` lack `return_exceptions=True`. When any single task raises an exception, Python immediately cancels all other in-flight tasks, causing cascading data loss with no warning.

### 1. Cascading cancellation in multi-datapoint graph extraction (line 64)

**Before:** If one data point fails during `get_graph_from_model`, all data points in the batch are cancelled and lost.

**After:** Failing data points are logged and skipped. Successful extractions are preserved and stored.

### 2. Cascading cancellation in parallel graph + vector writes (lines 132, 143, 154)

**Before:** If vector indexing succeeds but graph node insertion fails (or vice versa), the successful operation is cancelled, creating inconsistent state. On retry the already-written side produces duplicates.

**After:** Both parallel operations run to completion. If either fails, the exception is re-raised after both finish, preventing partial writes.

## Impact

Any pipeline that passes multiple `DataPoint`s to `add_data_points` — which includes all high-level cognee pipelines (`cognify`, `memify`, etc.) — is vulnerable to silent data loss when even one item encounters a transient error (network blip, rate limit, malformed content).

## Fix

- All 4 `asyncio.gather` calls updated with `return_exceptions=True`
- Multi-datapoint gather: errors logged, processing continues
- Paired writes: both operations complete before re-raising

## Testing

- [x] Existing tests pass
- [ ] New test: batch with one intentionally broken data point should still process the rest


dexters1:
Thanks for the PR @HEETMEHTA18! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #3719, opened 2026-06-30
fix: thread wide_search_max_distance through all graph retriever subclasses

yubingz:
Fixes the TypeError reported by e-hosseini in PR #3006 review.

`get_search_type_retriever_instance` passes `wide_search_max_distance` to all six graph-based retrievers, but only `GraphCompletionRetriever` accepts it. The other five subclasses (decomposition, CoT, context-extension, summary, temporal) have closed `__init__` signatures with no `**kwargs`, so `retriever_cls(**retriever_args)` raises `TypeError: __init__() got an unexpected keyword argument 'wide_search_max_distance'` as soon as any of those search types are used.

## Changes

- Add `wide_search_max_distance` param (default 1.5) to `GraphCompletionRetriever` base class
- Thread through all 5 subclasses (same pattern as `wide_search_top_k`)
- Thread through `get_search_type_retriever_instance` factory (all 6 graph retrievers + AgenticRetriever)
- Thread through `brute_force_triplet_search` → `_get_top_triplet_importances` → `extract_relevant_node_ids`
- Add `max_distance` filter to `NodeEdgeVectorSearch.extract_relevant_node_ids()`
- Add parametrized test covering all 6 retriever types

## Verification

- 6/6 new test cases pass
- 21/21 existing `test_get_search_type_retriever_instance` tests pass
- 12/12 existing `test_node_edge_vector_search` tests pass (backward compatible: `None` disables filtering)

### DIFF
--- a/cognee/modules/retrieval/graph_completion_context_extension_retriever.py
+++ b/cognee/modules/retrieval/graph_completion_context_extension_retriever.py
@@ -1,6 +1,7 @@
 import asyncio
 from typing import Optional, List, Type, Union
 
+from cognee.base_config import get_base_config
 from cognee.modules.graph.cognee_graph.CogneeGraphElements import Edge
 from cognee.modules.retrieval.utils.query_state import QueryState
 from cognee.modules.retrieval.utils.validate_queries import validate_retriever_input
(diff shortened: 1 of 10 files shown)


dexters1:
Thanks for the PR @yubingz! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

yubingz:
Thanks for the response and for the pointer to `dev` — happy to rebase onto it.

Before I do, I want to flag something that might save us both some time. I checked the current `dev`:

- `wide_search_max_distance` appears **0 times** in `get_search_type_retriever_instance.py` and **0 times** in `graph_completion_retriever.py` on `dev`.
- The five subclasses (decomposition / CoT / context-extension / summary / temporal) still have **closed `__init__` signatures** that don't accept the parameter.

But the factory has been refactored since I opened this PR — it now forwards via `**kwargs` (`async def get_search_type_retriever_instance(query_type, query_text, **kwargs)`, and community retrievers are constructed with `retriever(**kwargs)`).

So I want to make sure I'm fixing the right thing:

**1.** Is the underlying `TypeError` (passing `wide_search_max_distance` to a subclass that doesn't accept it) still reproducible on `dev` — or did the `**kwargs` refactor change the code path such that it no longer occurs?

**2.** If it's still live, which shape would you prefer?
   - **(a)** explicit `wide_search_max_distance` parameter on each subclass (what this PR does), or
   - **(b)** `**kwargs` pass-through in the subclasses, matching the current style of the factory?

I'd rather rebuild the patch against `dev` in the shape you actually want than rebase a June diff onto a heavily-changed tree. I'll hold off on pushing anything until you reply.

(Also noting for context: this PR was opened against `main` because at the time I wasn't aware `dev` was the active branch — thanks for clarifying that.)

PR #3729, opened 2026-07-01
fix(search): return empty result when knowledge graph is empty

### LINKED ISSUE #3728: [Bug]: GRAPH_COMPLETION search returns a phantom LLM non-answer instead of an empty result when the knowledge graph is empty
### Bug Description

When the knowledge graph is empty (e.g. immediately after `forget()` deletes a dataset, or before any data is ingested), `search()` with `SearchType.GRAPH_COMPLETION` does not return an empty result. Instead, it sends empty context to the LLM and returns the LLM's polite deflection (e.g. *"There is no context provided. Please provide the context to answer the question."*) as a normal search result.

This makes an empty-graph search **indistinguishable from a real answer** to the caller — `search()` returns a non-empty list, so callers that check `if results:` are misled into thinking memory was found.

The retriever already detects the empty-context condition (it logs a warning about it) but proceeds to call the LLM anyway instead of short-circuiting.

### Steps to Reproduce

```python
import asyncio
import cognee

async def main():
    await cognee.add("Jane proposed SQLite for a 10M req/day system.")
    await cognee.cognify()
    r1 = await cognee.search("What did Jane propose?", query_type=SearchType.GRAPH_COMPLETION)
    print("BEFORE forget:", "NON-EMPTY" if r1 else "EMPTY")   # NON-EMPTY (correct)

    await cognee.forget(dataset="main_dataset")               # graph is now empty

    r2 = await cognee.search("What did Jane propose?", query_type=SearchType.GRAPH_COMPLETION)
    print("AFTER forget:", r2)   # BUG: non-empty phantom answer

asyncio.run(main())
```

1. Ingest a document and build the graph (`add` + `cognify`).
2. Confirm `GRAPH_COMPLETION` search returns a real answer.
3. Call `forget()` to empty the graph.
4. Run the same `GRAPH_COMPLETION` search again — it returns a phantom "no context" string instead of an empty result.

### Expected Behavior

When the knowledge graph / retrieved context is empty, `GRAPH_COMPLETION` search should return an empty result (`[]`) rather than calling the LLM and returning a phantom "no context" string as if it were data. Callers should be able to distinguish "no memory found" from "here is an answer."

### Actual Behavior

After `forget()`, the graph is confirmed empty (logs show `No nodes found in the database` and `Search attempt on an empty knowledge graph`), yet `search()` returns a non-empty list containing a phantom LLM deflection:
['There is no context provided. Please provide the context to answer the question.']
Because the list is non-empty, callers cannot tell this apart from a genuine answer.

### Environment

- Cognee: 1.2.1
- OS: Windows 11
- Python: 3.12.4
- Backend: SQLite + LanceDB + Ladybug (embedded KuzuDB)
- Embeddings: fastembed (BAAI/bge-small-en-v1.5)
- LLM: Groq (llama-3.3-70b-versatile)

### Logs/Error Messages

```shell
Immediately after forget(), the retriever logs that the graph is empty but still calls the completion:
[warning] No nodes found in the database [cognee.shared.logging_utils]
[warning] Search attempt on an empty knowledge graph [cognee.shared.logging_utils]
[warning] Search attempt on an empty knowledge graph [GraphCompletionRetriever]
[warning] Empty context was provided to the completion [GraphCompletionRetriever]
forget() itself returned success and did purge correctly:
forget() returned: {'dataset_id': '2d1edab5-...', 'status': 'success'}
The problem is downstream: the retriever emits the "Empty context was provided to the completion" warning and then calls the LLM anyway, returning its reply.
```

### Additional Context

`forget()` works correctly here — the graph is genuinely emptied. The bug is purely in the graph-completion retriever's handling of an empty graph.

**Suggested fix:** short-circuit in the retriever — when the retrieved context/graph is empty, return an empty result before invoking the completion LLM. The retriever already detects this condition (the "Empty context was provided to the completion" warning), so it just needs to return early instead of proceeding.

I hit this during the WeMakeDevs hackathon while testing `forget()` + recall, and I have the full reproduction and logs above. I'd like to work on the fix — @Vasilije1990 @dexters1 could this be assigned to me?

### Pre-submission Checklist

VignanNallani:
## What

`GRAPH_COMPLETION` search returned a phantom LLM non-answer (e.g. *"There is no context provided. Please provide the context to answer the question."*) instead of an empty result when the knowledge graph was empty — for example right after `forget()`, or before any data is ingested.

Because `search()` returned a non-empty list, callers doing `if results:` could not distinguish "no memory found" from a real answer.

## Root cause

In `get_retriever_output`, `graph_engine.is_empty()` was checked and the empty-graph warning was logged, but execution fell through instead of returning. It then called the retriever (which correctly returned `[]`), built empty context (`""`), and still passed that empty context to the completion LLM — whose polite "no context" reply became the search result.

## Fix

Early-return a `SearchResultPayload` with empty `result_object`/`completion` immediately after the `is_empty` warning, before any retriever instantiation or LLM call. This mirrors the existing early-return style used for the `turn_preparation.should_answer` branch.

`is_empty()` is a global, dataset-agnostic node check (`MATCH (n) ... LIMIT 1`), so this only short-circuits when the graph store is physically empty — it does not affect the separate case where a non-empty graph simply yields no results for a query (already handled by the retriever returning `[]`).

## Testing

- Manual repro: `add` + `cognify` → search returns a real answer → `forget()` → search now returns an empty result (no phantom answer), verified against the local KuzuDB/LanceDB backend.
- Added `cognee/tests/unit/search/test_get_retriever_output.py`: mocks `is_empty() -> True` and asserts the retriever is **never instantiated** and the payload is empty. Verified it **fails without the fix** (the retriever mock's `AssertionError` fires) and passes with it. `ruff format` and `ruff check` clean.

Fixes #3728

---

### DIFF
--- a/cognee/modules/search/methods/get_retriever_output.py
+++ b/cognee/modules/search/methods/get_retriever_output.py
@@ -34,6 +34,16 @@ async def get_retriever_output(
 
     if is_empty:
         logger.warning("Search attempt on an empty knowledge graph")
+        return SearchResultPayload(
+            result_object=[],
+            context=None,
+            completion=[],
+            search_type=query_type,
+            only_context=kwargs.get("only_context", False),
+            dataset_name=kwargs.get("dataset").name if kwargs.get("dataset") else None,
+            dataset_id=kwargs.get("dataset").id if kwargs.get("dataset") else None,
+            dataset_tenant_id=kwargs.get("dataset").tenant_id if kwargs.get("dataset") else None,
+        )
 
     retriever_instance = await get_search_type_retriever_instance(
         query_type=query_type, query_text=query_text, **kwargs
--- a/cognee/tests/unit/search/test_get_retriever_output.py
+++ b/cognee/tests/unit/search/test_get_retriever_output.py
@@ -1,0 +1,45 @@
+import importlib
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+from cognee.modules.search.types import SearchType
+
+
+@pytest.fixture
+def retriever_output_mod():
+    return importlib.import_module("cognee.modules.search.methods.get_retriever_output")
+
+
+@pytest.mark.asyncio
+async def test_empty_graph_returns_empty_payload_without_llm_call(
+    monkeypatch, retriever_output_mod
+):
+    """Empty knowledge graph must return an empty result without calling the completion LLM
+    (regression for the phantom 'no context' answer)."""
+    mock_engine = MagicMock()
+    mock_engine.is_empty = AsyncMock(return_value=True)
+
+    async def mock_get_graph_engine():
+        return mock_engine
+
+    monkeypatch.setattr(retriever_output_mod, "get_graph_engine", mock_get_graph_engine)
+    monkeypatch.setattr(
+        retriever_output_mod,
+        "get_search_type_retriever_instance",
+        AsyncMock(
+            side_effect=AssertionError(
+                "get_search_type_retriever_instance must not be called on an empty graph"
+            )
+        ),
+    )
+
+    result = await retriever_output_mod.get_retriever_output(
+        query_type=SearchType.GRAPH_COMPLETION,
+        query_text="What database mistake did Jane make?",
+    )
+
+    assert result.result_object == []
+    assert result.completion == []
+    assert result.context is None
+    assert result.search_type == SearchType.GRAPH_COMPLETION


dexters1:
Thanks for the PR @VignanNallani! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #3731, opened 2026-07-01
feat(llm): Ollama model validation — curated supported-models list + startup warning (#3647)

### LINKED ISSUE #3647: Hackathon: Validate & document supported Ollama models for extraction
## 🏆 Hackathon task: Validate & document supported Ollama models for end-to-end extraction

> **Ollama:** I couldn't get a satisfactory end-to-end extraction pipeline with qwen2.5 or mistral before switching to hosted providers. Better validation/documentation for supported Ollama models would help.

### What to build
- [ ] **Curated, tested list of Ollama models** known to work end-to-end (esp. structured graph extraction), with notes on which fail and why (structured-output/JSON-schema support varies by local model).
- [ ] **Startup validation/warning** — when `LLM_PROVIDER=ollama` with an unvalidated model, warn and point to the supported list (build on `ensure_env_vars_for_ollama` in `cognee/infrastructure/llm/config.py`).
- [ ] A runnable **local-only Ollama example** that works out of the box, plus a troubleshooting doc.

### Why
Local/Ollama is a headline use case (privacy, no cloud cost), but extraction quietly fails on models without solid structured-output support. Validating + documenting which models work removes a major frustration and prevents the "switch to hosted" fallback.

### Requirements
- ✅ Validation is advisory (warn, don't hard-block); docs list supported models + known-bad ones.
- ✅ **Test coverage** — validation warns on an unsupported model; the local example is covered via the mocked-LLM harness (#3601).

Related: #3605 (calm onboarding), #3604 (actionable errors), #3601 (mocked tests).

kartikjsonawane:
When LLM_PROVIDER=ollama, Cognee would silently fail during graph extraction if the model doesn't support JSON-schema mode. Users had no indication of why extraction was producing empty results. I wanted to fix this by adding a warning at startup so people know immediately if their model choice is problematic.

Added a curated ollama_models.py with SUPPORTED_MODELS (models I've verified work with structured output) and KNOWN_BAD_MODELS (models confirmed to break JSON extraction, like mistral:7b and llama2). The check_ollama_model() function warns but never blocks — the user can still proceed, they just get a clear message pointing to the docs.

Wired it into the existing ensure_env_vars_for_ollama validator in LLMConfig so it fires automatically at startup. Also added a local_ollama_example.py with full setup instructions and a troubleshooting section.

## Acceptance Criteria
- check_ollama_model() warns on known-bad models (e.g. mistral:7b) and unvalidated models
- check_ollama_model() is silent on supported models (e.g. llama3.1:8b)
- Warning fires through LLMConfig when LLM_PROVIDER=ollama is configured
- No hard block — advisory only
- 14/14 unit tests pass fully offline (no Ollama required)

Closes #3647
Screenshot From 2026-07-01 12-02-33
Screenshot From 2026-07-01 12-02-42

### DIFF
--- a/cognee/infrastructure/llm/config.py
+++ b/cognee/infrastructure/llm/config.py
@@ -1,7 +1,10 @@
 import json
+import logging
 import os
 from functools import lru_cache
 from typing import Any, ClassVar
+
+logger = logging.getLogger(__name__)
 
 from pydantic import model_validator
 from pydantic_settings import BaseSettings, SettingsConfigDict
@@ -121,6 +124,10 @@ class LLMConfig(BaseSettings):
             if isinstance(value, str) and len(value) >= 2:
                 if value[0] == value[-1] and value[0] in ("'", '"'):
                     setattr(self, field_name, value[1:-1])
+
+        # Warn if the chosen Ollama model is unvalidated or known-bad
+        from cognee.infrastructure.llm.ollama_models import check_ollama_model
+        check_ollama_model(self.llm_model)
 
         return self
 
--- a/cognee/infrastructure/llm/ollama_models.py
+++ b/cognee/infrastructure/llm/ollama_models.py
@@ -1,0 +1,139 @@
+"""Curated list of Ollama models validated for end-to-end graph extraction with Cognee.
+
+Structured-output / JSON-schema support varies widely across local models.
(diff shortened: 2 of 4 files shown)


Vasilije1990:
This will get outdated very quickly @kartikjsonawane , I won't merge it

Issue #3643, opened 2026-06-29
Hackathon: `--dry-run` token estimator + skip retries on quota_exceeded


Vasilije1990:
## 🏆 Hackathon task: `--dry-run` token estimator + skip retries on `quota_exceeded`

> _From user feedback during evaluation._

> **remember() token usage:** Ingesting a single ~108 KB text document exhausted my provider quota. After the quota error, the structured-output path retried several times, increasing latency and attempted requests. A `--dry-run` token estimator and skipping retries on `quota_exceeded` would make evaluation much easier.

### Two improvements
1. **`--dry-run` token estimator** — before ingesting, estimate the LLM token usage (and rough cost) of a `remember`/`cognify` over the input *without* making LLM calls, so users can size a job up front. Surface via the CLI (`--dry-run`) and SDK. Reuse the existing tokenizer (`cognee/infrastructure/llm/tokenizer/`) and the per-stage call shape (chunk summarization + structured graph extraction) — see the token-cost analysis in #3635/#3606.
2. **Don't retry on `quota_exceeded`** — the structured-output/instructor retry loop currently retries even on non-retryable provider errors (quota/insufficient_quota/429-billing). Classify these as terminal and **fail fast** with an actionable error (coordinate with #3604), instead of multiplying latency and request count.

### Requirements
- ✅ Token estimate is reasonably accurate vs the actual run (validate against a fixture); no LLM calls in dry-run mode.
- ✅ Quota/billing errors are non-retryable and produce a clear, actionable message.
- ✅ **Test coverage** — dry-run estimate path + retry-classification (mocked provider returning quota error retries 0 times) via the mocked-LLM harness (#3601).

Related: #3606 (usage metering), #3635 (token-cost bench), #3604 (actionable errors), #3601 (mocked tests).

RajdeepKushwaha5:
Hi @Vasilije1990, I would like to take this. Two contained improvements; plan grounded in the retry policy and tokenizer as they are today.

## What I confirmed in the code

- The retry policy is shared and time-based: `cognee/infrastructure/llm/retry_config.py` defines `llm_retry_stop_condition = stop_after_attempt(2) & stop_after_delay(240)`. The `&` means a structured-output call keeps retrying until BOTH floors are met, so it can
  retry for up to ~240s. The adapters' `@retry` use  `retry_if_not_exception_type((NotFoundError, AuthenticationError, ...))`, which excludes
[...]

Issue #3606, opened 2026-06-28
Hackathon: Expansion Signals + Usage Metering (built on existing OTEL)


Vasilije1990:
## 🏆 Hackathon task: Expansion Signals + Usage Metering

**Goal:** make usage and cost **visible** to a self-hosting user, and **calmly notify them when they're outgrowing the local setup** — so the path from "trying it locally" to "this is load-bearing, time to scale" is honest and obvious.

> Build on what's already there. Cognee is already **OTEL-native** and already tracks per-session usage; this task is mostly *aggregating, surfacing, and acting on* signals we already emit — not building telemetry from scratch.

---

### What already exists (build on this)

- **OTEL-native tracing** — [`cognee/modules/observability/tracing.py`](https://github.com/topoteretes/cognee/blob/main/cognee/modules/observability/tracing.py): `setup_tracing()`, `get_tracer()`, a `CogneeSpanExporter` in-memory span buffer, and rich semantic attributes already emitted, including `cognee.llm.model` / `cognee.llm.provider`, `cognee.db.row_count`, `cognee.data.size_bytes`, `cognee.data.item_count`, `cognee.vector.result_count`, `cognee.search.type`, `cognee.operation.mode`, `cognee.recall.scope/source`. The `@observe` decorator ([`get_observe.py`](https://github.com/topoteretes/cognee/blob/main/cognee/modules/observability/get_observe.py)) wraps functions in spans (e.g. `as_type="generation"`).
- **Per-session usage + cost** — [`session_lifecycle/usage_tracking.py`](https://github.com/topoteretes/cognee/blob/main/cognee/modules/session_lifecycle/usage_tracking.py): `track_session_usage()`, `record_llm_call`, a `_PRICING_PER_M_TOKENS` table, accumulating into `SessionRecord`. **Known gap (stated in the code):** token counts are a rough `chars/4` estimate and don't read real `response.usage` from litellm/instructor; the pricing table is "conservative and incomplete."
- **Usage logging + metrics** — [`shared/usage_logger.py`](https://github.com/topoteretes/cognee/blob/main/cognee/shared/usage_logger.py), `modules/metrics/operations/get_pipeline_run_metrics.py`, `session_lifecycle/metrics.py`, observers (`NONE`/`LLMLITE`/`LANGSMITH`).

---

### Part 1 — Usage & cost metering (make it accurate, then surface it)

[...]

PR #3762, opened 2026-07-01
feat(tokenizer): auto-select embedding tokenizer and warn on mismatch

### LINKED ISSUE #3646: Hackathon: Auto-select or warn on tokenizer/model mismatch
## 🏆 Hackathon task: Auto-select / warn on tokenizer–model mismatch

> **Tokenizer selection:** The default `bert-base-uncased` tokenizer didn't match the models I tested. Auto-selecting or warning about tokenizer/model mismatches could improve chunking efficiency.

### What to build
- [ ] **Auto-select the tokenizer to match the configured LLM** where possible (cognee already has tokenizer adapters under `cognee/infrastructure/llm/tokenizer/` — HuggingFace, Gemini, Mistral), instead of defaulting to `bert-base-uncased` regardless of model.
- [ ] When auto-selection isn't possible, **warn on mismatch** (configured model vs tokenizer) so chunk sizing isn't silently wrong.
- [ ] Document the mapping and how to override.

### Why
A tokenizer that doesn't match the model miscounts tokens → inefficient/incorrect chunking and token estimates (which also undermines the `--dry-run` estimator). Matching the tokenizer to the model fixes chunk efficiency at the source.

### Requirements
- ✅ Sensible model→tokenizer resolution with safe fallback; explicit warning on mismatch.
- ✅ **Test coverage** — resolution picks the right tokenizer per model and warns on mismatch; deterministic, no network.

Related: token estimator (#3601-tested), #3606 (metering).

RajdeepKushwaha5:
## What

Chunk sizing counts tokens with a tokenizer. When that tokenizer does not match the embedding model, the counts are wrong, so chunks are mis-sized and the `--dry-run` token estimate is skewed. This centralizes model to tokenizer resolution across all four embedding engines so the right tokenizer is picked automatically, with a safe fallback and a clear warning when a match is not possible.

Closes #3646.

## Framing note

The issue says "match the configured LLM", but in cognee the chunking tokenizer is tied to the **embedding** model (each engine's `get_tokenizer`), not the LLM. The mismatch the report describes (a generic default not matching the model) happens on the embedding side, so that is where this fixes it.

## What was wrong (verified in code)

- `FastembedEmbeddingEngine` hardcoded `TikTokenTokenizer(model="gpt-4o")` for every model, so BGE/MiniLM/E5 (BERT wordpiece) were counted with OpenAI BPE.  Concrete mismatch.
- The HuggingFace -> TikToken fallbacks in `LiteLLMEmbeddingEngine` and   `OpenAICompatibleEmbeddingEngine` logged a generic message, not a warning tied to chunk sizing being off.
- `OllamaEmbeddingEngine` built its tokenizer from `HUGGINGFACE_TOKENIZER` with no signal when that value is unset or does not match the embedding model, and no fallback (it would raise on a bad value).

## What this does

New `cognee/infrastructure/llm/tokenizer/resolver.py` with a single `resolve_embedding_tokenizer(provider, model, max_completion_tokens, huggingface_tokenizer=None)` that all four engines delegate to:

- `openai` / `gemini` -> TikToken; `mistral` -> Mistral tokenizer.
- `fastembed` -> the model's own HuggingFace tokenizer (resolved from the fastembed model map), instead of the hardcoded `gpt-4o` BPE tokenizer.
- ollama / openai-compatible / custom / other -> an explicit `HUGGINGFACE_TOKENIZER` override if set, otherwise the embedding model's own HuggingFace repo.
- Safe fallback: any failure to load a matching tokenizer falls back to TikToken and logs a clear, advisory warning ("token counts are approximate ... set HUGGINGFACE_TOKENIZER to a tokenizer matching your embedding model"). It never raises, since a wrong count is a degraded estimate, not a fatal error.
- Also logs an advisory when an explicit `HUGGINGFACE_TOKENIZER` plausibly differs from the embedding model, so a genuine mismatch is not silent.

The four engines (`LiteLLM`, `Fastembed`, `Ollama`, `OpenAICompatible`) now just call the resolver, so the mapping and the warnings live in one place.

## Tests

`cognee/tests/unit/infrastructure/llm/test_tokenizer_resolver.py` (12 tests). Deterministic and offline: the three tokenizer classes are patched with sentinels, so the tests assert the selection logic and the warning paths without optional deps or network. Covers per-provider resolution (openai ->TikToken,mistral -> Mistral, fastembed BGE -> its HF tokenizer, openai-compatible ->
model repo, ollama -> override), the fallbacks (unknown fastembed model, HF load failure, no resolvable target), the mismatch advisory, and that resolution never raises.

All 12 pass; `ruff check` / `ruff format` clean; the existing embedding-engine unit tests still pass.

## Backward compatibility

Resolution only improves the default and adds warnings; an explicit `HUGGINGFACE_TOKENIZER` override still wins. openai / gemini / mistral behaviour is unchanged. Ollama gains a safe TikToken fallback where it previously raised.

## Docs

`.env.template` documents the model to tokenizer mapping and the `HUGGINGFACE_TOKENIZER` override.

## Why it matters

Strengthens the `--dry-run` token estimator (#3643) and metering (#3606), whose accuracy depends on the tokenizer matching the model.

## Files

- `cognee/infrastructure/llm/tokenizer/resolver.py` (new)
- `cognee/infrastructure/databases/vector/embeddings/FastembedEmbeddingEngine.py`
- `cognee/infrastructure/databases/vector/embeddings/LiteLLMEmbeddingEngine.py`
- `cognee/infrastructure/databases/vector/embeddings/OllamaEmbeddingEngine.py`
- `cognee/infrastructure/databases/vector/embeddings/OpenAICompatibleEmbeddingEngine.py`
- `cognee/tests/unit/infrastructure/llm/test_tokenizer_resolver.py` (new)
- `.env.template` (tokenizer mapping + override docs)


cognee/tests/unit/infrastructure/llm/test_tokenizer_resolver.py
```diff
@@ -0,0 +1,140 @@
+"""Tests for embedding model to tokenizer resolution (issue #3646).
```
conorlazarparman, line 1: These tests cover the resolver thoroughly, but nothing asserts that the four engines actually delegate to it; if someone re-hardcodes a tokenizer in an engine's get_tokenizer(), no test fails. One small test per engine that patches resolve_embedding_tokenizer, & asserts that get_tokenizer() calls it with the engine's provider/model/override would lock in the regression this PR fixes (the fastembed gpt-4o hardcode was exactly this failure mode.)

Note: test_never_raises_and_returns_tokenizer only exercises an unknown provider, so it misses the issue  where known providers/models raise. Worth adding a test for an unknown openai model once fixed.

cognee/infrastructure/llm/tokenizer/resolver.py
```diff
@@ -0,0 +1,178 @@
...
+            model,
+            _MISMATCH_HINT,
+        )
+        return TikTokenTokenizer(model=None, max_completion_tokens=max_completion_tokens)
+
+    if huggingface_tokenizer and model and _bare_model(huggingface_tokenizer) != bare:
+        # An override that plausibly differs from the model is legitimate (an
+        # Ollama model id is not an HF repo), but flag it so a genuine mismatch
+        # is not silent.
+        logger.info(
```
conorlazarparman: This should be a warning, not logger.info, which is invisible under the default log config while all the fallback paths warn at warning. I suggest changing this to logger.warning. If you're concerned about noise for legitimate Ollama setups (where the override differing from the model ID is normal), keep the current soft wording but _still emit it at warning level_.

cognee/infrastructure/llm/tokenizer/resolver.py
```diff
@@ -0,0 +1,178 @@
...
+            for providers whose model id is not itself a HuggingFace repo.
+
+    Returns a ready tokenizer. Never raises: on any failure it warns and returns
+    a TikToken fallback.
+    """
+    provider_lower = (provider or "").lower()
+    bare = _bare_model(model)
+
+    if "openai" in provider_lower and "compatible" not in provider_lower:
+        return TikTokenTokenizer(model=bare, max_completion_tokens=max_completion_tokens)
```
conorlazarparman: What is supposed to happen: "never raises: on any failure it warns and returns a TikToken fallback." But this branch can raise: tiktoken.encoding_for_model throws KeyError for any model it doesn't recognize. Here is how to reproduce the error:
**resolve_embedding_tokenizer(provider="openai", model="openai/some-future-model")
# KeyError: 'Could not automatically map some-future-model to a tokeniser...'**
Same issue on line 137: MistralTokenizer raises if mistral-common isn't installed. 
Please wrap both in the same warn-and-fall-back-to-TikTokenTokenizer(model=None) pattern used in _huggingface_or_fallback, so the "never raises" guarantee holds everywhere.

conorlazarparman (changes requested):

RajdeepKushwaha5:
Thanks, I have addressed all three.

**Never-raises (blocking):** You were right, the guarantee was broken on two branches. `tiktoken.encoding_for_model` raises `KeyError` on an unknown model, and `MistralTokenizer` raises when `mistral-common` is not installed. Both now go through a shared `_load_or_tiktoken_fallback` helper that warns and falls back to the default `TikTokenTokenizer(model=None)`, the same pattern the HuggingFace path used (the HF helper now delegates to it, so there is a single fallback path). Added resolver tests for both cases, each asserting it falls back rather than raising.

**logger.info to logger.warning:** Done, so the override-mismatch advisory is visible under the default log config. Kept the soft wording for legitimate Ollama setups.

**Engine delegation coverage:** Added `test_embedding_tokenizer_delegation.py` with one test per engine (fastembed, litellm, ollama, openai-compatible) that patches `resolve_embedding_tokenizer` and asserts `get_tokenizer()` calls it with the engine's provider/model/override, so a re-hardcode like the fastembed `gpt-4o` bug now fails a test. The litellm one also checks the `hosted_vllm/` prefix is stripped before delegation.

Verified: ruff check + format clean; 46 unit tests pass (resolver, delegation, and the existing llm/embedding-engine suites). Rebased onto latest dev.

Vasilije1990 (approved):

PR #3781, opened 2026-07-02
fix(api): return 409 instead of 500 when deleting data during dataset processing (#3766)

### LINKED ISSUE #3766: [Bug]: Data deletion returns 500 while dataset pipeline is processing (forget dataId + DELETE data endpoints)
# [Bug]: Data deletion endpoints return 500 while dataset pipeline is processing

## Summary

On managed Cognee Cloud, deleting a single data item fails with an unexplained
`500 Internal Server Error` while the dataset's cognify pipeline is running
(`DATASET_PROCESSING_STARTED`). Both deletion surfaces are affected:

- `POST /api/v1/forget` with `dataId` (both `dataset` and `datasetId` addressing)
- `DELETE /api/v1/datasets/{dataset_id}/data/{data_id}`

Once the dataset reaches `DATASET_PROCESSING_COMPLETED`, both calls succeed:
`POST /api/v1/forget` with `dataId` returns
`200 {"data_id":"...","dataset_id":"...","status":"success"}` and
`DELETE /api/v1/datasets/{dataset_id}/data/{data_id}` returns 200, with the
item removed from `GET /datasets/{id}/data` in both cases.

The 500 gives the client no way to distinguish "retry later" from a genuine
server fault. A 409/423 with a "dataset is processing" message (or queuing the
deletion) would make the failure mode self-explanatory.

## Environment

- Endpoint: managed Cognee Cloud tenant (`https://tenant-<uuid>.aws.cognee.ai`)
- Server version reported by `/health`: `1.2.2.dev0`
- Auth: `X-Api-Key` header
- Date observed: 2026-07-01 / 2026-07-02

## Steps to reproduce

```bash
# 1. Remember a document (starts background cognify on the dataset)
curl -X POST "$BASE/api/v1/remember" -H "X-Api-Key: $KEY" \
  -F "data=@item.json;type=application/json" -F "datasetName=repro_ds" \
  -F "run_in_background=true"

# 2. Resolve dataset id and data id
curl -H "X-Api-Key: $KEY" "$BASE/api/v1/datasets/"                      # -> DS_ID
curl -H "X-Api-Key: $KEY" "$BASE/api/v1/datasets/$DS_ID/data"           # -> DATA_ID

# 3. Confirm the pipeline is still running
curl -H "X-Api-Key: $KEY" "$BASE/api/v1/datasets/status?dataset_ids=$DS_ID"
# -> {"<DS_ID>": "DATASET_PROCESSING_STARTED"}

# 4. Attempt deletion while processing — all three return 500
curl -X POST "$BASE/api/v1/forget" -H "X-Api-Key: $KEY" -H "Content-Type: application/json" \
  -d "{\"dataset\":\"repro_ds\",\"dataId\":\"$DATA_ID\"}"
# -> 500 {"error":"An error occurred during deletion."}

curl -X POST "$BASE/api/v1/forget" -H "X-Api-Key: $KEY" -H "Content-Type: application/json" \
  -d "{\"datasetId\":\"$DS_ID\",\"dataId\":\"$DATA_ID\"}"
# -> 500 {"error":"An error occurred during deletion."}

curl -X DELETE -H "X-Api-Key: $KEY" "$BASE/api/v1/datasets/$DS_ID/data/$DATA_ID"
# -> 500 Internal Server Error

# 5. Wait for DATASET_PROCESSING_COMPLETED, then retry either call — both succeed
curl -X POST "$BASE/api/v1/forget" -H "X-Api-Key: $KEY" -H "Content-Type: application/json" \
  -d "{\"dataset\":\"repro_ds\",\"dataId\":\"$DATA_ID\"}"
# -> 200 {"data_id":"<DATA_ID>","dataset_id":"<DS_ID>","status":"success"}

curl -X DELETE -H "X-Api-Key: $KEY" "$BASE/api/v1/datasets/$DS_ID/data/$DATA_ID"
# -> 200, item removed from GET /datasets/$DS_ID/data
```

## Expected behavior

Either of:

- Deletion requests during processing return a structured, retryable error
  (409 Conflict or 423 Locked) with a message like
  `"dataset <id> is processing; retry after completion"`, or
- The deletion is queued and applied when the pipeline finishes.

## Actual behavior

`500 Internal Server Error` with `{"error":"An error occurred during deletion."}`
(forget route) or an empty `Internal Server Error` body (datasets route), with
no indication that the dataset's processing state is the cause.

## Related

- #3526 covers a different deletion-lifecycle issue (recall still surfaces
  deleted facts after re-cognify); this issue is about the deletion call itself
  failing during pipeline processing.

## Context

Found while integrating Cognee Cloud memory (remember / recall / cognify /
forget) into a desktop Git client for The Hangover Part AI hackathon. Our
client now checks `GET /api/v1/datasets/status` before item deletion and
surfaces "still processing, retry in a moment" to users — but a structured
error from the API would make that workaround unnecessary.

Utkal059:
## What

Fixes #3766. Deleting a single data item while the dataset's cognify pipeline is
still running (`DATASET_PROCESSING_STARTED`) returned an opaque
**500 Internal Server Error** on both deletion surfaces:

- `POST /api/v1/forget` with `dataId`
- `DELETE /api/v1/datasets/{dataset_id}/data/{data_id}`

The 500 gave clients no way to distinguish "retry later" from a genuine server
fault. Both surfaces now return **409 Conflict** with a self-explanatory
message while processing is in progress.

## Root cause

Both endpoints funnel into the same chokepoint, `datasets.delete_data(...)`
(`cognee/api/v1/datasets/datasets.py`). It performed no pipeline-status check,
so the deletion raced the in-flight pipeline (relational/graph/vector state
mid-write) and blew up with an unhandled exception:

- the datasets `DELETE` endpoint has no local handler, so it surfaced as a raw 500;
- the forget router's `except Exception` catch-all masked it as
  `{"error": "An error occurred during deletion."}` with status 500.

## Fix

1. **New `DatasetProcessingInProgressError` (HTTP 409)** in
   `cognee/modules/data/exceptions/` — follows the existing exception pattern
   (`CogneeValidationError` subclass with a `status_code`).
2. **One guard at the chokepoint** — `datasets.delete_data` now checks
   `get_pipeline_status([dataset.id], "cognify_pipeline")` (the same mechanism
   `check_pipeline_run_qualification` already uses for cognify) right after
   authorization, and raises the new error when the status is
   `DATASET_PROCESSING_STARTED`. Because the guard sits at the shared
   chokepoint, **both** endpoints are fixed by one check.
3. **Routing the error:**
   - `DELETE /datasets/{id}/data/{id}` needs no changes — the app-level
     `CogneeApiError` handler in `api/client.py` already maps the exception's
     `status_code` (409) to the response.
   - The forget router needed an explicit
     `except DatasetProcessingInProgressError` **before** its generic
     `except Exception`, which would otherwise mask the 409 as a 500.
4. **Docs:** added the 409 to both endpoints' Error Codes docstrings.

## Behavior after the fix

```
# while DATASET_PROCESSING_STARTED
POST /api/v1/forget {dataId, datasetId}          -> 409 {"error": "Dataset ... is currently being processed. Retry once processing completes."}
DELETE /api/v1/datasets/{id}/data/{data_id}      -> 409 {"detail": "... [DatasetProcessingInProgressError]"}

# after DATASET_PROCESSING_COMPLETED
both -> 200, item removed (unchanged)
```

## Tests

Added `cognee/tests/unit/api/v1/datasets/test_delete_data_during_processing.py`:

- `test_delete_data_raises_409_while_dataset_is_processing` — asserts the
  guard raises with `status_code == 409` and fires **before** any deletion
  work touches the dataset.
- `test_delete_data_proceeds_when_processing_completed` — deletion still
  succeeds once the pipeline is `DATASET_PROCESSING_COMPLETED`.
- `test_delete_data_proceeds_when_no_pipeline_status` — datasets with no
  recorded pipeline run are not blocked.

## Notes

- Scope is intentionally limited to the two surfaces reported in #3766.
- The issue suggested 409/423 or queuing the deletion; 409 was chosen because
  the codebase already uses it for dataset conflicts and it requires no new
  semantics. Queuing deletions during processing would be a larger design
  change and can be layered on top of this error path later.


Utkal059:
@Vasilije1990 @dexters1

dexters1:
Thanks for the PR @Utkal059! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #3783, opened 2026-07-02
feat(db): add Turso as a relational backend

### LINKED ISSUE #3570: [hackathon]: Add Turso as a relational database option
## Task
Add Turso (libSQL) as a first-class relational backend in Cognee, selectable with `DB_PROVIDER=turso`. Implement it as a new `TursoAdapter` class extending the existing `SQLAlchemyAdapter`, wired into the engine factory so that `get_relational_engine()` returns a `TursoAdapter` when Turso is configured. Turso class methods should be based on the existing SQLAlchemyAdapter and should behave the same.

## Design details

1. Factory branch: add `elif db_provider == "turso":` in
   `cognee/infrastructure/databases/relational/create_relational_engine.py`
   (the `if/elif/else` at lines 49-79). Build the libSQL connection string from
   config and return `TursoAdapter(...)`. Raise the existing
   `ConnectionError("unsupported DB type: ...")` only for truly unknown providers.
2. Config: extend `RelationalConfig`
   (`cognee/infrastructure/databases/relational/config.py`) so `DB_PROVIDER`
   accepts `turso`, and add the fields Turso needs that aren't covered by the
   existing `DB_*` set 
3. Optional dependency: gate the import behind an extras group (e.g.
   `cognee[turso]`) and raise a clear `ImportError` with install instructions if
   the libSQL driver is missing — mirroring how the `postgres` branch guards
   `import asyncpg` in `create_relational_engine.py`.

- Reuse the existing `@lru_cache` engine caching in `create_relational_engine`
  (pass hashable args exactly as the sqlite/postgres branches do).
- Alembic migrations under `cognee/alembic/versions/` already branch on the
  sqlite dialect — these apply to libSQL unchanged.

## Acceptance Criteria
- Setting `DB_PROVIDER=turso` (with a valid Turso URL + auth token) makes
  `get_relational_engine()` return a `TursoAdapter`, and the full
  add → cognify → search flow works end-to-end against Turso.
- Both embedded/local libSQL and a remote Turso database are supported and
  documented in `.env.template`.
- Missing libSQL driver produces an actionable `ImportError`
  (e.g. "pip install cognee[turso]"), matching the Postgres precedent.
- Unit test added under
  `cognee/tests/unit/infrastructure/databases/relational/` (mirroring
  `test_create_relational_engine.py`) verifying the `turso` branch builds the
  right connection string and returns a `TursoAdapter`; lru_cache cleared
  between cases.

Make sure all relevant E2E tests in the test directory (that do not force a specific database) pass against Turso as the backend, and provide success information.

hassaansaleem28:
## Description

Resolves #3570 

This pr adds Turso (libSQL) as a relational backend, selectable with DB_PROVIDER=turso. I built it on the aioturso async engine rather than aiosqlite, following @dexters1's steer that the point of Turso here is real async, not the thread-queue model aiosqlite uses.

My approach was to keep it as non invasive as possible. The relational layer has hundreds of call sites using async SQLAlchemy sessions, so instead of replacing anything I added a TursoAdapter that subclasses the existing SQLAlchemyAdapter. It reuses all the current behavior and only specializes what Turso actually needs, which means everything downstream keeps working unchanged. It's wired into the engine factory the same way sqlite and postgres are, behind a cognee[turso] optional extra with a clear install error if the driver is missing.

For remote, a plain remote connection isn't truly async through SQLAlchemy, so I used Turso's embedded replica model: reads and writes hit a fast local replica and a background task syncs it to the cloud primary. I tested this against a real Turso database and confirmed writes actually reach the cloud.

Along the way I hit and handled a few real things by running it rather than assuming libSQL behaves exactly like SQLite: the engine rejects the timeout connect arg the base adapter hardcodes, its SQLAlchemy dialect was missing an attribute our version expects (fix also going upstream), and it's single writer so concurrent same row writes race. For that last one I serialize writes in process with an async lock and retry busy errors, since I confirmed busy_timeout is ignored by this engine.

**Note**: schema, CRUD, remote sync, and the write handling all work, but the full add to cognify to search pipeline is currently blocked by one engine limitation. cognee's many to many joinedload generates a right nested join that the Rust engine can't parse yet, even on the latest build, and SQLAlchemy 2.0 has no toggle to avoid it. The same pipeline passes on sqlite, so it's the engine, not the integration. I've laid out the options in the [issue](https://github.com/topoteretes/cognee/issues/3570#issuecomment-4880223922) thread and would like a steer on direction.

## Type of Change

## Screenshots
image

## Pre-submission Checklist

## DCO Affirmation


rshkarin:
Superseded by https://github.com/topoteretes/cognee/pull/4077

PR #3789, opened 2026-07-02
feat: bi-temporal DataPoint support (valid_to field + close_node)

### LINKED ISSUE #3700: Temporal awareness in the Cognee brain
## Context
The brain stores knowledge without a strong notion of time. Facts have no temporal context, so the brain can't tell when something was true, when it changed, or when it has gone stale — outdated data is treated as if it were still current.

## Request
Add a temporality layer so stored knowledge carries temporal context and can be reasoned about over time. Specifically:

- **Temporal tracking:** attach temporal context (when a fact was established / observed / last valid) to stored knowledge, and support temporal adjustments when data changes over time.
- **Staleness / freshness prediction:** estimate when a piece of data is likely to become outdated, so the brain can mark or down-weight stale knowledge instead of treating it as still-current.

## Acceptance criteria
- [ ] Stored knowledge carries temporal metadata that can be queried and updated.
- [ ] The system can identify (or predict) when data is outdated and expose that signal.
- [ ] Stale data can be marked or down-weighted rather than treated as current.

koopatroopa787:
## Summary

Closes #3700

Adds bi-temporal awareness to the graph so that stale or superseded facts can be detected, flagged, and filtered at query time — without deleting them from the graph.

### What's changed

**`cognee/infrastructure/engine/models/DataPoint.py`**
- Added `valid_to: int | None = None` — a millisecond-epoch timestamp. `None` means the node is still current; an integer means the fact was superseded at that time.
- The field is excluded from content-equality checks (alongside `created_at`, `updated_at`, `version`, `id`) so that stamping a node as closed does not trigger a contradiction warning.

**`cognee/modules/graph/utils/deduplicate_nodes_and_edges.py`**
- Added `_content_differs(existing, incoming)` — compares two DataPoints by their non-volatile fields.
- `deduplicate_nodes_and_edges` now logs a `WARNING` when two nodes with the same id carry different content in the same ingest batch ("Contradiction detected"), pointing users toward `close_node()` as the resolution path. The first version is kept (existing behaviour).

**`cognee/tasks/storage/close_node.py`** (new)
- `async close_node(node_id)` — sets `valid_to = now_ms` on a node in the graph backend. Best-effort: fails silently if the node doesn't exist or the backend doesn't support partial updates.
- `is_valid(node, at_ms=None)` — pure helper that checks whether a DataPoint or plain dict is valid at a given point in time. Works uniformly across DataPoint instances and raw graph-dict returns.

**`conftest.py`** (repo root)
- Stub for the `mistralai`/`instructor` version mismatch in some dev environments. Installs a minimal shim with the correct `__spec__` so `importlib.util.find_spec("mistralai")` inside `instructor` does not raise. This only affects local test runs; CI with pinned deps is unaffected.

**`cognee/tests/unit/test_temporal_awareness.py`** (new)
- 15 unit tests covering: `valid_to` field defaults and JSON round-trip, `is_valid()` with past/future/explicit timestamps and dict nodes, contradiction detection and warning text, first-version-wins on contradiction, content-differ ignoring timestamps, and normal dedup.

### Usage pattern

```python
from cognee.tasks.storage.close_node import close_node, is_valid

# When a fact changes: close the old node, then write the new one
await close_node(old_entity.id)
await add_data_points([new_entity])

# At query time, filter results to only current facts:
current_nodes = [n for n in results if is_valid(n)]
```

## Test plan

- [x] `pytest cognee/tests/unit/test_temporal_awareness.py` — 15/15 passing locally
- [ ] Integration test: full add → cognify → close → search flow with `valid_to` filter

### DIFF
--- a/cognee/infrastructure/engine/models/DataPoint.py
+++ b/cognee/infrastructure/engine/models/DataPoint.py
@@ -57,6 +57,7 @@ class DataPoint(BaseModel):
     ontology_valid: bool = False
     version: int = 1  # Default version
     topological_rank: int | None = 0
+    valid_to: int | None = None  # ms epoch when this fact was superseded; None = still valid
     metadata: MetaData = {"index_fields": []}
     type: str = Field(default_factory=lambda: DataPoint.__name__)
     belongs_to_set: "list[DataPoint] | list[str] | None" = None
--- a/cognee/modules/graph/utils/deduplicate_nodes_and_edges.py
+++ b/cognee/modules/graph/utils/deduplicate_nodes_and_edges.py
@@ -1,15 +1,40 @@
+import logging
+
 from cognee.infrastructure.engine import DataPoint
+
+logger = logging.getLogger(__name__)
+
+# Fields that change on every construction and must not trigger a contradiction warning.
+_VOLATILE_FIELDS = frozenset({"created_at", "updated_at", "valid_to", "version", "id"})
+
+
+def _content_differs(existing: DataPoint, incoming: DataPoint) -> bool:
+    """Return True when two same-id DataPoints have different content fields."""
+    ex_dict = existing.model_dump(exclude=_VOLATILE_FIELDS)
+    in_dict = incoming.model_dump(exclude=_VOLATILE_FIELDS)
+    return ex_dict != in_dict
 
 
 def deduplicate_nodes_and_edges(nodes: list[DataPoint], edges: list[dict]):
-    added_entities = {}
-    final_nodes = []
-    final_edges = []
+    added_entities: dict = {}
+    final_nodes: list = []
+    final_edges: list = []
 
     for node in nodes:
-        if str(node.id) not in added_entities:
+        node_key = str(node.id)
+        if node_key not in added_entities:
             final_nodes.append(node)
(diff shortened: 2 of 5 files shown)


koopatroopa787:
Hi @siillee @Vasilije1990 — this PR is complete and ready for review.

To summarise what's implemented:
- **DataPoint.valid_to** — optional int | None field (ms epoch) added to the base model. None = still valid, integer = superseded. Fully backward-compatible, all existing pipelines unaffected.
- **close_node(node_id)** — async helper that stamps alid_to = now_ms on any node, generalising the bi-temporal pattern from the Graphiti path into core storage.
- **is_valid(node, at_ms=None)** — predicate for freshness checks at query time.
- **Contradiction detection** in deduplicate_nodes_and_edges — logs a warning when conflicting content is detected on re-ingest.
- **15 unit tests** — covering valid_to semantics, is_valid(), and contradiction detection. All passing locally, no external dependencies.

This closes #3700. Happy to address any review feedback quickly.

dexters1:
Thanks for the PR @koopatroopa787! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #3797, opened 2026-07-02
feat: incremental re-cognify (cognee update) + git post-commit hook (#3669)

### LINKED ISSUE #3669: Incremental re-cognify (--update) + git post-commit hook
## Incremental re-cognify (`--update`) + git post-commit hook

**Hackathon idea.** Copy [graphify](https://github.com/safishamsi/graphify)'s incremental update: only re-process inputs that actually changed, instead of rebuilding the whole graph. Graphify hashes a manifest, re-extracts just the changed files on `--update`, installs a post-commit hook to auto-refresh, and ships a git merge driver so `graph.json` never ends up with conflict markers.

### Surface (the opt-in "option")
- `--update` flag (and/or `cognee.remember(..., incremental=True)`) that hash-diffs inputs against a stored manifest and only re-`cognify`s changed/new items, pruning nodes for removed sources.
- Optional `cognee hook install` writing a post-commit hook to refresh the graph.

### Why
- Big cost saver on large/evolving corpora — today every re-ingest reprocesses everything.
- Complements the md5 exact-dedup already in the trivago demo's `ingest_trv.py` (that skips byte-dupes within a run; this skips unchanged inputs *across* runs).

### Scope notes
- Needs a persisted per-dataset manifest (path → hash → node ids) to know what to prune vs add.
- Merge-driver / team-sync piece is a nice-to-have stretch goal.

Runner-up. Part of #3670.

theUtkarshRaj:
Closes #3669.

Adds an opt-in incremental re-cognify path so evolving corpora don't rebuild the whole graph.

**API / CLI**
- `cognee.incremental_update(paths, dataset_name=..., prune_removed=True)`
- `cognee update <paths...> [-d NAME] [--no-prune]`
- `cognee hook install/uninstall` — writes a *managed* block into the git post-commit
  hook (resolved via `git rev-parse --git-path hooks`, so worktrees/submodules work;
  existing hooks are preserved).

**How it works (reuses existing infra — no parallel manifest)**
- added/modified/unchanged → `add()` + `cognify(incremental_loading=True)`; unchanged
  inputs skip LLM extraction, modified ones re-cognify (content-hash reset). The `Data`
  table is the source of truth, so it can't drift from the graph.
- removed sources → pruned via the existing `delete_data_nodes_and_edges` + `delete_data`;
  entities other still-present sources contribute are preserved by the delete path's
  co-ownership check (shared nodes never orphaned).

**Safety**
- Removal is only inferred inside directories that currently exist and were scanned, so a
  missing/unmounted/mistyped path can never be read as a mass deletion. Never prunes a
  file that still exists. First run (dataset absent) prunes nothing.
- Runs `setup()` first, so it's safe as the very first cognee call (e.g. a git hook on a
  fresh checkout).

**Tests**
- 8 offline, deterministic unit tests in `cognee/tests/unit/test_incremental_update.py`
  (no keys), covering removal detection, scoping guards, and the hook writer.
- Verified end-to-end against a live graph (add → cognify → prune) as well.

Single-file removal is intentionally out of scope (use `cognee forget`/`delete`).


Vasilije1990:
Due to complexity, we decided to handle this internally

PR #3806, opened 2026-07-02
feat: add per-stage LLM model routing

### LINKED ISSUE #3645: Hackathon: Per-stage model routing (cheap/local extraction, strong summarization/query)
## 🏆 Hackathon task: Per-stage model routing (cheap/local for extraction, stronger for summarization & query)

> **Per-stage model routing:** It would be useful to use a cheap/local model for extraction and a stronger model only for summarization and query-time reasoning.

### What to build
- [ ] Let users configure **different LLMs per pipeline stage** — e.g. a cheap/local model (Ollama) for graph extraction, a stronger model for summarization and for query-time reasoning (`search`/`recall`).
- [ ] Extend `cognee/infrastructure/llm/config.py` (and the gateway) so a stage can resolve its own model/provider, with a sensible default that preserves current behavior.
- [ ] Surface in config + docs; make the routing inspectable (which model ran which stage — ties to OTEL #3606).

### Why
Big cost lever: extraction dominates token usage (it runs per chunk), so routing it to a cheap/local model while reserving a strong model for the few summarization/query calls cuts cost without hurting answer quality.

### Requirements
- ✅ Backward compatible (single-model config still works); per-stage override is additive.
- ✅ **Test coverage** — each stage uses its configured model; verify via the mocked-LLM harness (#3601) asserting per-stage routing.

Related: #3606 (usage metering / OTEL), #3635 (token-cost), #3601 (mocked tests).

chiruu12:
Closes #3645

## Summary

Adds optional per-stage LLM model routing so a different model can be used for each pipeline stage. Extraction runs once per chunk and dominates token spend, so it is often worth routing a cheaper or local model there while keeping a stronger model for summarization and query-time reasoning.

The change is fully backward compatible (a single-model configuration behaves exactly as before) and the routing is inspectable in traces (each LLM span records which stage it ran in).

## Configuration

Each stage reads an optional `LLM_<STAGE>_*` group. Any field left unset falls back to the base `LLM_*` value, so you only set what you want to override.

```env
# Route a cheap or local model to extraction, keep a stronger model elsewhere.
LLM_EXTRACTION_MODEL="ollama_chat/llama3.1"
LLM_EXTRACTION_PROVIDER="ollama"
LLM_EXTRACTION_ENDPOINT="http://localhost:11434"

LLM_SUMMARIZATION_MODEL="openai/gpt-5-mini"
LLM_SUMMARIZATION_PROVIDER="openai"

LLM_QUERY_MODEL="openai/gpt-5-mini"
LLM_QUERY_PROVIDER="openai"
```

Stages: `extraction`, `summarization`, `query`. Each supports `MODEL`, `PROVIDER`, `ENDPOINT`, `API_KEY`, and `API_VERSION`.

## Design

The routing reuses the existing config seam rather than changing the gateway or any call site.

Every LLM call on the instructor path already flows:

```
LLMGateway.acreate_structured_output -> get_llm_client() -> get_llm_context_config() -> cached client
```

`get_llm_context_config()` reads the `llm_config` ContextVar and falls back to the global singleton, and the client cache key is derived from that config. So the feature is:

1. `LLMConfig.stage_config(stage)` returns a copy with the base `llm_*` fields overridden by any set `llm_<stage>_*` fields. With no overrides it returns an equivalent config.
2. A small `pipeline_stage(stage)` context manager sets the `llm_config` ContextVar to that merged config for the duration of the stage, and sets a second ContextVar with the stage label for tracing. Both reset on exit.
3. The three stage boundaries are wrapped: the extraction fan-out, the summarization fan-out, and `generate_completion` (the single completion path all retrievers delegate to).

Because the merged config only changes the client cache key, each stage transparently gets its own cached client. No change to `acreate_structured_output` or to any extraction, summarization, or retrieval call signature.

## Inspectability

`_enrich_llm_span` now stamps a `cognee.pipeline.stage` attribute on every LLM span when a stage is active, so traces show which model ran which stage.

## Included fix

`_record_session_usage_after` in `LLMGateway` read the global singleton (`get_llm_config().llm_model`) when recording usage. With per-stage routing the recorded model would not match the stage that actually ran, so it now reads `get_llm_context_config().llm_model`. This is a small latent correctness fix that per-stage routing makes visible.

## Backward compatibility

- All new fields default to empty; an unset stage returns the base config unchanged.
- `stage_config` with no overrides returns an equivalent config, so the client cache key and the chosen client are identical to today.
- No signature changes to the gateway or to any task call site.

## Tests

New no-network unit tests in `cognee/tests/unit/infrastructure/llm/test_stage_routing.py`:

- routing applies the stage override inside the block and reverts outside it,
- the client cache key is stage-specific,
- with no overrides the effective config and cache key are identical inside and outside (single-model behavior locked in),
- the usage-recording fix reads the context config,
- the stage label is set inside the block and cleared after.

The existing LLM unit suite stays green (`uv run pytest cognee/tests/unit/infrastructure/llm/ -v` -> 26 passed). These tests use deterministic `monkeypatch` in the same style as `test_llm_global_cache.py`.

## Documentation

`.env.template` documents the new settings inline. The narrative docs at docs.cognee.ai appear to be maintained outside this repo, so I have not added a page there in this PR. I am happy to open a follow-up PR adding a short "Per-stage model routing" page if you point me at the docs source, or to fold additional in-repo documentation into this PR if you would prefer that instead. Let me know which you would like.

## Out of scope (possible follow-ups)

- Per-stage routing on the BAML path. BAML builds its client registry at config init and bypasses `get_llm_client`, so it needs a separate approach.
- The custom extraction-callback branch in `extract_graph_from_data` (when a caller supplies `calculate_chunk_graphs`) is not wrapped; only the default fan-out is. Can be added if desired.
- Automatic model selection or cost optimization. This PR is configuration plus routing plus inspectability only.


rshkarin:
Thanks @chiruu12! Superseded by #3961 so full CI can run.

PR #3640, opened 2026-06-29
feat(tests): add LLM mocking harness and mocked tests for examples/guides

harshitsaxena214:
## Description

Currently, all example scripts under `examples/guides/` require real LLM API keys
to run and cannot be tested in CI. This PR adds a reusable LLM/embedding mocking
harness and mocked tests for all `examples/guides/` scripts so they can run in CI
fast, free, and deterministically.

The approach:
- Built a mock harness in `cognee/tests/utils/mock_llm.py` that patches cognee's
  LLM and embedding layers to return deterministic canned responses instead of
  hitting any real provider
- For structured outputs, the mock inspects the requested Pydantic `response_model`
  and builds a minimal valid instance instead of returning a plain `MagicMock` —
  this prevents `AttributeError` on field access inside pipelines
- Each test uses an isolated temp directory per run to avoid file lock conflicts
  on the Ladybug graph database
- Added `cognee/tests/examples/guides/test_*.py` for every script under
  `examples/guides/`

Closes #3601 

## Acceptance Criteria

- [x] A reusable mocking harness lives in `cognee/tests/utils/mock_llm.py`
- [x] Every script under `examples/guides/` has a corresponding mocked test
- [x] Tests run with zero real API keys and are fully deterministic
- [x] Tests fail loudly on import errors, signature changes, or runtime exceptions
- [x] Full suite result: **13 passed, 1 skipped, 0 failed**

## Type of Change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots

image

image

## Pre-submission Checklist

- [x] **I have tested my changes thoroughly before submitting this PR** (See `CONTRIBUTING.md`)
- [x] **This PR contains minimal changes necessary to address the issue/feature**
- [x] My code follows the project's coding standards and style guidelines
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if applicable)
- [x] All new and existing tests pass
- [x] I have searched existing PRs to ensure this change hasn't been submitted already
- [x] I have linked any relevant issues in the description
- [x] My commits have clear and descriptive messages

## DCO Affirmation
I affirm that all code in every commit of this pull request conforms to the terms of the Topoteretes Developer Certificate of Origin.

Issue #3623, opened 2026-06-28
Hackathon: Eval harness + LongMemEval runner; make eval/DeepEval an optional addon (propose the design)


Vasilije1990:
## 🏆 Hackathon task: Productize the eval harness — LongMemEval runner + DeepEval as an optional addon

**Goal:** make reproducible **memory-quality benchmarking** a first-class, optional **addon component** of cognee: add a **LongMemEval** runner, and turn the existing eval + **DeepEval** integration into a clean, packaged add-on with a one-command runner. **Ignore BEAM here** — that adapter already exists; this ticket is about LongMemEval + productizing the harness.

### 📣 Propose the design first

Comment with the best approach before coding: the public API/CLI surface for running an eval end to end, how the addon is packaged/optional (extra), and the LongMemEval adapter shape. Converge, then build.

### What already exists (extend this — don't rebuild)

cognee has a real eval framework at [`cognee/eval_framework/`](https://github.com/topoteretes/cognee/tree/main/cognee/eval_framework):
- **Benchmark adapters** (`benchmark_adapters/`): `base_benchmark_adapter.py`, plus `beam`, `hotpot_qa`, `musique`, `twowikimultihop`, `logistics_system`, `dummy` — registered in `benchmark_adapters.py`. **There is no LongMemEval adapter yet.**
- **Evaluation engines** (`evaluation/`): `base_eval_adapter.py`, `deep_eval_adapter.py` (DeepEval), and a `DirectLLM` path. `eval_config.py` defaults `evaluation_engine = "DeepEval"`.
- **Metrics** (`evaluation/metrics/`): `exact_match`, `f1`, `context_coverage`, `rubric`.
- **Pipeline pieces**: `corpus_builder/`, `answer_generation/`, `analysis/` (`metrics_calculator.py`, `dashboard_generator.py`).
- DeepEval is an **optional extra** today, but it's tangled with heavyweight `evals/` scripts + Modal/Docker images (`deepeval==3.2.x`).

### What to build

- [ ] **LongMemEval benchmark adapter** — subclass `base_benchmark_adapter`, register it, load the dataset, build the corpus, run QA, and score with the existing metrics/engines (mirror the HotPotQA/MuSiQue adapters).
[...]

Issue #3605, opened 2026-06-28
Hackathon: Calm, Honest First Run — evaluate onboarding & propose improvements


Vasilije1990:
## 🏆 Hackathon task: Calm, Honest First Run

**Goal:** polish the first few minutes for a brand-new user so they **get to a result quickly** and **always know the next step** — without surprises. This ticket is deliberately scoped as **evaluate the current onboarding first, then propose and implement improvements.**

> "Calm" = no scary stack traces, no dead ends, clear expectations. "Honest" = tell the truth up front about prerequisites, what's happening, and what it costs. The first run should feel like the product is guiding you, not testing you.

---

### Part 1 — Evaluate the current onboarding (deliverable: a short audit)

Walk the real first-run paths as a brand-new user (fresh machine, no prior cognee config) and document the friction. The entry points to cover:

- **Python SDK quickstart** (README): `uv pip install cognee` → set `LLM_API_KEY` → `remember` / `recall` / `forget` / `improve`.
- **CLI**: `cognee-cli remember/recall/forget`, plus the many other commands (`add`, `cognify`, `config`, `datasets`, `search`, `sessions`, …).
- **Local UI**: `cognee-cli -ui` (requires Docker/Colima).
- **Docker**: Compose (`cp .env.template .env` → `docker compose up`, plus `ui`/`mcp`/`postgres`/`neo4j` profiles) and the prebuilt image.

For each path, measure/record:
- [ ] **Time-to-first-result** — how long until the user sees a real `recall` answer?
- [ ] **Decisions/forks** — how many choices must a newcomer make before step 1 (which install tool? which entry point? which DB? Docker or not?).
- [ ] **Failure modes** — what happens with **no/invalid `LLM_API_KEY`**, no Docker for `-ui`, wrong Python version, calling `recall` before any `remember`/`cognify`? Capture the actual error text and whether it tells the user what to do next.
- [ ] **"What now?" gaps** — after install, after a successful `remember`, after a `recall` — does the output point to the obvious next step, or does the user have to go hunting?
- [ ] **Honesty gaps** — are prerequisites (API key required, token cost, Docker requirement, how long ingestion takes) stated *before* the user hits them?

Output: a concise findings doc (markdown in the repo or the issue) ranking the friction points by impact.

### Part 2 — Propose improvements

Based on the audit, propose a "calm, honest first run." Ideas to consider (not prescriptive — justify with the findings):
[...]

Issue #3601, opened 2026-06-28
Hackathon: build LLM-mocked tests for all examples in the repo


Vasilije1990:
### The problem

We ship a lot of example scripts under [`examples/`](https://github.com/topoteretes/cognee/tree/main/examples) — quickstarts, demos, guides, custom pipelines, database integrations, and POCs (~70 scripts in total). They're the first thing new users run, but right now nothing guarantees they keep working. They:

- require real LLM API keys (and burn real tokens / money) to run,
- are slow and non-deterministic, so they can't run in CI,
- silently rot when an API, import path, or signature changes.

### The goal

Add **LLM-mocked tests** for **all examples** so every example can run in CI, fast and for free, with **deterministic, mocked LLM responses** — catching breakage before users hit it.

The idea: mock the LLM (and embedding) layer so calls return canned, deterministic responses instead of hitting a provider. The example should then execute end to end, and the test should assert it ran without error and produced the expected shape of output (e.g. a graph was built, a recall returned results, no exceptions raised).

### Scope — examples to cover

Everything under `examples/`:

- `examples/guides/` — quickstarts (`agent_memory_quickstart.py`, `recall_core.py`, `improve_quickstart.py`, `ontology_quickstart.py`, `temporal_recall.py`, custom models/prompts/tasks, …)
- `examples/demos/` — the largest folder (`simple_cognee_example.py`, `remember_recall_improve_example.py`, session/temporal/multimedia demos, migration examples, …)
- `examples/custom_pipelines/` — custom cognify pipelines, agentic reasoning, org hierarchy, etc.
- `examples/database_examples/` + `examples/configurations/` — Chroma, Neo4j, pgvector, Neptune, Ladybug, permissions/tenancy setups
- `examples/pocs/` — disambiguation and canonicalization POCs
- `examples/python/` — provenance, references, schema inventory, reranking demos

### Suggested approach

1. Build a small, reusable mocking fixture/harness that intercepts cognee's LLM + embedding calls (e.g. patch the LiteLLM / LLM client layer) and returns deterministic canned responses, including valid structured outputs for graph extraction.
2. Put example tests under `cognee/tests/` (mirror the `examples/` folder structure), discoverable by the existing pytest setup.
3. For each example, add a test that runs it with mocks and asserts it completes successfully and produces the expected output shape.
4. Database-backed examples: mock or use an in-memory/local backend so no external services are required.
5. Wire the suite into CI so examples are checked on every PR.

### Acceptance criteria

[...]

PR #3817, opened 2026-07-03
feat(eval): one-command eval runner + optional eval/DeepEval addon (#3623)

Taimoo-r:
## What this does

This picks up the two chunks I was assigned on #3623: a one-command eval runner/CLI, and turning the eval + DeepEval harness into a proper optional addon. @siillee approved the design on the issue before I started.

I left the LongMemEval adapter out on purpose since that is a separate chunk. But I built everything so that the day someone registers a `LongMemEval` adapter in `BenchmarkAdapter`, `cognee eval --benchmark LongMemEval` just works with no extra wiring.

## The real problem I wanted to fix first

The addon wasn't actually optional. `evaluation/evaluator_adapters.py` imported `DeepEvalAdapter` at module load, and `deep_eval_adapter.py` imports `deepeval` at the top. So just importing the evaluator registry (or the `EvaluationExecutor`) pulled in `deepeval`, even on a run that only uses the DirectLLM engine. Everything else here builds on top of fixing that.

## What changed

**Making the addon optional**
- `EvaluatorAdapter` now resolves each engine lazily from an import path instead of importing the class up front. Importing the registry, or running the DirectLLM engine, no longer touches `deepeval`.
- If you pick DeepEval without the extra installed, you get a clear message telling you to run `pip install "cognee[eval]"` instead of a raw `ImportError`.
- Added an umbrella `eval` extra (evals deps + deepeval) so one install covers the whole harness. I checked that hatchling expands the self-referential extra correctly.

**One-command runner and CLI**
- `run_eval(config) -> EvalResult` runs corpus, answer, evaluate and dashboard in order and hands back the artifact paths and aggregate metrics, instead of leaving you to go hunting for files. The pipeline imports happen inside the call, so importing the runner (and `cognee eval --help`) needs no extras.
- `cognee eval` and `python -m cognee.eval_framework` share one set of flags that map onto `EvalConfig`. No Modal or Docker on this path.

```bash
cognee eval --benchmark HotPotQA --engine direct_llm --limit 5
python -m cognee.eval_framework --benchmark Dummy --no-dashboard
```

**Reproducibility**
- Added a `seed` config and actually threaded it through the corpus builder into the benchmark adapters. The adapters already accepted `seed`, it just was never passed down before.
- Added `results_dir`. Artifacts get namespaced under `<benchmark>_<engine>` and the resolved config is written next to them, so two runs don't clobber each other and you can compare them.

**Docs**: added `cognee/eval_framework/README.md`, a short "run a memory benchmark in one command" guide, in line with #3605.

## Tests

New `cognee/tests/unit/eval_framework/test_eval_runner.py`. These are deterministic, need no API keys, and pass even without the eval extra installed, which is the whole point of the decoupling:
- importing the registry does not import `deepeval` (checked in a clean subprocess)
- DirectLLM engine loads without the extra
- missing deepeval gives the actionable error
- flags map onto `EvalConfig` correctly
- artifact path resolution, both legacy flat and namespaced
- the runner wiring, on mocked steps, so no LLM or DB is needed

Result: 8 passed on their own, and the full `cognee/tests/unit/eval_framework/` suite is 59 passed with the extra installed. `ruff check` and `ruff format` are clean.

## Notes for reviewers

- The full corpus to answer path still needs a real LLM. Deterministic CI coverage of that end to end fits with the mocked-LLM harness from #3601 (PR #3640) once it lands. The logic this PR actually owns is unit tested without it.
- Backwards compatible. `run_eval.py` still runs, and when `results_dir` is unset the old flat artifact filenames are kept.

Closes the runner/CLI and optional-addon chunks of #3623.


rshkarin:
@Taimoo-r thanks for your contribution! Your changes will now go through our internal review process and will be merged afterward.

PR #3848, opened 2026-07-03
feat(benchmarks): Add Kafka streaming ingestion bridge and cross-SDK benchmark

### LINKED ISSUE #3635: Hackathon: Benchmark cognee (Python) under Kafka-fed streaming ingestion — cross-SDK with cognee-rs
## 🏆 Hackathon task: Benchmark cognee (Python) under Kafka-fed streaming ingestion

**Goal:** the Python-side companion to the cognee-rs Kafka benchmark. Feed a Kafka topic into cognee's ingestion (`add`/`cognify` / `remember`) and benchmark **ingestion throughput**, **end-to-end load times with the cloud LLM**, and **search latency as the dataset grows** — using the **same topic, corpus, and metrics** as the Rust ticket so results are directly comparable across SDKs.

> 🔗 **Paired with the Rust implementation + bench:** **topoteretes/cognee-rs#26**. Keep the producer, corpus, rate/volume matrix, and metric definitions identical so the two are a fair cross-SDK comparison.

### What to build

1. **Kafka → cognee bridge** — a consumer (e.g. `aiokafka`/`confluent-kafka`) reading the topic and feeding messages into cognee ingestion, with configurable batch size + concurrency and backpressure handling. Reuse the existing DLT/streaming ingestion paths where possible.
2. **Shared deterministic producer + corpus** — ideally the *same* synthetic corpus/producer as cognee-rs#26 (share a fixture) publishing at configurable rate/volume.
3. **Benchmark harness** measuring, separately:
   - **Throughput / performance:** sustained messages/sec topic→graph; backpressure behavior.
   - **Load times with cloud LLM:** end-to-end add+cognify latency per message/batch (p50/p95/p99) — dominated by cloud LLM round-trips; report with/without `improve` and across batch sizes.
   - **Search vs increasing data:** `search`/`recall` latency at growing corpus sizes (1k / 10k / 100k+) to chart search scaling with stream volume.
4. **Cross-SDK report:** present cognee (Python) vs cognee-rs (#26) on identical streams — throughput, latency percentiles, and the search-vs-scale curve.

### Requirements

- ✅ Reproducible: dockerized Kafka, fixed corpus + seeds (shared with cognee-rs#26); cloud-LLM cost called out.
- ✅ **Test coverage** — bridge + harness covered; deterministic smoke mode in CI via the mocked-LLM harness (#3601), full load run manual/opt-in.
- ✅ Results committed; metric definitions identical to cognee-rs#26.

### Good first chunks (claim one 👇)
- Kafka consumer → cognee ingestion bridge (do first)
- Shared producer/corpus fixture (coordinate with cognee-rs#26)
- Harness + metrics (throughput / latency percentiles / search-vs-scale)
- Cross-SDK comparison report

🔗 Rust counterpart: topoteretes/cognee-rs#26 · Related: #3601 (mocked tests), #3626 (DLT ingestion perf).

AliRana30:
### Descrption

The idea for this PR was to create something that runs side-by-side with `cognee-rs` so we can actually compare ingestion performance between the Python and Rust implementations using the same corpus and the same metrics.

### What I built

- **The Kafka consumer bridge** lives at `cognee/integrations/kafka/consumer.py`. It uses `aiokafka` to stay consistent with cognee's asyncio architecture and feeds messages into the existing `add()` and `cognify()` pipeline through the public API without touching any internals. Backpressure is handled through `asyncio.Semaphore` so there is a real concurrency ceiling rather than a sleep loop. Each batch returns a typed `BatchResult` dataclass with nanosecond timing broken down into add latency, cognify latency, and total latency separately.
- **The corpus generator** at `benchmarks/kafka/corpus.py` uses a fixed seed of 42 and a local `random.Random` instance rather than the global random module, so two calls always produce identical output regardless of what else ran before them. This matches what `cognee-rs` uses so the cross-SDK comparison is fair.
- **The benchmark harness** at `benchmarks/kafka/harness.py` measures sustained ingestion throughput, end-to-end latency percentiles at p50, p95 and p99 broken down by pipeline stage, and search latency at different corpus sizes. Results are written to a JSON file in `benchmarks/results/` with full metadata so runs are reproducible.
- **CI / Local Infrastructure**: Included a `docker-compose.yml` for spinning up Kafka locally, and a CI workflow that runs the harness in `--mock-llm` mode so the pipeline is tested on every push without needing any API keys.

### 1. Corpus Generator Determinism

Corpus generator producing deterministic documents. Running it twice gives the exact same output, confirming seed isolation.

### 2. Producer Dry Run

Producer dry run on 10 documents without opening any Kafka connection. This is what CI uses to validate the pipeline without infrastructure.

### 3. Integration Tests

All 6 integration tests passing on Python 3.12.10. No API keys were needed. The mock LLM harness handles the LLM layer so the tests focus purely on the bridge and pipeline logic.

### 4. Harness Execution (Mock LLM)

Harness running against a real local Kafka instance in `--mock-llm` mode. Cognee initialized correctly and the consumer connected and processed messages successfully.

### 5. Final Benchmark Artifact

Results file written to `benchmarks/results/kafka_benchmark_...json`.

*Note on Mock Mode:* The `#3601` `_MockVectorEngine` harness does not implement an `embedding_engine`. To gracefully handle this and ensure the JSON artifacts are cleanly generated in CI, the harness explicitly bypasses the `search_vs_scale` query loop and records `0` latencies when the `--mock-llm` flag is active. Real search benchmarking requires connecting a real LLM endpoint.

### What this does not change

The public API in `cognee/__init__.py` is untouched. The pipeline runner is untouched. The bridge calls `add()` and `cognify()` through the existing public interface. All existing tests still pass. This is purely additive.

I affirm that all code in this pull request conforms to the terms of the Topoteretes Developer Certificate of Origin.

### DIFF
--- a/.github/workflows/kafka-benchmark-ci.yml
+++ b/.github/workflows/kafka-benchmark-ci.yml
@@ -1,0 +1,77 @@
+name: Kafka Benchmark CI
+
+on:
+  push:
+    paths:
+      - 'benchmarks/kafka/**'
+      - 'cognee/integrations/kafka/**'
+      - '.github/workflows/kafka-benchmark-ci.yml'
+  pull_request:
+    paths:
+      - 'benchmarks/kafka/**'
+      - 'cognee/integrations/kafka/**'
+      - '.github/workflows/kafka-benchmark-ci.yml'
+
+permissions:
+  contents: read
+
+env:
+  COGNEE_SKIP_CONNECTION_TEST: 'true'
+  RUNTIME__LOG_LEVEL: ERROR
+  ENV: 'dev'
+  LLM_PROVIDER: openai
+  LLM_MODEL: gpt-4o-mini
+  LLM_ENDPOINT: https://api.openai.com/v1
+  LLM_API_KEY: stub-key-not-used-mocked
+  EMBEDDING_PROVIDER: openai
+  EMBEDDING_MODEL: text-embedding-3-small
+  EMBEDDING_DIMENSIONS: 300
+  EMBEDDING_API_KEY: stub-key-not-used-mocked
+
+jobs:
+  kafka-benchmark-e2e:
+    name: Kafka Benchmark E2E
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Cognee Setup
+        uses: ./.github/actions/cognee_setup
+        with:
+          python-version: '3.11.x'
+
+      - name: Start Kafka
+        run: |
+          docker compose -f benchmarks/kafka/docker-compose.yml up -d
+          echo "Waiting for Kafka to be healthy..."
+          for i in {1..15}; do
+            STATUS=$(docker inspect --format='{{.State.Health.Status}}' $(docker compose -f benchmarks/kafka/docker-compose.yml ps -q kafka) 2>/dev/null || echo "starting")
+            if [ "$STATUS" == "healthy" ]; then
+              echo "Kafka is healthy!"
+              break
+            fi
+            echo "Status: $STATUS. Waiting 2s..."
+            sleep 2
+          done
+
+      - name: Run Kafka Benchmark (Mocked LLM Mode)
+        run: |
+          uv run python benchmarks/kafka/harness.py \
+            --messages 10 \
+            --batch-size 5 \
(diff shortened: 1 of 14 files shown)


AliRana30:
@siillee From my side it's done. If you still think something needs to be improved or fixed. Then you can comment your opinion. Thanks

Issue #3526, opened 2026-06-27
[Bug]: forget() removes data/graph nodes but recall() can still surface deleted facts after re-cognify


garimapahwa:
## Summary

On the managed Cognee Cloud endpoint, after deleting a single data item via `POST /api/v1/forget` (with `dataId`) and re-running `POST /api/v1/cognify`, the deleted fact is confirmed gone from both the raw data list (`GET /datasets/{id}/data`) and the graph endpoint (`GET /datasets/{id}/graph`) — but `POST /api/v1/recall` (`GRAPH_COMPLETION`) continues to surface the deleted fact, sometimes indefinitely.

## Environment

- **Endpoint:** managed Cognee Cloud (`https://tenant-ae39df34-cbc2-4087-bcc6-71254933079b.aws.cognee.ai`)
- **Auth:** `X-Api-Key` header
- **Date observed:** 2026-06-26 / 2026-06-27

## Steps to reproduce

```bash
# 1. Remember two distinct facts into a shared dataset
curl -X POST "$BASE/api/v1/remember" -H "X-Api-Key: $KEY" \
  -F "data=@-;filename=f1.txt;type=text/plain" -F "datasetName=test_ds" <<< "My name is Pahwa."
curl -X POST "$BASE/api/v1/remember" -H "X-Api-Key: $KEY" \
  -F "data=@-;filename=f2.txt;type=text/plain" -F "datasetName=test_ds" <<< "I prefer Python over JavaScript."

# 2. Confirm recall combines both correctly
curl -X POST "$BASE/api/v1/recall" -H "X-Api-Key: $KEY" -H "Content-Type: application/json" \
  -d '{"query":"What is the user'"'"'s name and language preference?","datasets":["test_ds"],"searchType":"GRAPH_COMPLETION"}'
# => "Name: Pahwa / Language preference: Python" (correct)

# 3. Forget fact 1 specifically by its data_id
curl -X POST "$BASE/api/v1/forget" -H "X-Api-Key: $KEY" -H "Content-Type: application/json" \
  -d '{"dataset":"test_ds","dataId":"<fact-1-data-id>","memoryOnly":false}'
# => {"data_id":"...","dataset_id":"...","status":"success"}

# 4. Re-cognify
curl -X POST "$BASE/api/v1/cognify" -H "X-Api-Key: $KEY" -H "Content-Type: application/json" \
  -d '{"datasets":["test_ds"],"runInBackground":false}'

# 5. Recall again
[...]

PR #3854, opened 2026-07-04
fix: auto-recover DATASET_PROCESSING_ERRORED state before cognify re-run

### LINKED ISSUE #3853: [Bug]: POST /api/v1/cognify returns HTTP 500 RetryError[ProgrammingError] on Cloud tenant after forget() - dataset stuck in DATASET_PROCESSING_ERRORED
## Summary

On Cognee Cloud, calling `POST /api/v1/cognify` after a prior (failed or successful) run on the same dataset name consistently returns HTTP 500 with `RetryError[<Future state=finished raised ProgrammingError>]`. The dataset gets stuck in `DATASET_PROCESSING_ERRORED` state, which also causes cascading 409 errors on `recall CHUNKS` and 500 on `forget`.

## Minimum Viable Reproduction

```python
import httpx

BASE = 'https://<your-tenant>.aws.cognee.ai'
HEADERS = {
    'X-Api-Key': '<key>',
    'X-Tenant-Id': '<tenant>',
    'Content-Type': 'application/json',
}

# Works fine:
httpx.post(f'{BASE}/api/v1/add_text', headers=HEADERS, json={
    'text_data': ['Test sentence.'],
    'datasetName': 'my_dataset',
})

# Fails with HTTP 500 every time:
httpx.post(f'{BASE}/api/v1/cognify', headers=HEADERS, json={
    'datasets': ['my_dataset']
}, timeout=120)
```

## Error Response

```json
{
  "error": "Internal server error",
  "detail": "RetryError[<Future state=finished raised ProgrammingError>]"
}
```

## Cascade

After `cognify` 500:
- `recall CHUNKS` → HTTP 409 `{"error": "An error occurred during recall."}`
- `recall GRAPH_COMPLETION` → HTTP 200 (still works via typed memories)
- `forget` → HTTP 500

## Root Cause (observed)

Suspected related to the same race condition described in #3526: calling `forget` then `cognify` on the **same dataset name** within a short time window leaves the dataset stuck in an errored state. The `ProgrammingError` inside Cognee's graph processing pipeline suggests a DB cursor/connection issue on stale dataset state.

## Workaround

Using a **unique dataset name per session** (e.g. `f'dataset_{int(time.time())}'`) avoids the issue because it bypasses the stale state entirely.

## Environment

- Platform: Cognee Cloud (AWS-hosted tenant)
- API: REST v1
- Client: Python httpx 0.27 / requests
- Python: 3.12
- Reproducible: 100% when re-using dataset name across runs

AmanM006:
## Problem

When a `cognify` run fails (transient DB error, network issue, etc.), the `pipeline_runs` table retains a row with `status = DATASET_PROCESSING_ERRORED`. A subsequent `POST /api/v1/cognify` call on the **same dataset** then raises:

```
RetryError[<Future state=finished raised ProgrammingError>]
```

This is returned as HTTP 500 and the dataset is permanently stuck — every re-run on that dataset name fails with the same error. The only workaround is to use a completely different dataset name.

Fixes: #3853
Related: #3526

## Root Cause

`run_pipeline_per_dataset` in `pipeline.py` calls `check_pipeline_run_qualification` and then `run_tasks` directly. When the dataset status is `DATASET_PROCESSING_ERRORED`, `check_pipeline_run_qualification` returns `None` (the ERRORED case falls through without being handled), and `run_tasks` is invoked against a dataset whose stale state in the relational DB causes a `ProgrammingError` inside the async DB session.

## Fix

Add a status check for `DATASET_PROCESSING_ERRORED` in `run_pipeline_per_dataset` **before** `run_tasks` is called. When detected, call the existing `reset_pipeline_run_status()` helper (already used by `reset_dataset_pipeline_run_status` layer), which inserts a fresh `DATASET_PROCESSING_INITIATED` row and clears the stale state.

```python
task_status = await get_pipeline_status([dataset.id], pipeline_name)
if task_status.get(str(dataset.id)) == PipelineRunStatus.DATASET_PROCESSING_ERRORED:
    logger.warning(
        "Dataset %s has a previous errored run for pipeline '%s'. "
        "Auto-resetting status to allow re-run. (GitHub Issue #3853)",
        dataset.id,
        pipeline_name,
    )
    await reset_pipeline_run_status(user.id, dataset.id, pipeline_name)
```

## Why This Approach

- Uses **existing recovery infrastructure** — `reset_pipeline_run_status` and `get_pipeline_status` are already part of the codebase
- **Minimal surface area** — two new imports, one guard block added to one function
- **No behaviour change for healthy datasets** — the check adds one DB read (`get_pipeline_status`)
- **Idempotent** — safe to call multiple times; each call inserts a fresh INITIATED row that overrides the ERRORED status

## How to Reproduce (before fix)

1. Call `POST /api/v1/add_text` for a dataset
2. Call `POST /api/v1/cognify` and let it fail (or interrupt it)
3. Call `POST /api/v1/cognify` again on the same dataset → HTTP 500 `RetryError[ProgrammingError]`

After this fix, step 3 succeeds by auto-resetting the ERRORED state first.

### DIFF
--- a/cognee/modules/pipelines/operations/pipeline.py
+++ b/cognee/modules/pipelines/operations/pipeline.py
@@ -25,6 +25,9 @@ from cognee.modules.pipelines.layers.check_pipeline_run_qualification import (
 from cognee.modules.pipelines.models.PipelineRunInfo import (
     PipelineRunStarted,
 )
+from cognee.modules.pipelines.models import PipelineRunStatus
+from cognee.modules.pipelines.operations.get_pipeline_status import get_pipeline_status
+from cognee.modules.pipelines.methods import reset_pipeline_run_status
 from typing import Any
 
 logger = get_logger("cognee.pipeline")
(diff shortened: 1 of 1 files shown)


dexters1:
@AmanM006 can you try to reproduce this with the local SDK? If it is not reproducible in the local SDK the fix should be something else

PR #3872, opened 2026-07-04
fix: remove shell=True and prevent zombie subprocess in modal_eval_dashboard

### LINKED ISSUE #3856: [Bug]: Zombie subprocess and shell=True in modal_eval_dashboard.py
## Bug Description

modal_eval_dashboard.py spawns a background Streamlit process via `subprocess.Popen(cmd, shell=True)` but discards the return value. This creates two problems:

1. **Zombie process** — The spawned process is never tracked, waited on, or cleaned up. On process exit it becomes an orphaned zombie that persists until the parent terminates.
2. **Security: `shell=True`** — The command string is passed through the system shell, which is unnecessary here since it's a hardcoded string. If the command ever incorporates user-controlled input (e.g., via env vars or config), this becomes a command injection vector.

## Steps to Reproduce

1. Open `cognee/eval_framework/modal_eval_dashboard.py`
2. Look at line 41:
   `python
   subprocess.Popen(cmd, shell=True)
   `
3. Run any code path that invokes `serve_dashboard()` — a background Streamlit process is spawned with no tracking.

## Expected Behavior

- Store the `Popen` return value so the process can be monitored/cleaned up.
- Use `shell=False` (pass args as a list) since shell features aren't needed.
- Optionally call `.wait()` or use a context manager to prevent zombie processes.

## Actual Behavior

- The return value of `Popen` is discarded.
- The spawned Streamlit process runs untracked.
- On shutdown, the process becomes a zombie or orphan.
- `shell=True` is used despite no shell features being needed.

## Suggested Fix

`python
# Before (line 41):
subprocess.Popen(cmd, shell=True)

# After:
import shlex
cmd_args = shlex.split(cmd)
subprocess.Popen(cmd_args, shell=False)
`

Or if the process should be tracked:
`python
proc = subprocess.Popen(cmd_args, shell=False)
# ... later ...
proc.wait()
`

## Environment

- Cognee version: latest main branch
- Python: 3.x
- Platform: Any (Linux, macOS, Windows)

## Additional Context

`cognee/eval_framework/modal_eval_dashboard.py`, line 41

This is a low-severity issue in an eval dashboard (not user-facing API), but it sets a bad pattern. The `shell=True` usage is flagged by security linters (bandit, semgrep) and the zombie process can accumulate if the dashboard is started/stopped repeatedly.

rahulrao85:
## Summary

Fixes two issues in `modal_eval_dashboard.py`: unnecessary `shell=True` (security) and discarded Popen handle (zombie process).

## Changes

- **Security**: Replaced `subprocess.Popen(cmd, shell=True)` with `subprocess.Popen(shlex.split(cmd), shell=False)`. The command is a hardcoded string with no shell features needed (no pipes, redirects, or variable expansion). `shell=True` is flagged by bandit (B602) and semgrep.
- **Zombie prevention**: Using `shlex.split()` ensures clean argument handling.
- Added `import shlex` to support safe command splitting.

The command behavior is identical - only the execution mechanism changes.

Closes #3856


Vasilije1990:
Resolved the merge conflict by merging `dev` into this branch. The conflict came from #3900, which landed an equivalent `shell=False` fix on dev in the meantime — I kept your simpler one-line `subprocess.Popen(shlex.split(cmd), shell=False)` call, so this PR now also cleans up the redundant comments and unused `proc` reference that #3900 introduced.

Vasilije1990:
Handled already

Issue #3636, opened 2026-06-28
Hackathon: Ingest video data in cognee (Python) — propose a solution, best wins


Vasilije1990:
## 🏆 Hackathon task: Ingest video data in cognee (Python)

**Goal:** let cognee ingest **video** into queryable memory. cognee already ingests audio and images via the loader engine — video is the missing modality. This is the Python counterpart to **topoteretes/cognee-rs#27**; keep the approach aligned so both engines behave the same.

> 🔗 Rust counterpart: **topoteretes/cognee-rs#27**. 🖼️ See also the image-ingestion improvements ticket: **#3637**.

### What already exists (extend this — don't rebuild)

- Loader engine: `LoaderInterface` / `LoaderEngine` / `create_loader_engine`, with registered loaders in [`cognee/infrastructure/loaders/core/`](https://github.com/topoteretes/cognee/tree/main/cognee/infrastructure/loaders/core) (`text`, `csv`, `image`, `audio`) and `external/`. Registration in `supported_loaders.py`. **There is no `video_loader` yet** — a new loader slots in here.
- `audio_loader.py` → `LLMGateway.create_transcript(file_path)` (ASR). `image_loader.py` → `LLMGateway.transcribe_image(file_path)` (vision caption). **Both already exist**, which makes a combined video approach very feasible.
- Reference: [`examples/demos/multimedia_processing/`](https://github.com/topoteretes/cognee/tree/main/examples/demos/multimedia_processing) (audio+image) and `notebooks/cognee_multimedia_demo.ipynb`.

### Proposed solutions (pick / propose the best — best wins)

1. **Audio-track transcription (ASR).** ffmpeg-extract the audio track → `LLMGateway.create_transcript` → text pipeline. Cheapest; best for talking-head/lecture/meeting video.
2. **Keyframe sampling + vision captioning.** Sample frames (interval or scene-change) → `LLMGateway.transcribe_image` → ingest **timestamped captions**. Captures visual content.
3. **Combined, time-aligned.** Merge transcript (1) + keyframe captions (2) into one time-aligned document — richest memory. Very feasible since both LLMGateway methods already exist.
4. **Native multimodal embeddings.** Embed video segments directly into the vector store; raw-video similarity search without captioning. Most experimental.
5. **Pluggable external-preprocess hook.** A `video_loader` that delegates extraction to a configurable external tool (ffmpeg + ASR/vision provider), keeping core thin.

[...]

PR #3912, opened 2026-07-05
feat(loaders): add video ingestion via audio-track transcription

Akshats-git:
## Description

Related to #3636 

This adds a `video_loader` so cognee can ingest video, which was the one multimedia type it could not handle yet (it already does audio and images). This is the Python side of [cognee-rs#27](https://github.com/topoteretes/cognee-rs/issues/27).

I went with Solution 1 from the issue: transcribe the video's audio track and feed that text into the normal pipeline. The loader reuses the existing `LLMGateway.create_transcript` path that the audio loader already uses. The transcript is written out with `[HH:MM:SS]` markers on each segment, so the timing is part of the text and survives chunking. Because the output is just text, a video becomes a normal `TextDocument` and needs no new document type.

ffmpeg is optional, as the issue asked. `.mp4` and `.webm` are sent straight to the transcription endpoint and need no ffmpeg. Other containers (`.mov`, `.mkv`, `.avi`, `.m4v`) need ffmpeg to pull out the audio track first, and if it is missing the loader raises a clear error telling you how to install it (`pip install cognee[video]`).

I kept it aligned with cognee-rs: same loader name, same format set, text output and fail-fast on bad input. Timestamps are the one thing that runs ahead of the Rust transcriber, which does not expose segments yet. I documented the format so [cognee-rs#27](https://github.com/topoteretes/cognee-rs/issues/27) can match it.

Keyframe captioning (Solution 2) is intentionally left for a follow-up PR. The loader is structured so it layers on without rework.

## Acceptance Criteria

Requirements from #3636 and how they are met:

- New `video_loader` in the loader engine with extensions and MIME types, reusing the audio/image LLMGateway paths. 
- Heavy deps (ffmpeg) optional with graceful degradation when absent. (mp4/webm work without it; clear error otherwise).
- Deterministic tests over a tiny fixture clip with the mocked-LLM harness, no keys. (19 tests).
- Example and docs under `examples/demos/multimedia_processing/`.

**Proof it works:**

- 19 unit tests pass (detection, timestamp building, both load paths, degradation, fallback).
- I also ran it against a real 59s speech video with a real key. The direct mp4 path and the ffmpeg mkv path both produced the same timestamped transcript, and `remember` → `recall` returned a time-anchored summary.

## Type of Change

## Screenshots

image

## Pre-submission Checklist

## DCO Affirmation


rshkarin:
@Akshats-git thanks for your contribution! Your changes will now go through our internal review process and will be merged afterward.

PR #3928, opened 2026-07-06
feat(ontology): preserve ontology URIs + RDF/SPARQL export + RDF ingestion

Vasilije1990:
## Motivation

cognee currently *consumes* ontologies as extraction scaffolding: an entity that matches is tagged with an `ontology_valid` boolean, but the stable URI is flattened to a local label at persistence, and the memory graph can't be queried as RDF or linked out to other domains. This PR closes that gap for teams that maintain knowledge natively as RDF and want cognee as a **complementary agentic-memory/retrieval layer** over their RDF KB — while staying fully backward compatible.

Three composable phases:

### Phase 0 — URI preservation
- Add `DataPoint.ontology_uri: str | None = None`.
- `expand_with_nodes_and_edges` now threads the matched `AttachedOntologyNode.uri` onto the created `Entity`/`EntityType` instead of discarding it. The stable IRI survives end-to-end onto the persisted node.

### Phase 1 — RDF/SPARQL read surface (`cognee.modules.graph.rdf`)
- `graph_data_to_rdf(nodes, edges)` — pure builder that turns the property-graph tuples from `GraphDBInterface.get_graph_data` into an `rdflib.Graph`. Nodes emit under their preserved `ontology_uri`, or a minted `https://cognee.ai/graph/...` IRI when ungrounded (so the RDF stays well-formed, nothing is dropped). `is_a` resolves to `rdf:type` (individual→class) or `rdfs:subClassOf` (class→class); other relationships become predicate IRIs.
- `export_memory_graph_to_rdf` / `serialize_memory_graph` / `query_memory_graph_sparql` — async wrappers that read the live graph and let consumers run **SPARQL** over an RDF view, decoupled from any one graph engine's query language.

### Phase 2 — RDF ingestion (`cognee.modules.ontology.rdf_xml.rdf_ingest`)
- `build_datapoints_from_rdf(rdf_graph)` — parse an RDF T-Box + A-Box into `EntityType`/`Entity` nodes that **keep the verbatim external IRIs** (open-world; no fuzzy canonicalization into a local vocabulary). Node identity is derived from the IRI, so distinct IRIs stay distinct and re-ingestion is idempotent.
- `load_rdf_graph` (reuses the resolver's multi-format parsing) and `ingest_rdf` (thin persist wrapper over `add_data_points`).

## Compatibility
Additive: `ontology_uri` defaults to `None`; existing extraction/matching/persistence behavior is unchanged. No changes to identity for the existing text-extraction path.

## Tests
`cognee/tests/unit/modules/graph/test_rdf_roundtrip.py` — Phase 0 field + serialization; Phase 2 RDF→DataPoints (IRI preservation, labels, is_a, object properties); Phase 1 property-graph→RDF semantics + SPARQL; and a Phase 2→Phase 1 round-trip asserting external IRIs survive the trip through cognee datapoints and back to RDF.

- New tests: **5 passed**
- Existing ontology suite: **37 passed**
- DataPoint/expand/identity/ontology selection: **151 passed**
- `ruff format` + `ruff check` clean

## Not in this PR (discussed as later work)
RDFS/OWL RL **entailment** (Phase 3) — would run `owlrl` over the Phase-1 RDF view. Deferred pending whether reasoning is needed at query time vs. materialized once. Write-back of inferred triples to the caller's store is also out of scope.


lxobr (changes requested):
Nice work. I like that this moves us toward better RDF compatibility. I think we should probably tighten one part before merging, though: how we handle RDF predicates on edges. Right now the flow preserves node IRIs, but non-is_a predicates are turned into local relationship names and exported as cognee-generated predicate IRIs, so the exported graph loses the original RDF property vocabulary. We could  preserve RDF-ingested predicate IRIs through export, add a test with same-local-name predicates from different namespaces, and document the supported RDF round-trip scope in the PR description and RDF module docstrings. What do you think?

lxobr (approved):

PR #3962, opened 2026-07-08
fix(config): prevent silent embedding defaults

### LINKED ISSUE #3383: [hackathon] Simplify embedding configuration
Part of #3381.

## Problems (confirmed in `embeddings/config.py` + `get_embedding_engine.py`)
1. **Silently defaults to OpenAI when only the LLM is configured.** Two mechanisms:
   - Static defaults `embedding_provider="openai"`, `embedding_model="openai/text-embedding-3-large"` — if you set only LLM env vars, embeddings silently use OpenAI regardless of your LLM choice.
   - Key fallback in the engine factory: `api_key = embedding_api_key or llm_api_key` (with a confusing `custom`-provider tautology at `get_embedding_engine.py:120-121`). The embedder borrows the LLM's key when its own is unset.
   This is the documented "configure only one and the other defaults to OpenAI" footgun — it surfaces as confusing runtime behavior, not a clear error.
2. **`embedding_dimensions` resolution fails silently.** `model_post_init` probes fastembed then `litellm.model_cost`; on miss it logs a warning and uses `_FALLBACK_DIMENSIONS = 3072`. The code comment itself flags this as "the source of the silent `Vector(3072)` mismatch on every non-OpenAI embedder" — the wrong dimension only blows up on first vector write, not at config time.
3. **Dead-branch default:** `embedding_batch_size` both branches return 36 (lines 104-107).
4. **Naming drift:** the field is `embedding_max_completion_tokens`, but docs/people expect `EMBEDDING_MAX_TOKENS`.

## Proposed simplification
- **Make embedding provider derive from the LLM when truly unset, OR require it explicitly — but never silently pick OpenAI.** At minimum, when `LLM_PROVIDER` is non-OpenAI and no embedding provider is set, emit a clear warning/error rather than defaulting to OpenAI.
- **Fail loud on unknown dimensions:** if `embedding_dimensions` can't be resolved and isn't set, raise a config-time error naming the model, instead of defaulting to 3072.
- **Collapse the dead batch-size branch** to a single default.
- **Accept `EMBEDDING_MAX_TOKENS` as an alias** for discoverability.

## Acceptance
- Configuring a non-OpenAI LLM without embedding settings produces a clear message (not a silent OpenAI embedder).
- An unresolvable embedding model fails at config time with an actionable error, not on first write.

Venkat-Kolasani:
## Summary
- fail fast when embedding dimensions cannot be resolved without explicit EMBEDDING_DIMENSIONS
- reject the silent default OpenAI embedding path when LLM_PROVIDER is non-OpenAI and embeddings were not explicitly configured
- avoid borrowing a mismatched non-OpenAI LLM key for embeddings while preserving same-provider/default OpenAI key reuse
- add EMBEDDING_MAX_TOKENS as an alias and collapse the duplicate embedding batch-size default

Fixes #3383

## Notes
This keeps the default OpenAI quickstart working and preserves valid explicit embedding setups, including intentionally using OpenAI embeddings with a non-OpenAI LLM when the embedding model/key are configured explicitly.

## Tests
- uv run pytest cognee/tests/unit/infrastructure/databases/vector/test_embedding_config.py -q
- uv run pytest cognee/tests/unit/infrastructure/llm/test_embedding_connection_dimensions.py -q
- uv run ruff check cognee/infrastructure/databases/vector/embeddings cognee/tests/unit/infrastructure/databases/vector

## DCO

### DIFF
--- a/cognee/infrastructure/databases/vector/embeddings/config.py
+++ b/cognee/infrastructure/databases/vector/embeddings/config.py
@@ -1,19 +1,13 @@
 from typing import Optional
 from functools import lru_cache
+from pydantic import AliasChoices, Field
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
+from cognee.exceptions import CogneeConfigurationError
 from cognee.shared.logging_utils import get_logger
 
 
 logger = get_logger("embedding_config")
-
-
-# Hard fallback when neither litellm nor fastembed knows the model. This used
-# to be the unconditional default and was the source of the silent
-# "Vector(3072)" mismatch on every non-OpenAI-text-embedding-3-large embedder.
-# Keep it for back-compat (the OpenAI default model still resolves to this via
-# litellm), but log a warning when we hit it without a real lookup.
-_FALLBACK_DIMENSIONS = 3072
 
 
 def _resolve_embedding_dimensions(provider: Optional[str], model: Optional[str]) -> Optional[int]:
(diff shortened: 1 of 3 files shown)


Venkat-Kolasani:
Closing this draft in favor of a replacement PR from the same commits on branch fix-embedding-config-3383.

PR #4006, opened 2026-07-10
fix(ci): regenerate uv.lock for gmail extra + fix fastembed test on py3.14 (SDK-145)

rshkarin:
## What & why

Three independent failures were keeping the `dev` branch CI red. This PR fixes all three.

### 1. Stale `uv.lock` — `basic checks / Lockfile and Pre-commit Hooks`

The Gmail DLT connector (#3752) added a `gmail` extra to `pyproject.toml` but did not regenerate `uv.lock`, so `uv lock --check` fails:

```
error: The lockfile at `uv.lock` needs to be updated, but `--check` was provided.
```

Regenerated with **uv 0.11.28** — the version `astral-sh/setup-uv@v7` installs in CI — adding only the connector's Google API dependencies.

### 2. Unit tests abort on Python 3.14 — `OS and Python Tests Ubuntu / Unit tests 3.14.x`

`test_embedding_tokenizer_delegation.py` (from tokenizer PR #3762) imports `FastembedEmbeddingEngine` unconditionally, and that module raises `ImportError` at import time when `fastembed` is missing. fastembed reaches CI only via the `codegraph` extra, which is pinned `; python_version < '3.14'`, so on 3.14 the whole unit-test collection aborted:

```
ModuleNotFoundError: No module named 'fastembed'
!!! Interrupted: 1 error during collection !!!
```

### 3. `ty check .` fails — `Community Tests / Code Quality`

The type checker could not resolve the optional `fastembed` import in `resolver.py`, plus a real `str | None` argument-type error at the `MistralTokenizer` call.

## Fix

**fastembed stays an optional package extra** (no `pyproject.toml` change). It is installed in CI only in the jobs that import it, via the existing `extra-dependencies` plumbing — not by editing the shared `cognee_setup` action. The `fastembed` extra carries the `onnxruntime` version split (incl. the cp314 path), so it installs on 3.10–3.14:

- **`test_suites.yml`** `different-os-tests-basic`: `extra-dependencies: 'postgres'` → `'postgres fastembed'`. This is the bare-runner ubuntu unit-test matrix; only 3.14 lacked fastembed (3.10–3.13 get it transitively via `codegraph`).
- **`community_tests.yml`** `code-quality` (ty) and mocked unit-tests: `--extra fastembed` (these jobs roll their own `uv sync`, not `cognee_setup`).
- **`integration_tests.yml`** `integration-retrieval`: fastembed added to its `extra-dependencies`.

The engine's own `try/except` guard is unchanged, so end users who skip the extra still get a clear "install cognee[fastembed]" message.

Separately, `MistralTokenizer(model=bare)` passed `str | None` where `str` is required. Guarded the `None` case (nothing to load → default TikToken tokenizer), which also narrows the type.

## Verification

- `uv lock --check` passes with uv 0.11.28; the lock adds only the gmail connector's deps.
- `ty check` passes against an env with fastembed installed; the Mistral type error is fixed and the fastembed import resolves once installed (verified in a scratch venv).
- CI: the previously-red bare-runner jobs (`Unit tests 3.14.x`, `Code Quality`, `Unit Tests (Mocked)`) plus the lockfile check go green; `Integration - Retrieval` stays green.

Linear: SDK-145

---


.github/workflows/test_suites.yml
```diff
@@ -168,7 +168,9 @@ jobs:
       os: '["ubuntu-22.04"]'
       # postgres so the SQL cache adapter tests run instead of skipping; omitted
       # on macos/windows (extended) where psycopg2 has no prebuilt wheels.
-      extra-dependencies: 'postgres'
+      # fastembed so the 3.14 unit tests can import FastembedEmbeddingEngine;
+      # 3.10-3.13 already get it transitively via the codegraph extra.
+      extra-dependencies: 'postgres fastembed'
```
dexters1: Not needed here if it is an integration test

.github/workflows/community_tests.yml
```diff
@@ -77,8 +79,9 @@ jobs:
           enable-cache: true
 
       - name: Install dependencies
-        # postgres is needed so the SQL cache adapter tests run instead of skipping
-        run: uv sync --extra dev --extra postgres
+        # postgres is needed so the SQL cache adapter tests run instead of skipping;
+        # fastembed so the infrastructure/llm tests can import FastembedEmbeddingEngine.
+        run: uv sync --extra dev --extra postgres --extra fastembed
```
dexters1: Won't be needed here when the test in question gets moved to integration tests

.github/workflows/community_tests.yml
```diff
@@ -42,7 +42,9 @@ jobs:
         uses: pre-commit/action@v3.0.1
 
       - name: Install dependencies
-        run: uv sync --extra dev
+        # fastembed so `ty check` can resolve the optional import in
+        # FastembedEmbeddingEngine / the tokenizer resolver.
+        run: uv sync --extra dev --extra fastembed
```
dexters1: uhm not sure about this one, I would remove it here as well if it doesn't break the type check

.github/workflows/test_suites.yml
dexters1:
```suggestion
      extra-dependencies: 'postgres'
```

.github/workflows/community_tests.yml
dexters1:
```suggestion
        run: uv sync --extra dev --extra postgres
```

.github/workflows/community_tests.yml
dexters1:
```suggestion
        run: uv sync --extra dev
```

dexters1 (approved):

PR #4016, opened 2026-07-11
fix: partition DLT chunks in a single pass instead of O(n^2) scan

### LINKED ISSUE #4015: [Bug]: Chunk deduplication in extract_graph_from_data scans a list instead of a set, causing quadratic slowdown on large DLT ingests
### Bug Description

`extract_graph_from_data` filters out chunks already produced by DLT using Python list membership (`c not in dlt_chunks`) instead of a set lookup. Since `dlt_chunks` is a list of `DocumentChunk` (Pydantic) objects, each `in` check triggers a full linear scan with Pydantic `__eq__` comparisons. For DLT-based ingestion of large tabular sources, this makes chunk filtering scale as O(len(data_chunks) * len(dlt_chunks)) instead of O(N).

`cognee/tasks/graph/extract_graph_from_data.py:153-156`

    non_dlt_chunks = [c for c in data_chunks if c not in dlt_chunks]

### Steps to Reproduce

1. Configure a DLT-based ingestion source that produces a large number of row-chunks (e.g. a CSV/DB table with tens of thousands of rows), so `dlt_chunks` ends up large.
2. Run `cognee.add(...)` followed by `cognee.cognify(...)` on that dataset.
3. Profile (or time) the `extract_graph_from_data` task as chunk count grows — the filtering step's time cost grows quadratically with row/chunk count rather than linearly.

### Expected Behavior

Chunk filtering should be O(N) (linear in the number of chunks), independent of how it's implemented,
since it is logically a "remove already-processed items" operation with no reason to be quadratic.

### Actual Behavior

Filtering time grows proportionally to `len(data_chunks) * len(dlt_chunks)` because of the list-based
`not in` check, which becomes a measurable bottleneck on the extraction hot path for large DLT sources.

### Environment

- OS: N/A (source-level performance issue, reproducible on any platform)
- Python version: N/A
- Cognee version: current `main`/`dev` branch (verified 2026-07-11)
- LLM Provider: N/A (unrelated to this code path)
- Database: N/A (unrelated to this code path)

### Logs/Error Messages

```shell
N/A — this is not a crash/error, it's an algorithmic performance issue. No screenshots apply since
there's no visible UI or error output; the bug is only observable via profiling/timing chunk counts
against wall-clock time on the `extract_graph_from_data` task.
```

### Additional Context

Proposed fix: build a `set` of `dlt_chunks` identifiers up front and filter against that set instead
of the list, e.g.:

    dlt_chunk_ids = {c.id for c in dlt_chunks}
    non_dlt_chunks = [c for c in data_chunks if c.id not in dlt_chunk_ids]

This is a self-contained, low-risk fix isolated to one function — will open a PR shortly.

### Pre-submission Checklist

joaquinhuigomez:
`extract_graph_from_data` filtered non-DLT chunks with `c not in dlt_chunks`, rescanning the list for every chunk with full Pydantic `__eq__` comparisons — `O(len(data_chunks) * len(dlt_chunks))` on the extraction hot path for large DLT ingestions. This partitions both lists in a single pass over `data_chunks` using the same predicate: identical output, ~9,000× faster at 4k chunks in a micro-benchmark (2.75s → 0.3ms), and it avoids accidental value-equality exclusions between distinct chunks.

Adds unit tests covering the all-DLT short-circuit and the partition routing only non-DLT chunks to LLM extraction, alongside the existing suite (9 passed). Fixes #4015

### DIFF
--- a/cognee/tasks/graph/extract_graph_from_data.py
+++ b/cognee/tasks/graph/extract_graph_from_data.py
@@ -150,10 +150,16 @@ async def extract_graph_from_data(
     # deterministically by extract_dlt_fk_edges from schema metadata.
     from cognee.modules.data.processing.document_types import DltRowDocument
 
-    dlt_chunks = [
-        c for c in data_chunks if isinstance(getattr(c, "is_part_of", None), DltRowDocument)
-    ]
-    non_dlt_chunks = [c for c in data_chunks if c not in dlt_chunks]
+    # Partition in a single pass: a list-membership check against dlt_chunks
+    # rescans the list for every chunk (O(n^2) with Pydantic __eq__ comparisons),
+    # which becomes a bottleneck on the extraction hot path for large DLT sources.
+    dlt_chunks = []
+    non_dlt_chunks = []
+    for c in data_chunks:
+        if isinstance(getattr(c, "is_part_of", None), DltRowDocument):
+            dlt_chunks.append(c)
+        else:
+            non_dlt_chunks.append(c)
 
     if not non_dlt_chunks:
         return data_chunks
--- a/cognee/tests/unit/tasks/graph/test_extract_graph_from_data.py
+++ b/cognee/tests/unit/tasks/graph/test_extract_graph_from_data.py
@@ -179,3 +179,52 @@ async def test_non_knowledge_graph_model_unchanged():
 
     assert chunk.contains == custom_graph
     assert result == [chunk]
+
+
+@pytest.mark.asyncio
+async def test_all_dlt_chunks_short_circuits_llm_extraction():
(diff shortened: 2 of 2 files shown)


dexters1:
@joaquinhuigomez thank you for creating a PR for the issue! Your fix was merged along with this branch:

https://github.com/topoteretes/cognee/pull/4017

we had to make a new branch to run our CI against it and to change it to point to the dev branch, but your commit is a part of it and you will be listed as a contributor!

joaquinhuigomez:
Thanks @dexters1 — glad the fix landed, and the dev-branch retarget makes sense. Appreciate the contributor credit!

Issue #3613, opened 2026-06-28
Hackathon: Personal cross-platform memory bot (second brain) (propose the implementation)


Vasilije1990:
## 🏆 Hackathon task: Personal cross-platform memory bot ("second brain")

One memory graph reachable from multiple front-ends (e.g. Slack DM + Telegram + web) sharing the **same user's** dataset/session — capture anywhere, recall anywhere. Directly demonstrates cognee's persistence-across-sessions story.

### 📣 We want your proposal first

**Before writing code, propose the best implementation approach in a comment** so we can align on one design. Cover: how conversations map to cognee `session_id`/dataset and the memory boundary (per-user / per-channel / per-workspace); whether you build on the Python SDK, the TypeScript client [`@cognee/cognee-ts`](https://www.npmjs.com/package/@cognee/cognee-ts), or **MCP**; how/where ingestion runs (sync vs background); how replies surface **citations**; and the opt-in/opt-out + `forget` story. The clearest, simplest, most reusable proposal gets adopted.

### What cognee gives you

`remember` / `recall` / `forget` / `improve` with per-conversation **`session_id`** isolation; Python + TypeScript clients; MCP server; rankable/citable results.

### Scope
- [ ] User-identity → shared dataset/session mapping across 2+ transports.
- [ ] Unified `remember`/`recall` with citations regardless of entry point.
- [ ] `forget me` across all front-ends.

### Requirements

- ✅ Build on the shared chat-memory adapter (#3608) rather than reinventing session handling.
- ✅ **Test coverage** with the mocked-LLM harness (#3601) — deterministic, no real keys.
- ✅ **Easy onboarding** — copy-paste "run your own in 5 minutes" guide (#3605).
- ✅ Reply with **citations** to source messages (#3604), and support a `forget` command.

Related: #3608 (adapter core), #3604 (citable results), #3601 (mocked tests), #3605 (onboarding), #3603 (Integrations Hub).

Flux07-gezz:
Hey @Vasilije1990 ,
 I'd love to claim this one! Here is my implementation proposal for the cross-platform second brain:
Memory Boundaries (#3608): Dataset = User ID (shared across all front-ends so you capture anywhere, recall anywhere). The session_id tracks individual chat threads (Slack DM vs Telegram chat) to preserve local conversation flow under the same user graph.
Tech Stack: Python SDK combined with a lightweight unified routing layer to standardize inbound webhooks from Slack, Telegram, etc., into a single payload structure.
[...]

Issue #3610, opened 2026-06-28
Hackathon: Telegram bot powered by cognee memory (propose the implementation)


Vasilije1990:
## 🏆 Hackathon task: Telegram bot with cognee memory

A Telegram bot where each chat (DM or group) is a session: it `remember`s what's sent and answers via `/ask`. Great for personal capture — "forward me articles and links, ask me about them later." No OAuth, so it's an approachable first contributor bot.

### 📣 We want your proposal first

**Before writing code, propose the best implementation approach in a comment** so we can align on one design. Cover: how conversations map to cognee `session_id`/dataset and the memory boundary (per-user / per-channel / per-workspace); whether you build on the Python SDK, the TypeScript client [`@cognee/cognee-ts`](https://www.npmjs.com/package/@cognee/cognee-ts), or **MCP**; how/where ingestion runs (sync vs background); how replies surface **citations**; and the opt-in/opt-out + `forget` story. The clearest, simplest, most reusable proposal gets adopted.

### What cognee gives you

`remember` / `recall` / `forget` / `improve` with per-conversation **`session_id`** isolation; Python + TypeScript clients; MCP server; rankable/citable results.

### Scope
- [ ] Telegram bot (python-telegram-bot or grammY), per-chat session mapping.
- [ ] Ingest messages/forwarded content/links; `/ask` queries memory with citations.
- [ ] `/forget` and per-chat opt-out.

### Requirements

- ✅ Build on the shared chat-memory adapter (#3608) rather than reinventing session handling.
- ✅ **Test coverage** with the mocked-LLM harness (#3601) — deterministic, no real keys.
- ✅ **Easy onboarding** — copy-paste "run your own in 5 minutes" guide (#3605).
- ✅ Reply with **citations** to source messages (#3604), and support a `forget` command.

Related: #3608 (adapter core), #3604 (citable results), #3601 (mocked tests), #3605 (onboarding), #3603 (Integrations Hub).

Flux07-gezz:
Hey @Vasilije1990 , I'd love to jump on this one! Here is my implementation proposal for the Telegram Memory Bot:

Memory Boundaries (#3608): Dataset = Chat ID (works seamlessly for both private DMs and groups). The session_id maps directly to this Chat ID using the shared core adapter to maintain explicit chat isolation.

Tech Stack: Python SDK via python-telegram-bot (PTB) for its native async event/update loop processing.

[...]

Issue #3608, opened 2026-06-28
Hackathon: Shared chat-memory adapter core for bots (propose the design)


Vasilije1990:
## 🏆 Hackathon task: Shared "chat memory adapter" core for bots

**Goal:** a small, framework-agnostic layer that every cognee-powered bot (Slack, Telegram, Discord, …) plugs into, so each bot is thin (~100 lines) and they all share one consistent memory model. This is the **foundation** for the bot tickets — build it first.

### 📣 We want your proposal first

**Before coding, please propose the best implementation approach in a comment**, and let's converge on one. Specifically, make the case for:
- The **session-scoping convention** — how a platform conversation maps to a cognee `session_id` / dataset (e.g. `slack:{team}:{channel}:{thread}`), and the default memory boundary: per-user vs per-channel vs per-workspace.
- The **adapter API shape** — e.g. `ingest(message, conversation)` → background `remember`, `answer(query, conversation)` → `recall` with citations, `forget(conversation|user)`.
- Language/runtime: Python SDK vs the TypeScript client [`@cognee/cognee-ts`](https://www.npmjs.com/package/@cognee/cognee-ts) vs talking to cognee over **MCP** — and why.
- Where ingestion runs (sync vs background queue) and how opt-in/opt-out works.

The strongest proposal (clearest, simplest, most reusable across platforms) gets adopted as the base.

### What cognee gives you to build on

- The memory API: `remember` / `recall` / `forget` / `improve`, with **`session_id`** for per-conversation isolation (session memory + background sync to the graph).
- Clients: Python SDK, TypeScript client, and the **MCP** server.
- Rankable/citable results (see #3604) so replies can cite sources.

### Scope

- [ ] A documented adapter interface + reference implementation.
- [ ] Session/dataset mapping convention with configurable scoping.
- [ ] `ingest` / `answer` (with citations) / `forget` primitives.
- [ ] Opt-in/opt-out + a `forget me` path (privacy).
- [ ] At least one platform bot built on it as proof (coordinate with the Slack/Telegram/Discord tickets).

### Requirements

- ✅ **Test coverage** with the mocked-LLM harness from #3601 (deterministic, no real keys).
- ✅ **Easy onboarding** — copy-paste "build your own bot in 5 minutes" guide (consistent with #3605).
- ✅ Natural entry in the Integrations Hub (#3603).

Related: #3604 (citable results), #3601 (mocked tests), #3605 (calm onboarding), #3603 (Integrations Hub).

[...]

PR #4018, opened 2026-07-11
feat(integrations): shared chat-memory adapter core for bots (SDK-149)

rshkarin:
Rework of community PR #3754 (by @Taimoo-r) for GitHub hackathon issue #3608 — a shared **chat-memory adapter core** that every cognee bot (Slack, Telegram, Discord, a personal "second brain") plugs into, so each bot stays ~100 lines and they all share one memory model over `remember` / `recall` / `forget`. This is the foundation the bot tickets build on (#3610, #3613, and the web-widget bot).

Linear: SDK-149.

## What the adapter provides

- **Two-axis `Scope { dataset, session }`** — `dataset` is the durable graph + recall + forget boundary (what is shared); `session` is the conversation identity. One adapter serves both a per-channel team bot (`dataset = chat:{platform}:{workspace}:{channel}`) and a per-user cross-transport brain (`dataset = brain:{user}`); the collapsed case is just `dataset == session`.
- **Three primitives** — `ingest` (background remember), `answer` (recall with citations), `forget` (whole-scope or per-user "forget me").
- **Consent gate** — deny-until-opt-in in groups, allow in DMs, explicit override always wins.
- **Pluggable backend** — `CogneeMemoryBackend` (in-process SDK reference) behind a four-method `MemoryBackend`; a dependency-free `InMemoryMemoryBackend` runs the whole contract with no LLM/DB/keys.
- **Proof + onboarding** — a keyless console bot, a ~100-line Telegram bot, a "build your own bot in 5 minutes" guide, and an Integrations Hub entry.

## Rework

The original commit is preserved verbatim as the base (authorship intact). Every change below lands as a separate, single-concern commit on top.

**Correctness — the reference backend was non-functional.** It passed `session_id` to `remember()`, which routes into cognee's session-cache-only branch. That branch stringified the `DataItem` to the literal `"[DataItem]"` (losing the message text, `external_metadata`, and `data_id`) and created no `Data` row — so every message was stored as `"[DataItem]"`, per-user "forget me" was a silent no-op, and citations never resolved. All of it was masked by the `InMemoryMemoryBackend` the tests run against. Fixes:

- ingest through the durable `add()` + `cognify()` path (`run_in_background=True` keeps it fire-and-forget) so the text, the `external_metadata` stamp, and the deterministic `data_id` actually persist onto the `Data` row.
- per-user forget resolves the user's rows by `external_metadata["user"]` and *full*-deletes them (not `memory_only`, so the raw record goes, not just its graph projection).
- citations resolve via the `data_id` cognee reports in recall provenance, looked up against the Data-row stamp; they degrade to text-only when a result carries none.
- `_stamp` merges free-form `message.metadata` before the identity keys so a stray extra can't overwrite the resolved author; `answer` falls back to a non-empty snippet; scoping rejects a conversation whose identity segment sanitizes to empty (it was silently aliasing distinct conversations into one dataset).

**Simplification — less code, same abstractions.** Merged the field-identical `RecalledItem` into `Citation`; dropped `per_workspace_scope` (the issue asks only for per-user / per-channel; a bot needing more writes a one-line strategy); dropped the dead `Conversation.metadata` and the unused `Citation.metadata`; trimmed over-broad exports and the two-axis rationale that was repeated near-verbatim across several docstrings. The core (non-test) code is smaller than the original despite the correctness rework.

**Tests / conventions.** Added fast, keyless wire-contract tests for the real backend (they fail against the old session-path / `memory_only` code); package `__init__.py` for the new test dirs, matching the `tests/unit/api` subtree; corrected two inaccurate docstrings.

## Verification

- Keyless unit suite: **32 passed** — the adapter against `InMemoryMemoryBackend`, plus the new `CogneeMemoryBackend` wire-contract tests.
- Keyless real-cognee end-to-end: `add` stamped `DataItem`s → `list_data` sees `external_metadata` → `forget_user` removes only the target user's rows, others survive.
- Console proof bot runs the full contract (consent → ingest → answer-with-citations → forget-me) with no keys; `ruff` check + format clean.

## DCO Affirmation

### DIFF
--- a/cognee/integrations/__init__.py
+++ b/cognee/integrations/__init__.py
@@ -1,0 +1,14 @@
+"""First-party integrations built on top of the cognee memory API.
+
+Sub-packages here are thin, framework-agnostic layers that map an external
+system onto cognee's ``remember`` / ``recall`` / ``forget`` primitives. They
+never reach past those primitives into cognee internals, so they stay small
+and stable across cognee releases.
+
+Currently ships:
+
+* :mod:`cognee.integrations.chat_memory` — a shared "chat memory adapter"
+  core that every cognee-powered chat bot (Slack, Telegram, Discord, …)
+  plugs into, so each bot stays thin (~100 lines) and they all share one
+  consistent memory model.
+"""
--- a/cognee/integrations/chat_memory/README.md
+++ b/cognee/integrations/chat_memory/README.md
@@ -1,0 +1,123 @@
+# Chat Memory Adapter
+
+A small, framework-agnostic layer that every cognee-powered chat bot (Slack,
+Telegram, Discord, a personal "second brain", and so on) plugs into. Each bot
+stays thin, around 100 lines, and they all share one consistent memory model
+built on cognee's `remember` / `recall` / `forget` primitives.
+
+```
+platform event  ->  Conversation + Message  ->  ChatMemoryAdapter  ->  cognee
+                                                (ingest / answer / forget)
+```
+
+## Conversation scoping
+
+A conversation maps to a [`Scope`](./models.py) that holds two keys, matching
+cognee's two storage knobs:
+
+| key | what it is | used for |
+| --- | --- | --- |
+| `dataset` | durable graph + recall boundary; the unit `forget` wipes | what is shared |
+| `session` | fast per-conversation cache | what is recent |
+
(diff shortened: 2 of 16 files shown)


rshkarin:
Migrated to the cognee-integrations monorepo: **topoteretes/cognee-integrations#242** (`integrations/chat-memory/`).

PR #4019, opened 2026-07-11
feat(slack): Slack bot powered by cognee memory (SDK-150)

### LINKED ISSUE #3609: Hackathon: Slack bot powered by cognee memory (propose the implementation)
## 🏆 Hackathon task: Slack bot with cognee memory behind the scenes

A Slack app (Bolt) that silently ingests messages/threads from channels it's invited to into cognee, and answers questions like `@cognee what did we decide about X?` or a `/recall` slash command — **citing the source messages**. Bonus: thread summarization and a "catch me up" command using graph recall across the channel's history.

### 📣 We want your proposal first

**Before writing code, propose the best implementation approach in a comment** so we can align on one design. Cover: how conversations map to cognee `session_id`/dataset and the memory boundary (per-user / per-channel / per-workspace); whether you build on the Python SDK, the TypeScript client [`@cognee/cognee-ts`](https://www.npmjs.com/package/@cognee/cognee-ts), or **MCP**; how/where ingestion runs (sync vs background); how replies surface **citations**; and the opt-in/opt-out + `forget` story. The clearest, simplest, most reusable proposal gets adopted.

### What cognee gives you

`remember` / `recall` / `forget` / `improve` with per-conversation **`session_id`** isolation; Python + TypeScript clients; MCP server; rankable/citable results.

### Scope
- [ ] Slack app (events API + slash command), invited-channels-only ingestion.
- [ ] Map team/channel/thread → session; `remember` on message, `recall` on ask.
- [ ] Cited answers; "catch me up" / thread summary; `forget` + opt-out.

### Requirements

- ✅ Build on the shared chat-memory adapter (#3608) rather than reinventing session handling.
- ✅ **Test coverage** with the mocked-LLM harness (#3601) — deterministic, no real keys.
- ✅ **Easy onboarding** — copy-paste "run your own in 5 minutes" guide (#3605).
- ✅ Reply with **citations** to source messages (#3604), and support a `forget` command.

Related: #3608 (adapter core), #3604 (citable results), #3601 (mocked tests), #3605 (onboarding), #3603 (Integrations Hub).

rshkarin:
Review + rework of the community Slack bot (#3845, by @Yashsainani123) for hackathon issue #3609 — "Slack bot powered by cognee memory". Closes #3609.

A [Slack Bolt](https://slack.dev/bolt-python/) app (Socket Mode — no public URL) where **each Slack channel is a cognee memory**: it silently ingests opted-in channels' messages, answers `@cognee …` mentions and a `/recall …` slash command from that channel's memory **with clickable citations back to the source Slack messages**, and supports `/cognee-forget` + per-channel `/cognee-optin` / `/cognee-optout`. Self-contained under `examples/slack_cognee_bot/`; the only changes outside `examples/` are additive — the optional `slack` extra in `pyproject.toml`, its `uv.lock` entry (slack-bolt + slack-sdk only), and a documented env block in `.env.template`. No cognee core logic is touched.

## Design (verified against cognee `dev`)

The hard part is **citations to source messages**. cognee strips arbitrary metadata from the vector payload on `search()` — the `DocumentChunk_text` collection stores an `IndexSchema` that excludes `metadata` and nested models — so `channel`/`ts`/`permalink` cannot ride through a search. The only per-message identifier that survives is `document_id`, and the original author correctly exploits that: each message is ingested with a controllable `DataItem.data_id = uuid5("channel:ts")`, which cognee reuses as `Data.id` → `Document.id` → `DocumentChunk.document_id`. A `CHUNKS` search returns that `document_id`, which is joined back to the message's permalink/author via a small adapter-owned index. The whole chain (`data_id` → `document_id`, and metadata-stripping) was verified against the real cognee source before any change. Ingestion is batched (a full `cognify()` per message would be prohibitively expensive) with an on-demand flush before every answer, so a question reflects everything said so far. `forget` is dataset-level (cognee has no per-user delete), and the reply says so plainly.

The bot talks to cognee through a thin local `ChatMemory` adapter (`ingest`/`flush`/`answer`/`forget`) — the shared chat-memory adapter pattern from #3608, implemented locally until that core lands; only the concrete adapter swaps when it does.

## Rework (maintainer)

@Yashsainani123's original **7 commits are preserved verbatim** (authorship + DCO sign-off intact); the design was verified sound before any change. Improvements are layered on top as separate single-concern commits.

**Correctness fixes**
- **fix:** reply gracefully instead of failing silently — asking in a channel with nothing cognified yet raised `DatasetNotFoundError`/`NoDataError` with no reply; `/cognee-forget` on a never-used channel raised `AttributeError`; the answer path had no catch for transient LLM/rate-limit errors. All three now degrade to a calm reply.
- **fix:** stop ingesting the bot's own messages and @mentions — Slack delivers a plain `message` event alongside every `app_mention`, so the question was being fed back into channel memory. `bot_user_id` (previously accepted but never supplied, so the own-message guard was dead) is now sourced from Bolt's `context`.
- **fix:** answer inside the thread when asked in one, and ignore empty `@cognee` / `/recall` (no empty-query search, no needless cognify).
- **fix:** serialize `cognify` per channel so a size-triggered flush and an on-demand flush never run on the same dataset at once (overlapping cognify can corrupt/deadlock the store).

**Simplification (removing what issue #3609 doesn't need, without sacrificing abstractions)**
- **refactor:** drop the out-of-scope time-interval flush trigger (freshness is covered by the on-demand flush; cost by the size threshold).
- **refactor:** replace the SQLite + `threading.Lock` citation index with a bounded in-memory map folded into the adapter — an example needs the `document_id` join, not cross-restart persistence; a missing row already degrades to a plain-text citation (noted in the README).
- **refactor:** pass the cognify batch size directly instead of a one-field settings object; drop the unused `session_id` property and the redundant display-level dedupe.
- **refactor:** package-relative imports and a fixed 5-minute run command (`python -m src`).
- **test:** drop tests that assert Python's `abc`/dataclass behaviour; assert `top_k` is forwarded; make the mocked `CHUNKS` payload faithful to the real `IndexSchema`.

Net: production code (`src/`) goes **1182 → 1097 lines** and 9 → 8 modules *while adding* the fixes above; every real abstraction is kept — the `#3608` `ChatMemory` seam, the batch + on-demand buffer, the `document_id` citation join, the access-control result-shape normalization, the `forget`/opt-in/opt-out commands, and the Block Kit renderer's real Slack 3000-char / source-cap guards.

## Verification

- Keyless mocked suite: **69 passed** (Slack SDK and every cognee call mocked; no sockets, no keys). `ruff check` + `ruff format` clean.
- Offline: `python -m src` composes the full object graph and reaches the actionable "install the slack extra" guard (proves the wiring imports and relative imports resolve).
- The `data_id → document_id` citation chain and the metadata-stripping behaviour were verified against the cognee source in this tree.

## Scope / limitations (deliberate)

- Answers reflect the last flush, not every message instantly (batch design, documented).
- Citations aren't persisted across restarts (in-memory map; degrades to plain text until re-ingested).
- `forget` is channel/dataset-level; per-user "forget me" isn't supported by cognee (noted as a follow-up).
- Citations are the channel's most relevant messages to the question (a separate `CHUNKS` search), so they may include context the prose answer didn't directly quote.
- Out of scope (issue bonus, not built): "catch me up" / thread summarization; other transports.

## Type of change
- [x] New feature (non-breaking; example + additive packaging only)

## DCO Affirmation


rshkarin:
Migrated to the cognee-integrations monorepo: **topoteretes/cognee-integrations#244** (`integrations/slack/`).

PR #4027, opened 2026-07-11
feat(vector): add Turso (libSQL) vector database adapter (SDK-152)

### LINKED ISSUE #3583: [hackathon] Add Turso as a vector database option

rshkarin:
Continues #3639 (by @therealharshit) for hackathon issue #3583. The original contributor's commits are preserved as the branch base; maintainer commits are added on top.

## What it does

Adds Turso / libSQL as a selectable vector backend (`VECTOR_DB_PROVIDER=turso`), directly mirroring how the PGVector adapter was added:

- **`TursoVectorAdapter(VectorDBInterface)`** — each collection is a libSQL table with an `F32_BLOB` vector column; similarity via `vector_distance_cos`; PGVector's `belongs_to_set` (NodeSet) OR/AND semantics reproduced with SQLite JSON1 instead of Postgres JSONB. Works both embedded (local file) and against a remote Turso cloud DB.
- **`create_vector_engine` routing** — `VECTOR_DB_PROVIDER=turso` routes to the adapter with a guarded `cognee[turso]` ImportError.
- **Multi-user** — `TursoVectorDatasetDatabaseHandler` gives each dataset its own libSQL file, so the existing permission system isolates users under `ENABLE_BACKEND_ACCESS_CONTROL`. Auto-mapped in `VectorConfig`.
- **Cache** — rides the existing `_create_vector_engine` LRU unchanged (no new config).
- **Env-vars only** — `.env.template` block + `fill_derived` auto-mapping make `VECTOR_DB_PROVIDER=turso` sufficient.

libSQL's only client with embedded-file vector support (`libsql-experimental`) is synchronous, so DB calls run through `asyncio.to_thread` behind a single lock, isolated to `_get_connection` / `_run` so a future native-async client is a one-spot swap.

## Maintainer commits on top of the preserved base

1. **`fix(turso)`: threading.Lock for cross-event-loop safety** — the adapter is cached process-globally, so its single `asyncio.Lock` outlived the loop that first contended it and crashed the common add → cognify → search flow (separate `asyncio.run()` calls) with "bound to a different event loop". Serialize the shared sync connection with a `threading.Lock` instead — loop-agnostic, exactly as LanceDB does and documents.
2. **`fix(turso)`: evict cached engines + remove the file on dataset delete** — the old open-and-`prune()` left one `{dataset_id}.turso.db` and an open connection per deleted dataset (DROP TABLE never removes the file). Mirror the current LanceDB handler: evict cached engines (awaiting their closes), then remove the on-disk store — a single file, so `remove` not `remove_all`.
3. **`refactor(turso)`: drop the in-Python batch dedup** — the per-row `ON CONFLICT DO UPDATE` already merges in-batch duplicates in SQL (last payload wins, `belongs_to_set` unioned, first vector kept). PGVector needs the pre-dedup only because it uses a single multi-row VALUES statement; this adapter does not. Verified byte-identical.
4. **`refactor(turso)`: flatten the no-op `_run_many` batch loop** — single-row executes made the `QUERY_BATCH_SIZE` chunking pointless there (still used for `IN(...)` lists in `retrieve`/`delete`).
5. **`perf(turso)`: skip the payload column in `search` when `include_payload=False`** — mirrors PGVector; the graph/RAG hot path fetches only id + distance.
6. **`fix(turso)`: surface write failures in `remove_belongs_to_set_tags`** — one broad `except → debug` hid UPDATE/DELETE errors; keep the quiet skip only for the "not a vector collection" probe and log real write failures at warning.
7. **`test(turso)`: pin cosine ranking + in-batch merge; remove orphaned `test_turso.py`** — the ordering assertion was tautological; add a real ranking + an in-batch-merge test. `test_turso.py` was run by no CI job and its coverage is a strict subset of the wired-in suites, so it only rots silently.
8. **`build(turso)`: regen `uv.lock` for the turso extra, align pyproject with #4077** — the missing lock regen made `uv lock --check` fail and CI silently skip installing `libsql-experimental` (the Turso job's `ModuleNotFoundError`). The extra now sits on the exact line #4077 adds, so whichever PR merges first the other still merges to a single `turso` key.
9. **`fix(turso)`: probe `libsql-experimental` directly** — the guarded import wrapped the adapter module, but the driver is imported lazily, so the actionable "install cognee[turso]" message never fired. Probe the driver itself (same pattern as #4077's `create_relational_engine`) and drop the untested `import libsql` fallback.
10. **`ci(turso)`: bake the turso extra into the CI image, key the image on `Dockerfile.ci`** — containerized jobs install nothing per-job (`cognee_setup`'s container path ignores `extra-dependencies`; extras must be baked into the image, which is how postgres gets its extra). Without it the adapter unit tests silently *skipped* and the e2e failed: the startup id-scheme migration hit the missing-driver ImportError while creating the dataset's vector engine and blocked all writes with `MigrationError`. Also include `Dockerfile.ci` in the image-tag hash — the build is skipped when the tag exists, so Dockerfile-only changes (like this one) were previously invisible for already-built dep hashes.

## Alignment with the sibling Turso backends (#4077 relational, #4080 graph)

- **Same package, same extra, same pin:** all Turso backends use `turso = ["libsql-experimental>=0.0.55,<0.1"]`; both PRs that need it add the identical pyproject line at the identical location, and both regenerate `uv.lock`.
- **Same missing-dependency UX:** identical probe pattern and identical ImportError message.
- **Why the query path differs deliberately:** #4077/#4080 read the libSQL file through SQLAlchemy + aiosqlite because a libSQL file is a SQLite file. The vector adapter cannot: `vector32()` / `vector_distance_cos()` are libSQL-engine builtins that plain SQLite lacks (verified: `sqlite3` opens the same file fine but raises `no such function: vector32`), and issue #3583 explicitly requires libSQL's native vector type + built-in similarity. Hence `libsql-experimental` on the query path, per the issue.
- **No registry collisions:** vector registers dataset handler key `"turso"`, graph (#4080) registers `"turso_graph"` — mirroring the existing `pgvector` / `postgres_graph` split.
- Remote mode follows the issue's env-var contract (`VECTOR_DB_URL` + `VECTOR_DB_KEY`, direct connection, author-tested against Turso cloud); #4077's embedded-replica model (`DB_TURSO_URL`/`DB_TURSO_AUTH_TOKEN` sync) is specific to the relational engine's aiosqlite query path.

## Testing

- **13 adapter unit tests** against a real embedded libSQL DB (no LLM, no network): CRUD, cosine ordering, NodeSet OR/AND, `belongs_to_set` merge, in-batch dedup, tag removal (incl. untagged-row preservation), delete, batch, prune — all green.
- CI's vector-DB suite runs the full search-type e2e (`test_search_db.py`) and the multi-user permissions suite (`test_permissions.py`) against `VECTOR_DB_PROVIDER=turso` (jobs run under the default `databases: all`).
- `ruff check .`, `ruff format --check .`, and `ty check .` are green on the whole repo.

Closes #3583.

---


therealharshit:
@rshkarin Thanks for continuing this work! I appreciate you building on my earlier commits and adding the required changes.

PR #4033, opened 2026-07-12
fix(kuzu): reclaim stale Ladybug lock left by a dead worker (#3708)

### LINKED ISSUE #3708: [Bug]: Kuzu worker crash leaves orphaned database lock, blocking all subsequent recall() calls
### Describe the bug

When calling `cognee.recall()` or any operation that accesses the Kuzu graph database, a worker process crashes and leaves an orphaned lock on the graph database file. This blocks ALL subsequent `recall()` calls with:

```
RuntimeError: IO exception: Could not set lock on file:
/.../cognee/.cognee_system/databases/cognee_graph_ladybug 
(Lock is held by PID XXXXX)
```

### To Reproduce

1. Install cognee v1.2.2
2. Ingest some data (e.g., via `cognee.add()` or `cognee.remember()`)
3. Call `cognee.recall("any query")`
4. Observe: first call may work, but if a worker crashes (e.g., due to LLM API key issue or transient error), all subsequent calls fail with the lock error
5. The PID holding the lock is a dead/zombie worker process spawned by `cognee_db_workers/kuzu_worker.py`

### Error trace

```
File "cognee_db_workers/harness.py", line 411, in run_worker_loop
    result = handler(registry, msg)
File "cognee_db_workers/kuzu_worker.py", line 61, in _open_database
    db = ladybug.Database(**req.kwargs)
File "ladybug/database.py", line 136, in __init__
    self.init_database()
...
RuntimeError: IO exception: Could not set lock on file:
  .../cognee_graph_ladybug (Lock is held by PID XXXXX)
```

### Environment

- **cognee version**: 1.2.2
- **Python version**: 3.11.15
- **OS**: Linux 6.17.0-1016-oracle (Oracle ARM64, Ubuntu 24.04)
- **Database backend**: Kuzu (ladybug) + LanceDB + SQLite
- **LLM**: DeepSeek deepseek-chat
- **Embedding**: Local sentence-transformers all-MiniLM-L6-v2 (via monkey-patched `LiteLLMEmbeddingEngine.embed_text`)

### Root cause analysis

The worker subprocess architecture in `cognee_db_workers/kuzu_worker.py` does not appear to have proper crash recovery:

1. Worker process crashes (e.g., due to LLM API key error)
2. The Kuzu/ladybug database lock is NOT released on crash
3. The harness retries by spawning a new worker
4. New worker hits the same lock → crashes again
5. Repeat indefinitely — no self-healing mechanism

The lock file is a single file (not a directory):
```
$ file cognee_graph_ladybug
cognee_graph_ladybug: data
```

### Workaround

- Manually `kill -9 <PID>` the dead worker process before each recall call
- Or avoid `cognee.recall()` entirely and query the SQLite/LanceDB stores directly

### Suggested fix

1. Add a lock file watchdog/timeout in the worker harness (`harness.py`) — if a lock is held for more than N seconds by a process that no longer exists, force-release it
2. Wrap `_open_database()` in `kuzu_worker.py` with proper exception handling that releases the lock on any error path
3. Consider using Kuzu's read-only mode for recall queries to avoid lock contention entirely

abankar1:
## Description

Fixes #3708.

When a Ladybug/Kuzu worker process crashes (the reporter hit it via an LLM/embedding error mid-open, but OOM or SIGKILL do the same), it can leave a `<db_path>.lock` file behind on disk. After that, every attempt to open the graph DB fails with:

```
RuntimeError: IO exception: Could not set lock on file:
  .../cognee_graph_ladybug (Lock is held by PID XXXXX)
```

for a PID that is already dead. The worker already has a backoff-retry (`_retry_open_locked`), but that only helps the *transient* case where a still-live sibling worker is finishing its shutdown — it waits and re-tries the same open. Nothing ever removes an **orphaned** lock, so once a worker dies uncleanly the DB stays wedged and every later `recall()` fails until someone deletes the lock file by hand.

I added a reclamation fallback that runs only after the existing backoff retry is exhausted, so the transient path is completely unchanged. The new step:

1. Parses the holder PID out of the error message.
2. Checks whether that PID is still alive (`os.kill(pid, 0)`, mapped: `ProcessLookupError` → dead, `PermissionError`/other `OSError` → treat as alive).
3. **Only if the holder is confirmed dead**, removes `<db_path>.lock` (the same lock-file path `ladybug_migrate.py` already uses) and reopens once.

The safety bias is deliberate — a live holder, an unparseable PID, our own PID, or a platform that can't answer liveness (e.g. Windows, where signal 0 isn't supported) all leave the lock untouched. We never yank a lock out from under a healthy worker; the worst case is the pre-existing behavior (the lock error propagates).

I kept `kuzu_worker.py` stdlib-only (just added `import re`) so the no-cognee-import invariant that `test_worker_import_hygiene.py` enforces still holds.

## Acceptance Criteria

- A lock orphaned by a dead PID is removed and the open succeeds on retry, instead of wedging all subsequent opens.
- A lock held by a live PID (or an unknown/unparseable holder) is never removed.
- Transient contention with a live sibling worker keeps its existing backoff-retry behavior.

Proof — new regression tests, all passing locally (13 in the new file; the existing `test_kuzu_worker_open_retry.py` and `test_worker_import_hygiene.py` still pass):

```
TestPidAlive:  current-alive / nonpositive-conservative / lookup-error-dead / permission-error-alive
TestReclaimStaleLock:  dead-holder-removed / alive-holder-kept / own-pid-kept /
                       no-pid-kept / missing-lock-file / empty-path
TestOpenLockedWithRecovery:  reopen-after-reclaim / raises-when-not-reclaimable /
                             non-lock-error-not-reclaimed

13 passed  (test_kuzu_worker_stale_lock.py)
20 passed  (+ test_kuzu_worker_open_retry.py + test_worker_import_hygiene.py)
```

## Type of Change

## Pre-submission Checklist
- [x] My code follows the project's coding standards and style guidelines (ruff format + ruff check pass)

## DCO Affirmation


abankar1:
Heads-up for triage: #3708 was closed as resolved by #3720. That PR reworked the async graph/vector engine lifecycle (`closing_lru_cache` / stale-pin handling) and reduces *transient* lock contention between live workers.

This PR is narrower and complementary to that: it doesn't touch the engine lifecycle, only the worker-side open path. #3720 doesn't reclaim a Ladybug `.lock` file that a **hard-crashed worker PID** (OOM / SIGKILL) leaves on disk — the current `_retry_open_locked` only backs off, on the assumption the OS frees the lock when the holder dies, which doesn't hold for a persisted lock file. This adds stale-lock reclamation for that residual case, and only when the holder PID is confirmed dead (a live/unknown holder is never touched).

Totally fine to close if you consider that scenario out of scope now that #3720 has landed — just flagging that it's a different code path so it isn't mistaken for a dup.

Issue #3691, opened 2026-06-29
cognee-mcp image is 27.8GB: ~9GB wasted by chown -R /app duplicating a layer, plus ~6.6GB CUDA/torch via cognee[docs] (no slim/CPU variant)


michabbb:
## Summary

The `cognee/cognee-mcp:1.2.2` Docker image is **27.8 GB**. Two independent issues account for almost all of it. On an API-only deployment (remote LLM + remote embeddings, no GPU) running the MCP in **proxy mode** (`API_URL` set), none of the heavy ML stack is ever used.

## Issue 1 — `chown -R /app` duplicates the entire 9 GB `/app` layer (~9 GB wasted)

`docker history cognee/cognee-mcp:1.2.2 --no-trunc`:

```
9.02GB  COPY /app /app # buildkit
9.02GB  RUN ... groupadd ... && useradd ... && chown -R cognee:cognee /app # buildkit
```

The `chown -R cognee:cognee /app` runs in a separate layer **after** `COPY /app /app`. Changing ownership rewrites metadata on every file, so overlayfs copies up the whole ~9 GB tree into a second layer — the venv is effectively stored twice.

**Fix:** use `COPY --chown=cognee:cognee /app /app` (or perform the chown within the same layer as the COPY). Saves ~9 GB.

## Issue 2 — `cognee[docs]` pulls torch + the full CUDA stack (~6.6 GB) with no opt-out

`site-packages` breakdown inside the image:

| package | size |
| --- | --- |
| `nvidia` (CUDA runtime libs) | 4.3 GB |
| `torch` (CUDA build) | 1.7 GB |
| `triton` | 639 MB |
| onnxruntime / opencv / transformers | ~0.5 GB |

Dependency chain (from `cognee-mcp/uv.lock`):

```
cognee[docs]  ->  unstructured[pdf]  ->  unstructured-inference  ->  torch / torchvision / nvidia-*-cu12
```

`torch` is a 3-levels-deep transitive dependency (it is not a direct dep and not a cognee extra by itself), pulled in solely by the `docs` extra declared in `cognee-mcp/pyproject.toml`:

```toml
dependencies = [
    "cognee[postgres-binary,docs,neo4j]>=1.1.0,<2.0.0",
    ...
]
```

The Dockerfile installs from the default PyPI index, so the CUDA `torch` wheel (15× `nvidia-*-cu12` libs: cublas, cudnn, nccl, cusparse, …) is bundled.

[...]

PR #4045, opened 2026-07-13
fix(cognee-mcp): image is twice its content size — hand /app ownership to COPY --chown

gorynychzmey:
## Description

The cognee-mcp image is ~18GB while its filesystem holds ~8.6GB. The final stage COPYs /app (~8.6GB) and then runs a separate `chown -R cognee:cognee /app`: overlayfs materialises a second full copy of the tree in the chown layer.

Fix: create the user **before** the COPY, hand ownership to `COPY --chown`, and keep a non-recursive `chown` of the still-empty WORKDIR — /app doubles as HOME, so the runtime user must be able to write dotfiles/caches there (Kuzu extension cache).

Measured on podman: 18GB -> 9.12GB, one 8.9GB layer instead of two; container still runs as uid 1000 `cognee`; /app writable; MCP smoke test (health, .well-known, tool call) runs green.

This addresses the first half of #3691 (the layer duplication); the CUDA/torch weight discussed there is left untouched — this PR is Dockerfile-only, no dependency changes.

## Acceptance Criteria

* Image size halves; `podman history` shows a single big layer instead of two.
* Container starts as non-root `cognee`; /app stays writable as the runtime HOME.

```
$ podman images | grep cognee-mcp
localhost/cognee-mcp   0.1.2   9.12 GB   (was 18 GB)

$ uv run pytest tests/ -q
22 passed in 3.51s
```

## Type of Change

## Pre-submission Checklist
- [x] **I have tested my changes thoroughly before submitting this PR** (running in my self-hosted deployment)

### DIFF
--- a/cognee-mcp/Dockerfile
+++ b/cognee-mcp/Dockerfile
@@ -62,18 +62,22 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 
 WORKDIR /app
 
+# Run as a non-root user. Create it BEFORE the COPY so --chown can resolve
+# the name and the tree lands owned by cognee within the COPY layer itself:
+# a separate `chown -R` afterwards makes overlayfs duplicate the whole ~9GB
+# /app into a second layer, doubling the image size.
+RUN groupadd --system --gid 1000 cognee \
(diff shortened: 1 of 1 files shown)


Vasilije1990:
Handled this separately

Issue #3604, opened 2026-06-28
Hackathon: Make memory agents can trust — rankable, citable results + actionable errors


Vasilije1990:
## 🏆 Hackathon task: Make the memory something agents can *trust*

Agents don't read prose the way humans do. They need memory that returns **results they can rank and cite**, and **errors they can act on** — programmatically. cognee already has the building blocks; this task is to make them first-class, uniform, and tested across the whole retrieval surface.

> Why now: as agents consume cognee over the SDK, CLI, and especially **MCP**, an opaque result blob or a stack-trace error forces the agent to guess. Rankable + citable results let an agent weigh and attribute evidence; structured errors let it self-correct instead of failing.

---

### Pillar A — Results agents can rank and cite

**Where we are today**
- [`ScoredResult`](https://github.com/topoteretes/cognee/blob/main/cognee/infrastructure/databases/vector/models/ScoredResult.py) exposes `score`, but it's the *raw backend distance* where **lower is better** — unintuitive and not comparable across retrievers/backends.
- [`SearchResult`](https://github.com/topoteretes/cognee/blob/main/cognee/modules/search/types/SearchResult.py) wraps the payload as `search_result: Any` — an opaque blob with no per-item score or citation at the API surface.
- Citations exist but are **opt-in and uneven**: `search(..., include_references=False)` plus `cognee/modules/retrieval/utils/references.py` and the `references_example.py` / `memory_provenance` demos. Coverage across the ~15 retrievers (graph completion, RAG, chunks, summaries, BM25, hybrid, temporal, cypher, natural-language, triplet, agentic…) is inconsistent.

**What to build**
- [ ] **Calibrated relevance score**: expose a normalized `relevance` (0–1, **higher is better**) alongside the raw backend score, with documented semantics, so agents can rank and threshold consistently across retrievers and DB backends.
- [ ] **First-class, structured citations**: every recalled item carries provenance an agent can cite — source document id/name, chunk id + text span/offsets, contributing graph node/edge ids, and dataset. Promote references from opt-in to a uniform, structured field returned by **all** retrievers.
- [ ] **Typed, rankable result items**: turn `SearchResult.search_result: Any` into a structured list of items, each with `{content, relevance, citations[], source_ids}` — so an agent gets a ranked, attributable list instead of a blob. Keep backward-compat (e.g. behind a flag or additive field).
[...]

PR #4061, opened 2026-07-14
feat(cli): calm, honest first run quick wins (SDK-169)

rshkarin:
## Summary

Reworks and hardens the hackathon PR #4002 by @Labi-Joy for issue #3605 ("Calm, Honest First Run"). The original PR is preserved verbatim as its six commits (attribution and `Signed-off-by` intact); every change of mine lands as a separate commit on top.

The audit (Part 1) and proposal (Part 2) are genuinely useful and the four quick wins (Part 3) target the right first-run pain. Adversarial review found one correctness bug that defeated the flagship case, a bogus debug guard, a doc that promised more than the code delivered, and some indirection the issue never asked for. This branch fixes those and trims the incidental code (811 insertions vs the original 904, while adding correctness).

Linear: SDK-169.

## What the original PR delivers (kept)

- **W1** — fail fast on terminal auth errors in `LiteLLMEmbeddingEngine` so an invalid key no longer walks the full ~2‑minute retry ladder.
- **W2** — a first-run error remediation table printed by the CLI, plus suppression of aiohttp's shutdown-time `Unclosed ...` `ResourceWarning`s.
- **W3** — a copy-pastable `Next: ...` hint after `remember` / `cognify` / `recall` / `forget`.
- **W4** — a `--sample-data` flag on `remember` that ingests a bundled fixture.
- The onboarding audit + proposal doc, and unit tests for each quick win.

## What this rework changes (separate commits)

1. **`fix(cli)`: remediation matches the real errors.** The missing-key row keyed off `llm_api_key` (underscore), but the real runtime error is `LLMAPIKeyNotSetError: LLM API key is not set.` — so the single most common first-run failure got no hint. The unit test masked this with a synthetic string. Repointed the needles at the real message/class, anchored the tests on the real strings and on the `str()` of the actual raised exceptions, fixed a dead endpoint-timeout needle, and replaced the `Remediation` dataclass + helper with a plain table.
2. **`refactor(cli)`: trim next-step hints.** Dropped the `--quiet` flag and its helpers (not requested by #3605; it also made `--qu` ambiguous with `--query-type` on `recall`), collapsed two byte-identical hint functions, and removed a dead `had_results` parameter.
3. **`fix(cli)`: harden the aiohttp suppression.** Moved the filter out of import time into the CLI run path, gated it on the real `--debug` flag (the previous `-x` guard matched no cognee flag), and covered all three `Unclosed ...` warnings with one pattern.
4. **`refactor(embedding)`: tidy the fast-fail.** Dropped a redundant `CancelledError` re-raise (it already propagates as a `BaseException`), corrected an inline comment, and relabelled a misleading "quota exhausted" test to a real 403 case.
5. **`docs`: reconcile the audit with what shipped.** The proposal described `--sample-data` as an offline, no-API-call demo with a `demo` sub-command; the shipped flag needs a key and makes real calls. Rewrote it honestly and listed the truly keyless path as a follow-up gated on the (unmerged) mocked-LLM harness in #3601.

## Test plan

- [x] `cognee/tests/cli_tests/cli_unit_tests/` — 138 passed (incl. the rewritten hint/remediation tests).
- [x] `cognee/tests/unit/test_litellm_embedding_auth_fastfail.py` — 3 passed.
- [x] Remediation verified against the real `str()` of `LLMAPIKeyNotSetError` and `litellm.AuthenticationError`.
- [x] End-to-end, keyless: `cognee-cli remember --sample-data` with an empty key resolves the fixture, fails fast, and prints the missing-key remediation hint (no stack trace, no token spend). Before this fix that case produced only the generic docs pointer.
- [x] `ruff format --check` and `ruff check` clean on all changed files.
- [ ] Pre-existing, unrelated: two timing-based tests in `test_embedding_rate_limiting_realistic.py` fail identically on `origin/dev` (verified by reverting the engine change).

## Notes for reviewers

- The `--sample-data` flag still requires `LLM_API_KEY`; a fully keyless first result depends on the mocked-LLM harness in #3601 and is deferred (documented in the audit).
- Full alignment of the error-message contract (stripping leaked HTTP internals such as `(Status code: NNN)`) is coordinated with #3604 and left as a follow-up.

---


Labi-Joy:
Thanks @rshkarin for the review and rework. I checked claim #1: str(LLMAPIKeyNotSetError()) renders as "LLMAPIKeyNotSetError: LLM API key is not set. (Status code: 418)", so neither "llm_api_key" nor "no api key" matches. Claims on #2 and #3 also check out.

PR #3980, opened 2026-07-08
feat(vscode): project-memory extension — remember / recall / ask my project memory

Anjaligarhwal:
## Description

Closes #3622.

When you work in an editor you constantly re-explain the same project context - to teammates, to yourself after a break, and to AI assistants that forget everything between sessions. Cognee already solves the "remember it once" problem, but today you have to leave the editor (CLI, notebooks, the API) to put things in or get them out. This PR closes that gap: it brings Cognee's remember/recall into VS Code so the project's memory lives next to the code, and most importantly for me — an answer can point back to the exact file it came from.

I was assigned this issue after proposing the design in the thread, so this PR is that design built out. A few decisions I want to be explicit about, because they drove the whole shape:

- **Transport = HTTP, not MCP.** The heart of "ask my project memory" is a *citable* answer, and `include_references` (the Evidence block) is only reachable over the HTTP `recall` endpoint. So the extension talks to Cognee over HTTP (`recall` / `remember` / `forget` / `improve`) with `X-Api-Key` auth — the same contract Cognee's own Cloud client uses and the same code works against a local server or a Cognee Cloud tenant.
- **Per-repo isolation with no server change.** Each workspace is pinned to a deterministic dataset `vscode_<sha256(gitRemote || workspaceRoot)>`, so two repos open in the same editor never bleed into each other, and the dataset is stable across clones/machines when a git remote exists. It's overridable via a setting.
- **Editor-agnostic core.** `src/core/` has zero `vscode` imports — the VS Code layer depends only on a `CogneeClient` interface. That's deliberate so a JetBrains client (the other half of the issue) can reuse the exact same tested core instead of a rewrite, and so the whole core is unit-testable against a mock with no editor and no network.
- **Ingestion is opt-in, always.** Nothing goes to Cognee unless you ask — Remember Selection, Remember File, Remember Note, or an explicit *Index Workspace* that respects`.gitignore`/`.cogneeignore`, skips oversized files, and shows a preflight count before it runs.
- **Same-name citation resolution.** Cognee's recall citations carry a document *basename*, so a repo with several `README.md`/`package.json` files is ambiguous. The extension prepends a `Source: <path>` header at ingest and reads it back from the Evidence snippet to open the exact file deterministically — no picker — falling back to a per-workspace path index, then snippet-content matching, then a user pick only when genuinely ambiguous. Sources are ranked by relevance to the query and de-duplicated per file. (The clean upstream complement —surfacing the stored source path in structured citations — is tracked separately as #3844.)

[...]

Issue #3622, opened 2026-06-28
Hackathon: VS Code / JetBrains extension — remember/recall + ask my project memory (propose the design)


Vasilije1990:
## 🏆 Hackathon task: VS Code / JetBrains extension — "ask my project memory"

**Goal:** bring cognee into the editor so developers can `remember` and `recall` without leaving their IDE, and "**ask my project memory**" — query everything cognee knows about the current repo/workspace, with citations.

### 📣 Propose the implementation first

Before coding, comment with the best approach so we converge on one design:
- **How the extension talks to cognee** — via the **MCP** server (editors increasingly speak MCP), the **CLI** (`cognee-cli`), the **TypeScript client** [`@cognee/cognee-ts`](https://www.npmjs.com/package/@cognee/cognee-ts), or the HTTP API — and why.
- **Workspace → memory scoping** — map the open project/workspace to a cognee `session_id`/dataset so "project memory" is isolated per repo.
- **Target first** — VS Code and JetBrains are separate ecosystems; propose which to build first (and whether a shared core/LSP-style service can back both).
- What gets ingested (selection, file, whole repo?) and how opt-in works.

The clearest, most reusable proposal gets adopted.

### What cognee gives you to build on

- Memory API: `remember` / `recall` / `forget` / `improve`, with **`session_id`** for per-project isolation.
- **MCP server**, **CLI** (`cognee-cli remember/recall`), Python + **TypeScript** clients.
- Rankable, **citable** results (#3604) so answers can link back to source files/notes.
- Prior art: there's already a [Claude Code plugin](https://github.com/topoteretes/cognee-integrations/tree/main/integrations/claude-code) to learn from.

### Scope

- [ ] Editor commands: **Remember** (selection/file/note) and **Recall** (query) from the command palette / context menu.
- [ ] An **"Ask my project memory"** panel/quick-pick that queries the workspace's memory and shows answers **with citations** to source files.
- [ ] Workspace → session/dataset mapping; a way to ingest the repo (coordinate with the GitHub-repo connector idea).
- [ ] Config for cognee endpoint/keys; `forget` for the project; opt-in ingestion.

### Requirements

- ✅ **Test coverage** — extension logic tested against a mocked cognee backend; reuse the mocked-LLM harness (#3601) where the flow hits an LLM. No live keys in CI.
- ✅ **Easy onboarding** — install → point at cognee → first recall in minutes; copy-paste setup (consistent with #3605).
[...]

PR #4067, opened 2026-07-15
feat(vscode): project-memory extension — remember/recall + ask my project memory (SDK-173)

rshkarin:
## What this is

Review + rework of @Anjaligarhwal's community VS Code extension (#3980) for hackathon issue #3622 — bring cognee's `remember`/`recall` into the editor and **"ask my project memory"** with answers that link back to the exact source file. Base `dev`. Supersedes #3980.

The extension (`cognee-vscode/`) is self-contained TypeScript: an **editor-agnostic core** (`src/core/`, no `vscode` imports) behind a `CogneeClient` interface, plus a thin VS Code layer. It talks to cognee over the HTTP API (`recall` / `remember` / `forget`) with `X-Api-Key`, so the same code works against a local server or a Cognee Cloud tenant.

## Original design (preserved)

- **Transport = HTTP;** citations come from the `include_references` Evidence block, parsed into clickable sources.
- **Per-repo isolation:** `vscode_<sha256(gitRemote || workspaceRoot)>`, overridable via a setting.
- **Opt-in ingestion:** Remember Selection / File / Note, plus an explicit Index Workspace with a preflight count.
- **Same-name citation resolution** via a `Source: <path>` header, with a per-workspace path-index / snippet-content fallback.

The **6 original commits are preserved verbatim** (authorship and sign-off intact). Every improvement below is a separate single-concern commit on top.

## Review + rework

I verified the HTTP wire contract against the live server routers (`get_recall_router.py`, `get_remember_router.py`, `get_forget_router.py`) and the Evidence format in `modules/retrieval/utils/references.py` — endpoints, field names (including the camelCase `datasetName` multipart field), and the citation parser all match. An adversarial review (find → independently verify each finding) then surfaced the following, fixed as separate commits:

**Correctness**
- **Citation reveal never scrolled to the cited region for code.** The server collapses each snippet's whitespace, but the reveal did a literal `indexOf` against the file's real (newlined / indented) text, so it silently no-op'd for source code. Replaced with a whitespace-tolerant `findSnippetRange` that maps the match back to a real range (unit-tested).
- **`dedupeByFile` listed a file twice** when its first (header) chunk and a later (headerless) chunk both grounded the answer. It now collapses them without merging genuinely distinct files that share a basename.
- **Off-by-one** in the recorded `(lines a-b)` provenance for full-line selections.

**Safety / correctness of Index Workspace**
- **`.gitignore` / `.cogneeignore` now actually respected.** The `respectGitignore` setting previously only toggled a hardcoded blocklist and never read either ignore file; turning it off passed `null` to `findFiles`, disabling *all* excludes so the indexer would sweep `node_modules`/`.git` and upload their contents. It now filters candidates through the project's `.gitignore` and `.cogneeignore` with full git syntax (globs, negation, anchoring) via the `ignore` library, and **always** skips dependency/build/VCS directories regardless of the setting.
- **Citation path traversal.** The server-provided `Source:` path was joined onto a workspace folder and opened with no containment check, so a compromised or MITM'd backend could open arbitrary local files (`../../etc/passwd`). `normalizeRelativePath` now drops `..`/`.` segments.
- **Crypto-grade webview nonce** (was `Math.random()`).

The editor-agnostic `CogneeClient` keeps its full typed wire contract (including the not-yet-surfaced `improve` endpoint) — it is the reusable seam for the planned JetBrains client and future features, so it is retained deliberately rather than trimmed to today's call sites.

## Verification

- `npm run typecheck` clean, **`npm test` → 68/68 passing** (mocked backend, no live keys, no LLM), `npm run build` produces `dist/extension.js`.
- **End-to-end (no live keys):** drove the real `HttpCogneeClient` against a mock server mimicking the routers — `remember` (multipart `datasetName`) → `recall` → Evidence parse → rank/dedupe → whitespace-tolerant reveal selecting the function body across multi-line source → path-traversal guard. Separately verified the ignore path against real `.gitignore` + `.cogneeignore` files (glob / directory / negation / combined). All green.

Linear: SDK-173.

## DCO Affirmation

### DIFF
--- a/.github/workflows/cognee_vscode_tests.yml
+++ b/.github/workflows/cognee_vscode_tests.yml
@@ -1,0 +1,33 @@
+name: cognee-vscode tests
+
+# Scoped CI for the VS Code extension: runs only when the extension changes.
+# The suite runs against a mocked backend — no live keys, no LLM calls.
+
+on:
+  push:
+    paths:
+      - "cognee-vscode/**"
+  pull_request:
+    paths:
+      - "cognee-vscode/**"
+
+permissions:
+  contents: read
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    defaults:
+      run:
+        working-directory: cognee-vscode
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+          cache: npm
+          cache-dependency-path: cognee-vscode/package-lock.json
+      - run: npm ci
+      - run: npm run typecheck
+      - run: npm test
+      - run: npm run build
--- a/cognee-vscode/.gitignore
+++ b/cognee-vscode/.gitignore
@@ -1,0 +1,7 @@
+node_modules/
+dist/
+out/
+*.vsix
+.vscode-test/
+coverage/
+*.log
--- a/cognee-vscode/.vscode/launch.json
+++ b/cognee-vscode/.vscode/launch.json
@@ -1,0 +1,16 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Run Cognee Extension",
+      "type": "extensionHost",
+      "request": "launch",
+      "args": [
+        "--extensionDevelopmentPath=${workspaceFolder}",
+        "${workspaceFolder}"
+      ],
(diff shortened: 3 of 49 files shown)


rshkarin:
Migrated to the cognee-integrations monorepo: **topoteretes/cognee-integrations#247** (`integrations/vscode/`).

PR #3788, opened 2026-07-02
feat: AST-based code graph ingestion and CODE_GRAPH search type

koopatroopa787:
## Summary

Closes #3668

Adds a new `codegraph` extra that uses tree-sitter to walk a Python source tree, builds a structured knowledge graph of the code, and makes it searchable via a new `CODE_GRAPH` retrieval type.

### What's added

**`cognee/tasks/codegraph/`**
- `models.py` — Pydantic `DataPoint` subclasses: `CodeFile`, `CodeClass`, `CodeFunction`, `CodeImport`. Each uses `identity_fields` for deterministic UUID deduplication across repeated ingestions.
- `parse.py` — tree-sitter 0.24.x parser. Walks a `.py` file's AST and emits nodes + typed edges (`defines_class`, `defines_function`, `has_method`, `imports`). Methods are qualified as `ClassName.method_name`. Docstrings are extracted from the first string literal in a body block. Falls back to `([], [])` gracefully if tree-sitter is not installed.
- `pipeline.py` — `async ingest_code_graph(path, extensions={".py"})` walks a directory, calls `parse_file`, stores nodes via `add_data_points`, and persists edges via the graph engine.

**`cognee/modules/retrieval/code_graph_retriever.py`**
- `CodeGraphRetriever(BaseRetriever)` — vector-searches the `CodeFunction_name`, `CodeClass_name`, and `CodeFile_file_path` collections, then returns graph neighbours of the hits for richer context.

**`cognee/modules/search/types/SearchType.py`**
- Added `CODE_GRAPH = "CODE_GRAPH"` to the `SearchType` enum.

**`cognee/modules/search/methods/get_search_type_retriever_instance.py`**
- Wired `SearchType.CODE_GRAPH` → `CodeGraphRetriever`.

**`cognee/tests/unit/test_codegraph.py`**
- 10 unit tests covering: file/class/function/import detection, method qualification, edge emission, deterministic IDs across two parses, graceful missing-file handling, and docstring extraction. Fully offline — no LLM or DB calls.

### Usage

```bash
pip install cognee[codegraph]
```

```python
from cognee.tasks.codegraph import ingest_code_graph
import asyncio

asyncio.run(ingest_code_graph("/path/to/src"))

# Then query:
results = await cognee.search("find functions that handle authentication", query_type=SearchType.CODE_GRAPH)
```

## Test plan

- [x] `pytest cognee/tests/unit/test_codegraph.py` — 10/10 passing locally
[...]

Issue #3668, opened 2026-06-29
Local AST code-graph ingestion (tree-sitter, no API cost)


Vasilije1990:
## Local AST code-graph ingestion (tree-sitter, no API cost)

**Hackathon idea.** Copy [graphify](https://github.com/safishamsi/graphify)'s biggest differentiator: build the structural code graph **locally via tree-sitter AST**, with no LLM calls. Graphify supports 36 grammars and extracts files/classes/functions plus `imports`/`calls`/`references` edges deterministically; only docs/PDFs/media go through the model.

### Surface (the opt-in "option")
A code loader / pipeline, e.g. `cognee.add(path, loader="code")` or a `CodeGraphPipeline`, producing typed nodes (`File`, `Class`, `Function`, `Symbol`) and edges (`imports`, `calls`, `references`) directly via `add_data_points` — bypassing the LLM extraction path for code.

### Why
- **Cheap + deterministic** — structural layer costs nothing and doesn't hallucinate edges.
- Cognee's `repo_processor` task package is **stubbed/empty** in 1.2.1 and there's a `CODING_RULES` SearchType already — this looks like previously-intended surface to fill.
- Opens the codebase-memory use case to complement document memory.

### Scope notes
- Start with a few high-value grammars (`.py`, `.ts/.tsx`, `.js`, `.go`) before chasing graphify's full 36.
- Pairs naturally with the Graph Insight Report (#3667) — call-flow hubs become god nodes.

Higher effort than the report; flagged as a runner-up. Part of #3670.

Tanya-garg10:
Hi @Vasilije1990 @dexters1! I'd like to work on this issue as part of the hackathon.

This feature looks like a great addition to Cognee's memory pipeline, especially since it enables deterministic code graph generation without relying on LLM calls. My initial approach would be to build a dedicated code ingestion pipeline using Tree-sitter, starting with a small set of high-value languages such as Python, JavaScript/TypeScript, and Go. The pipeline would parse source files locally, construct typed nodes like `File`, `Class`, `Function`, and `Symbol`, and establish structural relationships such as `imports`, `calls`, and `references` directly from the AST.

[...]

PR #4088, opened 2026-07-16
feat(codegraph): local AST code-graph ingestion, no LLM cost (SDK-187)

rshkarin:
## Summary

Maintainer-continuation of community PR #3788 (by @koopatroopa787) for hackathon issue #3668 (Linear SDK-187).

Adds a **local, deterministic** code-graph ingestion path: walk a Python source tree with tree-sitter, build cognee's structural code-graph nodes and edges, and store them via `add_data_points` — with **no LLM calls**. Once ingested, the graph is answerable through the existing default search (e.g. `GRAPH_COMPLETION`), so no new retrieval surface is introduced.

```python
from cognee.tasks.codegraph import ingest_code_graph
import cognee
from cognee.modules.search.types import SearchType

await ingest_code_graph("/path/to/repo")            # tree-sitter → graph, no LLM
await cognee.search("what verifies credentials?", query_type=SearchType.GRAPH_COMPLETION)
```

The original author's commit is preserved as the base; the rework lands as separate commits on top.

## What the ingestion does

* `cognee/tasks/codegraph/parse.py` parses each `.py` file with tree-sitter (the pre-existing `codegraph` extra) into cognee's **existing** shared entities (`cognee/shared/CodeGraphEntities.py`): `Repository`, `CodeFile`, `FunctionDefinition`, `ClassDefinition`, `ImportStatement` — wired through the model's relationship fields (`CodeFile.part_of` / `depends_on` / `provides_class_definition` / `provides_function_definition`).
* `cognee/tasks/codegraph/pipeline.py` exposes `async ingest_code_graph(path)`: it runs `setup()` (like `cognee.add`), parses every file under the path, and calls `add_data_points([code_files])` **once**. Because the nodes carry relationship fields, `get_graph_from_model` mints every File→Class / File→Function / File→Import edge natively — no hand-rolled edge tuples, no manual `add_edges`/`index_graph_edges`.
* Node ids are deterministic (`DataPoint.id_for(...)` over file path + symbol name), so re-ingesting the same tree upserts in place.
* Methods are captured as qualified `Class.method` functions.

## Why the original PR was reworked

* **Dropped the custom `CODE_GRAPH` SearchType + `CodeGraphRetriever` entirely.** The issue never asked for a new SearchType, and the retriever was both redundant and non-functional (all verified against the code):
  * `graph_completion_retriever._get_vector_index_collections()` already enumerates every `DataPoint` subclass's vector collections, so `GRAPH_COMPLETION` reaches the ingested code nodes automatically and produces a grounded, LLM-backed completion.
  * It called `graph_engine.get_neighbours` (UK spelling) — the real method is `get_neighbors` (US) — under a bare `except`, so graph neighbours were never fetched.
  * It read `.name`/`.file_path`/`.type` off `ScoredResult`, which exposes only `id`/`score`/`payload`, so context degraded to `[node] id=<uuid> score=<float> payload=None` — no code identifiers.
  * `get_completion_from_context` returned the raw context with no LLM call.
* **Reused the existing shared code-graph entities instead of a parallel model set.** The original added new `CodeFile`/`CodeClass`/`CodeFunction`/`CodeImport` DataPoints that duplicated concepts already modelled by the shared entities — and the new `CodeFile` collided on class name / graph label with the live shared `CodeFile` (used by `summarization/models.py` and `description_to_codepart_search.py`). Reuse removes the duplication and the collision, and lets edges be built natively (deleting the manual edge machinery, which also bypassed default edge enrichment).
* **Fixed relative-import capture.** `from . import x` / `from .mod import y` recorded the imported symbol as the module; the parser now reads the statement's `module_name` field.
* **Made `ingest_code_graph` self-sufficient.** It runs `setup()` first, so the code graph is queryable via `cognee.search` without a separate init step.
* **Reworked the tests.** Added offline (no-LLM, no-DB) tests that exercise the actual deliverable — `get_graph_from_model` emitting native File→Class/Function/Import edges, and `ingest_code_graph` persisting through a single `add_data_points` call — and removed a dead `mistralai` `sys.modules` stub whose comment referenced a file absent from the repo.

## Scope

* Python `.py` only, as the first high-value grammar (issue #3668 says to start with a few grammars before chasing graphify's full set).
* Structural layer: files, classes, functions, import edges — all deterministic and LLM-free.
* `calls` / `references` edges and a dedicated `Symbol` node type (mentioned in the issue) are **deferred**: the original PR never implemented them, and precise cross-file call/reference resolution needs a symbol resolver that pure AST can't provide without guessing. Follow-up.

## Verification

* `pytest cognee/tests/unit/test_codegraph.py` — 9/9 passing (parser + offline ingestion/native-edge tests).
* End-to-end on a sample repo (local SQLite + LanceDB + Ladybug, real embeddings + LLM): ingesting `auth.py` + `models.py` produced `CodeFile×2`, `FunctionDefinition×4`, `ClassDefinition×2`, `ImportStatement×4`, `Repository×1` with `part_of`/`depends_on`/`provides_*` edges (13 nodes, 12 edges); re-ingest was idempotent (13/12 unchanged); and `GRAPH_COMPLETION` answered *"which function verifies credentials and how does it hash the password?"* with *"AuthService.login … AuthService._hash uses hashlib.sha256"* — with no custom retriever.
* `ruff format` / `ruff check` clean; the codegraph package is outside the `ty` include scope.


rshkarin:
Superseded by https://github.com/topoteretes/cognee/pull/4037

PR #4092, opened 2026-07-16
feat(ingestion): connector-agnostic DLT document-mode + foreground orphan cleanup (SDK-193)

rshkarin:
## Connector-agnostic DLT document-mode + foreground orphan cleanup (SDK-193)

Adds a **document ingestion path** for DLT sources so a connector can route its rows through
normal cognify (LLM entity extraction) instead of the deterministic relational schema-context
path. This is the core seam that lets data-source connectors live **outside core** (in
`cognee-community`) while reusing the existing DLT pipeline — the enabler for the connector
repo-boundary migration.

### Changes
- **`dlt_utils`** — `DOCUMENT_SOURCE_ATTR` marker + `document_source_tag()`. A dlt source sets
  the attribute to opt in; its value becomes `external_metadata["source"]` (the source's own tag,
  **not `"dlt"`**), so `is_dlt_sourced()` returns False and `classify_documents` already routes the
  row to `TextDocument`/cognify. The shared engine **never imports the connector** — connectors
  declare their own nature.
- **`resolve_dlt_sources`** — split document vs relational items; `_build_document_data_item`
  builds a text `DataItem` from the row's `title`/`content`/`url`/`id`; orphan cleanup is
  **source-scoped** (deletes only same-source rows) so mixed sources in one dataset never wipe
  each other.
- **`add()`** — run the deferred `orphan_cleanup` after commit in the **foreground/blocking** path
  too. It was awaited only for `run_in_background=True`, so forget-on-source-deletion was silently
  skipped in the default path for every DLT connector. (Folds in **SDK-171**.)

`classify_documents` and `ingest_dlt_source` need no change — the routing fall-through and
content-hash already do the right thing.

### Verification
- `test_dlt_document_mode.py` — new, connector-agnostic unit tests for the marker, routing, and
  `_build_document_data_item` (**4 passed**).
- `test_add_foreground_orphan_cleanup.py` — foreground orphan-cleanup regression, drives the real
  add pipeline twice against local stores, no LLM (**1 passed**).
- Existing relational-DLT tests (`test_dlt_p0_correctness`, `test_extract_dlt_fk_edges_provenance`)
  — **10 passed, no regression**.
- **Real end-to-end** (OpenAI): an external document connector → `remember()` → cognify produced a
  22-node/37-edge knowledge graph with correct entities + relationships, and `GRAPH_COMPLETION`
  answered a cross-page multi-hop question correctly. Document rows tagged `source="notion"`,
  `is_dlt_sourced=False` (cognify path), while `extract_dlt_fk_edges` correctly found nothing.

### Design note
Standardizes on **Notion's `DOCUMENT_SOURCE_ATTR`** (the fresher, connector-agnostic design) over
the alternative `CONTENT_COLUMN_HINT_ATTR` + `dlt_mode` approach. Google Drive additionally needs
the core to **respect the connector's `write_disposition`** (merge/incremental) rather than the
document path's implicit full-snapshot `replace`, so its Changes-API incremental model survives —
tracked as a follow-up so GDrive can rework onto this seam.

### Unblocks
The community connector migration — **Notion** (SDK-170, topoteretes/cognee-community#132) and
**Google Drive** (SDK-166, topoteretes/cognee-community#135). Once this merges and a cognee release
ships, those community packages flip their `cognee==` pin and go ready.

### DCO
I have read the DCO and I sign off on all commits in this pull request.

### DIFF
--- a/cognee/api/v1/add/add.py
+++ b/cognee/api/v1/add/add.py
@@ -272,6 +272,12 @@ async def add(
         data_cache=data_cache,
     )
 
+    # Foreground runs: the fresh rows are committed by pipeline_executor_func
+    # above, so it's now safe to clean up orphans. (Background runs already ran
+    # this up front and set orphan_cleanup to None.)
+    if orphan_cleanup is not None:
+        await orphan_cleanup()
+
     # run_pipeline_blocking returns {dataset_id: PipelineRunInfo} but callers
     # expect a single PipelineRunInfo (add always processes one dataset).
     if isinstance(result, dict) and len(result) == 1:
--- a/cognee/tasks/ingestion/dlt_utils.py
+++ b/cognee/tasks/ingestion/dlt_utils.py
@@ -1,10 +1,25 @@
 """Shared utilities for DLT ingestion."""
 
 import json
+from typing import Optional
 
 from cognee.shared.logging_utils import get_logger
 
 logger = get_logger("dlt_utils")
+
+# A dlt source sets this attribute to opt into the "document" ingestion path:
+# each row becomes a text document that flows through normal cognify (LLM entity
+# extraction), instead of the default relational schema-context path. The value
+# is the ``external_metadata["source"]`` tag used for that source's rows. This
+# lets resolve_dlt_sources stay connector-agnostic — a connector declares its
+# own nature rather than the shared engine hard-coding connector names.
+DOCUMENT_SOURCE_ATTR = "cognee_document_source"
+
+
+def document_source_tag(item) -> Optional[str]:
+    """Return the document-source tag a dlt source opted into, else ``None``."""
+    tag = getattr(item, DOCUMENT_SOURCE_ATTR, None)
+    return tag if isinstance(tag, str) and tag else None
 
 
 def is_dlt_sourced(metadata) -> bool:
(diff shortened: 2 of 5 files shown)


rshkarin:
**Update — extended to enable incremental document connectors + fixed orphan graph-cleanup** (pushed `8499ece0e`):

- The document path now **respects the caller's `write_disposition`/`primary_key`** (default `replace`/`id`) instead of forcing `replace`, so both sync models work: snapshot (Notion/Slack, `replace`) and incremental merge (Google Drive's Changes-API delta + `_deleted` hard-delete). With `max_rows_per_table=0` the whole merged table is read back each run, so merge forget-on-delete is correct — unchanged rows survive.
- Orphan cleanup now **always** runs graph+vector deletion instead of gating on `has_data_related_nodes` (a relational-ledger check that skipped graph-provenance cleanup, leaving a forgotten row's chunks/entities retrievable).

Verified against a real dev+PR build: Google Drive **incremental add→delete→re-sync** forgets the deleted file's graph entity while keeping the surviving file's (no over-delete); Notion (replace) e2e + the relational-DLT tests unchanged. This unblocks **both** Notion (#132) and Google Drive (topoteretes/cognee-community#135, now reworked onto `DOCUMENT_SOURCE_ATTR`).

PR #4095, opened 2026-07-17
fix: simplify AgentContextExtraction schema + strip unpaired UTF-16 surrogates

GrowDev1:
## Summary

Two related reliability fixes surfaced while running the agent-context-extraction pipeline against a local Ollama model:

**1. `AgentContextExtraction.lessons` schema simplification**

`lessons` was typed as a section-specific discriminated union (`AgentCandidateContextUpdateVariant`). `instructor`'s structured-output modes ask the model to pick one of N pydantic classes via a discriminator field before the fields even validate — reliable for large/cloud models, but in local testing against `llama3.1:8b` the discriminated union was rejected outright on most trials (model frequently omits the discriminator field, or returns a shape the union can't resolve).

Switching `lessons` to the plain base model (`AgentCandidateContextUpdate`) fixes this: its own `section_valid` field validator already enforces the identical section-membership constraint without needing a discriminator match, and the only real consumer (`extract_batch_agent_context` → `apply_candidate_updates` → `_coerce_candidate_model`) accepts any `CandidateContextUpdate` subclass via `isinstance`, so nothing downstream depends on the specific subclass identity — only on `section` already being valid, which the base model still guarantees.

**Trade-off, called out explicitly:** this does loosen the JSON schema `instructor` emits for every provider — `section` becomes a free-form string instead of an enumerated discriminator, rather than only affecting the Ollama path. Correctness is unaffected (`section_valid` is the real enforcement point either way, so an invalid `section` is still rejected — just via the validator instead of the schema), but it's a real, deliberate trade: a slightly more permissive schema for all providers in exchange for reliability against smaller local models. Happy to gate this behind a config flag instead if maintainers would prefer providers keep the stricter discriminated schema by default.

**2. Unpaired UTF-16 surrogate crash**

An unpaired UTF-16 surrogate code point (e.g. from a mis-decoded text boundary) is a valid Python `str` but is not valid UTF-8. Left unstripped, it crashes request serialization with `UnicodeEncodeError` wherever such text reaches a provider client or the embedding tokenizer.

Added a `_strip_surrogates()` helper (UTF-8 round-trip with `errors="replace"`) and applied it at the two real choke points, rather than patching individual adapters:
- `LLMGateway.acreate_structured_output` — the single call site every LLM provider (OpenAI, Anthropic, Gemini, Ollama, ...) routes through, so this protects all of them, not just one.
- `sanitize_embedding_text_inputs` — already the single choke point all 4 embedding engines route through.

`errors="replace"` (→ U+FFFD) was chosen over `errors="ignore"` since it preserves one character per surrogate rather than silently deleting, and nothing downstream indexes by character position in a way that would be corrupted by the substitution. Verified paired surrogates (proper emoji) and normal multi-byte text round-trip unchanged.

## Test plan

- [x] `cognee/tests/unit/infrastructure/session/test_session_context_models.py` — 33 tests pass (3 new)
- [x] `cognee/tests/unit/infrastructure/session/test_agent_context_extraction.py` — existing suite unaffected
- [x] `cognee/tests/unit/infrastructure/llm/test_ollama_adapter.py` — existing test still passes
- [x] `cognee/tests/unit/infrastructure/llm/test_llm_gateway_sanitize.py` — new, 2 tests covering surrogate stripping at the gateway layer for both `text_input` and `system_prompt`
- [x] `cognee/tests/unit/infrastructure/databases/vector/embeddings/test_embedding_utils_sanitize.py` — new, 7 tests covering `_strip_surrogates`/`sanitize_embedding_text_inputs`, including that normal/emoji text is left unchanged
- [x] `cognee/tests/unit/test_litellm_embedding_sanitize.py` — existing suite unaffected
- [x] Full combined run: 63 passed, 0 failed

### DIFF
--- a/cognee/infrastructure/databases/vector/embeddings/utils.py
+++ b/cognee/infrastructure/databases/vector/embeddings/utils.py
@@ -22,6 +22,21 @@ def is_embeddable(s: str) -> bool:
     return False
 
 
+def _strip_surrogates(s: str) -> str:
+    """
+    Replace unpaired UTF-16 surrogate code points that cannot be encoded to UTF-8.
+
+    A lone/unpaired surrogate (e.g. a mis-decoded character from a Windows console or
+    clipboard boundary) is a valid Python `str` but is not valid UTF-8. Left unstripped it
+    crashes the tokenizer's `encode_batch()` with `TypeError: TextEncodeInput must be
+    Union[...]` (and equivalent 422 errors in other embedding engines this function feeds),
+    since embedding text is eventually encoded to bytes. Round-tripping through UTF-8 with
+    `errors="replace"` removes/replaces surrogates while leaving normal text -- including
+    valid multi-byte characters and properly paired surrogate emoji -- unchanged.
+    """
+    return s.encode("utf-8", errors="replace").decode("utf-8")
+
+
 def sanitize_embedding_text_inputs(text: Union[str, List[str]]) -> List[str]:
     """
     Transform invalid/empty inputs into a safe dummy to prevent API 422 embedding errors while
@@ -31,7 +46,7 @@ def sanitize_embedding_text_inputs(text: Union[str, List[str]]) -> List[str]:
     text_list = [text] if isinstance(text, str) else text
     dummy_value = "."
 
-    return [t if is_embeddable(t) else dummy_value for t in text_list]
+    return [_strip_surrogates(t) if is_embeddable(t) else dummy_value for t in text_list]
 
 
 def handle_embedding_response(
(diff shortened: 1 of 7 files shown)


dexters1:
Thanks for the PR @GrowDev1! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

GrowDev1:
Thanks @dexters1, done. Rebased onto dev and switched the base. The source conflicts came from the typing cleanup (List → list), so I kept dev's side. I also dropped a comment-only change to test_ollama_adapter.py that didn't belong here.

One caveat: I measured the schema change with instructor, and dev now defaults to litellm_native. The surrogate fix doesn't depend on that. If it's easier, I can split the surrogate fix into its own PR.

GrowDev1:
Hi, this PR is still active on our side. It is rebased onto `dev` and we are working through the review feedback, with an update planned before Monday. Could it please stay open (or get the `keep-open` label) so the stale bot does not close it in the meantime? Thanks for the reviews so far.

PR #3799, opened 2026-07-02
feat: detect contradictions between new and existing graph facts

asadjan4611:
## Summary

Closes #3699.

The brain currently stores knowledge without any consistency check, so facts that conflict
with each other silently coexist in the graph. This PR adds the ability to **detect
contradictions between newly ingested facts and existing ones, and flag them** instead of
storing them silently.

After the knowledge graph is built, a new **opt-in** `detect_contradictions` step:

1. Collects the entities/events touched by the current ingestion.
2. Renders every fact connected to them (`subject relationship object`) with stable ids.
3. Asks the LLM (structured `ContradictionList` output) to identify pairs of facts that
   directly contradict each other, with a reason and a confidence score.
4. **Surfaces** each contradiction in two ways:
   - logs a warning exposing both facts and the reason;
   - writes a `contradicts` edge into the graph carrying `first_fact`, `second_fact`,
     `reason`, and `confidence`, so the conflict is queryable and visible next to the data.

### Acceptance criteria

- ✅ The brain can detect contradictions between new and existing data.
- ✅ Detected contradictions are flagged/surfaced (log + `contradicts` edge) rather than
  stored silently.
- ✅ Both conflicting facts **and** the reason they conflict are exposed.

## Usage

```python
import cognee

await cognee.add("Alice was born in 1985.")
await cognee.cognify()

await cognee.add("Alice was born in 1990.")
await cognee.cognify(detect_contradictions=True)   # opt-in
# -> logs a warning and adds a `contradicts` edge with the reason
```

## Design notes

- **Opt-in and backward compatible.** `cognify(detect_contradictions=False)` by default;
  no change to existing behaviour.
- **Non-destructive & fail-safe.** The task never rewrites/deletes data, returns its input
  unchanged, and swallows any error so contradiction detection can never break ingestion.
- **"Touched neighbourhood" comparison.** Entity ids are deterministic (`Entity:<name>`),
  so a re-mentioned entity's id is identical to the stored one. Comparing every fact
  connected to a newly touched entity guarantees both sides of a contradiction are compared
  together, while bounding the work (with a `max_facts` cap and per-batch execution).
- **Reuses existing infrastructure** only: `Task`, `LLMGateway.acreate_structured_output`,
  `render_prompt` / `read_query_prompt`, `get_graph_engine`. No new dependencies.

## Files changed

| File | Change |
|------|--------|
| `cognee/tasks/graph/detect_contradictions.py` | New task, models, and helpers |
[...]

Issue #3699, opened 2026-06-29
Contradiction detection in the Cognee brain


Vasilije1990:
## Context
The brain currently stores knowledge without a consistency check. As data accumulates, it ends up holding facts that conflict with each other, with no built-in way to surface those conflicts.

## Request
Add the ability to detect when information in the brain contradicts other information. Specifically:

- Identify when newly ingested information conflicts with existing information in the brain (e.g., a fact that directly negates or is incompatible with a stored one).
- Flag the contradiction so it can be reviewed or resolved rather than silently coexisting in the brain.

## Acceptance criteria
- [ ] The brain can detect contradictions between new and existing data.
- [ ] Detected contradictions are flagged/surfaced rather than stored silently.
- [ ] Both conflicting facts (and ideally the reason they conflict) are exposed.

AviArora02-commits:
Hi! I'd like to work on this issue.

My approach:
1.) Hook into the graph ingestion/construction step (during the remember/cognify lifecycle) to perform a consistency check.
2.) When new entities/triplets are extracted, query the database for existing matching entities to compare properties.
3.) Use a lightweight LLM validation or semantic similarity heuristic to flag direct negation (e.g., "A lives in X" vs. "A lives in Y").
4.) Create a contradiction register/log where these conflicting nodes and the reason for conflict are flagged and surfaced to the user.
5.) Avoid silently overwriting or coexisting without setting a ``flag or raising a validation warning.
6.) Add unit/E2E tests that ingest contradictory facts and verify that the conflict warning and both facts are correctly returned/logged.

@Vasilije1990 — could you please assign this to me?

Jia-Yee:
Interesting. learned a lot from the issues. 

I'm curious whether confidence scores should be embedded directly into the vector itself. At this point, I can only offer basic, preliminary thinking — this approach is mathematically incomplete. To make consistency detection truly robust, we need to add a rule engine and symbolic computation layers. I've been considering whether these should be built natively into the Cognee Brain.

Here are the 3 core consistency checks 
**1.  Factual Conflict Check**

**What it does:**
Checks whether the new fact directly contradicts a fact already stored in the graph by looking at the same subject and same relationship, but with a different value.

Example:

The brain already knows: "Elon Musk was born in 1971."

New information says: "Elon Musk was born in 1973."

Detection: Same person, same birth attribute, but the year differs → Conflict.

**Why it matters:**
These are the most obvious and serious contradictions that need immediate human review.

[...]

Issue #3631, opened 2026-06-28
Hackathon: Contradiction & dedup resolution — Approach E: temporal contradiction resolution


Vasilije1990:
# 🏆 Hackathon — Contradiction & dedup resolution, Approach E: temporal contradiction resolution

## The problem (shared across all 5 approach tickets)

As cognee ingests more data, the graph accumulates **duplicate entities** (same thing extracted under different names/spellings) and **contradictions** (conflicting property values or edges, often arriving at different times). Today, dedup is largely **identity-based** and contradiction handling is minimal.

**What exists to build on:**
- `cognee/modules/graph/utils/deduplicate_nodes_and_edges.py` — current node/edge dedup (used by `cognee/tasks/storage/add_data_points.py`).
- `cognee/infrastructure/engine/models/FieldAnnotations.py` — the `Dedup()` marker (UUID5 dedup key).
- POCs: `examples/pocs/disambiguation/`, `examples/pocs/post_extraction_canonicalization/`, `examples/pocs/prefetch_disambiguation/`.
- Related: contradiction-resolution is one of BEAM's memory abilities; issue #1831 (entity property conflicts during dedup).

**This is 1 of 5 parallel approach tickets** (#A–#E). Each explores a *different* strategy for contradiction + dedup resolution so we can compare them on the same fixtures and adopt the best (or a hybrid). Coordinate on a **shared evaluation set** of duplicate/contradiction cases.

### 📣 Propose before building
Comment with your design (where it hooks into ingestion/storage, merge/conflict policy, cost) before coding.

### Requirements (all approaches)
- ✅ Extend the existing dedup path; dont fork a parallel pipeline. Provenance preserved on merge (what merged into what, and why).
- ✅ **Test coverage** on a shared fixture of duplicates + contradictions (precision/recall of merges, correct conflict resolution); deterministic in CI via the mocked-LLM harness (#3601).
- ✅ Actionable, reversible merges where possible (audit trail; coordinate with #3604).

## Approach E — Temporal / provenance-based contradiction resolution
Focus on the **contradiction** half: when facts conflict (a property/edge changes over time), use cognee'\''s temporal awareness to **version the change and resolve by recency/provenance/confidence** instead of blindly merging or duplicating. Keeps "what was true when," surfaces the current value, and ties directly to BEAM'\''s contradiction-resolution ability.

Sibling approaches: A (deterministic), B (embeddings), C (LLM canonicalization), D (graph-structure).

[...]

PR #4104, opened 2026-07-17
feat(graph): contradiction detection between new and existing facts (SDK-199)

rshkarin:
## Summary

Reworks community hackathon PR #3799 by @asadjan4611 (GitHub issue #3699 — "Contradiction detection in the Cognee brain") into a shippable, opt-in feature. Linear: SDK-199.

After the knowledge graph is built, an opt-in `detect_contradictions` step compares the facts touched by the current ingestion against the facts already stored and flags any contradiction instead of letting it silently coexist — surfacing it both as a logged warning and as a queryable `contradicts` edge that records both facts, the reason, and a confidence.

```python
import cognee

await cognee.add("Alice was born in 1985.")
await cognee.cognify()

await cognee.add("Alice was born in 1990.")
await cognee.cognify(detect_contradictions=True)   # opt-in
# -> logs a warning and writes a `contradicts` edge carrying both facts + the reason
```

## Why this rework

The original design is sound (opt-in, non-destructive, fail-safe, "touched neighbourhood" comparison), and the original author's commit is preserved as the base of this branch. But as wired the feature did not work, so the fixes are layered on top as separate commits:

- **It was a silent no-op in the real pipeline.** The task is appended after `add_data_points`/`extract_dlt_fk_edges`, so the objects flowing into it are the `TextSummary` list returned by `extract_graph_and_summarize` — not `DocumentChunk`s. `TextSummary` has no `.contains`, so the "touched entities" set was always empty and the task early-returned on every run. The 14 original unit tests passed only because they fabricated objects carrying a `.contains` attribute, so they never exercised the real data shape. Proven end-to-end: on the original commit, "Alice born 1985" vs "Alice born 1990" produced **0** `contradicts` edges.
- **Make it run:** resolve the underlying chunk from each item (`made_from` when present, else the item itself) before reading `.contains`, so touched entity ids are collected from the real `TextSummary` stream and from raw chunks alike.
- **Scope the graph read:** replace the whole-graph `get_graph_data()` scan with `get_neighborhood(touched_ids, depth=1)` — same return shape, but only the region the ingestion actually touched (the original PR itself flagged the full scan as a scalability follow-up). Entity node ids are deterministic (`Entity:<name>`), so a re-mentioned entity's node id matches the stored one and both sides of a cross-ingestion contradiction land in the same neighbourhood.
- **Simplify:** the structured-output model now returns only the fact ids that conflict; the caller reconstructs the canonical fact text locally, so the stored text always matches the graph (fewer fields for the model to get wrong). Prompts and the module docstring were trimmed.
- **Tests** now exercise the real `TextSummary`-wrapped-chunk shape (the case that was silently broken), the neighbourhood/endpoint logic, confidence filtering, non-destructive pass-through, and error swallowing.

## Acceptance criteria (gh #3699)

- ✅ Detect contradictions between new and existing data.
- ✅ Surface them (log + `contradicts` edge) instead of storing silently.
- ✅ Expose both conflicting facts and the reason.

## Verification

- 15/15 new unit tests pass; `ruff check` / `ruff format` clean.
- End-to-end against a live LLM (isolated data dir):
  - opt-in (`detect_contradictions=True`): writes a `contradicts` edge (`1985` ↔ `1990`) with the reason, both facts, and confidence 0.95.
  - default (`detect_contradictions=False`): 0 `contradicts` edges — existing behaviour unchanged.

## Design notes

- Opt-in and backward compatible: `detect_contradictions=False` by default, and the task is only appended to the pipeline when enabled.
- Non-destructive and fail-safe: only adds edges, returns its input unchanged, and swallows its own errors so contradiction detection can never break ingestion.

## Deferred (future work)

- Vector-similarity candidate selection for very large neighbourhoods.
- A dedicated search type / API to list flagged contradictions.
- Resolution strategies (keep-latest, mark-uncertain), coordinated with the sibling contradiction/dedup reworks (SDK-184, SDK-186 for gh #3631).

---

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -492,6 +492,7 @@ WEB_SCRAPER_MAX_DELAY=10.0
 # -- Contradiction detection (opt-in LLM check at the end of cognify) ---------
 # When on, cognify compares the facts this ingestion touched against the facts
 # already stored around them and records each conflict as a "contradicts" edge.
+# Applies to remember() too, which builds its graph through cognify().
 # Default off — when off the cognify pipeline is unchanged.
 #CONTRADICTION_DETECTION=false
 # Minimum LLM confidence required to flag a pair as contradictory.
--- a/.env.template
+++ b/.env.template
@@ -489,6 +489,16 @@ WEB_SCRAPER_MAX_DELAY=10.0
 # -- Triplet embedding (extra triplet-level vectors during cognify) -----------
 #TRIPLET_EMBEDDING=false
 
+# -- Contradiction detection (opt-in LLM check at the end of cognify) ---------
+# When on, cognify compares the facts this ingestion touched against the facts
+# already stored around them and records each conflict as a "contradicts" edge.
+# Default off — when off the cognify pipeline is unchanged.
+#CONTRADICTION_DETECTION=false
+# Minimum LLM confidence required to flag a pair as contradictory.
+#CONTRADICTION_CONFIDENCE_THRESHOLD=0.5
+# Cap on the facts sent to the LLM in a single check (the rest are logged and skipped).
+#CONTRADICTION_MAX_FACTS=500
+
 # -- Session cache (Redis backend + session/usage tuning) ---------------------
 # Used when CACHE_BACKEND=redis; also the host/port for a remote cache.
 #CACHE_HOST="localhost"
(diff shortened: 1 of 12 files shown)


Vasilije1990:
@rshkarin I don't think named params are the best way. Also, should be wired to cognee.remember

asadjan4611:
@Vasilije1990 @rshkarin  I have a  question first of all sorry if i disturbing you 
can i jump into this PR because i Love to work on it and also i have already work on it,also i have better understanding about this feature

Vasilije1990 (approved):

PR #4105, opened 2026-07-17
feat(graph): bi-temporal node validity with a working close_node (SDK-200)

rshkarin:
Review + rework of the community hackathon PR #3789 (by @koopatroopa787) for hackathon issue #3700 — "Temporal awareness in the Cognee brain". Linear: SDK-200.

Gives the native graph a notion of **fact validity over time**: a `valid_to` timestamp on every node, a `close_node()` task that stamps it when a fact is superseded, and an `is_valid()` staleness check — without deleting the old node. This fills a real gap: cognee's temporal stack records *when events happened* (reified `Timestamp`/`Interval`/`Event` nodes, `TemporalRetriever`, `SearchType.TEMPORAL`), but has no concept of a fact *ceasing to be valid* — today that exists only by delegating to Graphiti.

## Why the original PR didn't solve #3700

- **`close_node()` was a silent no-op on every backend.** It called `graph_engine.update_node(...)`, but `update_node` exists on no adapter and not on `GraphDBInterface`; the resulting `AttributeError` was swallowed by a bare `except Exception: pass`. So `valid_to` could only ever be set at node *construction*, never on an existing node — defeating the whole "close a superseded fact" purpose. No test covered the write path, so CI couldn't catch it.
- **`is_valid()` was unreferenced**, and the dedup "Contradiction detected" warning was **unreachable** on the production path (`get_graph_from_model` already deduplicates by id, first-wins, before nodes reach `deduplicate_nodes_and_edges`).
- A root `conftest.py` shipped an unrelated `mistralai`/`instructor` dev-environment import shim (also duplicated inline in the test).

## What this rework does

The original author's commit is preserved as the base (attribution intact); each improvement is a separate commit on top.

- **`fix(graph)`: make `close_node` actually persist.** Add `update_node(node_id, values)` to `GraphDBInterface` as an optional extension (base raises `NotImplementedError`, like `set_node_frequency_weights`) and implement it on the default `LadybugAdapter` using the same read-modify-write blob pattern as `set_node_truth_state`, so fields the caller did not name are never disturbed. `close_node(node_id) -> bool` returns whether it persisted; an unsupported backend logs a warning and returns `False` instead of silently succeeding. This follows the codebase norm where per-node patch methods roll out Ladybug-first.
- **`refactor(graph)`: drop the dormant dedup contradiction warning** — `deduplicate_nodes_and_edges` is restored to its original form (the added branch never executed). Node-level contradiction detection overlaps SDK-90 / SDK-184 and belongs in that design track.
- **`test(temporal)`: prove behavior end-to-end and remove the dev-env stub** — deletes `conftest.py` and the inline shim; the tests now include the coverage the original lacked: `close_node` must call a real `update_node` and return its result (regression guard for the no-op), and `LadybugAdapter.update_node` merges a patch without touching unnamed fields.

## Verification

- `pytest cognee/tests/unit/test_temporal_awareness.py` — 13 passed; adjacent suites (`test_add_data_points.py`, `test_ladybug_temporal.py`) — 57 passed together.
- `ruff format --check` + `ruff check` clean; `ty check` clean on all changed files.
- **Real end-to-end on the default Ladybug backend:** add node (`valid_to=None`, valid) → `close_node()` returns `True` → re-read shows `valid_to` written and `is_valid()` reports it stale; a control node stays valid; `close_node()` on a missing node returns `False` without raising.

## Out of scope (deferred)

- Retrieval-side consumption of `valid_to` (projecting it through `get_memory_fragment` and filtering/down-weighting in the graph-completion path behind an opt-in flag) — deferred, consistent with SDK-184 and E6's "design & measure" scope. Relates to SDK-90 / SDK-85.
- `update_node` on Neo4j / Postgres / Neptune — follow-up, mirroring the incremental Ladybug-first rollout of the sibling setters.
- Interop with the reified `Timestamp`/`Interval`/`Event` temporal search — relates to SDK-120.

Reworks #3789. Addresses #3700 (partial — see Out of scope).


GrowDev1:
Tested this branch in isolation (fresh clone at `9c822f01b`, separate from any live install) because we run Cognee as a memory substrate and bi-temporal validity matters a lot to us. The core primitive holds up well — `close_node()` genuinely persists, verified by an independent `get_node()` read rather than by trusting its return value, and it survives re-close, concurrent calls, and nonexistent-node input without corruption. `is_valid()`'s boundary semantics are correct, including the exactly-equal-to-`at_ms` case the PR's own tests don't cover.

Two things I think block merge, both reproduced from scratch twice (once while testing, once by an independent re-run):

### 1. The test file doesn't collect at all — 0/15 tests run

`cognee/tests/unit/test_temporal_awareness.py` still imports `_content_differs` from `cognee.modules.graph.utils.deduplicate_nodes_and_edges` at module level. `f18ce57d7` reverted that module back to its pre-PR form, so the name no longer exists in source (`grep -rn '_content_differs' cognee/` shows it only in the test file). Python fails the whole module on a top-level `ImportError`, so this takes down **all 15 tests**, not just the ~6 dedup-related ones — including the ones proving `close_node`/`is_valid`.

```
pytest cognee/tests/unit/test_temporal_awareness.py -v
# collected 0 items / 1 error
# ImportError: cannot import name '_content_differs' from 'cognee.modules.graph.utils.deduplicate_nodes_and_edges'
```

Worth flagging because `9c822f01b`'s own message says "Drop the tests for the removed dedup contradiction warning" — that commit's diff only deletes `conftest.py` and never touches the test file, so it reads as done when it isn't. Deleting the dead import plus the 6 dedup tests unblocks the other 8, which I re-ran in isolation and confirmed all pass on their own merits.

### 2. `LadybugAdapter.update_node()` silently drops `created_at`/`updated_at` on every call

This is the persistence path `close_node()` depends on, and it contradicts the PR's own documented guarantee ("leaving every field not named in *values* untouched").

`update_node()` excludes `{'id','name','type','created_at','updated_at'}` from the properties dict it reconstructs from `get_node()` before merging the patch and writing back — treating the two timestamps as native graph columns that survive separately. They aren't: `get_node()`'s query is

```cypher
RETURN {id: n.id, name: n.name, type: n.type, properties: n.properties}
```

which never requests native `n.created_at`/`n.updated_at`. The values that do appear come from inside the properties JSON blob, where `DataPoint.model_dump()` puts them in `add_nodes`. Stripping them from that blob before rewriting it erases them permanently.

Reproduction against an isolated `LadybugAdapter(tmp_path)`:

```python
await adapter.add_nodes([node])
before = await adapter.get_node(str(node.id))   # created_at = 1784623794966
await close_node(str(node.id))                  # or update_node(id, {"valid_to": 123})
after  = await adapter.get_node(str(node.id))   # created_at = None
```

Unconditional, reproduces every time. For a bi-temporal feature specifically, losing creation timestamps is more than cosmetic — recency, decay, and "what did we know when" all key on them.

Fix either way works, but they have different follow-on cost: stop excluding `created_at`/`updated_at` from the reconstructed properties (only exclude the genuinely native `id`/`name`/`type`), **or**, if the intent really is for the graph engine to own those columns, then `get_node()` needs to start returning the native columns and `add_nodes` needs to stop duplicating them into the blob — otherwise there are two divergent sources of truth for the same field names.

### 3. Minor, probably intentional — no guard on re-closing

`close_node()` on an already-closed node silently overwrites `valid_to`, including with an *earlier* timestamp, moving the supersession point backward with no warning. Last-write-wins is a defensible design; it's just neither documented nor tested, so worth confirming it's the intended contract.

---

Recommendation is request-changes rather than reject — the bi-temporal mechanism itself is sound, and notably this doesn't repeat the silent-no-op bug from the predecessor PR. Happy to share the isolated harness (7 new tests plus the corrected working subset of the existing file) if that's useful.

rshkarin:
Thanks for the exceptionally thorough review — testing in isolation and verifying persistence via an independent `get_node()` read rather than trusting the return value is exactly right, and it caught two real bugs. Both are fixed and pushed on top as `4507d88` and `e6f6e69` (no force-push, so your review anchors stay intact).

### 1. Test collection (0/15) — fixed

Your diagnosis of the mechanism is spot on. The rewrite commit (`9c822f01b`) was meant to replace the test file, but its `git add <testfile> conftest.py` step hit a pathspec error on `conftest.py` (already `git rm`-ed), which aborts the *entire* `git add` — so only the earlier-staged `conftest.py` deletion was committed, and the stale file, still importing `_content_differs` (removed in `f18ce57d7`), stayed on the branch. My local run passed only because it exercised the working-tree copy, not the committed one — which is precisely the gap you flagged.

`e6f6e69` lands the intended rewrite: the `mistralai` stub and the dedup tests are gone, and it now collects — 14 tests, all passing.

### 2. `update_node` dropping `created_at`/`updated_at` — fixed

Reproduced exactly (`created_at` → `None` after every call). Root cause is as you described: `get_node()` returns those timestamps from *inside* the JSON blob (its `RETURN` selects only the native `id`/`name`/`type` columns plus `n.properties`), so excluding them when rebuilding the blob erased them permanently.

`4507d88` now excludes only `{id, name, type}` and carries `created_at`/`updated_at`/`version`/… through untouched. I also removed the separate native `n.updated_at` write, so there is a single source of truth — the blob that `get_node` actually reads — rather than a native column that silently diverges from it (the second option in your note). The real end-to-end check on the default Ladybug backend now asserts both timestamps survive a `close_node()`.

Related, for maintainers: `set_node_truth_state`'s `_build_node_truth_state_updates` excludes the same `{id,name,type,created_at,updated_at}` set, so it has the same latent strip — it just hasn't surfaced because nothing reads `created_at` after a truth-state write. Out of scope for this PR, but worth a follow-up.

### 3. Re-close (last-write-wins) — confirmed intentional, now documented + tested

It is last-write-wins, including moving the supersession point earlier. `close_node`'s docstring now states that contract and suggests guarding with `is_valid` first when idempotency is needed, with a test pinning the behavior.

Yes, please — I'd be glad to see your isolated harness; if any of those 7 cases aren't covered by what I added, I'll fold them in.

Issue #3629, opened 2026-06-28
Hackathon: Contradiction & dedup resolution — Approach C: LLM-judge canonicalization


Vasilije1990:
# 🏆 Hackathon — Contradiction & dedup resolution, Approach C: LLM-judge canonicalization

## The problem (shared across all 5 approach tickets)

As cognee ingests more data, the graph accumulates **duplicate entities** (same thing extracted under different names/spellings) and **contradictions** (conflicting property values or edges, often arriving at different times). Today, dedup is largely **identity-based** and contradiction handling is minimal.

**What exists to build on:**
- `cognee/modules/graph/utils/deduplicate_nodes_and_edges.py` — current node/edge dedup (used by `cognee/tasks/storage/add_data_points.py`).
- `cognee/infrastructure/engine/models/FieldAnnotations.py` — the `Dedup()` marker (UUID5 dedup key).
- POCs: `examples/pocs/disambiguation/`, `examples/pocs/post_extraction_canonicalization/`, `examples/pocs/prefetch_disambiguation/`.
- Related: contradiction-resolution is one of BEAM's memory abilities; issue #1831 (entity property conflicts during dedup).

**This is 1 of 5 parallel approach tickets** (#A–#E). Each explores a *different* strategy for contradiction + dedup resolution so we can compare them on the same fixtures and adopt the best (or a hybrid). Coordinate on a **shared evaluation set** of duplicate/contradiction cases.

### 📣 Propose before building
Comment with your design (where it hooks into ingestion/storage, merge/conflict policy, cost) before coding.

### Requirements (all approaches)
- ✅ Extend the existing dedup path; dont fork a parallel pipeline. Provenance preserved on merge (what merged into what, and why).
- ✅ **Test coverage** on a shared fixture of duplicates + contradictions (precision/recall of merges, correct conflict resolution); deterministic in CI via the mocked-LLM harness (#3601).
- ✅ Actionable, reversible merges where possible (audit trail; coordinate with #3604).

## Approach C — LLM-as-judge canonicalization (productize the POC)
Promote `examples/pocs/post_extraction_canonicalization/` (and the disambiguation POCs) into a real ingestion task: an LLM decides whether candidate entities are the same and **reconciles conflicting properties**, with confidence + rationale. Highest semantic accuracy; gate cost with blocking/batching so the LLM only judges genuine candidates.

Sibling approaches: A (deterministic), B (embeddings), D (graph-structure), E (temporal).

[...]

Issue #3628, opened 2026-06-28
Hackathon: Contradiction & dedup resolution — Approach B: embedding-similarity clustering


Vasilije1990:
# 🏆 Hackathon — Contradiction & dedup resolution, Approach B: embedding-similarity clustering

## The problem (shared across all 5 approach tickets)

As cognee ingests more data, the graph accumulates **duplicate entities** (same thing extracted under different names/spellings) and **contradictions** (conflicting property values or edges, often arriving at different times). Today, dedup is largely **identity-based** and contradiction handling is minimal.

**What exists to build on:**
- `cognee/modules/graph/utils/deduplicate_nodes_and_edges.py` — current node/edge dedup (used by `cognee/tasks/storage/add_data_points.py`).
- `cognee/infrastructure/engine/models/FieldAnnotations.py` — the `Dedup()` marker (UUID5 dedup key).
- POCs: `examples/pocs/disambiguation/`, `examples/pocs/post_extraction_canonicalization/`, `examples/pocs/prefetch_disambiguation/`.
- Related: contradiction-resolution is one of BEAM's memory abilities; issue #1831 (entity property conflicts during dedup).

**This is 1 of 5 parallel approach tickets** (#A–#E). Each explores a *different* strategy for contradiction + dedup resolution so we can compare them on the same fixtures and adopt the best (or a hybrid). Coordinate on a **shared evaluation set** of duplicate/contradiction cases.

### 📣 Propose before building
Comment with your design (where it hooks into ingestion/storage, merge/conflict policy, cost) before coding.

### Requirements (all approaches)
- ✅ Extend the existing dedup path; dont fork a parallel pipeline. Provenance preserved on merge (what merged into what, and why).
- ✅ **Test coverage** on a shared fixture of duplicates + contradictions (precision/recall of merges, correct conflict resolution); deterministic in CI via the mocked-LLM harness (#3601).
- ✅ Actionable, reversible merges where possible (audit trail; coordinate with #3604).

## Approach B — Vector/embedding similarity clustering
Use cognee'\''s existing embeddings to find fuzzy duplicates: blocking + similarity threshold to cluster candidate-duplicate entities, then merge. Handles spelling/phrasing variants the deterministic keys miss. Tune threshold for precision/recall; scales to large graphs.

Sibling approaches: A (deterministic), C (LLM canonicalization), D (graph-structure), E (temporal).

Anasyaseen92:
[...]

PR #4106, opened 2026-07-17
feat(cognify): entity canonicalization via LLM judge (SDK-198)

rshkarin:
Reworks community hackathon PR #3808 (by Yash Sainani) for hackathon issue #3629 — "Contradiction & dedup resolution — Approach C: LLM-judge canonicalization". The original author's 7 commits are preserved as the base (attribution intact); the review fixes are layered as 3 separate commits on top.

Linear: SDK-198

## What this adds

Opt-in (`ENTITY_CANONICALIZATION`, default **off**) LLM-judge entity canonicalization that runs during `cognify()`, between entity extraction and storage. Today dedup is identity-based — it only merges entities whose normalized names hash to the same UUID5, so "OpenAI" and "OpenAI Inc." survive as two nodes. This step:

1. **Blocks** candidate near-duplicate pairs by transient name-embedding cosine similarity (threshold-gated, capped, and logged — never silently truncated), so the LLM only judges genuine candidates.
2. **Judges** each pair with a batched, structured `CanonicalizationJudgment` (same/different + canonical name + reconciled description + confidence + rationale).
3. **Merges** confident matches by collapsing each duplicate onto a canonical winner, reusing the existing `deduplicate_nodes_and_edges` path (no parallel pipeline) — the losing entities take the winner's identity so the existing UUID5 dedup collapses them at storage. Provenance (`merged_aliases`, `merge_confidence`) is stamped on the survivor plus a structured `entity_canonicalization_merge` audit log.

When the flag is off, the cognify task list is byte-for-byte identical to today (there's a regression test asserting exactly that).

## Review findings addressed (the 3 rework commits)

An adversarial review of the merge primitive found that it was correct only for two-entity merges and only when the surviving object happened to be the first one `get_graph_from_model` traversed — but the persisted node is built from whichever component member is reached first in `chunk.contains` (extraction order, unrelated to winner selection). That left three silent, order-dependent defects, now fixed:

- **Dropped edges on 3+ member merges** — the pairwise mirror left every non-final loser with a stale relation union, so a stale member traversed first dropped later-merged members' edges.
- **Lost audit provenance** — `merged_aliases`/`merge_confidence`/`source_task`/`version` were stamped on the winner only and never mirrored, so they vanished from the persisted node whenever a loser was traversed first.
- **Order-dependent `is_a` type** — a survivor could inherit the loser's type and drop its own.

The fix collapses each component onto a deterministic winner and makes **every** member byte-identical, so the result is the same node regardless of traversal order. Two further robustness fixes: the canonical id is derived from the winner's own name (so a judge-chosen `canonical_name` can't collide with an unrelated entity's id and conflate entities never compared), and a failed embedder/judge call now falls back to persisting the unmerged graph instead of aborting the pipeline and discarding an already-extracted batch. A stale-docs/dead-code cleanup commit rounds it out.

Five new regression tests drive the **real** `get_graph_from_model` + `deduplicate_nodes_and_edges` with a loser traversed first; the order-dependent ones fail on the pre-fix code.

## Scope / limitations (deferred follow-ups)

- Reconciliation is **description-only** (`description` is the one reconcilable non-identity text field on `Entity` today); general per-property reconciliation is a follow-up.
- Blocking is **within the current cognify batch** (no cross-batch/cross-run dedup); exact-name duplicates still collapse cross-run via UUID5 identity.
- The **temporal cognify path** is not wired.
- Merges are auditable (survivor metadata + structured log) but not yet programmatically reversible; a full un-merge needs the dedicated audit table from #3604.

## Verification

- 42 deterministic unit tests for the feature (LLM and embedder both mocked, no API keys), plus the surrounding graph/engine/cognify suites — all green.
- End-to-end run through the **real** `extract_graph_and_summarize → canonicalize_entities → get_graph_from_model + deduplicate_nodes_and_edges` seam (only the two LLM calls + the DB edge-lookup mocked): "OpenAI" and "OpenAI Inc." collapse to one node that carries **both** entities' edges (incoming `leads` from Sam Altman and outgoing `created` to GPT-4) and the merge provenance — confirming the data flow is real, not a no-op.

## Type of Change

- New feature (non-breaking, opt-in, default off)
- Bug fix (merge correctness) + code refactoring

## DCO Affirmation


rshkarin:
CI status note (from automated maintenance): all substantive checks are green — build, the full OS/Python matrix (3.10–3.13), soft/custom-graph delete, pre-commit + lockfile, DCO, and `linear-issue-check`. The secret-gated heavy suites show `skipping` and the `Test Completion Status` / `Claude Code Review` aggregators show `cancelled`, which is the normal draft-PR pattern here.

Two infra flakes appeared and were diagnosed:
- **Kuzu Relational DB Migration Test** — flaked once, **passed on re-run**.
- **Distributed Cognee test (Modal)** — failing on a migration-startup race, not this diff: `asyncpg UndefinedTableError: relation global_database_version ... does not exist` → `MigrationError: database migration failed for the global database` (the framework notes it "retries automatically on the next call"). This change is opt-in (`ENTITY_CANONICALIZATION`, default off), adds no relational schema/migration, and leaves the default pipeline byte-identical, so it cannot cause an undefined migrations-bookkeeping table. The same Modal job fails intermittently on sibling draft PRs (e.g. #4104) and passes on others (#4105, #4107). Left as-is rather than re-run a third time.

lallenlowe:
Adding some real-world data here, in case it helps with a decision.

**Who we are.** MX runs cognee 1.5.4 as shared memory for our engineering teams. Agents call `remember` when they learn something (a decision, a root cause, an incident). There's no ontology and no alias list, just one dataset with about 4,800 entities. We've sent two fixes upstream from this deployment already (#4569, #4856).

**What we found when we ran dedup.**

The default `consolidate_entities` config (threshold 0.85) merged the wrong things 30–46% of the time. It collapsed ticket IDs like `mlp-333` and `mlp-332`, tool names like `cognee_remember` and `cognee_recall`, opposite phrases like `signal worth alarming on` and `signal not worth alarming on`, and every date. Those are exactly the entities that matter most in engineering memory.

We tried other thresholds. None helped. Embeddings put negations and neighboring IDs at 0.92+ similarity no matter what, so raising the bar just trades false merges for missed ones.

Exact name matching (threshold 1.0, plus type protection) was safe: 194 merges, zero mistakes, about 3 seconds. That's what we run today.

But it leaves a lot behind. There are still 576 pairs of same-type entities with near-identical names. Some are real duplicates (`institution` / `institutions`, `mx data centers` / `mx datacenters`). Some are traps (`big-cat` / `big-nat`, `aiil-1095` / `aiil-1094`). Nothing short of reading the descriptions can tell them apart.

So we tried having an LLM read the descriptions. On the 12 hardest pairs it got 12 out of 12 right, in about 4 seconds, for about 1,200 tokens. A full pass over all 576 pairs would cost around $0.14.

**Why we ended up on this PR.** Of the five hackathon approaches, this is the only one that handles both kinds of pair correctly: merge `institution` / `institutions`, keep `plum-2069` / `plum-2089` apart. It was built, tested, and green, then closed on 08-08 with #4136, #4107, and #4087, and I couldn't find a note saying why. The `prefetch_disambiguation` example (feed nearby entity names into the extraction prompt) was removed a few days later on the grounds that ontology canonicalization and `consolidate_entities` cover the problem. For a graph with no ontology, they don't. @moritzscheele made the same point on #3628.

**The question:** is entity resolution at write time still on the roadmap? If so, which shape do you prefer: this PR's post-extraction judge, or the prefetch approach? We're happy to do the work either way. We could rebase this PR onto current `dev`, bring back prefetch as an opt-in cognify step, or do both (prefetch as a cheap first filter, judge for what gets through). We can also contribute the 576-pair set as a real-world test fixture.

cc @rshkarin @lxobr

PR #4108, opened 2026-07-17
feat(provenance): add low-overhead edge evidence

Vasilije1990:
## Summary

- add an append-only `provenance_edge_evidence` sidecar that links graph assertions to source chunks, data revisions, and pipeline runs
- capture evidence in a bounded per-context buffer and bulk-persist it once per data item
- expose structured evidence for RAG and graph results while keeping graph-native source references authoritative for delete and rollback
- preserve original source URIs and add the relational migration/model wiring

## Architecture and performance

- the sidecar has no foreign keys or cascades, so graph deletion remains independent of the relational database
- capture adds no LLM or embedding calls; entries are deduplicated in memory and written with one `executemany` operation per data item
- the default flush threshold is 10,000 entries and capture can be disabled with `PROVENANCE_ENABLED=false`
- evidence lookup runs concurrently with completion, and graph completion no longer performs a post-answer vector search
- a synthetic capture benchmark measured about 14.5 ms for 1,000 evidence links

## Validation

- [x] 43 focused provenance/graph/search tests after rebasing onto current `dev`
- [x] 148 graph, provenance, and migration tests
- [x] 345 retrieval tests
- [x] 52 search and recall tests
- [x] targeted Ruff checks
- [x] `git diff --check`

## Follow-up before merge

- add an end-to-end ingest/query/delete contract test across the supported graph backends
- benchmark persistence and evidence lookup on production-sized documents and tune the default threshold if needed

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -610,6 +610,21 @@ WEB_SCRAPER_MAX_DELAY=10.0
 # -- Triplet embedding (extra triplet-level vectors during cognify) -----------
 #TRIPLET_EMBEDDING=false
 
+# -- Provenance (three independent systems, three flags) ----------------------
+# Which pipeline run / task wrote each graph node, stamped as source_* fields
+# on the node itself. lightweight (default) | deep | disabled.
+#COGNEE_PROVENANCE_MODE=lightweight
+# Audit ledger: append-only, hash-chained provenance_entries table written by
+# an opt-in cognify task. For compliance/lineage history, not retrieval.
+#PROVENANCE_TRACKING=false
+# Edge evidence: which document chunk supports each graph edge, stored in
+# provenance_edge_evidence and returned as citations when include_references
+# is on. Captured only for edges extracted from document chunks.
+#EDGE_EVIDENCE_ENABLED=true
+# Pending evidence rows per data item before an early bulk flush (min 100);
+# below the threshold everything is written once per data item.
+#EDGE_EVIDENCE_FLUSH_THRESHOLD=10000
+
 # -- Contradiction detection (opt-in LLM check at the end of cognify) ---------
 # When on, cognify compares the facts this ingestion touched against the facts
 # already stored around them and records each conflict as a "contradicts" edge.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -762,6 +762,21 @@ Supported code files (`.py`, `.go`, `.ts`, `.java`, `.rs`, … — the extension
 - **Opt-out per add**: `preferred_loaders={"text_loader": {}}` treats a code file as a plain document (chunking + LLM extraction).
 - **Whole repositories**: `remember(content_type="code")` remains the repo-level path (cross-file edges); the CODE route is per-file.
 
+### Provenance
+Cognee has five provenance mechanisms. They answer different questions and are controlled by three unrelated flags — do not confuse them:
+
+| # | Mechanism | Question it answers | Stored where | Flag (default) |
+|---|---|---|---|---|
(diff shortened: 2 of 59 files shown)


NMZivkovic:
The migration at `f3a7b9c1d2e4_add_provenance_edge_evidence.py:15` sets `down_revision = "b3d5f7a9c1e2"`, but dev already has `d1e2f3a4b5c6_add_pipeline_runs_status_index.py` on that parent. Merging gives two Alembic heads, so `command.upgrade(config, "head")` in `migrations/startup.py` raises "Multiple head revisions are present" and every process start fails with `MigrationError` — fresh installs included, which also means `provenance_edge_evidence` never gets created. Point it at the current head during the rebase. Note that Alembic files never conflict textually, so a plain rebase will leave this in place silently, and `test_alembic_chain_has_a_single_head` is green here only because the branch is stale.

Related and more dangerous: this branch is `CONFLICTING` on `run_tasks_data_item.py`, and dev has since replaced that function's body with `_drain_item_events`/`progress_state` (a0a8dc0c6, CLO-557). Resolving in this PR's favour — the obvious choice since it's the side that adds the flush — drops the `progress_state` parameter that `run_tasks.py::_run_item` passes positionally, so every cognify raises TypeError, and the add/cognify progress ticks stop being forwarded. Re-apply the flush inside dev's structure rather than taking either side wholesale, then re-run CI; the current green checks are all against the pre-rebase tree.

While you're rebasing, `run_tasks_data_item.py:329` — `await flush_context_provenance(ctx)` on the success path isn't guarded, so a locked SQLite file, a transient Postgres error or a deployment that applies migrations out of band turns a fully-ingested dataset into an errored run. The exception escapes into `asyncio.gather` (no `return_exceptions=True`) and takes the whole run down after every graph and vector write already landed. Same for the threshold flush in `capture.py`. The read path in `lookup.py` deliberately fails open with a comment saying so; the write path should match.

Third: `include_references=True` on GRAPH_COMPLETION silently loses its Evidence block for every corpus ingested before this PR, since `_append_graph_evidence` is now a no-op and the sidecar has no rows for those edges. It's also skipped whenever `dataset` is None, which `search.py:388` sets on any multi-dataset search. And when it does fire the format changed — the snippet is dropped and `chunk_id:` became `data_id: ..., chunk_id: ...`, which breaks anyone parsing it. Keeping the answer-grounded builder as a fallback when the sidecar returns nothing would make this safe for existing users; at minimum log a warning when references were asked for and none could be produced.

Also worth fixing: `_source_uri_from_input` at `ingest_data.py:107` resolves client-supplied basenames against the server cwd, so an HTTP upload of `report.pdf` gets stored as `file:///<server cwd>/report.pdf`, and `cognee.add("/etc/hosts is the file that maps names to addresses.")` stores `file:///etc/hosts%20is%20the%20file...` because the string is absolute. Nothing reads these URIs, so it's corrupted lineage rather than file access, but lineage is the point of the PR. Require `path.is_file()` for strings and an already-absolute filename for the object branch.

Last, `linear-issue-check` is failing because neither the title nor `feat/provenance-edge-evidence` carries an issue key. Also flagging that `COGNEE_PROVENANCE_MODE=disabled` does not turn the new capture off — the sidecar reads a separate `PROVENANCE_ENABLED`, and `@lru_cache` on the new `get_provenance_config` (which shadows the existing same-named pair) means it can't be flipped without a restart. For a default-on write path that's the escape hatch users will reach for.

Vasilije1990:
Thanks — confirmed. `f3a7b9c1d2e4` now revises `d1e2f3a4b5c6` (rebased onto current dev; single head verified by walking every revision's down_revision). Also found while writing the e2e test that dev's merged `extract_graph_and_summarize` task made capture a silent no-op (chunks nested under `TextSummary.made_from`); fixed in the second commit.

PR #4116, opened 2026-07-18
fix(windows): correct parent-death detection in cognee_db_workers workers

GrowDev1:
## Problem

`cognee_db_workers/harness.py`'s Windows fallback for detecting "did my parent die"
(`start_parent_liveness_watchdog`) polls `os.getppid()` waiting for it to change. That's
POSIX-only reparenting behavior (death of the parent causes reparenting to init/launchd, and
a ppid change is a reliable signal there).

**Windows never reparents an orphaned child and never updates the reported parent PID** —
`os.getppid()` on Windows returns the original parent's PID forever, dead or alive. This
means the watchdog's trigger condition can structurally never fire on Windows.

In practice this orphans every worker (`kuzu_worker.py`, `lancedb_worker.py`) spawned under
this harness the instant its parent is killed by anything other than a graceful exit — a
hard `TerminateProcess`/`taskkill /F`/`Stop-Process -Force`, which is exactly how a hung
parent process gets recovered from in the field. The worker just sits there holding
whatever exclusive resource lock it opened (e.g. a Kuzu/LanceDB database file lock), with
no self-termination path, until a human manually finds and kills it.

This was root-caused live during a real ~9-hour production outage: a parent process was
force-killed, its `cognee_db_workers` children never noticed, and each one held an
exclusive graph-DB file lock indefinitely, blocking every subsequent pipeline run until the
orphans were found and killed by hand.

## Fix

Add `start_windows_parent_death_watchdog(original_ppid)`: opens a Windows `HANDLE` to the
original parent PID once, at worker startup (called from `run_worker_loop` within
milliseconds of the worker's own creation, while the parent is provably still alive — it's
the one that just called `Process.start()`). Blocks on that `HANDLE` with
`WaitForSingleObject` instead of polling a numeric PID.

This is also strictly more correct than a `getppid()`-based check even where `getppid()`
otherwise works: a Windows `HANDLE` refers to one specific kernel process object for its
entire life. If the PID number is later reused by an unrelated process (a real risk on any
long-running system), the handle is untouched — it keeps pointing at the original process
object, and `WaitForSingleObject` only signals when *that* object terminates.

Only a confirmed-gone PID (`ERROR_INVALID_PARAMETER`, what `OpenProcess` returns for a
nonexistent PID) or a signaled handle (`WAIT_OBJECT_0`) triggers exit. Any other outcome —
`OpenProcess` failing for another reason (e.g. access denied in a hardened environment), or
`WaitForSingleObject` returning something other than `WAIT_OBJECT_0` — falls back to the
existing portable polling watchdog rather than assuming the parent is dead and killing a
worker whose parent may still be alive.

The POSIX/macOS path (`set_pdeathsig()` / the existing `start_parent_liveness_watchdog`
polling fallback) is untouched — `os.getppid()` polling is genuinely valid there.

## Testing

Real (non-mocked) subprocess spawn/hard-kill tests in
`cognee_db_workers/tests/test_windows_parent_death_watchdog.py`, Windows-only
(`skipif(sys.platform != "win32")`):

- Self-exits quickly (sub-second) when the parent is hard-killed
- Does NOT self-exit while the parent is still alive
- An already-dead watched PID exits immediately rather than hanging
- PID reuse does not confuse the handle-based wait (a fresh unrelated process reusing the
  original PID does not cause a false "parent alive" read)
- Returns `False` on non-Windows platforms (no-op, falls through to the existing watchdog)

All 5 pass locally on Windows. Also verified end-to-end against a live-deployed copy of
this fix (a real worker process importing the patched module, hard-killing its real OS
parent, confirming self-exit in ~0.05s versus never, before).

### DIFF
--- a/cognee_db_workers/harness.py
+++ b/cognee_db_workers/harness.py
@@ -312,6 +312,118 @@ def start_parent_liveness_watchdog(poll_interval: float = 1.0) -> None:
     t.start()
 
 
+def start_windows_parent_death_watchdog(original_ppid: int) -> bool:
+    """Windows-correct replacement trigger for ``start_parent_liveness_watchdog``.
+
+    Windows never reparents an orphaned child (unlike POSIX, where death of
+    the parent causes reparenting to init/launchd and a ppid change is a
+    reliable signal) — ``os.getppid()`` on Windows returns the ORIGINAL
+    parent's PID forever, dead or alive, so the polling-ppid-change trigger
+    ``start_parent_liveness_watchdog`` relies on can structurally never fire
+    on Windows. In practice this orphans every worker spawned under this
+    harness the instant its parent is killed anything other than gracefully
+    (a hard ``TerminateProcess``/``taskkill /F``/``Stop-Process -Force``,
+    exactly how a hung parent gets recovered from in the field) — the worker
+    just sits there holding whatever exclusive resource lock it opened
+    (e.g. a Kuzu/LanceDB database file lock) with no self-termination path,
+    until a human manually finds and kills it.
+
+    Opens a Windows HANDLE to the original parent PID once, at worker
+    startup (called from ``run_worker_loop`` within milliseconds of this
+    process's own creation, while the parent is provably still alive — it's
+    the one that just called ``Process.start()``). Then blocks on that
+    HANDLE with ``WaitForSingleObject``, not on the numeric PID.
+
+    This survives PID reuse (a real risk on any long-running system): a
+    Windows HANDLE refers to one specific kernel process object for its
+    entire life. If the PID number is later reused by an unrelated process,
+    the handle is untouched — it keeps pointing at the ORIGINAL process
+    object, and ``WaitForSingleObject`` only signals when THAT object
+    terminates. This is an OS guarantee, not a heuristic, and is strictly
+    stronger than a ``getppid()``-based check even on platforms where
+    ``getppid()`` works correctly.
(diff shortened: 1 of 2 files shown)


dexters1:
Thanks for the PR @GrowDev1! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

GrowDev1:
Thanks @dexters1, done. Rebased onto dev and switched the base.

While I was in there I made the fix smaller. On Windows, multiprocessing's spawn bootstrap already holds a handle to the parent process (`parent_process().sentinel`), so the watchdog now waits on that handle instead of opening its own through ctypes. Before exiting, it checks the parent's exit code through a fresh handle. A stray handle can only leave a worker running, never kill one whose parent is alive.

I also moved the test into cognee/tests/unit/infrastructure/databases/ so the Windows leg of the OS matrix runs it. The old cognee_db_workers/tests/ location isn't picked up by any workflow, same as the point raised on #4649. This touches the same few lines in run_worker_loop as #4649, so whichever lands second will need a small rebase. I'll do that one.

GrowDev1:
Hi, this PR is still active on our side. It is rebased onto `dev` and we are working through the review feedback, with an update planned before Monday. Could it please stay open (or get the `keep-open` label) so the stale bot does not close it in the meantime? Thanks for the reviews so far.

PR #4149, opened 2026-07-20
fix(ingestion): derive basename from Windows paths in classify() [SDK-235]

rshkarin:
## What

Follow-up to #3588 (Windows basename fix for Mistral transcription, gh #3587).

While reviewing #3588 I found the **same separator bug still live** in `classify()` — the exact function the original issue believed was "already fixed". `classify()` derived the `BinaryData` name with:

```python
str(data.name).split("/")[-1]
```

On Windows, an opened file's `.name` is a backslash path (e.g. `C:\Users\me\audio\clip.mp3`) with no forward slashes, so the split kept the **entire path** as the `BinaryData` name instead of the basename `clip.mp3`.

## Fix

```diff
-        return BinaryData(data, filename if filename else str(data.name).split("/")[-1])
+        derived_name = str(data.name).replace("\\", "/").split("/")[-1]
+        return BinaryData(data, filename if filename else derived_name)
```

Normalizes both `\` and `/` before taking the basename — identical to the `_normalize_filename` helper in `cognee/tasks/ingestion/utils.py` and the Mistral fix in #3588. Deliberately not `os.path.basename`, which is platform-dependent (on Linux it does not split backslashes).

A repo-wide grep confirmed this was the only other genuine instance of this class of bug (the `.split("/")` uses in `measure.py` and `RDFLibOntologyResolver.py` are model-id / URI splits where `/` is the correct separator).

## Testing

Adds `cognee/tests/unit/modules/ingestion/test_classify_basename.py`:

- Windows-style `.name` (`C:\Users\me\audio\clip.mp3`) → name == `clip.mp3`
- POSIX-style `.name` (`/home/me/audio/clip.mp3`) → name == `clip.mp3`
- Explicit `filename` arg takes precedence (backslash `.name` ignored)

The test builds a `BufferedReader` over a minimal raw stream whose `.name` returns a Windows-style path regardless of host, so a revert to splitting on `/` only fails on any OS. Verified locally: 3 pass with the fix; reverting fails the Windows case on macOS.

Relates to #3587. Follow-up to #3588.

### DIFF
--- a/cognee/modules/ingestion/classify.py
+++ b/cognee/modules/ingestion/classify.py
@@ -14,7 +14,11 @@ def classify(
         return TextData(data)
 
     if isinstance(data, BufferedReader) or isinstance(data, SpooledTemporaryFile):
-        return BinaryData(data, filename if filename else str(data.name).split("/")[-1])
+        # Normalize Windows ("\") and POSIX ("/") separators before taking the
+        # basename; on Windows `data.name` is a backslash path, so splitting on
+        # "/" alone would keep the full path as the file name.
+        derived_name = str(data.name).replace("\\", "/").split("/")[-1]
+        return BinaryData(data, filename if filename else derived_name)
 
     try:
         from s3fs import S3File
--- a/cognee/tests/unit/modules/ingestion/test_classify_basename.py
+++ b/cognee/tests/unit/modules/ingestion/test_classify_basename.py
@@ -1,0 +1,56 @@
+"""Regression test for filename derivation in ``classify``.
+
+``classify`` derived the ``BinaryData`` name with ``str(data.name).split("/")[-1]``.
+On Windows, an opened file's ``.name`` is a backslash path (e.g.
+``C:\\audio\\clip.mp3``) with no forward slashes, so the split kept the *entire
+path* as the name instead of the basename.
+
+The fix normalizes both ``\\`` and ``/`` before taking the basename (matching the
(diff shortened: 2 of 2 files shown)


rshkarin:
Closing as **redundant and superseded**.

While reviewing #3588 I flagged `classify.py` as a follow-up and opened this PR — but I missed that a dedicated issue (#3341) and community PR **#3342** by @YashasviThakur already cover this exact line, and that the canonical fix is now #4157.

Two reasons this PR should not land:

1. **Superseded by #4157**, which is a strict superset: it fixes the *actual* stored-name derivation path (`get_file_metadata` — the real user-facing defect: percent-encoding leak + non-OS-agnostic basename), folds in the same `classify.py` normalization, and credits @YashasviThakur (#3341/#3342). This PR touches no file absent from #4157 and would merge-conflict with it on the shared `classify.py` region.
2. **The `classify.py` split is latent.** In the real ingestion pipeline every stream reaching `classify()` is a `FileBufferedReader` named with a forward-slash `file://`/`s3://` URI (`Path(...).as_uri()`), and the persisted `Data.name` comes from `get_file_metadata`'s `Path(file.name).stem`, not from classify's name (which is only a `None`-fallback that never fires). So the original `split("/")` already returns the correct basename here.

Canonical fix: **#4157**. Credit for the classify normalization stays with @YashasviThakur (#3341 / #3342).

PR #3750, opened 2026-07-01
Fix: Call super().__init__() in EntityNotFoundError and NodesetFilterNotSupportedError

Kkt04:
**Fixes: #3749**
**Description: **
**EntityNotFoundError** and **NodesetFilterNotSupportedError** in `cognee/infrastructure/databases/exceptions/exceptions.py` were `overriding __init__()` without calling `super().__init__()`, which:
1. Left self.args as an empty tuple () — breaking repr(), str(), and exception chaining
2. Skipped the centralized logging in CogneeApiError.__init__(), so these errors were never logged

**Fix:**
Added `super().__init__(message, name, status_code, log=False)` to both classes — consistent with all other exceptions in the file. `log=False` is used since these are validation/configuration errors that shouldn't produce log noise.

**Changes:**
- `cognee/infrastructure/databases/exceptions/exceptions.py:46 `— EntityNotFoundError.__init__ now calls super().__init__
- `cognee/infrastructure/databases/exceptions/exceptions.py:81` — **NodesetFilterNotSupportedError.__init__** now calls `super().__init__`

**Testing:**
- uv run python `test_exceptions_fix.py` — all 6 checks pass (args, repr, chaining, consistency)
- Existing unit tests using **EntityNotFoundError** pass unchanged
image
image

Issue #3749, opened 2026-07-01
[Bug]: EntityNotFoundError and NodesetFilterNotSupportedError bypass super().__init__(), breaking logging and exception chain


Kkt04:
### Bug Description

Two exception classes in `cognee/infrastructure/databases/exceptions/exceptions.py` — **EntityNotFoundError** and **NodesetFilterNotSupportedError** — override` __init__()` but do not call `super().__init__()`. The **`super()`** call is either commented out or simply omitted.
The parent class `CogneeApiError.__init__()` (`cognee/exceptions/exceptions.py:10-32`) does three things: 
(1) `sets self.message`, `self.name`,` self.status_code`
(2) logs every error via `logger.error(...)`
(3) calls `Exception.__init__`(self.message, self.name) to properly initialize the base exception. By skipping `super().__init__()`, both the logging and the Exception initialization are bypassed.

Note: **CogneeApiError** already has a `log=False` parameter for cases where logging is not desired. The proper fix is `super().__init__`(message, name, status_code, log=False), not skipping super() entirely.

### Steps to Reproduce

1. Raise an **EntityNotFoundError**: 
```
from cognee.infrastructure.databases.exceptions.exceptions import EntityNotFoundError
try:
    raise EntityNotFoundError()
except EntityNotFoundError as e:
    print(e.args)   # Empty tuple — broken
    print(repr(e))  # No message in Exception repr
```
2. Check application logs — no **EntityNotFoundError** or **NodesetFilterNotSupportedError** will appear, even though **CogneeApiError** is designed to log every exception at creation time.
3. Compare with sibling **EntityAlreadyExistsError** or any other exception in the same file — they all work correctly.

### Expected Behavior

1. **EntityNotFoundError** and **NodesetFilterNotSupportedError** should call `super().__init__()` so `Exception.__init__()` sets `self.args` properly.
2. If logging is deemed undesirable for these specific exceptions, `log=False` should be passed to the parent: `super().__init__(message, name, status_code, log=False).`
3. Behavior should be consistent with all other exceptions in the file (e.g., EntityAlreadyExistsError, MissingQueryParameterError, CacheConnectionError).

### Actual Behavior

-  `self.args` is an empty tuple () instead of (message, name) — breaks `repr()`, `str()` (though __str__ is overridden), and exception chaining.
- Errors are never logged, defeating the purpose of the centralized logging in `CogneeApiError.__init__()`.
[...]

PR #4151, opened 2026-07-20
fix(exceptions): call super().__init__ in CogneeError subclasses [SDK-240]

rshkarin:
## Summary

Fixes GitHub #3749 (Linear **SDK-240**, related to **SDK-217**). Three `CogneeApiError` subclasses overrode `__init__` but never called `super().__init__()`, so `Exception.args` stayed empty `()` — breaking `repr()`, exception chaining, and the centralized logging in `CogneeApiError.__init__`.

An AST sweep of every family subclass in the repo confirms **exactly three** offenders on `dev`:

| Class | File | In #3749? |
|-------|------|-----------|
| `EntityNotFoundError` | `infrastructure/databases/exceptions/exceptions.py` | ✅ named |
| `NodesetFilterNotSupportedError` | `infrastructure/databases/exceptions/exceptions.py` | ✅ named |
| `WrongTaskTypeError` | `modules/pipelines/exceptions/tasks.py` | ❌ same bug, not named in the issue |

## The `log=False` nuance

`EntityNotFoundError` is raised in **routine control flow** and caught in ~10 places (`cli/user_resolution.py`, `migrations/runner.py`, `data/deletion/prune_system.py`, `brute_force_triplet_search.py`, `CogneeGraph` empty-graph/neighborhood cases, `create_authorized_dataset.py`, …). The original `# TODO … we shouldn't log error` comment was intentional. So the two database exceptions pass `super().__init__(..., log=False)` — restoring `args`/chaining **without** flooding logs with ERROR lines for expected conditions (matching the existing `UnsupportedProvenanceCapability` idiom). `WrongTaskTypeError` is a genuine programming error and keeps default logging.

## Regression guard

Adds `cognee/tests/unit/exceptions/test_cognee_error_contract.py` with two complementary checks:

1. **Static AST sweep** of every `.py` in the package — import-free, so it also covers `CogneeApiError` subclasses that live **outside** `*/exceptions/` (e.g. `modules/retrieval/code_retriever.py`, `tasks/web_scraper/ssrf_protection.py`, `tasks/code_graph/*`) and modules behind optional extras.
2. **Runtime contract** — instantiates every importable subclass and asserts `args` is populated, `str()` is non-empty, and `raise … from cause` preserves `__cause__`.

Verified both guards **fail** on the pre-fix code and **pass** after; existing `test_user_resolution.py` still green.

## Credit

Builds on the approach in #4138 by @silasbrookshaha, extending it to the third affected class (`WrongTaskTypeError`) and closing a coverage gap in the original reflection test's importer (subclasses outside `*/exceptions/`). Supersedes the partial fixes in #3750, #3803, #3837 (the latter two drop `log=False`, reintroducing ERROR-log noise for routinely-caught `EntityNotFoundError`).

## Checklist
- [x] Bug fix (non-breaking)
- [x] Tests added; verified they fail on the buggy code
- [x] `ruff format` / `ruff check` clean

I have read the CLA Document and I hereby sign the CLA.

### DIFF
--- a/.github/workflows/community_tests.yml
+++ b/.github/workflows/community_tests.yml
@@ -109,6 +109,7 @@ jobs:
             cognee/tests/unit/infrastructure/test_rate_limiting_retry.py \
             cognee/tests/unit/tasks/storage/test_add_data_points.py \
             cognee/tests/unit/shared/ \
+            cognee/tests/unit/exceptions/ \
             -v --tb=short
 
   cli-unit-tests:
--- a/cognee/infrastructure/databases/exceptions/exceptions.py
+++ b/cognee/infrastructure/databases/exceptions/exceptions.py
@@ -60,10 +60,11 @@ class EntityNotFoundError(CogneeValidationError):
         name: str = "EntityNotFoundError",
         status_code=status.HTTP_404_NOT_FOUND,
     ):
-        self.message = message
-        self.name = name
-        self.status_code = status_code
-        # super().__init__(message, name, status_code) :TODO: This is not an error anymore with the dynamic exception handling therefore we shouldn't log error
+        # log=False: this is raised in routine control flow (caught in user
+        # resolution, migrations, pruning, graph projection, etc.), so logging
+        # every occurrence at ERROR would be noise. super() still populates
+        # Exception.args and preserves chaining.
+        super().__init__(message, name, status_code, log=False)
 
 
 class EntityAlreadyExistsError(CogneeValidationError):
@@ -98,9 +99,9 @@ class NodesetFilterNotSupportedError(CogneeConfigurationError):
         name: str = "NodeSetFilterNotSupportedError",
         status_code=status.HTTP_404_NOT_FOUND,
     ):
-        self.message = message
-        self.name = name
-        self.status_code = status_code
+        # log=False: a missing nodeset filter is a capability difference between
+        # backends, not an application error worth logging on every raise.
+        super().__init__(message, name, status_code, log=False)
 
 
 class EmbeddingException(CogneeConfigurationError):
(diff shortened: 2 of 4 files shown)


rshkarin:
## Verification pass (adversarial, multi-lens)

Ran a 4-lens audit (completeness / test-robustness / behavioral-regression / CI-gates) with adversarial verification of each material finding.

**Result: fix is complete and correct; no merge blockers.**
- **Completeness** — independently re-derived the offender set via AST fixpoint (Name + dotted bases, multi-level chains): `dev` has exactly these 3 offenders, worktree has **0**. The `cognee/modules/graph/exceptions/EntityNotFoundError` is a *distinct*, already-correct class (calls `super()`), rightly out of scope.
- **Regression** — 104 tests across every dependent suite pass (`test_user_resolution`, `test_remove_user_from_tenant`, `test_create_authorized_dataset_parent_share`, `cognee_graph_test`, `test_brute_force_triplet_search`, contract). `str()`/`repr()` are byte-identical pre/post (super sets the same attributes), so `pytest.raises(match=…)` sites are unaffected.
- **CI/lint/type** — `ruff` clean; `ty` include-list excludes these paths (can't break the gate); no `__init__.py` needed.

**Hardening applied in `3f8354b8` from the audit:**
1. The static guard now only accepts a base call whose target is `super()` or a *family* class — a future subclass calling the grandparent `Exception.__init__` or a mixin's `__init__` (which would bypass `CogneeApiError.__init__`) is now flagged. Proven against synthetic bypass shapes.
2. Family discovery + base-call check now also recognise dotted `module.Base` references (future-proofing; none exist today).
3. Runtime assertion tightened to `args == (message, name)` (the exact tuple `CogneeApiError.__init__` forwards).
4. Added `cognee/tests/unit/exceptions/` to the **keyless** `community_tests` unit allowlist so the guard runs on fork PRs — where new exceptions actually arrive — not only in the secrets-bearing full-tree job.

PR #4157, opened 2026-07-20
fix(ingestion): derive clean cross-platform document names [SDK-237]

rshkarin:
## What & why

Follow-up to #3342 / #3341 (Windows document-name handling), scoped to the code
that actually determines the stored document name.

**Root cause (deeper than #3341).** The symptom in #3341 — "on Windows the document
name is the full `C:\...` path" — points at `classify()`, but on current `dev` the
stored `Data.name` is derived in `get_file_metadata` as `Path(file.name).stem`. In
the ingestion pipeline `file.name` is a percent-encoded `file://` URI (every file is
wrapped in a `FileBufferedReader` named `Path(full_path).as_uri()`), so:

* `classify()`'s `split("/")` derivation is effectively latent — it only ever sees a
  forward-slash URI, so the buggy split already returns the right basename;
* the real derivation, `Path(file_path).stem`, has two platform-independent defects:
  * percent-escapes leak into the name — `Annual Report.pdf` is stored as `Annual%20Report`;
  * it is not OS-agnostic — a Windows-style backslash path on POSIX yields the whole
    path as the "stem" (and vice versa).

## Change

* **`get_file_metadata`**: new `_derive_basename` helper — percent-decodes `file://`
  URIs and extracts the basename treating both `/` and `\` as separators, regardless
  of host OS. The extension-less stem semantics are preserved (the extension is stored
  separately in `FileMetadata["extension"]`).
* **`classify()`**: folds in the defensive `\`-normalization from #3342
  (credit @YashasviThakur) so the fallback name derivation is correct on Windows too.
* **Tests**: cross-platform-strict unit tests for both — they fail on the pre-fix
  behavior on *any* OS (backslash-literal / percent-encoded inputs), so CI catches a
  regression regardless of the runner's platform.

## Verification (local)

* New unit tests: **15 passed**. The `get_file_metadata` integration test fails on
  pre-fix behavior with `'Annual%20Report%202025' != 'Annual Report 2025'`; the
  `classify` Windows test fails on pre-fix behavior even on macOS.
* Real `cognee.add()` end-to-end: `"Annual Report 2025.txt"` → `Data.name =
  "Annual Report 2025"` (was `Annual%20Report%202025`); `"Natural_language_processing_copy.txt"`
  → `"Natural_language_processing_copy"` (stem preserved — `test_deduplication`'s
  assertion still holds).
* `_derive_basename` is byte-identical to the old `Path().stem` for every non-buggy
  input (10 parametrized cases).
* `ruff format --check` and `ruff check` clean. No new failures in the ingestion/files
  unit sweep (2 pre-existing `unstructured`-extra failures also fail on `dev`).

Linear: **SDK-237**. Folds in and supersedes #3342.

### DIFF
--- a/cognee/infrastructure/files/utils/get_file_metadata.py
+++ b/cognee/infrastructure/files/utils/get_file_metadata.py
@@ -1,6 +1,6 @@
 import io
 import os.path
-from pathlib import Path
+from pathlib import PureWindowsPath
 from typing import BinaryIO, Optional, TypedDict
 from urllib.parse import unquote, urlparse
 
@@ -24,7 +24,8 @@ def _derive_basename(file_path: str) -> Optional[str]:
     "stem", and vice versa). This normalizes both cases:
 
     * percent-decodes ``file://`` URIs (so "Report%20Q1.pdf" -> "Report Q1"),
-    * treats both "/" and "\\" as separators regardless of host OS,
+    * resolves the basename + stem with ``PureWindowsPath``, which treats both "/"
+      and "\\" as separators on every host OS (so it is OS-agnostic),
     * strips a single trailing extension, preserving the prior ``Path(...).stem``
       semantics (the extension is stored separately in ``FileMetadata["extension"]``).
     """
@@ -32,9 +33,11 @@ def _derive_basename(file_path: str) -> Optional[str]:
     if candidate.startswith("file://"):
         candidate = unquote(urlparse(candidate).path)
 
-    basename = candidate.replace("\\", "/").rstrip("/").rsplit("/", 1)[-1]
-    # ``basename`` is now separator-free, so Path(...).stem is OS-independent here.
-    return Path(basename).stem or None
+    # ``PureWindowsPath`` treats both "/" and "\\" as separators regardless of host
+    # OS, so basename + stem resolution is OS-independent (a plain ``Path`` /
+    # ``PurePosixPath`` would not split "\\"). Percent-decoding above is URL work
+    # that pathlib cannot do.
+    return PureWindowsPath(candidate).stem or None
 
 
 class FileMetadata(TypedDict):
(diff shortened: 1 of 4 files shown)


YashasviThakur:
Thanks for digging into this, @rshkarin — good catch that the real issue lives in `get_file_metadata`'s `Path(file.name).stem` rather than `classify()`. The percent-encoding leak explains behavior I hadn't traced that far back.

Happy to have the `\`-normalization folded in here rather than landing as a separate patch — this is the more complete fix. I'll close #3342 in favor of this once it merges, unless you'd prefer I close it now.

cognee/infrastructure/files/utils/get_file_metadata.py
```diff
@@ -11,6 +12,31 @@
...
+    * percent-decodes ``file://`` URIs (so "Report%20Q1.pdf" -> "Report Q1"),
+    * treats both "/" and "\\" as separators regardless of host OS,
+    * strips a single trailing extension, preserving the prior ``Path(...).stem``
+      semantics (the extension is stored separately in ``FileMetadata["extension"]``).
+    """
+    candidate = file_path
+    if candidate.startswith("file://"):
+        candidate = unquote(urlparse(candidate).path)
+
+    basename = candidate.replace("\\", "/").rstrip("/").rsplit("/", 1)[-1]
```
dexters1: Can we use the Path library to handle path and stem resolution instead of manually handling it like here? (after the file:// prefix was removed)

dexters1 (changes requested):
Lets use the Path library to handle file paths and basename derivation (after the file:// prefix is removed)

cognee/infrastructure/files/utils/get_file_metadata.py
rshkarin: Done 👍

Vasilije1990 (approved):

PR #4167, opened 2026-07-20
fix(pipelines): lock datasets across workers

GautamSharma99:
## Summary

  Pipeline execution currently uses an asyncio.Lock keyed by dataset ID. Because this registry exists only within one Python
  process, separate FastAPI workers can run pipelines against the same dataset concurrently.

  This PR introduces cross-worker locking for per-dataset pipeline execution while retaining an in-process lock as a lightweight
  first layer.

  ## Problem

  Concurrent pipeline runs for the same dataset can currently be started by different FastAPI workers. The process-local lock
  does not coordinate those workers, which may lead to:

  - duplicate extraction and processing;
  - concurrent graph and vector writes;
  - inconsistent pipeline run statuses;
  - database write conflicts.

  The existing _dataset_locks registry also retains one asyncio.Lock for every dataset processed and never removes unused
  entries.

  ## Changes

  - Added a database-aware dataset_pipeline_lock abstraction.
  - Added PostgreSQL cross-worker locking using session-level advisory locks.
      - Lock keys are deterministic, signed 64-bit values derived from dataset UUIDs.
      - The advisory lock uses a dedicated database connection for its lifetime.
      - The transaction opened while acquiring the lock is committed immediately.
      - The lock is released in finally, including when pipeline execution fails.

  - Added host-wide locking for file-backed SQLite and Turso using OS advisory file locks.
      - Lock acquisition uses non-blocking polling so the event loop is not blocked.
      - Lock paths use 256 deterministic stripes, preventing an unbounded number of lock files.
      - Stripe collisions may serialize unrelated datasets but cannot weaken same-dataset exclusion.

  - Retained an in-process asyncio.Lock as an optimization so only one task per worker contends for the cross-worker resource.
  - Made the local lock registry reference-counted.
      - Entries are removed after the final holder or waiter exits.
      - Cancelled waiters do not leave stale entries.

  - Preserved the existing re-entrant behavior for nested pipelines operating on the same dataset.
  - Updated the concurrent cognify test documentation to reflect cross-worker locking.

  ## Backend Behavior

   Relational backend          Coordination scope
  ━━━━━━━━━━━━━━━━━━━━━━━━━━  ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
   PostgreSQL                  Across workers and hosts sharing the same database
  ──────────────────────────  ─────────────────────────────────────────────────────────────────
   File-backed SQLite/Turso    Across workers on the same host
  ──────────────────────────  ─────────────────────────────────────────────────────────────────
   In-memory SQLite            Current process only, because the database itself is not shared

  ## Testing

  Added tests covering:

  - serialization of concurrent local callers;
  - cleanup of local registry entries;
  - cleanup when a waiting task is cancelled;
  - SQLite advisory-lock contention;
  - cancellation without orphaning a SQLite lock;
  - bounded SQLite lock-file stripes;
  - PostgreSQL advisory-lock release after pipeline failure;
  - lock lifetime around the complete pipeline generator;
  - nested same-dataset pipeline execution without lock reacquisition.

  Commands run:

  uv run ruff check cognee/infrastructure/locks/dataset_pipeline_lock.py \
    cognee/modules/pipelines/operations/pipeline.py \
    cognee/tests/unit/infrastructure/locks/test_dataset_pipeline_lock.py \
    cognee/tests/unit/modules/pipelines/test_pipeline_dataset_lock.py \
    cognee/tests/test_concurrent_cognify_e2e.py

  uv run ruff format --check cognee/infrastructure/locks/dataset_pipeline_lock.py \
    cognee/modules/pipelines/operations/pipeline.py \
    cognee/tests/unit/infrastructure/locks/test_dataset_pipeline_lock.py \
    cognee/tests/unit/modules/pipelines/test_pipeline_dataset_lock.py \
    cognee/tests/test_concurrent_cognify_e2e.py

  uv run pytest \
    cognee/tests/unit/modules/pipelines \
    cognee/tests/unit/modules/cognify \
    cognee/tests/unit/modules/memify_tasks \
    cognee/tests/unit/memify_pipelines \
    -q --timeout=60

  Result: `132 passed`.

  The complete unit suite could not be collected in the current local environment because the optional `plotly` and `neo4j`
  packages are not installed. The failure occurs during test collection and is unrelated to these changes.

  ## Impact

  - Prevents simultaneous same-dataset pipeline execution across FastAPI workers.
  - Reduces the risk of duplicate processing and conflicting graph/vector writes.
  - Prevents unbounded growth of the process-local dataset lock registry.
  - Does not change the public API, MCP server, or frontend.
  - Different datasets can still execute concurrently, except for occasional SQLite lock-stripe collisions.

  ## Checklist

  - [x] Code formatted with Ruff
  - [x] Ruff checks pass
  - [x] Targeted regression tests added
  - [x] Relevant pipeline, cognify, and memify tests pass
  - [x] Existing nested pipeline behavior preserved
  - [x] Commit signed
  - [x] DCO sign-off included

### DIFF
--- a/cognee/infrastructure/locks/dataset_pipeline_lock.py
+++ b/cognee/infrastructure/locks/dataset_pipeline_lock.py
@@ -1,0 +1,142 @@
+"""Cross-worker locking for pipeline runs that target the same dataset."""
+
+import asyncio
+import os
+import threading
+from collections.abc import AsyncGenerator
+from contextlib import asynccontextmanager
+from dataclasses import dataclass
+from hashlib import sha256
+from pathlib import Path
+from uuid import UUID
+
+from sqlalchemy import text
+
+from cognee.infrastructure.databases.relational import get_relational_engine
+
+
+_SQLITE_LOCK_STRIPES = 256
+_SQLITE_LOCK_POLL_INTERVAL_SECONDS = 0.05
+
+
+@dataclass
+class _LocalLockEntry:
+    lock: asyncio.Lock
+    users: int = 0
+
+
+# Protecting this small registry with a threading lock avoids binding the guard to
+# one event loop. The registry key includes the loop because asyncio locks cannot
+# safely be shared between loops (which is common in SDK users and unit tests).
+_local_locks: dict[tuple[asyncio.AbstractEventLoop, UUID], _LocalLockEntry] = {}
+_local_locks_guard = threading.Lock()
+
+
+def _postgres_lock_key(dataset_id: UUID) -> int:
+    """Return a deterministic signed bigint key namespaced to pipeline locks."""
+    digest = sha256(b"cognee:dataset-pipeline:" + dataset_id.bytes).digest()
+    return int.from_bytes(digest[:8], byteorder="big", signed=True)
+
+
+def _sqlite_lock_path(engine, dataset_id: UUID) -> Path | None:
+    """Return one of a bounded number of lock files next to the SQLite database."""
+    database = engine.url.database
+    if not database or database == ":memory:":
+        return None
+
+    # A fixed number of stripes prevents an unbounded lock-file leak. A hash
+    # collision only serializes two unrelated datasets; it cannot compromise
+    # same-dataset exclusion.
+    stripe = sha256(dataset_id.bytes).digest()[0] % _SQLITE_LOCK_STRIPES
+    database_path = Path(os.path.abspath(database))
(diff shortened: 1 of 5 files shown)


dexters1:
Thanks for the PR @GautamSharma99! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #4168, opened 2026-07-20
fix(sessions): scope session data by dataset

Vasilije1990:
## Summary

- namespace session cache, trace, context, vectors, locks, and lifecycle state by dataset
- require live dataset ACLs for remember, recall, SDK, agent detail, dashboard rows, and aggregates
- quarantine legacy unscoped sessions as owner-only with no scoped fallback
- preserve public session IDs through a trusted lifecycle marker and migration
- route session vectors through the selected dataset owner database context
- make dashboard identity composite across owner, dataset, and public session ID

## Security behavior

- the same public session ID can safely exist in multiple datasets
- scoped reads never fall back to legacy storage
- revoked dataset access hides session content and metadata
- ambiguous cross-owner session IDs fail closed unless an owner is selected
- deletion keeps lifecycle markers when cache content cannot be proven deleted

## Tests

- 703 unit tests passed
- 41 session adapter integration tests passed
- 28 agent API end-to-end tests passed
- Ruff format and lint passed
- changed frontend files passed ESLint
- Alembic upgrade and downgrade round-trip passed


dexters1:
Closing PR, opened a simpler PR that doesn't add complex backwards compatibility

PR #4170, opened 2026-07-21
COG-5929 fix(mcp): cloud client works against Cognee Cloud (redirects, unscoped recall/search, env)

LStromann:
## Problem
`cognee-mcp` connected to a Cognee **Cloud** tenant reports it empty even when it has data. Confirmed against a live tenant: the raw REST API returns 8 datasets and a dataset-scoped recall/search returns content, but the MCP tools see nothing.

The MCP's exposed memory tools on `dev` are the **new** pipeline — `remember`, `recall`, `forget` (`cognify`/`search` are not registered as tools). So the failures were in the cloud client path those tools use.

Four issues in `cognee-mcp/src/cognee_client.py` + `server.py` (API mode):

1. **Redirects not followed.** The httpx client had no `follow_redirects=True` and built collection URLs without a trailing slash (`/api/v1/datasets`). The cloud serves `/api/v1/datasets/` and 307-redirects the slash-less form, so `list_datasets()` read the empty redirect body as "no datasets" (3xx doesn't trip `raise_for_status`).
2. **Unscoped `recall` hits the empty default dataset.** A bare `recall` (no dataset, no session) → `404 "Recall prerequisites not met"`, so "what do you know from cognee?" (which routes to the `recall` tool) returned nothing.
3. **Unscoped `search`** had the same empty-default problem (legacy path, kept in sync).
4. **`COGNEE_BASE_URL` env ignored.** `server.py` only entered API mode via `--api-url`; env `COGNEE_BASE_URL` was ignored, so the env-based config the cloud onboarding cards ship silently fell back to direct/local mode.

## Fix
- `follow_redirects=True` on the client.
- `recall()` and `search()`: when no `datasets` (and, for recall, no `session_id`) is given, default to every dataset the caller can see. Unauthorized/empty datasets are filtered server-side (verified a mixed authorized+unauthorized query returns the authorized results, not a 404). Session-scoped recalls are left untouched.
- Default `--api-url` to `COGNEE_BASE_URL` and `--api-token` to `COGNEE_API_KEY`, matching the published package + onboarding cards.

## Verification (live cloud tenant, via the patched client)
- `list_datasets()` → **8 datasets** (was `[]`).
- unscoped `recall(query, "CHUNKS")` → **returns results** (was `404`).
- unscoped `search(query, "CHUNKS")` → **returns results** (was `404 NoDataError`).
- with only `COGNEE_BASE_URL` set, server logs `Cognee client initialized in API mode: <tenant url>`.
- End-to-end in Claude Desktop pointed at this branch.

Note: the earlier "recall timed out" symptom was **scale-to-zero cold start** (first calls hit a cold pod; ~10s once warm, well within the client's 300s timeout), not a recall bug.

## Notes for maintainers
Arguably the server's `recall(datasets=None)` / `search(datasets=None)` should themselves search all *authorized* datasets rather than defaulting to an empty one; the client-side default here is a pragmatic fix. Also, `/datasets` lists other users' datasets in the tenant while recall/search authorize per-owner — worth reconciling.

Fixes COG-5929.


cognee-mcp/src/cognee_client.py
```diff
@@ -230,6 +236,16 @@ async def search(
...
             payload = {"query": query_text, "search_type": query_type.upper(), "top_k": top_k}
+            if not datasets:
+                # Cloud search with no dataset targets the (usually empty) default
+                # dataset and 404s with NoDataError. Default to every dataset the
+                # caller can see so an unscoped query ("what do you know?") searches
+                # real memory instead of nothing. Unauthorized/empty datasets are
+                # filtered server-side, so passing them all is safe.
+                try:
+                    datasets = [d["name"] for d in await self.list_datasets() if d.get("name")]
+                except Exception:
```
dexters1: Lets not do except Exception:

and handle expected issues explicitely so other issues won't be masked

cognee-mcp/src/cognee_client.py
```diff
@@ -544,6 +560,16 @@ async def recall(
...
                 payload["search_type"] = search_type.upper()
+            if not datasets and not session_id:
+                # A bare recall (no dataset and no session) targets the empty
+                # default dataset and 404s ("Recall prerequisites not met"). Fall
+                # back to every dataset the caller can see so an unscoped
+                # "what do you know?" recalls from real memory. A session-scoped
+                # recall is left untouched so it can search the session cache.
+                try:
+                    datasets = [d["name"] for d in await self.list_datasets() if d.get("name")]
+                except Exception:
```
dexters1: Here as well

cognee-mcp/src/server.py
```diff
@@ -1914,15 +1914,17 @@ async def main():
     # Cognee API connection options
     parser.add_argument(
         "--api-url",
-        default=None,
-        help="Base URL of a running Cognee FastAPI server (e.g., http://localhost:8000). "
-        "If provided, the MCP server will connect to the API instead of using cognee directly.",
+        default=os.getenv("COGNEE_BASE_URL"),
```
dexters1, line 1917: Lets not add this as the default, they might unexpectedly use the URL when running the MCP without the api url argument

LStromann:
Closing in favor of #4173, which ports the same fix to `main` and also applies review feedback: a failed `list_datasets()` now propagates instead of being swallowed into an unscoped query. Superseded by #4173.

PR #4192, opened 2026-07-23
fix(neo4j): bind filter values as parameters in get_filtered_graph_data

### LINKED ISSUE #4191: hardening(neo4j): get_filtered_graph_data interpolates filter values into Cypher instead of binding parameters
### Summary

**Classification: defensive hardening — not exploitable through current call sites.** Every in-repo caller of `get_filtered_graph_data` passes constant attribute lists (`{"type": ["Entity"]}`, `{"type": ["DocumentChunk"]}`, retrieval's memory-fragment filters), so no user input reaches this path today. Filing it because the Neo4j implementation is the only one of the adapters implementing this interface method without guards, and it also carries a plain correctness bug.

### The three defects (`cognee/infrastructure/databases/graph/neo4j_driver/adapter.py::get_filtered_graph_data`)

```python
for attribute, values in attribute_filters[0].items():
    values_str = ", ".join(
        f"'{value}'" if isinstance(value, str) else str(value) for value in values
    )
    where_clauses.append(f"n.{attribute} IN [{values_str}]")
...
WHERE {where_clause} AND {where_clause.replace("n.", "m.")}
```

1. **String values get bare `'...'` wrapping with no escaping** — a value containing a quote (`"it's quoted"`) produces invalid Cypher and the query throws; a crafted value could inject clauses.
2. **Attribute names are interpolated unconstrained** (`n.{attribute}`) — no whitelist.
3. **The edge clause is derived via `where_clause.replace("n.", "m.")`** — any *value* containing the literal `"n."` substring (e.g. `"Section n.1"`) gets rewritten to `"Section m.1"` inside the query text, silently corrupting the filter. This one is a correctness bug independent of the injection concern.

### Sibling comparison

Postgres (`postgres/adapter.py`) validates every attribute against `_ALLOWED_FILTER_ATTRS = {"id", "name", "type"}` (raising `ValueError` otherwise) and binds values as parameters (`n.{attr} = ANY(:param)`); Turso does the equivalent with placeholders. Neo4j has neither guard.

### Why a public issue rather than a security report

Not exploitable today: all call sites pass constants, and the attribute whitelist the fix introduces matches what the sibling adapters already enforce for the same interface — this is bringing an outlier in line, not patching a live vulnerability.

### Environment

- cognee: `dev` @ `a717756`

Fix ready (same whitelist as postgres/turso, values bound as query parameters, edge clause built explicitly per alias — behavior unchanged for all valid input) with regression tests; PR follows right after this issue.

microbluey:
## Description

Fixes #4191 — `get_filtered_graph_data` in the Neo4j adapter interpolated attribute names and values directly into the Cypher text (bare quote wrapping broke on values containing quotes and permitted injection), and derived the edge clause via `where_clause.replace("n.", "m.")`, which silently corrupted any value containing the literal `"n."` substring.

The method now mirrors the guards the postgres and turso adapters already enforce for this same interface method:

- attribute names validated against `_ALLOWED_FILTER_ATTRS = {"id", "name", "type"}` (raising `ValueError` otherwise — the same whitelist and behavior as postgres);
- values bound as query parameters (`n.type IN $filter_values_0`) instead of interpolated;
- the edge clause built explicitly per alias (`n.` and `m.`) instead of string replacement.

Behavior is unchanged for all valid input: every in-repo caller passes whitelisted attributes (`type`/`name`/`id`) already, and since postgres enforces this exact whitelist for the same method, any caller that works there works here.

## Verification

- 3 new tests in `test_neo4j_filtered_graph_data.py`: quote-containing values are parameterized (and absent from the query text), values containing `"n."` survive the per-alias edge clause intact, and non-whitelisted attributes raise `ValueError` without touching the database.
- The existing endpoint-fallback test passes unchanged; full graph test directory: 39 passed.
- `ruff format --check` clean; commit is DCO signed-off.

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
+++ b/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
@@ -157,6 +157,10 @@ class Neo4jAdapter(GraphDBInterface):
     managing sessions and projecting graphs.
     """
 
+    # Attribute names allowed in get_filtered_graph_data, mirroring the
+    # postgres/turso adapters. Values are always bound as query parameters.
+    _ALLOWED_FILTER_ATTRS = {"id", "name", "type"}
+
     def __init__(
         self,
         graph_database_url: str,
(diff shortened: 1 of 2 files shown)


dexters1:
Thanks for the PR @microbluey! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #4198, opened 2026-07-23
[CLO-330] Approximate agent-session token usage from transcript

LStromann:
Companion to the cognee-saas frontend PR `fix/session-cost-display` (see CLO-330). This is the **backend half** — it populates token counts for the session paths that never entered the LLM-completion usage scope, so agent sessions stop showing $0.

## Why
Session costs showed $0 because agent-facing operations don't run under `track_session_usage`, so their session rows recorded zero tokens. Only UI search (which runs completions) was tracked. Cost itself is priced downstream in the cloud UI from token counts, so this PR only needs to populate tokens.

## What
- **`record_transcript_usage()`** (`usage_tracking.py`) — estimates tokens (chars/4, the same heuristic the completion path uses) from text and accumulates into the existing `tokens_in/out` columns. No `cost_usd` written (the UI prices it).
- **`SessionManager.add_agent_trace_step`** — calls it for every `remember` and `@agent_memory`-decorated tool call.
- **`recall` (`only_context=True`)** — attributes query + served context to the session; these run no completion so nothing else tracks them.

All three paths are **disjoint** from `record_llm_call` (completion retrievers neither emit trace steps nor take the `only_context` branch), so there is **no double-counting**.

## Scope / limitations
- Figures are **estimates** — char-based token counts, not real `response.usage`. The exact-accounting follow-up is described in CLO-330.
- Reflects cognee's own processing spend per session, not the agent's own model spend.

## Tests
- New unit tests for `record_transcript_usage` and the recall hook; a wiring test on `add_agent_trace_step`. 91 session/recall unit tests pass; ruff clean.

### DIFF
--- a/cognee/api/v1/recall/recall.py
+++ b/cognee/api/v1/recall/recall.py
@@ -633,4 +633,50 @@ async def recall(
             session_id or "-",
         )
 
+        # Context-only recalls run no LLM completion, so the completion usage
+        # scope records nothing and the session would show $0 for its reads
+        # (the common claude-code / agent pattern). Attribute the query + the
+        # served context to the session as a usage proxy. Guarded on
+        # only_context so a completion recall — already tracked by the usage
+        # scope — is never double-counted.
+        if session_id and only_context and merged:
+            await _record_recall_context_usage(
+                session_id=session_id, user=user, query_text=query_text, entries=merged
+            )
+
         return merged
+
+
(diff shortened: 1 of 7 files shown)


dexters1:
@LStromann should this one be closed in the end? Or is still needed?

LStromann:
Closed it. @goranradonic took that part over

PR #4200, opened 2026-07-23
fix: use dataset-membership check instead of owner-only get_data in raw-download endpoint

### LINKED ISSUE #4162: [Bug]: Users with dataset read access can't open raw data files (only data owners can)
### Bug Description

The dataset read endpoints apply inconsistent authorization rules and do not correctly handle missing or inaccessible
  datasets.

  Affected endpoints:

  GET /api/v1/datasets/{dataset_id}/data
  GET /api/v1/datasets/{dataset_id}/data/{data_id}/raw

  Both endpoints initially verify dataset-level read access using:

  dataset = await get_authorized_existing_datasets(
      [dataset_id],
      "read",
      user,
  )

  This permits owners as well as users, roles, or tenants with a valid dataset-level read grant.

  However, the raw-download endpoint later retrieves the data using:

  data = await get_data(user.id, data_id)

  get_data() independently requires the authenticated user to be the data owner:

  if data and data.owner_id != user_id:
      raise UnauthorizedDataAccessError(...)

  Consequently, a user with a valid read grant can list the dataset’s data but cannot download the raw content of those same
  data items. The endpoint returns 401 UnauthorizedDataAccessError even though dataset-level authorization succeeded and the
  route already verified that the requested data belongs to that dataset.

  There is also broken missing-dataset handling in both endpoints:

  if dataset is None:

  get_authorized_existing_datasets() returns a list when successful. For UUID-based requests, a missing or inaccessible dataset
  normally causes get_specific_user_permission_datasets() to raise PermissionDeniedError before this condition is reached. The
  advertised 404 branch is therefore effectively unreachable, and callers receive 403 even though the endpoint documentation
  states that missing or inaccessible datasets return 404.

  If the authorization helper returns an empty list, both endpoints proceed to:

  dataset[0]

  which raises IndexError.

  The list-data endpoint’s intended 404 response is additionally malformed:

  ErrorResponseDTO(f"Dataset ({str(dataset_id)}) not found.")

  ErrorResponseDTO is a Pydantic model and does not accept positional arguments. This raises:

  TypeError: BaseModel.__init__() takes 1 positional argument but 2 were given

  Relevant locations:

  - cognee/api/v1/datasets/routers/get_datasets_router.py:371-379
  - cognee/api/v1/datasets/routers/get_datasets_router.py:526-546
  - cognee/modules/data/methods/get_data.py:20-28
  - cognee/modules/data/methods/get_authorized_existing_datasets.py:25-39
  - cognee/modules/users/permissions/methods/get_specific_user_permission_datasets.py:29-45

#### Shared dataset raw-download failure

  1. Create two users, User A and User B, in the same tenant.
  2. Authenticate as User A.
  3. Create a dataset and add a file to it.
  4. Grant User B read permission on the dataset.
  5. Authenticate as User B.
  6. Request the dataset’s data:

     GET /api/v1/datasets/{dataset_id}/data

  7. Observe that the request succeeds and returns the uploaded data item.
  8. Using the returned data_id, request its raw content:

     GET /api/v1/datasets/{dataset_id}/data/{data_id}/raw

  9. Observe that the request fails with UnauthorizedDataAccessError because User B is not the data owner, despite having read
     access to the containing dataset.

  #### Missing dataset response mismatch

  1. Authenticate as any valid user.
  2. Generate a random dataset UUID that does not exist.
  3. Request:

     GET /api/v1/datasets/{missing_dataset_id}/data

  4. Repeat with:

     GET /api/v1/datasets/{missing_dataset_id}/data/{data_id}/raw

  5. Observe that authorization raises PermissionDeniedError and produces a 403 response before either route’s documented 404
     branch is reached.

  The empty-list failure can also be reproduced in a unit test by mocking get_authorized_existing_datasets() to return []. Both
  routes then attempt to access dataset[0].

### Expected Behavior

A user with a valid dataset-level read grant should be able to:

  - List the data contained in that dataset.
  - Download the raw content of data items belonging to that dataset.

  The raw endpoint should authorize access based on both conditions:

  1. The user has read permission on the requested dataset.
  2. The requested data item belongs to that dataset.

  It should not additionally require the user to own the data record.

  The existing owner-only behavior of get_data() may still be appropriate for other operations. The raw-download route could
  instead use the data object already obtained from the authorized dataset membership check, or use a new dataset-aware
  accessor, rather than weakening get_data() globally.

  Missing or inaccessible datasets should return a consistent response matching the endpoint contract. The current documentation
  specifies:

  404 Not Found: Dataset doesn't exist or user doesn't have access

  The response body should be JSON-serializable and constructed using valid Pydantic syntax or a plain dictionary.

### Actual Behavior

 - A shared reader can list a dataset’s data.
  - The same reader receives a 401 when downloading a listed data item’s raw content.
  - Missing or inaccessible UUIDs raise PermissionDeniedError and return 403 before the intended 404 handling runs.
  - Empty authorization results are indexed using [0], which can raise IndexError.
  - The list-data endpoint’s intended 404 branch constructs ErrorResponseDTO positionally and raises TypeError.

  Existing raw-download unit tests mock both authorization and get_data() to return successful results. They do not cover:

  - A non-owner with a valid dataset read grant.
  - An empty authorization result.
  - A missing dataset.
  - The intended 404 response body.

### Environment

  - OS: All platforms
  - Python version: Python 3.10–3.13
  - Cognee version: Current main / observed on 1.4.0-local
  - Component: FastAPI dataset routes
  - Authentication: Multi-user access control enabled
  - LLM Provider: Provider-independent
  - Database: Relational-backend independent
  - Storage: Reproducible with local files and S3-backed raw data

### Logs/Error Messages

```shell
For a shared reader attempting to download raw data:

  UnauthorizedDataAccessError:
  User <user_id> is not authorized to access data <data_id>
  (Status code: 401)

  For a missing or inaccessible dataset UUID:

  PermissionDeniedError:
  Request owner does not have necessary permission: [read]
  for all datasets requested.
  (Status code: 403)

  When the authorization helper returns an empty list:

  IndexError: list index out of range

  If execution reaches the malformed list-data 404 branch:

  TypeError: BaseModel.__init__() takes 1 positional argument but 2 were given
```

### Additional Context

The raw route already confirms that the data item belongs to the authorized dataset:

  dataset_data = await get_dataset_data(dataset[0].id)
  matching_data = [data for data in dataset_data if data.id == data_id]

  if len(matching_data) == 0:
      raise DataNotFoundError(...)

  After this check, calling the owner-only get_data(user.id, data_id) introduces a second, conflicting authorization policy.

  A possible fix is to use:

  data = matching_data[0]

  after dataset authorization and membership verification. Alternatively, a dedicated accessor could retrieve a data item only
  when it belongs to a dataset on which the caller has the required permission.

  Regression tests should verify that:

  - A dataset owner can list and download data.
  - A user with a dataset read grant can list and download data.
  - A user without a read grant cannot access the dataset or its raw data.
  - A data ID from another dataset cannot be downloaded through an authorized dataset ID.
  - Missing datasets return the documented response.
  - Empty authorization results never cause IndexError.
  - Error responses contain valid JSON.
  - Local-file and S3 download behavior continues to work.

### Pre-submission Checklist

rkfshakti:
## Problem

When a user with a dataset-level **read** grant (via ACL) tries to download a raw data file, the `GET /api/v1/datasets/{dataset_id}/data/{data_id}/raw` endpoint calls `get_data(user.id, data_id)`, which checks `data.owner_id == user_id` and raises `UnauthorizedDataAccessError` (401) for non-owners.

This breaks the ACL permission model: a user who can **list** the dataset's data cannot **download** the actual files.

## Root cause

The route already verifies dataset-level read access via `get_authorized_existing_datasets()` and confirms the data item belongs to that dataset via `get_dataset_data()` + membership filter. But then it calls `get_data(user.id, data_id)` which imposes an **additional** owner-only gate that bypasses the ACL system entirely.

## Fix

Use the already-verified `matching_data[0]` instead of calling `get_data()`. The data object has already been confirmed to:
1. Belong to a dataset the user has read permission on
2. Be the specific data item requested

## Secondary fixes

1. **`ErrorResponseDTO(f"...")`** — uses positional Pydantic construction, which raises `TypeError: BaseModel.__init__() takes 1 positional argument but 2 were given`. Changed to a plain `dict`.

2. **`if dataset is None`** — unreachable because `get_authorized_existing_datasets()` raises `PermissionDeniedError` for missing datasets. Changed to `if not dataset` to also guard against empty-list returns that would cause `dataset[0]` IndexError.

## Testing

- A dataset owner can still list and download data (unchanged)
- A user with a dataset read grant can now list **and** download data (fixed)
- A user without a read grant still cannot access the dataset (unchanged)
- Missing datasets return 404 with valid JSON (fixed)
- Empty authorization results no longer cause IndexError (fixed)

Fixes #4162

### DIFF
--- a/cognee/api/v1/datasets/routers/get_datasets_router.py
+++ b/cognee/api/v1/datasets/routers/get_datasets_router.py
@@ -691,7 +691,6 @@ def get_datasets_router() -> APIRouter:
             },
         )
 
-        from cognee.modules.data.methods import get_data
         from cognee.modules.data.methods import get_dataset_data
 
         # Verify user has permission to read dataset
@@ -699,7 +698,7 @@ def get_datasets_router() -> APIRouter:
 
         if not dataset:
             return JSONResponse(
-                status_code=404, content={"detail": f"Dataset ({dataset_id}) not found."}
+                status_code=404, content={"message": f"Dataset ({dataset_id}) not found."}
             )
 
         dataset_data = await get_dataset_data(dataset[0].id)
@@ -715,12 +714,10 @@ def get_datasets_router() -> APIRouter:
                 message=f"Data ({data_id}) not found in dataset ({dataset_id})."
             )
 
-        data = await get_data(user.id, data_id)
-
-        if data is None:
-            raise DataNotFoundError(
-                message=f"Data ({data_id}) not found in dataset ({dataset_id})."
-            )
+        # Use the data object already verified to belong to the authorized dataset,
+        # rather than calling get_data() which checks owner_id and would reject
+        # ACL-granted readers who are not the data owner.
+        data = matching_data[0]
 
         raw_location = data.raw_data_location
         parsed_uri = urlparse(raw_location)


rkfshakti:
Hi maintainers — gentle bump on this fix for #4162. Replaces the owner-only get_dataset check with a dataset-membership check so that non-owner members can access datasets they belong to. CI is passing. Would appreciate a review when time allows. Thanks!

rkfshakti:
Friendly ping — this PR has been open for over a week. Would appreciate a review when time allows.

rkfshakti:
Friendly ping — this PR has been open for over a week. The fix is minimal and tested. Would appreciate a review when time allows. Thanks!

rkfshakti:
Hi maintainers — just a friendly nudge on this one. The fix for #4162 uses a dataset-membership check instead of owner-only get_data in the raw-download endpoint so authorized members can access data. I'm excited to see it land. Would appreciate a review when you have a moment. Thanks!

dexters1:
@rkfshakti thank you for your contribution! I've made a PR and merged to dev based on your PR. We had to make some changes to it due to some internal changes in Data access.

Your commit is preserved in the new branch and it will be part of our git history

PR #4122, opened 2026-07-18
fix(mcp): forward recall system prompts

silasbrookshaha:
Fixes #4120.

Summary:
- Add optional system_prompt to the MCP recall tool.
- Forward system_prompt through CogneeClient.recall in API mode to /api/v1/recall.
- Forward system_prompt through direct/SDK recall kwargs.
- Add regression coverage for API payload forwarding and MCP tool forwarding.

Testing:
- python3 -m py_compile cognee-mcp/src/cognee_client.py cognee-mcp/src/server.py cognee-mcp/tests/test_mcp_server_hardening.py
- git diff --check -- cognee-mcp/src/cognee_client.py cognee-mcp/src/server.py cognee-mcp/tests/test_mcp_server_hardening.py

Attempted but not completed locally:
- python3 -m pytest cognee-mcp/tests/test_mcp_server_hardening.py -q (collection failed because this sparse local environment is missing the structlog dependency).

I affirm that all code in every commit of this pull request conforms to the terms of the Topoteretes Developer Certificate of Origin.

Issue #4120, opened 2026-07-18
[Feature]: Expose a system_prompt parameter on the MCP recall tool


zakhounet:
### Problem Statement

The recall MCP tool (cognee-mcp/src/server.py) and the underlying CogneeClient.recall (src/cognee_client.py) do not accept or forward a system_prompt, even though the backend HTTP endpoint POST /v1/recall already supports one (RecallPayloadDTO.system_prompt), and the completion layer honors a raw-text system_prompt over the file-based system_prompt_path (modules/retrieval/utils/completion.py: system_prompt = system_prompt if system_prompt else read_query_prompt(system_prompt_path)).

Consequence: MCP clients cannot override the answer/synthesis prompt per call. In API mode the effective prompt is the terse RecallPayloadDTO.system_prompt default ("Answer the question using the provided context. Be as brief as possible."), which also means answer_simple_question.txt is never consulted through the HTTP/MCP path. Use cases that need provenance-aware or temporal-validity-aware synthesis (count distinct source documents, flag single-source facts, prefer decommission over an earlier active description) have no supported lever through MCP.

### Proposed Solution

Requested change: add an optional system_prompt: str | None = None parameter to the recall tool and to CogneeClient.recall, forwarded into the /v1/recall payload (and into kwargs for the local/SDK path). No behavior change when omitted.

Workaround in use: a local thin image (FROM the pinned digest) patches CogneeClient.recall to inject system_prompt from a file pointed to by COGNEE_RECALL_SYSTEM_PROMPT_FILE. Happy to open a PR along these lines if useful.

### Use Case

We run cognee as a shared long-term memory for a personal infrastructure knowledge base, accessed by multiple MCP clients (Claude Desktop, Claude Code, an autonomous agent). We need the synthesis step to be provenance-aware and temporal-validity-aware: count distinct source documents behind each fact, explicitly flag single-source facts, mark reasoning that goes beyond the retrieved chunks as inference, and prefer a dated decommission/migration statement over an earlier "active" description of the same fact.

All of this is achievable with a custom system prompt — and the backend HTTP endpoint already supports one — but there is currently no way to pass it through the MCP tool, so the terse DTO default always applies.

### Pre-submission Checklist

- [x] I have searched existing issues to ensure this feature hasn't been requested already
- [x] I have provided a clear problem statement and proposed solution
- [x] I have described my specific use case

silasbrookshaha:
[...]

PR #4205, opened 2026-07-23
feat(mcp): add opt-in server-side default for recall system_prompt

zakhounet:
Follow-up to #4120 / #4122.

First of all, thank you for taking #4120 into account so quickly — the per-call `system_prompt` shipped in #4122 within a day, and it already unblocks clients that can set the parameter themselves. Much appreciated.

While adopting it, I ran into one remaining case that #4122 does not cover. When a caller omits `system_prompt`, the request falls back to the terse `RecallPayloadDTO` default, which also short-circuits `answer_simple_question.txt`. So a deployment cannot define a **default** synthesis policy.

This matters for a cognee instance used as shared long-term memory by several MCP clients: there is currently no way to set a synthesis policy server-side — every client would have to resend the full prompt text on every call, and any client that forgets silently falls back to the terse default.

**Would it be possible to also support an opt-in server-side default?** This PR proposes one, kept as small and conservative as I could make it. I would of course be glad to reshape it if you would rather see it done differently.

### Change

- Add an opt-in, env-driven default, resolved once at the top of `CogneeClient.recall`:
  - `COGNEE_MCP_RECALL_SYSTEM_PROMPT` — inline prompt text
  - `COGNEE_MCP_RECALL_SYSTEM_PROMPT_FILE` — path to a file holding the prompt
- Precedence is **explicit caller argument > env default > backend default**.
- Because the default is resolved before the API/SDK branches, both paths are covered without touching either.
- When neither variable is set, behaviour is unchanged: no `system_prompt` is added to the payload.
- `search()` is deliberately left untouched; this only affects `recall()`.

### Testing

Four tests added to `cognee-mcp/tests/test_mcp_server_hardening.py`, following the `httpx.MockTransport` pattern introduced by #4122:

- the env default is injected when the caller omits `system_prompt`;
- an explicit caller prompt takes precedence over the env default;
- the default can be supplied through a file (`..._FILE`);
- with nothing configured, no `system_prompt` is added to the payload.

The tests are written against the existing fixtures but I could not run the suite locally: my environment is missing the dependencies needed to import `src.cognee_client`. Happy to iterate if CI surfaces anything.

**On naming and placement:** the variable names are only a suggestion — happy to rename them, or to move the resolution to the server layer instead of the client, if that fits your conventions better. Thanks again for the quick follow-up on #4120, and thanks in advance for taking a look at this one.

### DIFF
--- a/cognee-mcp/src/cognee_client.py
+++ b/cognee-mcp/src/cognee_client.py
@@ -6,6 +6,7 @@ This module provides a unified interface for interacting with Cognee, supporting
 - API mode: Makes HTTP requests to a running Cognee FastAPI server
 """
 
+import os
 import sys
 import hashlib
 from typing import Optional, Any, List, Dict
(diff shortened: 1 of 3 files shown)


Vasilije1990 (approved):

Vasilije1990:
@zakhounet keep bringing them on!

PR #4217, opened 2026-07-25
SDK-319 feat(sessions): attribute sessions to agent connections via agent_id

Vasilije1990:
## Description

Nothing in the relational store attributes a session (or the searches recorded under it) to the agent that drove it. `session_records` has only `(session_id, user_id)` + counters; the agent registry's `AgentConnection.session_id` is the sole bridge, and it lives outside the relational schema — so provenance views reverse-map it at read time and the attribution is lost for unregistered or expired connections.

This adds:

- **`session_records.agent_id`** — nullable, indexed `String` (agent-connection ids are `slug-sha1[:16]` strings, not UUIDs), exposed in `to_dict()`.
- **Fill-if-null stamping**: `ensure_and_touch_session(agent_id=...)` carries it on upsert (back-fills a null like `dataset_id`); new `set_session_agent()` claims an existing row. First agent wins — re-registration is a no-op and a second agent cannot overwrite the attribution.
- **Registry hook**: `register_agent_connection` stamps the session best-effort when registered with `user_id` + `session_id`; failures never break registration.

**Schema migration is intentionally NOT included — it will be applied by hand.** Fresh installs get the column from the model's `create_all`; existing databases need the equivalent of:

```sql
ALTER TABLE session_records ADD COLUMN agent_id VARCHAR NULL;
CREATE INDEX ix_session_records_agent_id ON session_records (agent_id);
```

Note on multiple agents in one session: `agent_id` records the FIRST agent to claim the `(session_id, user_id)` row — a deliberate v1 simplification. Agents authenticating as their own agent users get separate rows anyway (the PK is per user); only same-user multi-agent sessions collapse to first-wins. Finer-grained attribution (per-QA-entry agent, or a session↔agent association table) can layer on top without changing this column's meaning ("primary/initiating agent").

## Type of Change

## How was this tested

- 6 new unit tests (`test_agent_attribution.py`): carry-on-create, back-fill, no-overwrite, fill-if-null claim, missing-row no-op, `to_dict` exposure — plus the existing session-visibility suite (12 passed total) and agents registry suite (6 passed).
- `ruff format --check` / `ruff check` clean.

## DCO Affirmation


cognee/modules/agents/registry.py
```diff
@@ -170,6 +170,18 @@ async def register_agent_connection(
...
+    if user_id and session_id:
+        # Attribute the session to this connection in the relational store
+        # (fill-if-null), so agent↔session attribution survives registry
+        # restarts and is joinable by SQL. Best-effort: the session_records
+        # table is optional for registry correctness.
+        try:
+            from cognee.modules.session_lifecycle.metrics import set_session_agent
+
+            await set_session_agent(session_id=session_id, user_id=user_id, agent_id=connection.id)
+        except Exception as error:  # noqa: BLE001 — attribution must not break registration
```
dexters1, line 186: Can we avoid general except Exception as error clauses? Makes it very hard to debug issues later on

cognee/modules/session_lifecycle/models.py
```diff
@@ -33,6 +33,12 @@ class SessionRecord(Base):
 
     dataset_id = Column(UUID, nullable=True, index=True)
 
+    # Agent-connection id (registry string like "my-agent-ab12cd34ef56",
+    # not a UUID) — attributes the session to the agent that drove it.
+    # Stamped best-effort when an agent connection registers with this
+    # session_id; null for sessions no agent claimed.
+    agent_id = Column(String, nullable=True, index=True)
```
dexters1, line 40: Migration has to be done when adding new relational DB columns/tables and etc.

To be able to maintaint the relational DB state through migrations

dexters1 (changes requested):
A migration for adding the new agent_id column is needed

cognee/modules/session_lifecycle/models.py
Vasilije1990: Added in a04b39130: Alembic revision `a7c2e9f4b8d1_add_session_records_agent_id.py`, chained from the current dev head `f3a7b9c1d2e4`. It adds `agent_id` and `ix_session_records_agent_id` with existence guards (no-op on DBs created by `create_all()`, idempotent on re-run) and drops both on downgrade via batch mode for SQLite. Covered by `cognee/tests/unit/test_add_session_records_agent_id_migration.py` against a real SQLite DB, and `test_alembic_chain_has_a_single_head` still passes. The branch is also merged up to dev, which clears the stale unit-test failures.

dexters1 (approved):

Issue #4129, opened 2026-07-19
Add adapter-parametrized contract tests for CacheDBInterface (empty-collection and last_n edge cases)


Vasilije1990:
## Context

Issue #3930 reported `RedisAdapter.get_latest_qa_entries` returning `None` instead of `[]` for an empty session with `last_n=1`, which crashes iteration downstream. This is a class of bug, not a one-off: `cognee/infrastructure/databases/cache/cache_db_interface.py` declares return types like `list[SessionQAEntry]`, and three concrete adapters implement it independently:

- `cognee/infrastructure/databases/cache/redis/RedisAdapter.py`
- `cognee/infrastructure/databases/cache/fscache/FsCacheAdapter.py`
- `cognee/infrastructure/databases/cache/sql/SqlCacheAdapter.py`

Each adapter can drift from the interface contract (return `None`, raise instead of returning empty, order results differently) and nothing catches it until a user hits it in production.

## Task

1. Write a **shared, adapter-parametrized contract test suite** for `CacheDBInterface` (pytest fixtures parametrized over fscache + sql; redis behind a marker if no redis is available in CI). For every method that returns a collection, assert:
   - empty state → empty collection, never `None`;
   - `last_n=0`, `last_n=1`, `last_n > available` edge cases;
   - ordering guarantees stated in the docstring (e.g. SqlCacheAdapter documents "chronological; [] when none").
2. Fix every violation the suite finds in the three adapters.
3. Align docstrings on `cache_db_interface.py` so the contract (empty-collection semantics, ordering) is stated once at the interface level.

## Acceptance criteria

- [ ] One test module exercises all cache adapters through the same contract cases.
- [ ] The #3930 scenario (empty session, `last_n=1`) is an explicit test case and passes on all adapters.
- [ ] No interface method on any adapter can return `None` where the signature says `list[...]`.

## Out of scope

- The `tapes/` cache backend and performance changes to the adapters.

Darshan3690:
# Implementation Plan - CacheDBInterface Shared Contract Test Suite & Adapter Fixes (#3930)

Align `CacheDBInterface` implementations (`RedisAdapter`, `FsCacheAdapter`, `SqlCacheAdapter`) to enforce interface contracts for all collection-returning methods:
- **Empty state**: return empty list `[]` (never `None`).
- **Edge cases**: handle `last_n=0`, `last_n=1`, `last_n > available`, `last_n < 0`.
[...]

Issue #4029, opened 2026-07-11
[Bug]: Batch add() with byte-identical files crashes: UNIQUE constraint failed: data.id (concurrent per-item ingestion races on content-derived data_id)


rstover-fo:
### Bug Description

Calling `cognee.add()` once with a batch that contains two files whose contents are **byte-identical** (same content, different filenames) crashes with:

```
sqlite3.IntegrityError: UNIQUE constraint failed: data.id
```

Both files resolve to the same content-derived `data_id` (`ingestion.identify()` → uuid5 over the content hash + owner). The pipeline schedules **each data item as its own concurrent task** — `run_tasks()` wraps every item in `run_tasks_data_item()` and runs them via `asyncio.gather()` with `Semaphore(data_per_batch)` (default 20):

https://github.com/topoteretes/cognee/blob/v1.2.2/cognee/modules/pipelines/operations/run_tasks.py#L105-L132

So each file gets its **own** `ingest_data()` invocation with a single-item list. The in-batch dedup guard inside `ingest_data.store_data_to_dataset()` (`if str(data_id) in dataset_data_map: continue`) never sees the sibling file, and the race unfolds:

1. Both invocations build `existing_data_map` from the DB (`SELECT ... WHERE data.id IN (...)`) **before either has committed** — both find nothing:
   https://github.com/topoteretes/cognee/blob/v1.2.2/cognee/tasks/ingestion/ingest_data.py#L110-L115
2. Both take the "new data" path and construct a `Data` row with the **same primary key**:
   https://github.com/topoteretes/cognee/blob/v1.2.2/cognee/tasks/ingestion/ingest_data.py#L201-L229
3. The first flush wins; the second invocation's flush (query-invoked autoflush at `session.merge(dataset)`, ingest_data.py#L245) raises the `UNIQUE constraint failed: data.id`.

Classic check-then-insert (TOCTOU) race — deterministic with 2 identical files, because both tasks always start before either commits.

The same per-item `asyncio.gather` scheduling exists on current `main` (only verified by reading the source there; the crash itself was reproduced on 1.2.2).

### Steps to Reproduce

```python
"""Repro: cognee 1.2.2, embedded defaults (sqlite/lancedb/kuzu).
add() makes no LLM calls, so dummy keys are fine."""

import asyncio, os, tempfile
from pathlib import Path

BASE = Path(tempfile.mkdtemp(prefix="cognee_dup_repro_"))
[...]

PR #4229, opened 2026-07-26
COG-5975 fix(cache): unique constraint + upsert for cache_session_context so retried writes can't poison a session

### LINKED ISSUE #4226: [Bug]: `cache_session_context` has no unique constraint — a retried write creates duplicate rows and permanently 503s the session
### Bug Description

`update_session_context_entry` resolves a session-context row with `scalar_one_or_none()`, but the `cache_session_context` table has no unique constraint on `(user_id, session_id, entry_id)`. Any client that retries a write after a timeout can therefore insert a duplicate row instead of updating the existing one.

Once two or more rows exist for the same key, `scalar_one_or_none()` raises and **every subsequent write for that session returns 503 — permanently**. There is no self-healing path: the session stays poisoned until the extra rows are deleted by hand in the database.

The trigger is an ordinary client retry, not anything exotic:

1. A write succeeds server-side but the client times out and retries.
2. The retry inserts a second row (no constraint to collapse it into an update).
3. All later updates for that key now 503.
4. A client with a retry loop keeps retrying forever, and each retry can add another duplicate.

I hit this with the Claude Code `cognee-memory` plugin, whose buffered-replay loop drove it to 77 copies of a single entry. That client's retry behaviour is filed separately in `topoteretes/cognee-integrations` (see Additional Context), but the server-side defect is independent of any one client — **any** caller that retries a timed-out write can reach this state, and several in-flight integration PRs are actively adding retry-on-timeout behaviour.

Likely introduced in #3072, which added the `cache_session_context` table with session and partial-expiry indexes and an `entry_id` column promoted for direct `UPDATE` — no uniqueness guarantee on the update key.

### Steps to Reproduce

1. Run a Cognee server with the default SQL cache backend (`SqlCacheAdapter`, sqlite or postgres).
2. Write a session-context entry for some `(user_id, session_id, entry_id)` — e.g. any client that calls `POST /remember/entry` for a session.
3. Make the client time out and retry that same write — e.g. delay or block the response past the client's socket timeout so the retry fires while the first write is still committing.
4. Inspect `cache_session_context`: there are now two rows for the same `(user_id, session_id, entry_id)`.
5. Issue any further write to that same entry.

In my case step 3 happened naturally: the client replays a buffer of entries sequentially, each one triggering a full LLM graph-extraction pipeline server-side (~15s), which exceeded the client's timeout and retried.

### Expected Behavior

Writing the same `(user_id, session_id, entry_id)` twice should be idempotent: exactly one row, updated in place, and a successful response. A client retry after a timeout should be safe — it's the normal thing for an HTTP client to do.

Failing that, the server should at minimum degrade gracefully rather than entering a permanent, unrecoverable failure state for the affected session.

### Actual Behavior

The retry inserts a duplicate row, and every subsequent write for that session returns 503 for as long as the duplicates exist.

Observed on my instance for a single session (`99576d70`): the internal bookkeeping entry `__agent_context_extraction_state__` was present **77 times** in `cache_session_context`, and every write to that session failed — **216 such errors in roughly 6 hours**, a steady ~36/hour. Reads and recall were unaffected; only writes failed.

Because the client retried the failed writes indefinitely, this also became sustained pointless load on the server: a stuck client-side sync process ground against the same poisoned session for 6.5 hours.

The only recovery was direct database surgery. I deleted 76 of the 77 duplicate rows, keeping the newest. `Multiple rows were found` errors dropped to zero within 10 minutes, and the client's stuck 62-entry backlog then drained cleanly in 105s with no failures.

### Environment

- **OS**: Debian 12 (bookworm) — official `cognee/cognee` + `cognee/cognee-mcp`
  container images, self-hosted on a k3s Kubernetes cluster (x86_64)
- **Python version**: 3.12.13 (backend container) / 3.12.12 (MCP container)
- **Cognee version**: `cognee` 1.4.0 (backend), image `cognee/cognee:main-6df1eda`
  — note: the same-tagged `cognee/cognee-mcp:main-6df1eda` image ships
  `cognee` **1.2.2** + `cognee-mcp` 0.5.4, i.e. the two images at the identical
  tag are two minor versions apart
- **LLM Provider**: Fireworks.ai via OpenAI-compatible endpoint
  (`https://api.fireworks.ai/inference/v1`)
  - `LLM_PROVIDER=custom` (LiteLLM 1.91.1),
    `LLM_MODEL=openai/accounts/fireworks/models/deepseek-v4-flash`
  - BAML on the same model (`BAML_LLM_PROVIDER=openai`)
  - `EMBEDDING_PROVIDER=openai_compatible`,
    `EMBEDDING_MODEL=nomic-ai/nomic-embed-text-v1.5`, `EMBEDDING_DIMENSIONS=768`
- **Cache backend**: `SqlCacheAdapter` (default — no cache adapter env override set)
- **Database**: single external PostgreSQL 17.9 for all three layers —
  `DB_PROVIDER=postgres`, `VECTOR_DB_PROVIDER=pgvector` (pgvector 0.8.5),
  `GRAPH_DATABASE_PROVIDER=postgres`. Connections go through **pgbouncer in
  transaction mode on port 6432**, not directly to Postgres. SQLAlchemy 2.0.51,
  asyncpg. NVMe-backed node storage; file staging on a Longhorn RWX PVC.
- **Client**: Claude Code + `cognee-memory` plugin 1.0.0, HTTP mode via
  `COGNEE_BASE_URL=http://cognee.k3s.internal` (hits the backend pod, so the
  1.4.0 side)
- **Other config**: `ENABLE_BACKEND_ACCESS_CONTROL=false`,
  `LITELLM_DROP_PARAMS=true`
- **Networking**: local/LAN; `/health` responds in ~40ms, so no latency or
  resource pressure is involved

### Logs/Error Messages

```shell
Server-side, repeated ~36x/hour for the affected session:

    Multiple rows were found when one or none was required

Duplicate state in cache_session_context for session 99576d70:

    entry_id = __agent_context_extraction_state__   ->  77 rows

Client-side symptom (Claude Code hook), caused by the retry loop that can never succeed:

    UserPromptSubmit hook timed out after 30s — output discarded.
    Raise the hook's "timeout" to allow more time.

After deleting 76 of the 77 rows:

    Multiple rows were found  ->  0 occurrences in 10 minutes
    62-entry client backlog   ->  drained in 105s, 0 failures
```

### Additional Context

**Suggested fix**

- Add a unique constraint / index on `(user_id, session_id, entry_id)` for `cache_session_context`.
- Make the write path an upsert (`ON CONFLICT DO UPDATE`) so a retried write is idempotent.
- Ship a migration that collapses existing duplicate groups (keep newest) before applying the constraint.
- Defensive fallback: if the multiple-row case still occurs, prefer the newest row and prune the rest rather than 503 forever. This is a TTL'd cache table, so recovery is cheap and safe.

**Related**

- #3072 — implemented session-context CRUD on `SqlCacheAdapter` and introduced the `cache_session_context` table; likely where the missing constraint originates.
- #4029 / #4034 — `UNIQUE constraint failed: data.id` on concurrent `add()`. Same class of problem (non-idempotent write under retry/concurrency), mirrored symptom: there the constraint exists and the write crashes loudly; here the constraint is missing and the read-back silently ambiguates.
- #4129 — adapter-parametrized contract tests for `CacheDBInterface`. This looks like the natural home for a regression test: writing the same `(user_id, session_id, entry_id)` twice yields exactly one row and a successful update, parametrized across the SQL, FS, and Redis adapters. I've only reproduced this on `SqlCacheAdapter`; the other two implement `update_session_context_entry` independently and may or may not share the flaw.
- #3930, #4030 — other recent session-cache correctness bugs, for context on the area.

**Companion issue (client side, Claude Code plugin)**

https://github.com/topoteretes/cognee-integrations/issues/298

**Discord thread with the full investigation**

https://discord.com/channels/1120795297094832337/1529611685290184764

### Pre-submission Checklist

Vasilije1990:
## Description

Fixes #4226 — `cache_session_context` had no uniqueness on `(user_id, session_id, entry_id)` (unlike `cache_qa_entries`, which has `uq_cache_qa_entry`). The session-context write path is update-then-create across two transactions, so an ordinary client retry after a timeout — or two concurrent writers — inserted a duplicate row. Once duplicates existed, `scalar_one_or_none()` in `SqlCacheAdapter.update_session_context_entry` raised on **every** subsequent write for that session: a permanent 503 recoverable only by manual row deletion (the reporter hit 77 copies of one bookkeeping entry and ~36 failed writes/hour for 6.5 hours).

All four parts of the fix suggested in the issue:

- **Unique index** `uq_cache_session_context_entry` on `(user_id, session_id, entry_id)` — declared as a unique `Index` rather than a `UniqueConstraint` so tables created before it existed can be upgraded in place.
- **Upsert on create** — `create_session_context_entry` now uses `ON CONFLICT DO UPDATE` (Postgres + SQLite dialects, same pattern as `set_value`), so a retried write replaces the payload in place. Id-less entries keep their synthetic UUID and are unaffected.
- **In-place migration** — `_ensure_initialized` collapses each duplicate group to its newest row, then runs `CREATE UNIQUE INDEX IF NOT EXISTS`. Migration failure is non-fatal: the adapter logs a warning and falls back to plain inserts so `ON CONFLICT` can never error against a missing index.
- **Self-healing update** — `update_session_context_entry` fetches all rows for the key ordered by `seq`, merges into the newest, and prunes older duplicates in the same transaction, so even an un-migrated database recovers on the next write instead of 503ing forever.

The Redis and FS adapters don't share the crash (they update the first list match), so this fix is scoped to the SQL adapter. Follow-up noted in COG-5975: their `create_session_context_entry` still appends duplicates on retried creates (no crash, but duplicated context) — the adapter-parametrized contract tests from #4129 would be the natural home for cross-adapter idempotency coverage.

## Testing

New regression tests in `test_sql_adapter_crud.py` (sqlite+aiosqlite):

- retried create with the same entry id yields exactly one row with the retried payload
- with legacy duplicates simulated (index dropped, rows stacked), update succeeds, merges into the newest row, and exactly one row remains
- a legacy database is deduped (newest kept per key) and re-indexed on next adapter startup, after which upserts work

`pytest cognee/tests/unit/infrastructure/databases/cache/` — 139 passed; `cognee/tests/unit/infrastructure/session/` — 181 passed. Ruff lint + format clean.

## Related

- Linear: COG-5975
- Introduced in #3072 (table added without a uniqueness guarantee on the update key)
- Same defect class as #4029 / #4034; contract-test home #4129

### DIFF
--- a/cognee/infrastructure/databases/cache/sql/SqlCacheAdapter.py
+++ b/cognee/infrastructure/databases/cache/sql/SqlCacheAdapter.py
@@ -9,7 +9,18 @@ from hashlib import sha256
 from typing import List, Optional
 
 from pydantic import ValidationError
-from sqlalchemy import create_engine, delete, event, func, insert, or_, select, text, update
+from sqlalchemy import (
+    create_engine,
+    delete,
+    event,
+    func,
+    insert,
+    inspect,
+    or_,
+    select,
+    text,
+    update,
+)
 from sqlalchemy.engine import make_url
 from sqlalchemy.exc import DBAPIError
 from sqlalchemy.ext.asyncio import async_sessionmaker, create_async_engine
@@ -26,6 +37,7 @@ from cognee.modules.storage.utils import JSONEncoder
 from cognee.shared.logging_utils import get_logger
 
 from .tables import (
+    UQ_CACHE_SESSION_CONTEXT_ENTRY,
     cache_kv,
     cache_metadata,
     cache_qa_entries,
@@ -161,6 +173,11 @@ class SqlCacheAdapter(CacheDBInterface):
         self._initialized = False
         self._init_lock = asyncio.Lock()
         self._last_purge = 0.0
+        # Whether the unique index backing session-context upserts is in place;
+        # when the in-place upgrade of a legacy table fails, writes fall back to
+        # plain inserts (the pre-index behavior) instead of erroring on
+        # ON CONFLICT with no matching index.
+        self._session_context_unique = False
 
     # --------------------------------------------------------------------- #
     # Initialization / shared helpers
@@ -180,7 +197,55 @@ class SqlCacheAdapter(CacheDBInterface):
                 error_msg = f"Failed to connect to SQL cache database: {error}"
                 logger.error(error_msg)
                 raise CacheConnectionError(error_msg) from error
+            self._session_context_unique = await self._migrate_session_context_uniqueness()
             self._initialized = True
+
+    async def _migrate_session_context_uniqueness(self) -> bool:
+        """Upgrade pre-existing ``cache_session_context`` tables to the unique index.
+
+        ``create_all(checkfirst=True)`` never alters an existing table, so
+        databases created before ``uq_cache_session_context_entry`` existed have
+        no uniqueness guarantee on (user_id, session_id, entry_id) and a retried
+        write could stack duplicate rows. Collapse each duplicate group to its
+        newest row, then add the index. Returns True when the index is in place.
+        Failure is non-fatal: writes fall back to plain inserts and
+        update_session_context_entry self-heals duplicates, so the next
+        initialization can retry.
+        """
+        try:
+            async with self.engine.begin() as connection:
+                index_names = await connection.run_sync(
+                    lambda sync_connection: [
+                        index.get("name")
+                        for index in inspect(sync_connection).get_indexes("cache_session_context")
+                    ]
+                )
+                if UQ_CACHE_SESSION_CONTEXT_ENTRY in index_names:
+                    return True
+                newest_per_entry = select(func.max(cache_session_context.c.seq)).group_by(
+                    cache_session_context.c.user_id,
+                    cache_session_context.c.session_id,
+                    cache_session_context.c.entry_id,
+                )
+                await connection.execute(
+                    delete(cache_session_context).where(
+                        cache_session_context.c.seq.not_in(newest_per_entry)
+                    )
+                )
+                await connection.execute(
+                    text(
+                        f"CREATE UNIQUE INDEX IF NOT EXISTS {UQ_CACHE_SESSION_CONTEXT_ENTRY} "
+                        "ON cache_session_context (user_id, session_id, entry_id)"
+                    )
+                )
+            return True
+        except Exception as error:
+            logger.warning(
+                "Could not enforce session-context uniqueness "
+                "(falling back to non-upsert writes until next initialization): %s",
+                error,
+            )
+            return False
 
     @staticmethod
     def _now() -> datetime:
@@ -839,6 +904,10 @@ class SqlCacheAdapter(CacheDBInterface):
 
         The caller validates the payload; we only promote its ``id`` to the
         ``entry_id`` column so updates can target a single row directly.
+
+        Idempotent under client retries: re-sending the same (user, session,
+        entry id) replaces the stored payload in place instead of stacking a
+        duplicate row.
         """
         await self._ensure_initialized()
         # Redis/FS append regardless of "id" (an id-less entry is simply never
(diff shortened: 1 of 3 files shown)


dexters1:
@Vasilije1990 added a PR with the fix here:

https://github.com/topoteretes/cognee/pull/4232

Fix should be handled with an alembic migration of the relational DB

That also resolves the issue for existing users who had it

Issue #4233, opened 2026-07-27
[Bug]: MCP `remember` always runs `improve()` — full-graph projection per call, whole-graph re-embed with triplet_embedding


plexus77:
## Current remaining scope — September 2026 triage

Reopened for the missing MCP `self_improvement` control only. The prior full-graph projection/re-embedding fix remains historical completed work; this ticket does not assert that those costs recur today.

Expose an optional boolean on MCP `remember` and forward explicit `False` through `CogneeClient` in direct and API modes. Extend the REST path if needed so the flag cannot be silently ignored. Preserve the existing default when omitted and document permanent/session-mode behavior.

Acceptance criteria:

- [ ] MCP `remember(..., self_improvement=False)` reaches the underlying operation unchanged in direct and API modes.
- [ ] Explicit `False` suppresses the automatic improvement stage without skipping ingestion.
- [ ] Omitted flag preserves current defaults; session-mode behavior is documented and tested.
- [ ] Regression tests cover true, false, omission, and background permanent writes.

The original community report and resolved performance history are retained below.

---

### Environment

* `cognee` 1.4.0, `cognee-mcp` 0.5.5 (stdio transport, direct mode — not API mode)
* Postgres 17 + pgvector, Neo4j 5 Community
* Graph: \~27,100 nodes / \~86,400 edges
* macOS 15, Python 3.12

### Summary

`remember()` defaults to `self_improvement=True`, so **every** call runs `improve()` → `memify()`. Because `memify` builds its memory fragment by projecting the **entire graph** out of the graph DB, every `remember` pays a full-graph projection — and `cognee-mcp` gives no way to turn it off.

On our graph that is \~1.1 GB RSS and several seconds per `remember` in the best case. In the worst case — `triplet_embedding` enabled — every `remember` re-embeds the whole graph: \~86k triplets, \~30 minutes, \~US$8 of embeddings **per remember**.

There are really three separable problems here; (2) is the one I'd most like to see fixed.

---

### 1. `self_improvement` is not reachable through the MCP server

`remember()` accepts it ([`remember.py#L643`](<https://github.com/topoteretes/cognee/blob/v1.4.0/cognee/api/v1/remember/remember.py#L643>)) and calls `improve()` when true ([`#L1126`](<https://github.com/topoteretes/cognee/blob/v1.4.0/cognee/api/v1/remember/remember.py#L1126>)), but the MCP client builds its kwargs without it, so the default always wins:

```python
[...]

PR #4234, opened 2026-07-27
fix(memify): skip the memory-fragment projection when nothing consumes it

plexus77:
Fixes item (2) of #4233.

## Problem

`memify()` builds its memory fragment whenever `data is None`, before anything considers whether the resolved task list can use it:

```python
if not data:
    async with set_database_global_context_variables(...):
        memory_fragment = await get_memory_fragment(node_type=node_type, node_name=node_name)
        data = [memory_fragment]
```

`get_memory_fragment()` reads the **entire graph** into memory. In the default configuration nothing consumes the result:

- `get_default_memify_extraction_tasks()` returns `[]` unless the `triplet_embedding` config flag is set, and it is `False` by default.
- With no extraction tasks the fragment goes straight to the default enrichment task, `index_data_points`, which already documents that it discards such objects:

```python
# Skip non-DataPoint objects (e.g. CogneeGraph) that may be
# passed through the memify pipeline without metadata.
if not hasattr(data_point, "metadata") or not data_point.metadata:
    continue
```

So every default-configuration `memify()` call paid a full-graph read to build an object that was immediately thrown away.

This is not a rare path: **`remember()` runs `improve()` → `memify()` by default** (`self_improvement=True`), so ordinary `remember()` calls inherit the cost. On our graph (~27k nodes / ~86k edges) that was several seconds and ~1 GB resident per call, and it scales with total graph size rather than with what was just remembered.

## Change

Build the fragment only when something can consume it:

```python
should_build_memory_fragment = bool(extraction_tasks) or caller_supplied_enrichment_tasks
```

`caller_supplied_enrichment_tasks` is captured **before** defaults are filled in. This deliberately preserves the one case that could otherwise regress: a caller who passes their own enrichment tasks and no extraction tasks may be consuming the fragment directly, so they keep receiving one. Passing `data` explicitly is unaffected.

Also documents the behaviour in the `data:` docstring entry, and logs at debug level when the projection is skipped.

If you'd rather narrow or widen the condition — e.g. skip whenever `extraction_tasks` is empty regardless of caller-supplied enrichment tasks — I'm happy to adjust; I chose the conservative version to guarantee no behaviour change for existing callers.

## Tests

New file: `cognee/tests/unit/modules/memify/test_memify_fragment_projection.py`. Fully mocked — no graph backend, no vector backend, no network. Four cases: the fix itself, plus three guards that pin the behaviour which must *not* change.

```
$ uv run pytest cognee/tests/unit/modules/memify/test_memify_fragment_projection.py -v
cognee/tests/unit/modules/memify/test_memify_fragment_projection.py::test_skips_fragment_projection_when_no_extraction_tasks PASSED
cognee/tests/unit/modules/memify/test_memify_fragment_projection.py::test_projects_fragment_when_extraction_tasks_exist PASSED
cognee/tests/unit/modules/memify/test_memify_fragment_projection.py::test_projects_fragment_for_caller_supplied_enrichment_only_pipeline PASSED
cognee/tests/unit/modules/memify/test_memify_fragment_projection.py::test_explicit_data_is_never_overwritten_by_a_projection PASSED
============================== 4 passed in 0.02s ===============================
```

**The new test fails without the fix** (verified by reverting `memify.py` and re-running), so it genuinely pins the behaviour rather than passing vacuously:

```
$ git stash push cognee/modules/memify/memify.py && uv run pytest <new test file> -q
E  AssertionError: Expected mock to not have been awaited. Awaited 1 times.
1 failed, 3 passed
```

The three guards pass either way, which is what you want from a no-regression test.

### Surrounding suites

```
$ uv run pytest cognee/tests/unit/memify_pipelines/ cognee/tests/unit/modules/memify_tasks/ \
    cognee/tests/unit/tasks/memify/ cognee/tests/unit/api/v1/improve/ \
    cognee/tests/unit/api/test_improve_global_context_index.py cognee/tests/unit/modules/memify/ -q
180 passed in 0.40s
```

### Full unit suite

```
$ uv run pytest cognee/tests/unit/ -q \
    --ignore=cognee/tests/unit/eval_framework/dashboard_test.py \
    --ignore=cognee/tests/unit/infrastructure/databases/graph/neo4j_deadlock_test.py
13 failed, 3369 passed, 51 skipped in 455.21s
```

**All 13 failures are pre-existing on `dev` and unrelated to this change.** I verified rather than assumed:

- Running those same 13 in isolation gives an identical result **with and without** this patch: `11 failed, 2 passed`, same test names both times.
- The two that pass in isolation but fail in the full suite (`test_contradiction_detection_wiring.py::TestRememberInheritsTheFlag`) are order-dependent, and cannot be affected by this change — that file contains no reference to `memify` or `get_memory_fragment` and calls `remember(self_improvement=False)`, so it never reaches the modified path.
- The remaining 11 are in `graph_completion_retriever_cot_test.py`, `temporal_retriever_test.py`, `triplet_retriever_test.py` and `test_cognee_error_contract.py`.

The two `--ignore`d files fail at collection with `ModuleNotFoundError: No module named 'neo4j'` (optional dependency not installed locally), also independent of this change.

### Checks

```
$ uvx ruff@0.15.11 check <changed files>          # All checks passed!
$ uvx ruff@0.15.11 format --check <changed files> # 2 files already formatted
```
Trailing-whitespace and end-of-file hooks verified clean on both files.

A note on your PR checklist: it asks for screenshots of tests passing. I've pasted the terminal output verbatim above instead, since it's copy-pasteable and greppable — happy to attach images too if that's a hard requirement.

## Notes

- Branched from `dev` per CONTRIBUTING; commit is DCO signed-off.
- No changelog entry added — happy to add one under `Unreleased` → `Fixed` if you'd like.
- Items (1) and (3) of #4233 (exposing `self_improvement` through `cognee-mcp`, and the `triplet_embedding` blast radius) are untouched here and want separate decisions.

### DIFF
--- a/cognee/modules/memify/memify.py
+++ b/cognee/modules/memify/memify.py
@@ -49,7 +49,10 @@ async def memify(
         enrichment_tasks: List of Cognee Tasks to handle enrichment of provided graph/data from extraction tasks.
         data: The data to ingest. Can be anything when custom extraction and enrichment tasks are used.
               Data provided here will be forwarded to the first extraction task in the pipeline as input.
-              If no data is provided the whole graph (or subgraph if node_name/node_type is specified) will be forwarded
+              If no data is provided the whole graph (or subgraph if node_name/node_type is specified) will be forwarded.
+              The graph projection is skipped entirely when there are no extraction tasks to consume it
+              and the caller did not supply its own enrichment tasks, since the default enrichment task
+              discards non-DataPoint input. Pass data explicitly to run an enrichment-only pipeline.
         dataset: Dataset name or dataset uuid to process.
         user: User context for authentication and data access. Uses default if None.
         node_type: Filter graph to specific entity types (for advanced filtering). Used when no data is provided.
@@ -86,6 +89,11 @@ async def memify(
         ```
     """
 
+    # Track whether the caller supplied enrichment tasks before defaults are
+    # filled in: a caller-provided enrichment-only pipeline may legitimately
+    # consume the memory fragment itself, so it must keep receiving one.
+    caller_supplied_enrichment_tasks = bool(enrichment_tasks)
+
     # Use default triplet embedding tasks if no tasks were provided
     if not extraction_tasks:
         extraction_tasks = get_default_memify_extraction_tasks()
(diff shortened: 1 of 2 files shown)


plexus77:
Closing this — v1.4.2 already removes the projection, fixed at the caller rather than here.

`improve()` now passes a non-empty sentinel when no custom tasks or data are supplied, so `memify()` never reaches its fragment build ([`improve.py#L251-252`](https://github.com/topoteretes/cognee/blob/v1.4.2/cognee/api/v1/improve/improve.py#L251)):

```python
if not any(kwargs.get(key) for key in ("extraction_tasks", "enrichment_tasks", "data")):
    kwargs["data"] = [{}]
```

Verified against the released 1.4.2 wheel: the sentinel is set, and `memify.py`'s `if not data:` branch at L101 is skipped. That covers the path #4233 was actually about — every `remember` → `improve` call — and does it without touching `memify()`'s contract. Thanks for sorting it.

What this PR still does that 1.4.2 doesn't is gate the projection inside `memify()` itself, so a **direct** `memify()` call with no data and no task that can consume a fragment also skips the full-graph read. That's defence-in-depth against a caller that forgets the sentinel, not the bug I reported — not worth holding an open PR for. Happy to re-open if you'd ever want the guard at that level.

For anyone landing here from #4233: item (2) is done as above, item (1) is not — `remember()` still defaults to `self_improvement=True` and `cognee_client.remember()` still builds its kwargs without it, so an MCP deployment has no per-call opt-out. `improve()` is much cheaper now, but it still runs on every `remember`.

PR #4240, opened 2026-07-28
feat(llm): auto-enable the RPM limiter on overload evidence

dexters1:
## What this does

Cognee runs **unbounded, at full speed, for every provider**. When there is evidence the LLM provider cannot keep up, **one warning is logged and the existing RPM limiter is enabled** (`LLM_RATE_LIMIT_REQUESTS` per `LLM_RATE_LIMIT_INTERVAL`, default 60/min) for the rest of the process. That is the whole mechanism.

| Setting | Behavior |
|---|---|
| default (`AUTO_RATE_LIMIT=true`) | full speed until overload evidence → warning + RPM limiter for a 900s (15 min) cooldown, extended while issues persist; lapses back to the configured state afterwards |
| `AUTO_RATE_LIMIT=false` | fully unbounded always; errors handled by retries as before |
| `LLM_RATE_LIMIT_ENABLED=true` | RPM limiter on from the start (unchanged behavior) |

`AUTO_RATE_LIMIT` is a new config flag, on by default, documented in `.env.template` next to the other rate-limiter settings.

The RPM budget itself is provider-aware: when `LLM_RATE_LIMIT_REQUESTS` is not explicitly configured, serial local inference servers (Ollama, LM Studio, llama.cpp) default to **10** requests per interval instead of the regular 60 — vLLM deliberately keeps the regular settings, since its continuous batching absorbs concurrency like a cloud endpoint — resolved in `LLMConfig` as a `model_validator(mode="after")` post-step, so an explicitly set value always wins.

### Overload evidence
1. **A rate-limit error** — how cloud providers signal overload; both RPM and TPM limits arrive this way.
2. **HTTP 503 / 529** — Ollama past its queue cap, Anthropic overloaded.
3. **A timeout** — how overwhelmed local servers usually surface (they never send rate limits).

Response time is deliberately **not** evidence: a successful completion, however slow, proves the request was served — some models legitimately take 500 s while accepting all requests.

Errors are recognized through the exception **cause chain**, not the top-level type — instructor wraps provider errors in `InstructorRetryException` (found via live testing). `openai.RateLimitError`/`APITimeoutError`/`APIStatusError` cover every client stack: litellm's exceptions subclass them and SDK-direct adapters raise them natively. Detection lives in the dispatch seam (`llm_rate_limiter_context_manager()`) all 8 instructor adapters wrap around each attempt, inside their tenacity retry — so once enabled, retried attempts are paced too.

### Also in this PR
- **Default `chunks_per_batch` raised to 2000** (main cognify pipeline; `CHUNKS_PER_BATCH` config still overrides; temporal pipeline default unchanged at 10).

## Validation (live)
- **Ollama phi4, healthy**: full speed, no warnings, no pacing.
- **Ollama phi4, flooded (~690 concurrent from a full War-and-Peace cognify)**: overload detected; warning logged; limiter enabled.
- **OpenAI gpt-5-mini**: 50 concurrent extraction calls, 50/50 in 21 s, mechanism never engaged — zero interference on healthy cloud traffic.
- **29 unit tests** across the policy (evidence taxonomy, cooldown episodes), the dispatch seam (wiring, opt-outs, explicit `LLM_RATE_LIMIT_ENABLED` unchanged), and the config defaults (provider-aware budget, explicit values win).

## Notes
- The reaction is a cooldown, not a latch: it lapses after 900s (15 min) without overload evidence, returning to the configured state (unbounded unless `LLM_RATE_LIMIT_ENABLED` is set). A fresh episode warns again.
- On an already-flooded serial local server the limiter paces retries and later work; it cannot un-queue requests that were already dispatched.
- Known follow-up discovered during testing: `litellm.UnsupportedParamsError` (a permanent config error) is retried as transient for the full 240 s tenacity floor per call — belongs in the terminal error classification (separate PR).

---
⚠️ Note for reviewers: needs a Linear issue key added to the title (`linear-issue-check`).

### DIFF
--- a/cognee/infrastructure/llm/config.py
+++ b/cognee/infrastructure/llm/config.py
@@ -32,6 +32,12 @@ KNOWN_LLM_PROVIDERS = frozenset(
     }
 )
 
+# Default RPM budget for local inference servers (Ollama, LM Studio, vLLM,
+# llama.cpp) when LLM_RATE_LIMIT_REQUESTS is not explicitly configured — they
+# process requests (near-)serially, unlike cloud providers which keep the
+# regular default of 60.
+LOCAL_DEFAULT_RATE_LIMIT_REQUESTS = 20
+
 
 class LLMConfig(BaseSettings):
     """
@@ -101,6 +107,8 @@ class LLMConfig(BaseSettings):
     temporal_graph_prompt_path: str = "generate_event_graph_prompt.txt"
     event_entity_prompt_path: str = "generate_event_entity_prompt.txt"
     llm_rate_limit_enabled: bool = False
+    # Default 60 requests per interval; local inference servers get
+    # LOCAL_DEFAULT_RATE_LIMIT_REQUESTS instead (see default_local_rate_limit_budget).
     llm_rate_limit_requests: int = 60
     llm_rate_limit_interval: int = 60  # in seconds (default is 60 requests per minute)
     llm_rate_limit_tokens: int = 0  # max tokens per interval (0 = disabled)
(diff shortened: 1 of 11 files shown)


cognee/infrastructure/llm/config.py
```diff
@@ -32,6 +32,29 @@
...
+# by first-class provider name, or by litellm model routing prefix (LM Studio
+# has no first-class provider in cognee). vLLM is deliberately NOT here: it
+# serves with continuous batching and absorbs concurrency like a cloud
+# endpoint, so it keeps the regular settings.
+LOCAL_LLM_PROVIDERS = frozenset({"ollama", "llama_cpp"})
+LOCAL_LLM_MODEL_PREFIXES = ("lm_studio/",)
+
+# Default RPM budget for local inference servers when LLM_RATE_LIMIT_REQUESTS
+# is not explicitly configured; cloud providers keep the regular default of 60.
+LOCAL_DEFAULT_RATE_LIMIT_REQUESTS = 20
```
dexters1:
```suggestion
LOCAL_DEFAULT_RATE_LIMIT_REQUESTS = 10
```

PR #4249, opened 2026-07-28
fix(workers): close the Linux parent-death spawn-window race that orphans DB workers holding locks

GrowDev1:
## The bug

`set_pdeathsig()` arms `PR_SET_PDEATHSIG` **inside the child**, at the top of `run_worker_loop`. `PR_SET_PDEATHSIG` is not retroactive — it fires only for a parent death occurring *after* the call. Workers are `spawn`-started, so there is a real window (re-exec, import chain, unpickle) between `proc.start()` and the arm. If the parent dies in that window — an OOM kill of the parent lands there readily — no signal is ever delivered.

Worse, `set_pdeathsig()` returns `True` because the *syscall* succeeded, so the call site

```python
if not set_pdeathsig():
    start_parent_liveness_watchdog()
```

**suppresses the fallback.** And that fallback could not have helped anyway: it reads its `getppid()` baseline *after* the reparenting has already happened, so it can never fire for a death that already occurred.

Net result: an orphaned worker blocked in `serve()`, holding an exclusive Kuzu/LanceDB file lock and blocking every subsequent open of that database until someone kills it by hand.

This is the symptom class in #3708 — *"Kuzu worker crash leaves orphaned database lock, blocking all subsequent recall() calls… Lock is held by PID XXXXX"* — which was closed by #3720 fixing a **different** root cause (an async engine-resolution race). The suppressing gate itself came from #2830, which is correct for the PID-1 case it was written for; the spawn window was simply never considered.

## Why not `os.getppid()`

The obvious fix is to compare `getppid()` against the launch-time parent pid. **That is wrong and outage-shaped.** Measured on CPython 3.14.4 with the parent alive throughout:

```
spawn        parent_process().pid=199682  getppid()=199682
fork         parent_process().pid=199682  getppid()=199682
forkserver   parent_process().pid=199682  getppid()=199706   <-- differ
```

Under `forkserver` the two differ **by design** (the real parent is the fork server), so that check reports "parent dead" for a healthy parent and `os._exit(0)`s **every worker at startup**. Python 3.14 makes `forkserver` the default start method on Linux.

`getppid()` is also self-defeating here: inside the worker it has already been rewritten to the reaper's pid in exactly the scenario being detected, so it cannot serve as its own baseline.

**The signal used instead is `multiprocessing.parent_process().is_alive()`** — the parent-sentinel pipe fd. The parent holds the far end open for the child's whole life, so the kernel releases it on *any* exit including `SIGKILL`. Kernel-guaranteed, start-method agnostic, PID-reuse immune. Confirmed to flip within ~10 ms of a parent `SIGKILL` under all three start methods.

## Changes

All POSIX-gated and fail-safe — exit only on **confirmed** death; every ambiguity returns `False`:

- `get_original_parent_pid()` — from `parent_process().pid`; `None` when not an mp child.
- `_parent_sentinel_alive()` — tri-state; `None` on *any* error, never "dead" from an error path.
- `parent_already_exited()` — `False` on win32 (deliberately POSIX-scoped); sentinel first; `getppid()` only as a fallback when no sentinel exists, **additionally gated on the start method being spawn/fork** so a sentinel error under `forkserver` can never reach the outage-shaped comparison.

`run_worker_loop` captures the parent pid **before** arming, leaves the existing block untouched, and checks `parent_already_exited()` **after** it — so one call site covers both paths, and the union with arming is exhaustive: a death before that line is caught there, a death after it by whatever armed above.

Also hardens the ctypes call: explicit `restype`/`argtypes` on `libc.prctl`, `use_errno=True` on the `CDLL(None)` fallback that dropped it, `set_errno(0)` before the call, and errno surfaced on failure — a seccomp-blocked `prctl` was previously completely silent.

## Tests

`cognee_db_workers/tests/test_linux_parent_death_race.py` — 10 tests, `skipif` non-Linux. **Run, not shipped unrun**, on two configurations:

| environment | Python | default start method | result |
|---|---|---|---|
| Ubuntu (WSL2) | 3.14.4 | `forkserver` | 10/10 |
| Ubuntu (bare metal) | 3.12.3 | `fork` | 10/10 |

Covers the race under spawn and forkserver, a no-false-positive guard under spawn/fork/**forkserver** (the explicit regression guard for the `getppid` trap), the normal pdeathsig path, and fail-safe unit semantics. Liveness is pinned via `/proc/<pid>/stat` field 22 (`starttime`) so PID reuse cannot fake a pass; state `Z` is treated as gone.

**Mutation-tested:** removing the new check turns exactly the two arm-window tests red — not a blanket failure — the mutant still compiles so the failure is behavioural rather than a `SyntaxError`, and restoring returns 10/10.

## Known, not addressed here

Under `forkserver`, `prctl` anchors to the OS parent, which is the **fork server** rather than the launching process, so a worker can outlive its real parent (observed surviving a full 20 s after the real parent was `SIGKILL`ed). That is a **distinct** gap from the arm-window race and latent today, since the kuzu and lancedb proxies both pin `spawn` explicitly. No test asserts that survival — doing so would lock the behaviour in and go red when someone fixes it.

## Note

Independent of my open #4116 (Windows parent-death), and does not depend on it merging.


Vasilije1990:
Handled

PR #3115, opened 2026-06-20
fix(security): restrict global settings and disable public registration

aniruddhaadak80:
Closes #3084

This PR addresses the security vulnerabilities reported in #3084:
- Requires superuser privileges for POST /api/v1/settings to prevent global configuration takeover.
- Fully masks LLM and VectorDB API keys in GET /api/v1/settings to prevent leaking key prefixes.
- Adds a COGNEE_PUBLIC_REGISTRATION_ENABLED environment variable to allow administrators to disable public self-registration.

PR #4252, opened 2026-07-28
fix(security): require superuser to modify LLM/vector DB settings

### LINKED ISSUE #4146: Security fix for CVE-2026-58473 (GHSA-49f7-whx5-4256) appears reverted 6 minutes after merge — all releases since still lack the guard, but advisories say "fixed in 1.2.0"
Hi — while evaluating cognee against CVE-2026-58473 / GHSA-49f7-whx5-4256 (unauthenticated read/overwrite of LLM configuration via `/api/v1/settings`, CWE-306/CWE-862, CVSS v3.1 9.1), I found what looks like an accidental revert of the fix, and I wanted to flag it because the public advisories currently tell users that upgrading remediates the issue.

**Timeline (all UTC):**

| Commit | When | Effect |
| --- | --- | --- |
| `d10b1b77e215` (PR #3115) | 2026-06-20 10:51 authored / 16:59 merged | adds the `is_superuser` guard to the settings routes |
| `10971db1b0bd` | 2026-06-20 **17:05** | **reverts it — 6 minutes after merge** |

`v1.2.0` was tagged 2026-06-21, i.e. **after** the revert, so the tag contains the revert, not the fix.

**Verification (repeatable):** grepping the release trees of 1.2.0, 1.2.1, 1.2.2, 1.3.0, and 1.4.0, plus `main` and `dev` HEAD (as of 2026-07-19), the `is_superuser` guard count in the settings router is **0** in every one of them, and `get_settings.py` still uses the older partial API-key masking (`llm_api_key[0:10] + "*" * (len - 10)`). I could not find any later commit that re-lands the change. Note that commit *ancestry* is misleading here — `d10b1b77e215` IS an ancestor of the release tags (a `compare` shows "behind: 0"), but its change was immediately reverted, so checking for the commit rather than the code gives a false "fixed" signal.

**The question:** was the revert in `10971db1b0bd` intentional? Its message is a bare `Revert "fix(security): ...)"` with no stated rationale, which made me suspect a merge-queue or automation accident rather than a deliberate decision to withdraw a security fix.

**Why it matters:** VulnCheck / NVD / GHSA-49f7-whx5-4256 all state the issue is fixed in ≥ 1.2.0. Anyone who "remediated" by upgrading is running unpatched while believing otherwise — arguably worse than a known-unpatched state, because the issue leaves their backlog. If the revert was accidental, it would be worth (1) re-landing the guard, and (2) correcting the advisory so the fixed-version claim matches a release that actually contains the fix.

Happy to provide the per-release grep transcript if useful. Reported in good faith; no exploitation details included beyond what the public advisory already documents.

SahilArate:
Description
Re-adds the is_superuser guard on POST /api/v1/settings that was introduced in #3115 but reverted 6 minutes later in 10971db1b0bd along with unrelated changes. This PR contains only the guard.

What was broken: POST /api/v1/settings only checked get_authenticated_user, not is_superuser so any registered account could overwrite the global LLM/vector DB config.

Why it was reverted before: I checked the original revert diff  PR #3115 bundled the guard together with two unrelated changes (a registration-disable env var, and API key masking changes). This PR restores only the guard, nothing else.

Addresses CVE-2026-58473 / GHSA-49f7-whx5-4256.

Testing: Added test_settings_authorization.py with two tests 

Screenshot 2026-07-29 005019

non-superuser POST returns 403, superuser POST still returns 200. Ran locally, both pass (screenshot below).

Note: Saw @nightcityblade express interest in #4146, but no PR had appeared after several days, so I went ahead given how long this CVE's been unpatched.

Fixes #4146 
Pre-submission Checklist
 [✅]I have tested my changes thoroughly before submitting this PR
 [✅]This PR contains minimal changes necessary to address the issue/feature
 [✅]My code follows the project's coding standards and style guidelines
 [✅]I have added tests that prove my fix is effective
 [✅]I have added necessary documentation (if applicable)
 [✅]All new and existing tests pass
 [✅]I have searched existing PRs to ensure this change hasn't been submitted already
 [✅]I have linked any relevant issues in the description
 [✅]My commits have clear and descriptive messages

### DIFF
--- a/cognee/api/v1/settings/routers/get_settings_router.py
+++ b/cognee/api/v1/settings/routers/get_settings_router.py
@@ -1,8 +1,8 @@
-from fastapi import APIRouter
+from fastapi import APIRouter, Depends
+from fastapi.responses import JSONResponse
 from cognee.api.DTO import InDTO, OutDTO
 from typing import Union, Optional, Literal
 from cognee.modules.users.methods import get_authenticated_user
-from fastapi import Depends
 from cognee.modules.users.models import User
 from cognee.modules.settings.get_settings import LLMConfig, VectorDBConfig
 
@@ -89,9 +89,16 @@ def get_settings_router() -> APIRouter:
         No content returned on successful save.
 
         ## Error Codes
+        - **403 Forbidden**: Caller is not a superuser
         - **400 Bad Request**: Invalid settings provided
         - **500 Internal Server Error**: Error saving settings
         """
+        if not user.is_superuser:
+            return JSONResponse(
+                status_code=403,
+                content={"error": "Superuser privileges required to modify settings"},
+            )
+
         from cognee.modules.settings import save_llm_config, save_vector_db_config
 
         if new_settings.llm is not None:
--- a/cognee/tests/api/test_settings_authorization.py
+++ b/cognee/tests/api/test_settings_authorization.py
@@ -1,0 +1,87 @@
+import os
+import pytest
+from unittest.mock import patch
+
+with patch("dotenv.load_dotenv"):
+    os.environ["REQUIRE_AUTHENTICATION"] = "true"
+    os.environ["ENABLE_BACKEND_ACCESS_CONTROL"] = "false"
+    os.environ["HASH_API_KEY"] = "false"
+
+    from fastapi.testclient import TestClient
+
+
+TEST_USER_EMAIL = "settings_test_user@example.com"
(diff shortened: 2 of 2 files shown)


dexters1:
@SahilArate thank you for your contribution! Your changes have been merged and will be part of our next release.

Your commit was also stored and will be part of our git history.

I had to make a new PR to run our full CI on the changes

PR #4254, opened 2026-07-28
feat(chunking): support nested arrays in JsonListChunker

### LINKED ISSUE #4236: [Feature]: JsonListChunker & Array-Root JSON Nested Schema Flattening
### Problem Statement

What is the issue?
Currently, JsonListChunker.py handles chunking lists of JSON objects. However, when ingesting large structured documents with deeply nested lists inside dictionary roots (e.g. {"records": [{"items": [...]}]} or complex JSON structures containing nested arrays), the chunker fails to flatten or preserve path metadata for the nested records.

Add support for nested JSON array flattening and root path metadata tracking in JsonListChunker.

Currently, structured document ingestion using JSON list chunking expects flat top-level JSON arrays. When ingesting JSON documents where lists of target objects are nested within top-level keys or deep dictionaries, the chunker fails to identify and extract child records, causing full document fallback or dropped records.

Implement root path traversal and array normalization in the JSON chunking module. When provided with a JSON document structure, the chunker should automatically detect nested array properties or accept a JSONPath/key selector string. Each emitted chunk should preserve its parent JSON key path as metadata (e.g., `json_path: "records.items[0]"`). 

When chunking nested lists, the chunker must preserve high-level contextual parent keys across all generated record chunks so that chunk payloads maintain their enclosing structural context during vector indexing.

### Proposed Solution

no

### Alternatives Considered

_No response_

### Use Case

no

### Implementation Ideas

no

### Additional Context

no

### Pre-submission Checklist

SahilArate:
Adds support for JSON documents where the target list is nested inside a dict (e.g. {"records": {"items": [...]}}) instead of only a flat top-level list. Auto-detects a single nested array, or accepts an explicit json_path when multiple arrays exist. Each chunk now carries json_path and, when present, json_context metadata. Flat top-level list behavior is unchanged. Fixes #4236.

## Description

Fixes #4236.

JsonListChunker only worked when the JSON was a plain list at the top, like [{...}, {...}]. If the list was hidden inside a dict, like {"records": {"items": [...]}}, the chunker had no way to find it and would just fail.

I fixed this by adding a way to look through the JSON and find a list even if it's nested inside other keys. If there's only one list found, it's used automatically. If there's more than one list in the document, the code doesn't guess it raises a clear error and asks the user to pass a json_path (like "records.items") to say exactly which list to use. I also added two new pieces of info to every chunk:

json_path shows exactly where the chunk came from in the original JSON, like "records.items[0]".
json_context if the list has other simple values sitting next to it (like a batch_id), those are carried along with each chunk so that info isn't lost. The old behavior for flat lists is exactly the same as before nothing changes for existing users.

## Acceptance Criteria

• Flat top-level JSON lists still work exactly like before.
• A dict with one nested list gets found and chunked automatically.
• A dict with more than one nested list raises a clear error instead of guessing.
• Passing json_path picks the correct list even when there are multiple.
• A missing list, or a bad json_path, gives a clear error message.
• Every chunk includes json_path in its metadata. json_context is added only when there's context to add.
Proof: all 36 tests in the chunking test folder pass, including 8 new tests I added for this feature (see screenshot).

## Type of Change

- [✅ ] New feature (non-breaking change that adds functionality)
- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots
Screenshot 2026-07-29 023609
Screenshot 2026-07-29 023643

## Pre-submission Checklist

- [✅ ] **I have tested my changes thoroughly before submitting this PR** (See `CONTRIBUTING.md`)
- [ ✅] **This PR contains minimal changes necessary to address the issue/feature**
- [ ✅] My code follows the project's coding standards and style guidelines
- [ ✅] I have added tests that prove my fix is effective or that my feature works
- [ ✅] All new and existing tests pass
- [ ✅] I have searched existing PRs to ensure this change hasn't been submitted already
- [✅ ] I have linked any relevant issues in the description
- [✅ ] My commits have clear and descriptive messages

## DCO Affirmation


Vasilije1990:
@SahilArate please read contributing.md and open a PR to dev

Kirtofu:
Thanks for this PR, @SahilArate — the structure is clean and the test matrix is a good start. I reviewed it against the current `JsonListChunker` and issue #4236, and also compared it with draft #4503, which targets the same issue. The flat-list path and the explicit-`json_path`/error handling are solid; below are the gaps I found, one of which means the feature as written doesn't satisfy the issue's canonical example.

## A. Ancestor scalar context is dropped (the context feature is incomplete)

`_find_arrays` computes `context` from the array's **immediate parent dict**, and `_resolve_json_path` does the same. For the shape:

```json
{"batch_id": "B1", "records": {"items": [...]}}
```

the root `batch_id` never reaches the chunks — `json_context` is empty at the `records` level, so the metadata ends up `{}` (or absent). The existing `test_sibling_context_is_preserved` only covers the case where `batch_id` is a sibling of the array at the same level (`{"records": {"batch_id": "B1", "items": [...]}}`), so the cross-level case is both broken and untested. This is the exact traceability case the feature promises ("info isn't lost").

Fix: accumulate ancestor scalar contexts along the traversal (merge root → leaf) in both `_find_arrays` and `_resolve_json_path`. Regression test to add:

```python
data = {"batch_id": "B1", "records": {"items": [{"name": "Alice"}]}}
# assert chunks[0].metadata["json_context"] == {"batch_id": "B1"}
```

## B. The issue's own example shape is still unsupported

Issue #4236's example is `{"records": [{"items": [...]}]}` — an array nested inside list items. `_find_arrays` only recurses into dicts, so for that document it detects `records` itself as the single array and chunks each `{"items": [...]}` object as one opaque chunk; the inner `items` array is never flattened element-wise. And `json_path` has no index syntax, so `records[0].items` cannot be expressed at all. Either add `[index]`/`[*]` selector support and list-item traversal, or explicitly scope the PR to dict-of-dict shapes and update the issue acceptance criteria — right now a reader would expect the issue's example to work.

## C. Context lives only in metadata, which isn't indexed

`metadata["json_context"]` is not part of the chunk `text`, and `index_fields` stays `["text"]`, so the carried context cannot be retrieved through the index. If the point is traceability only, that's fine — but please state it; if batch-level values should be queryable, they need to be woven into `text` (draft #4503 does exactly that, which is worth a look regardless).

## D. Chunk ids collide when the same document is chunked under different paths

`uuid5(NAMESPACE_OID, f"{document_id}-{index}")` and `chunk_index=index` are index-based. With multi-array documents it's now plausible to chunk the same document twice with different `json_path` values; overlapping indices would produce identical chunk ids/indices in the vector store. Consider deriving the id from the resolved path + index when a path is in play.

## E. Smaller edge cases worth a test or a decision

- Empty array (`{"items": []}`) auto-detects and silently yields zero chunks — decide whether that's an error or legitimate, and test it.
- Keys containing `.` can't be addressed by `json_path` (`{"a.b": [...]}`) — document the limitation or escape dots.
- `json_path` traversal that hits a list mid-path raises the generic "does not exist" error — fine, but a "path segment is a list, use [n]" hint would help.

## Summary

The approach is sound for dict-of-dict-of-list documents, but findings A and B mean the PR doesn't yet deliver the issue's stated behavior. Since #4503 covers A–C (ancestor context folded into text, `[index]`/`[*]` selectors, list-contained objects), it would be good to coordinate with that author (or the maintainers) before both of you invest further — otherwise reviewers have to pick between two overlapping implementations. Happy to re-review after the context-inheritance fix and the regression test land.

Kirtofu:
Follow-up on finding A (ancestor scalar context is dropped): I went ahead and prepared the fix so you can take it directly if you agree with the direction.

**Patch:** https://gist.github.com/Kirtofu/70350910a558f0fab4bdd153d62af3b8 (`git apply` on top of `feature/json-list-chunker-nested-arrays`)

What it does:

- `_find_arrays` gains an `inherited_context` parameter: scalar context from every ancestor dict is merged along the traversal (root → leaf, values closer to the array win on key collisions), so `{"batch_id": "B1", "records": {"items": [...]}}` now yields `json_context: {"batch_id": "B1"}` instead of dropping it.
- `_resolve_json_path` merges ancestor context the same way, so the explicit `json_path` branch behaves identically to auto-detection.
- Adds two regression tests (`test_ancestor_context_is_propagated`, `..._with_explicit_json_path`) covering both branches.

Verified locally against your branch: a regression suite extracted the helpers verbatim and went from `1 passed, 3 failed` (before) to `3 passed, 1 failed` (after) — the remaining failure is finding B (the `{"records": [{"items": [...]}]}` shape from the issue), which I deliberately left out since it involves the selector-syntax question that overlaps with #4503's `[index]`/`[*]` design.

If you'd rather implement it yourself, no problem — the patch is just the review finding made concrete. Either way I'm happy to re-review.

SahilArate:
Hey @Kirtofu, thanks a lot for this review, it really helped me understand what was missing.

You are right about both things.

For the first one (batch_id getting lost), I will use your patch and add the two tests you wrote. That part looks clear to me.

For the second one (the records: [{items: [...]}] shape from the issue), I want to try and fix that too, not just skip it. I know it means adding support for arrays inside list items and maybe adding something like [index] or [*] in json_path. I will work on it and see how far i will take this.

I also saw there is another PR (#4503) working on the same issue. I don't want us to do double work. Can you or the maintainers tell me a bit more about what that PR already covers? That way I know what parts I actually need to build, and what parts are already done by them.

Will update this PR soon with the context fix first, then work on the nested list item part.

Kirtofu:
@SahilArate glad the review helped, and thanks for the clear plan.

**Part one (batch_id context):** the patch is yours — go ahead and apply it with the two tests. Nothing to add on my side.

**Part two (arrays inside list items + selector syntax):** I pulled draft #4503 to answer your question precisely, because you asked before building. Short version: **that PR already implements what you planned**, so doing it here would be duplicated work:

- `_find_arrays` there traverses **into list-contained objects** and retains scalar context from **every ancestor level** (not just the immediate parent) — `cognee/modules/chunking/JsonListChunker.py`
- selector syntax `[index]` / `[*]` is implemented via `_parse_json_path` / `_path_matches` (e.g. `records[*].items`)
- it also folds ancestor context into the **indexed chunk text** (`_contextual_text`), not only metadata, and adds a `JsonListChunker.with_json_path(...)` class factory for `cognify(chunker=...)` callers
- 8 tests cover the traversal, normalized paths (`records[*].items`), ambiguity → explicit selector, the factory API, and invalid-selector rejection

It's a draft (+420/−31, untouched by reviewers so far), but the scope is exactly the second half of your plan plus more.

So my suggestion:

1. Land **only the context fix** here (our patch + your two tests) — it's small, reviewable, and independent of the selector design.
2. For the `{"records": [{"items": [...]}]}` shape, don't re-implement it — either wait for #4503 to be reviewed/merged, or reach out to its author (cucuwang) to coordinate (e.g. you could review their draft with the same lens, or they could split the selector machinery out). Duplicating it would just make the maintainers choose between two overlapping implementations.
3. Whatever you push, keep in mind Vasilije1990's earlier note to retarget this PR to `dev`.

Happy to re-review part one once it lands.

Issue #4281, opened 2026-07-30
`POST /v1/skills` returns generic 409 "Failed to ingest skill" for every exception, masking unrelated ingest bugs


SvetZitrka:
## Environment

- cognee 1.4.0, self-hosted (Docker) on an Apple M4 Mac mini
- Batch-ingesting 378 SKILL.md files via `POST /api/v1/skills`

## Observed

2 of 378 ingests failed. The API response in both cases was an identical `409 {"error": "Failed to ingest skill"}`, which reads like a duplicate/conflict. The server log shows the real causes were two different bugs, neither of which is a conflict:

- `ingest skill failed: 1, exception_type=KeyError` — raised while parsing one skill's SKILL.md
- `ingest skill failed: Graph edge indexing error, exception_type=RuntimeError` — raised during edge indexing for another skill

## Cause (verified against source)

`ingest_skill` in `cognee/api/v1/skills/routers/get_skills_router.py` wraps the whole pipeline in a catch-all:

```python
except Exception as exc:
    logger.error("ingest skill failed: %s", exc, exc_info=True)
    return JSONResponse(status_code=409, content={"error": "Failed to ingest skill"})
```

Every failure — parse errors, graph-store errors, genuine conflicts — collapses into the same 409, so callers cannot distinguish "already exists" from "server bug" and cannot report anything actionable without access to server logs.

## Suggestion

1. Map exceptions to meaningful statuses (400 for malformed skill input, 500 for internal pipeline errors, 409 only for actual conflicts) and include the exception type/message in the error body.
2. Separately, the two underlying failures (the `KeyError` in skill parsing and the edge-indexing `RuntimeError`) look like real ingest bugs — I have full tracebacks from `~/.cognee/logs` (2026-07-30T20:45:57Z and 20:59:14Z) and can attach them, plus the offending SKILL.md files, if useful. Happy to split these into their own issues if you prefer.

MohdMaaz000:
I’d like to contribute to this issue.

I reviewed the current `ingest_skill` error handling in
`cognee/api/v1/skills/routers/get_skills_router.py`. The broad
`except Exception` currently maps unrelated failures to HTTP 409,
which makes genuine server-side failures indistinguishable from
actual conflicts.

My proposed approach is to preserve 409 only for genuine conflict
conditions and route validation/input errors and unexpected internal
exceptions through the existing Cognee/FastAPI error-handling
patterns, rather than introducing a second error-handling mechanism.

[...]

Issue #4279, opened 2026-07-30
Skills page silently truncates at 200 skills — frontend sends no limit/offset while `GET /v1/skills` defaults to `limit=200`


SvetZitrka:
## Environment

- cognee 1.4.0, self-hosted (Docker) on an Apple M4 Mac mini
- Dataset with 378 ingested skills (SKILL.md files via `POST /api/v1/skills`)

## Observed

The Skills page permanently shows "200 skills" even though the dataset contains 378. There is no truncation indicator and no pagination control, so from the UI it is impossible to tell that ~half the skills are missing from the list.

## Cause (verified against source)

Frontend and backend disagree about pagination:

- `cognee-frontend/src/modules/skills/getSkills.ts` builds the query with only `dataset_id` and `include_inactive` — no `limit`/`offset` — and treats the response as the complete list.
- `cognee/api/v1/skills/routers/get_skills_router.py` (`list_dataset_skills`) declares `limit: int = Query(default=200, ge=1, le=1000)` and silently applies it.

So any dataset with more than 200 skills is silently cut off at exactly 200.

## Suggestion

Any of these would fix it (in increasing order of effort):

1. Frontend passes an explicit high `limit` (quick fix, still capped at the router's `le=1000`)
2. Frontend paginates / load-more using the existing `offset` parameter
3. Backend returns a total count (header or envelope) so the UI can at least show "200 of 378"

Related note: the same "truncated data with no indication" pattern also affects the Memory Activity feed and the session transcript view — a shared list-view convention (pagination or an explicit truncation indicator) might be worth considering.

ArmagedonFlamer:
Thanks for flagging this - it's been seen and noted. It's on us to address now - we'll let you know how it goes.

Issue #4280, opened 2026-07-30
Skills API has no delete/upsert — re-ingesting a skill creates a duplicate node with no API path to remove it


SvetZitrka:
## Environment

- cognee 1.4.0, self-hosted (Docker) on an Apple M4 Mac mini
- 378 skills ingested via `POST /api/v1/skills`

## Observed

`POST /api/v1/skills` is not idempotent with respect to `skill_name`: re-ingesting the same skill (e.g. after editing its SKILL.md) creates a second Skill node instead of updating the existing one. After a partial re-run of our ingest we now have 3 duplicate skills in the dataset.

There is no API path to clean this up:

- The skills router (`cognee/api/v1/skills/routers/get_skills_router.py`) exposes only `POST` (ingest), `GET /` (list) and `GET /{skill_id}` (detail) — no `DELETE`, no update.
- Skills are not data items, so `DELETE /api/v1/datasets/{dataset_id}/data/{data_id}` does not apply to them.

So duplicates accumulate permanently unless one edits the graph store directly.

## Suggestion

- `DELETE /api/v1/skills/{skill_id}` (dataset-scoped, mirroring the existing detail endpoint), and
- either upsert-by-name semantics on `POST` (replace when `skill_name` already exists in the dataset), or an explicit `PUT /api/v1/skills/{skill_id}`

Upsert matters for the natural workflow of keeping a skill library in sync with a directory of SKILL.md files — today every refresh has to be "wipe and re-ingest", and even the wipe half is not possible via the API.

Mefisto04:
Hi! I'd like to work on #4280.

**Proposed fix:** Add `DELETE /api/v1/skills/{skill_id}` and optionally upsert by `(dataset_id, skill_name)` so re-ingest updates instead of duplicating.

Please assign if you'd like a PR for this. Happy to discuss scope (delete-only vs full upsert). Thanks!

PR #4290, opened 2026-07-31
fix(skills): make ingest idempotent by name and add delete endpoint

ApexYash11:
## Description

Re-ingesting a skill through `POST /api/v1/skills` was creating a duplicate
Skill node instead of updating the existing one. Traced it down: skill IDs
are supposed to be deterministic (`_scoped_skill_id` hashes
`dataset_id + source_dir + skill.name`), and `add_data_points` already
upserts by node id — so the storage layer was never the problem. The actual
bug was that both the inline-text and file-upload ingestion paths in
`remember.py` materialized the skill into a fresh `TemporaryDirectory()` on
every single call. That randomized `source_dir` each time, which meant the
"deterministic" id was actually different on every POST for the same skill —
hence the duplicates.

Fix: stage materialized skills under a deterministic per-dataset root
(`cognee-skills-<sha256(dataset_id)[:16]>`) instead of a random tempdir, with
the skill's slug as a subfolder under it. That keeps `source_dir` stable per
`(dataset_id, skill_name)` so re-ingesting the same skill now upserts in
place — same node, refreshed content/embedding — instead of minting a new
one. Both the inline and upload paths had the bug, so both are fixed.

Also added the missing `DELETE /api/v1/skills/{skill_id}` endpoint (the
issue points out there was no way to remove a skill at all). It's
dataset-scoped, requires `delete` permission (matching how the other
dataset/data deletion endpoints are gated — the ingest path only needs
`write`, but deletion is a more destructive action and deserves its own
permission tier). It hard-deletes the graph node plus its
`Skill_search_text` vector embedding rather than soft-deleting via
`is_active=False` — soft delete would leave a resident node that a
re-ingest (now that ids are stable) would silently resurrect, which seemed
like a worse footgun than an irreversible delete.

Left untouched: the skills-router pagination bug (#4279) and the generic
409-on-ingest-failure issue (#4281) — same file, different bugs, out of
scope here.

## Acceptance Criteria

- Re-ingesting a skill with the same `skill_name` into the same dataset
  updates the existing Skill node (same id) instead of creating a duplicate.
- Re-ingesting the same `skill_name` into a *different* dataset produces a
  distinct id (no cross-dataset collision).
- `DELETE /api/v1/skills/{skill_id}` removes the graph node and its
  embedding; returns 403 for a dataset the caller doesn't have `delete`
  access to, 404 for an unknown skill id.
- No behavior change to path-based (folder) skill ingestion.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/api/v1/remember/remember.py
+++ b/cognee/api/v1/remember/remember.py
@@ -1,5 +1,6 @@
 import asyncio
 import time
+from pathlib import Path
 from uuid import UUID
 from typing import Union, BinaryIO, List, Optional, Any, Literal, TYPE_CHECKING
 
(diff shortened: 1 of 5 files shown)


ApexYash11:
@dexters1  @Vasilije1990  plss review this!!

Vasilije1990:
@ApexYash11 they should contribute in the MD and open a PR against the dev.  Too many files are changed right now.

ApexYash11:
I cleaned up the branch and force-updated this PR. It is now based on dev and contains only the five files relevant to #4280. The remaining +311/-59 includes the implementation and its unit tests. Could you please re-review the updated diff?

PR #4291, opened 2026-07-31
feat: Add opt-in per-operation activity log (COG-6037)

LStromann:
## What
Adds an **opt-in, per-operation activity log**: one event per pipeline run (`add`/`cognify`/`improve`/`remember`) carrying the acting **user + tenant**, **dataset** (id + name), **pipeline run id**, **token usage**, **status**, and **duration**. Powers a per-operation activity view for consumers like Cognee Cloud.

**Inert for open-source:** nothing is stored and there is zero overhead unless a consumer registers a sink via `register_activity_sink`. No table is added to core — persistence is the sink's responsibility.

## How (3 commits)
1. **Opt-in activity sink** (`session_lifecycle/usage_tracking.py`) — `ActivityEvent`, `register_activity_sink`, `track_operation_usage` / `begin_operation`+`end_operation` scope primitives. `record_llm_call` feeds the per-operation token accumulator independently of the session tracker.
2. **Real `response.usage`** — adapters call `report_llm_usage(prompt, completion)`; the gateway's usage hook consumes the total (accumulated across a call's internal retries) and passes it as the `record_llm_call` token overrides. **litellm-native** fully wired; falls back to the char estimate when unreported.
3. **Emit from `run_tasks`** — open an activity scope around each pipeline run (before the per-item tasks are created, so their LLM usage accrues), emitting the finished event on completion/error. No restructuring of the async generator.

## Follow-ups (noted, not in this PR)
- **instructor** (`generic_llm_api` is the OpenAI-compatible/`custom` path) + **BAML** real-usage capture (`create_with_completion` / a `Collector`) — until then those frameworks use the char estimate.
- Instrument non-pipeline ops: `forget`, `recall`/`search`, session-`remember`.

## Notes
- Cost is intentionally left to the consumer (tokens × rate); this PR stores tokens.
- Verified `py_compile` + `ruff format`; relies on CI for the full suite (local env can't build `ladybug`).

Part of COG-6037.

### DIFF
--- a/cognee/api/v1/forget/forget.py
+++ b/cognee/api/v1/forget/forget.py
@@ -10,10 +10,12 @@ from cognee.modules.observability import (
     COGNEE_FORGET_TARGET,
     COGNEE_RESULT_COUNT,
 )
+from cognee.modules.session_lifecycle.usage_tracking import log_activity
 
 logger = get_logger("forget")
 
 
+@log_activity("forget")
 async def forget(
     *,
     data_id: Optional[UUID] = None,
--- a/cognee/api/v1/remember/remember.py
+++ b/cognee/api/v1/remember/remember.py
@@ -171,6 +171,20 @@ async def _add_to_session(session_id: str, data, user):
         context="",
         answer=text,
     )
+
+    # Activity log: record the session write (no LLM cost). Inert without a sink.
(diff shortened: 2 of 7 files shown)


LStromann:
Closing for tracking — the activity-log work is tracked in COG-6037. Branch `feat/activity-log-usage-sink` is preserved; reopen when we pick this up.

PR #4231, opened 2026-07-27
fix(mcp): preserve original filename for API-mode file uploads

danceballos:
### Description

- Fixes #4230.
- `cognify_file` intentionally writes a tempfile whose basename is the sanitized original filename so `Data.name` stays readable.
- In API mode, `CogneeClient.add()` always called `_text_upload(data)`, which:
  1. stringified the path and uploaded it as text content, and
  2. forced a content-addressed `text_<md5>.txt` name.
- That content-addressed behavior is still correct for raw text (keeps #2747 fixed).
- When `data` is an existing filesystem path, `add()` now multipart-uploads the real file bytes with the original basename.

### Acceptance Criteria

- [x] `add("plain text")` in API mode still uploads as `text_<md5>.txt` (no `data.txt` regression).
- [x] `add("/tmp/.../meeting-notes.md")` for an existing file uploads bytes with `filename="meeting-notes.md"`.
- [x] Unit coverage for both paths.
- [x] No change to auth / Bearer vs X-Api-Key behavior.

### Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Code refactoring
- [ ] Other (please specify):

### Screenshots

```text
$ uv run --python 3.12 --project cognee-mcp pytest \
    cognee-mcp/tests/test_mcp_server_hardening.py::test_cognee_client_api_add_uses_content_addressed_filename \
    cognee-mcp/tests/test_mcp_server_hardening.py::test_cognee_client_api_add_uploads_existing_file_with_basename -q

2 passed, 10 warnings in 16.23s
```

(Existing Pydantic deprecation warnings from dependencies.)

### Testing

Commands run:

```bash
uv run --python 3.12 --project cognee-mcp pytest \
  cognee-mcp/tests/test_mcp_server_hardening.py::test_cognee_client_api_add_uses_content_addressed_filename \
  cognee-mcp/tests/test_mcp_server_hardening.py::test_cognee_client_api_add_uploads_existing_file_with_basename -q
```

Results: **2 passed**.

### Pre-submission Checklist

- [x] I have tested my changes thoroughly before submitting this PR
- [x] This PR contains minimal changes necessary to address the issue
- [x] My code follows the project's coding standards and style guidelines
- [x] I have added tests that prove the fix is effective
- [x] I have added necessary documentation (not applicable)
- [x] All new and existing MCP tests relevant to this change pass
[...]

PR #4301, opened 2026-08-01
fix(serve): Name raw-text cloud uploads by content hash

dexters1:
## Description

Every raw-text `add()`/`remember()` through `CloudClient` uploads its payload with the fixed multipart filename `data.txt`, and the server stores uploads under the client-supplied name — so **all text memories of a tenant collide on a single remote object** (`…/data/<tenant-id>/data.txt`). Rapid consecutive adds then race the server's content-hash read-back: the hasher pins the old Etag, the next add has already overwritten the object, and S3's "no longer exists" surfaces as `FileContentHashingError` 409s.

First surfaced by the nightly **cloud — 50 small documents** benchmark (3/3 runs failing at the add stage on three fresh tenants, while the single-document War-and-Peace cloud arm passed — no key collision with one document). Any real workload doing rapid small text adds to a cloud tenant hits the same failure.

## Fix

Name text uploads through `TextData` — the **same namer** local ingestion's `save_data_to_file` uses for nameless text (`text_<md5>.txt`) — rather than a placeholder. One naming source, so remote and local names can't drift: distinct content never collides, identical content dedupes to the same key.

- `_text_upload_filename()` delegates to `create_text_data(text).get_metadata()["name"]`
- All four upload sites (remember + add, single string + list) switched from `data.txt`
- Unit tests cover the format, distinctness across 50 texts, dedup of identical text, UTF-8 hashing, and exact equality with the `TextData` namer

Related: #4231 reworks filename preservation for file uploads; this PR covers the raw-text path. The server-side "trust client filename" behavior may still deserve hardening independently (a malicious/old client can still collide keys).

## Verification

- `cognee/tests/unit/api/v1/serve/` — 5 passed
- Import chain verified; pre-commit (ruff lint + format) green
- After merge, the nightly `cloud — 50_small_documents` arm (currently red on every run) is the live regression test

### DIFF
--- a/cognee/api/v1/serve/cloud_client.py
+++ b/cognee/api/v1/serve/cloud_client.py
@@ -7,9 +7,22 @@ from uuid import UUID
 
 import aiohttp
 
+from cognee.modules.ingestion.data_types.TextData import create_text_data
 from cognee.shared.logging_utils import get_logger
 
 logger = get_logger("serve.cloud_client")
+
+
+def _text_upload_filename(text: str) -> str:
+    """Content-hash filename for raw-text uploads, via local ingestion's namer.
+
+    Delegates to ``TextData`` — the same source ``save_data_to_file`` uses for
+    nameless text (``text_<md5>.txt``). A fixed placeholder here instead makes
+    every text upload for a tenant collide on one remote object, racing
+    concurrent adds against the server's content-hash read-back
+    (FileContentHashingError 409s).
+    """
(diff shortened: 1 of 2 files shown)


dexters1:
Folded into #4300 (the v1.4.1 nightly-CI follow-up) per release planning — same commit cherry-picked there (335f452a2). dev will receive it through the next main→dev merge.

PR #4302, opened 2026-08-01
CLO-398 feat: wire the capability layer over the default permission tables

goran-radonic:
## Description

The capability layer already existed as three tables that nothing read.
`TenantDefaultPermissions`, `RoleDefaultPermissions` and `UserDefaultPermissions`
carry no dataset column, which makes them a three-level capability system
mirroring how ACL grants dataset access to any principal. Three
`give_default_permission_to_*` methods write to them and nothing consumes them.

Because of that, the only way a non-owner could manage users was matching a role
literally named `admin`:

    USER_MANAGEMENT_ALLOWED_ROLE_NAMES: FrozenSet[str] = frozenset({"admin"})

So a customer group that happens to be called "admin" grants user management to
everyone in it, and adding a new level means editing a constant and redeploying
rather than changing data.

This PR reads those tables, lets capabilities be granted and revoked, and stops
`has_user_management_permission` from matching on names. The role-name set stays
as a deprecated fallback so tenants relying on an `admin` role today do not lose
access the moment they upgrade; it can be removed once those roles have been
granted the capability, which is now possible.

Three things I decided along the way:

**`UserDefaultPermissions` is deliberately not resolved.** It is keyed on the
user alone, so honouring it would grant the capability in every tenant in the
system, including ones the user does not belong to. Per-person capabilities need
a `(user, tenant, permission)` triple that does not exist yet, so the table is
left unread and the grant endpoint for it is not exposed.

**Resolution is gated on tenant membership once, at the top.** The alternative
was a join on every query, which only works as long as whoever adds the next
query remembers.

**No `kind` column on roles.** I had one in an earlier revision to let the UI
separate teams from levels, then dropped it: nothing writes or reads it, and
since a role can hold both members and capabilities, it is a label with no
invariant behind it. It belongs with the work that defines what the label means.

## Acceptance Criteria

- A capability granted at tenant or role level reaches members of that tenant
  and nobody else.
- Granting `manage_users` actually unlocks the endpoints it guards.
- Revoking is idempotent, so a retry is safe.
- Names outside the catalog are rejected, including dataset permissions
  (`read`/`write`/`delete`/`share`), which live in the same table and would
  otherwise write rows that resolution filters out again.
- The capabilities endpoint answers the same for a tenant you are not in and one
  that does not exist, so it cannot be used to discover real tenant ids.
- Tenants relying on an `admin` role keep user management after upgrading.

Verified end to end against a local instance with two users:

    1. before grant             member: []
    2. member self-grants       403  "not authorized to manage users"
    3. owner grants to tenant   200 → member: ["manage_users"]
    4. member lists users       200  ← the capability unlocks a real endpoint
    5. owner revokes            200 → member: []
    6. capability=read          400  "Unknown capability 'read'"

    outsider → real tenant        403, identical body
    outsider → nonexistent tenant 403, identical body

22 tests pass in `cognee/tests/unit/users`. The resolution tests run against a
real database in an isolated system root, because a mocked session cannot catch
a missing WHERE clause. I checked they fail for the right reason: removing the
membership gate fails only the non-member test, and moving the owner
short-circuit below the gate fails only the ordering test.

## Type of Change

## Pre-submission Checklist

## DCO Affirmation


cognee/modules/users/models/PrincipalCapability.py
```diff
@@ -0,0 +1,35 @@
+from datetime import datetime, timezone
+
+from sqlalchemy import UUID, Column, DateTime, ForeignKey, String
+
+from cognee.infrastructure.databases.relational import Base
+
+
+class PrincipalCapability(Base):
```
dexters1, line 8: Lets make sure the database creates this table when Cognee is used purely through SDK and not as a backend server

cognee/modules/users/permissions/methods/grant_capability.py
```diff
@@ -0,0 +1,66 @@
+from uuid import UUID
+
+from sqlalchemy import select
+
+from cognee.infrastructure.databases.relational import get_relational_engine
+from cognee.modules.users.exceptions import PermissionDeniedError
+from cognee.modules.users.models import PrincipalCapability, Role, Tenant
+
+
+async def grant_capability(principal_id: UUID, tenant_id: UUID, capability: str) -> None:
```
dexters1, line 10: Can we also expand the endpoints with the ability to batch add/remove capabilities? For simpler Frontend use

cognee/modules/users/permissions/methods/has_user_management_permission.py
dexters1, line 16: This function was not fully implemented across Cognee initially, I did a short scan and found this:

The finding: three endpoints are tenant-OWNER-only, not manager-level
- POST /roles — create_role: owner_id != tenant.owner_id → PermissionDenied
- POST /users/{id}/roles — add_user_to_role: tenant.owner_id != owner_id
- POST /users/{id}/tenants — add_user_to_tenant: tenant.owner_id != owner_id

We should gate these endpoints/methods behind this check as well instead of the owner level

cognee/modules/users/permissions/methods/grant_capability.py
dexters1, line 10: We also have to gate who has access to grant_capabilities to other users in the tenant.

Only tenant_owner + admin + users with GIVE_CAPABILITY permission type

users should be able to add capabilities

cognee/modules/users/permissions/methods/revoke_capability.py
```diff
@@ -0,0 +1,32 @@
...
+        existence checks either — deleting a row that is not there and deleting
+        on behalf of a principal that is not there are the same no-op.
+    Args:
+        principal_id: Id of the principal (user, role or tenant).
+        tenant_id: Id of the tenant the capability is scoped to.
+        capability: Name from the CAPABILITY_TYPES catalog.
+    """
+    db_engine = get_relational_engine()
+
+    async with db_engine.get_async_session() as session:
```
dexters1, line 24: We need to gate who can revoke capabilities:

Tenant owner + admin + REVOKE_CAPABILITIES access

for example

cognee/modules/users/models/PrincipalCapability.py
```diff
@@ -0,0 +1,35 @@
...
+
+    created_at = Column(DateTime(timezone=True), default=lambda: datetime.now(timezone.utc))
+
+    principal_id = Column(UUID, ForeignKey("principals.id", ondelete="CASCADE"), primary_key=True)
+
+    tenant_id = Column(UUID, ForeignKey("tenants.id", ondelete="CASCADE"), primary_key=True)
+
+    # A name from the CAPABILITY_TYPES catalog, not a foreign key: the catalog
+    # is code, because the code is what gives each name meaning.
+    capability = Column(String, primary_key=True)
```
dexters1, line 35: Let's also add a granted_by column so we it is possible to track how permission is given

cognee/modules/users/permissions/methods/has_user_management_permission.py
dexters1, line 16: Maybe we should opt for a more reusable function now? For all grants?

Like has_grant_permission(requester_id, tenant_id, grant_type)

So we can have a reusable function for different grant types

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -576,4 +576,190 @@ async def get_my_tenants(
...
+
+        ## Error Codes
+        - **403 Forbidden**: Caller is not a member of the tenant, or it does not exist
+        """
+        from cognee.modules.users.permissions.methods import (
+            get_effective_capabilities,
+            require_tenant_membership,
+        )
+
+        await require_tenant_membership(user.id, tenant_id)
```
dexters1, line 613: Lets move these permission checks (like require_tenant_membership) to be inside the method so SDK users have fully usable methods with permissions acounted for.

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -576,4 +576,190 @@ async def get_my_tenants(
...
+
+        capabilities = await get_effective_capabilities(user.id, tenant_id)
+
+        return JSONResponse(
+            status_code=200,
+            content={"capabilities": sorted(capabilities)},
+        )
+
+    def _validate_capability(capability: str) -> None:
```
dexters1, line 622: Routers should not contain any code/business logic, this should be moved to its own module

we can create a capabilities module in modules/users to have all the utils there

cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -576,4 +576,190 @@ async def get_my_tenants(
...
+          target tenant, or the principal does not exist
+        """
+        from cognee.modules.users.permissions.methods import (
+            has_user_management_permission,
+            revoke_capability,
+        )
+
+        _validate_capability(capability)
+        scope = await _capability_scope(principal_id, tenant_id)
+        await has_user_management_permission(requester_id=user.id, tenant_id=scope)
```
dexters1, line 760: There should be a different permission check for giving out and revoking capatibilities 

Something like MANAGE_CAPABILITIES

dexters1 (changes requested):
Marking as requesting changes until we align on the comments I made so it does not get merged by accident

cognee/modules/users/models/PrincipalCapability.py
goran-radonic: It's created by the migration, so SDK-only setups get it too. Tried a fresh setup() and an older db going through run_migrations(), the table is there in both

cognee/modules/users/permissions/methods/grant_capability.py
goran-radonic: Done, both endpoints take the capability param more than once now, e.g. ?capability=manage_users&capability=revoke_capabilities. It's all or nothing, one bad name and nothing gets written.

cognee/modules/users/permissions/methods/has_user_management_permission.py
goran-radonic: Done, all three check manage_users now instead of the owner
one extra thing on add_user_to_role: you can only hand out a role if you hold what it carries, otherwise manage_users could put you into a role with grant_capabilities

cognee/modules/users/permissions/methods/grant_capability.py
goran-radonic: Done, it is owner, admin role, or the grant_capabilities capability (went with that name instead of GIVE_CAPABILITY, and it's a capability rather than a permission type since those are per dataset)

also you can only grant what you hold yourself, so grant_capabilities alone can't be used to hand out manage_users

cognee/modules/users/capabilities/methods/revoke_capability.py
```diff
@@ -0,0 +1,32 @@
...
+        existence checks either — deleting a row that is not there and deleting
+        on behalf of a principal that is not there are the same no-op.
+    Args:
+        principal_id: Id of the principal (user, role or tenant).
+        tenant_id: Id of the tenant the capability is scoped to.
+        capability: Name from the CAPABILITY_TYPES catalog.
+    """
+    db_engine = get_relational_engine()
+
+    async with db_engine.get_async_session() as session:
```
goran-radonic: Done, revoking needs owner, admin role, or revoke_capabilities. The owner can't lose theirs, so nobody can lock them out

cognee/modules/users/models/PrincipalCapability.py
goran-radonic: Added in its own migration. If the same capability gets granted again, the first granted_by stays

cognee/modules/users/permissions/methods/has_user_management_permission.py
goran-radonic: Done, added has_grant_permission(requester_id, tenant_id, grant_type) and has_user_management_permission just calls it with manage_users now

cognee/api/v1/permissions/routers/get_permissions_router.py
goran-radonic: Done, the checks live in authorized_grant_capability, authorized_revoke_capability and authorized_get_effective_capabilities now, so calling them from the SDK gets the same checks as the endpoints

cognee/api/v1/permissions/routers/get_permissions_router.py
goran-radonic: Done, moved it all to modules/users/capabilities, the router just calls into it now

cognee/api/v1/permissions/routers/get_permissions_router.py
goran-radonic: Split it into two instead of one MANAGE_CAPABILITIES: grant_capabilities for giving out and revoke_capabilities for taking away, like in your other two comments

Can merge them into one if you'd rather have that

PR #3264, opened 2026-06-21
fix(api): return 404 (not 500) for missing dataset in data endpoints

abhay-codes07:
## Description

`GET /v1/datasets/{dataset_id}/data` and `GET /v1/datasets/{dataset_id}/data/{data_id}/raw` authorize the request like this:

```python
dataset = await get_authorized_existing_datasets([dataset_id], "read", user)
if dataset is None:
    return JSONResponse(status_code=404, ...)
dataset_id = dataset[0].id   # or dataset[0].id below
```

`get_authorized_existing_datasets` always returns a **list** — `[]` when the dataset doesn't exist or the user lacks permission — and never `None`. So `if dataset is None` is dead code, and `dataset[0].id` raises `IndexError: list index out of range`, surfacing as an **HTTP 500** instead of a clean **404** for a missing/unauthorized dataset.

The `.../data` endpoint had a second defect in that same dead branch: `ErrorResponseDTO(f"...")` passes a positional argument, which Pydantic v2 rejects, and the model instance was handed straight to `JSONResponse(content=...)` without `model_dump()` — so even if the branch were reached it could not produce a valid response.

The fix guards with `if not dataset` and returns `ErrorResponseDTO(message=...).model_dump()`.

## Acceptance Criteria

- Requesting a non-existent or unauthorized dataset on either endpoint returns **404** with a "not found" message, not a 500.

Regression tests (both 500 on `dev`, 404 with the fix):

```
--- BEFORE: dev code (bug present) ---
FAILED ...test_datasets_missing_dataset_404.py::test_get_dataset_data_missing_dataset_returns_404
  - IndexError: list index out of range  (get_datasets_router.py:533)
FAILED ...test_datasets_missing_dataset_404.py::test_get_raw_data_missing_dataset_returns_404
2 failed in 1.01s

--- AFTER: fix applied ---
2 passed in 0.11s
```

## Type of Change
- [x] Bug fix (non-breaking change that fixes an issue)

## Screenshots
image

## Pre-submission Checklist
- [x] **I have tested my changes thoroughly before submitting this PR**
- [x] **This PR contains minimal changes necessary to address the issue/feature**
- [x] My code follows the project's coding standards and style guidelines
- [x] I have added tests that prove my fix is effective
- [x] All new and existing tests pass
- [x] I have searched existing PRs to ensure this change hasn't been submitted already
- [x] My commits have clear and descriptive messages

## DCO Affirmation
[...]

PR #4307, opened 2026-08-01
fix(datasets): guard against empty list from get_authorized_existing_datasets

### LINKED ISSUE #4304: [Bug] GET /datasets/{id}/data and /data/{id}/raw return 500 (IndexError) instead of 404 when dataset is not found
## Bug description

`GET /api/v1/datasets/{dataset_id}/data` and `GET /api/v1/datasets/{dataset_id}/data/{data_id}/raw` return HTTP 500 instead of 404 when the dataset does not exist or the authenticated user lacks read permission.

## Steps to reproduce

1. Start cognee with any backend
2. Send `GET /api/v1/datasets/<nonexistent-uuid>/data` with a valid auth token
3. Observe HTTP 500 Internal Server Error

Expected: HTTP 404 Not Found

The same applies to `GET /api/v1/datasets/<nonexistent-uuid>/data/<any-data-id>/raw`.

## Root cause

Both endpoints call `get_authorized_existing_datasets()`, which always returns `list[Dataset]` (annotated `-> list[Dataset]`, never `None`). The guard at both call sites reads:

```python
dataset = await get_authorized_existing_datasets([dataset_id], "read", user)

if dataset is None:       # always False — a list is never None
    return JSONResponse(status_code=404, ...)

dataset_id = dataset[0].id   # IndexError: list index out of range when dataset == []
```

The `if dataset is None` check is dead code. When the dataset is missing or the user lacks permission, `get_authorized_existing_datasets` returns `[]`, the guard is skipped, and `dataset[0]` raises `IndexError` which surfaces as an unhandled HTTP 500.

There is also a secondary issue on that dead-code path at the `/data` endpoint: `ErrorResponseDTO(f"Dataset ({str(dataset_id)}) not found.")` passes a positional argument to a Pydantic `BaseModel`, which requires keyword arguments — it should be `ErrorResponseDTO(message=f"Dataset ({str(dataset_id)}) not found.")`. This would raise a `TypeError` if the None check were ever reached.

The correct pattern is already used in the same file for the `/schema` endpoints:

```python
# line 602 and 628 — correct:
if not dataset:
    return JSONResponse(status_code=404, ...)
```

## Affected lines (current `dev` branch)

- `GET /{dataset_id}/data`: `if dataset is None` guard (and the `ErrorResponseDTO` positional-arg call on the dead path)
- `GET /{dataset_id}/data/{data_id}/raw`: `if dataset is None` guard

## Fix

```python
# Both affected sites — change:
if dataset is None:
# to:
if not dataset:

# /data endpoint dead-code path — change:
content=ErrorResponseDTO(f"Dataset ({str(dataset_id)}) not found."),
# to:
content=ErrorResponseDTO(message=f"Dataset ({str(dataset_id)}) not found."),
```

I'll open a PR with the fix and regression tests for both endpoints.

Romil2112:
`get_authorized_existing_datasets()` always returns `list[Dataset]`, never `None`, so the two `if dataset is None` checks in `GET /{dataset_id}/data` and `GET /{dataset_id}/data/{data_id}/raw` were dead code. When the dataset is absent or the user lacks permission the function returns `[]`, which then falls through to `dataset[0]` and raises `IndexError` → HTTP 500 instead of the intended 404.

**Root cause:** the return type is `list[Dataset]` (confirmed in `get_authorized_existing_datasets.py`), never `Optional`. The correct guard is `if not dataset`, already used in the proposals and skills routers.

**Changes:**
- Replace both `if dataset is None` guards with `if not dataset`
- Fix `ErrorResponseDTO(f"...")` positional-arg call → `ErrorResponseDTO(message=f"...").model_dump()` (Pydantic v2 requires keyword args on BaseModel)

Fixes #4304

### DIFF
--- a/cognee/api/v1/datasets/routers/get_datasets_router.py
+++ b/cognee/api/v1/datasets/routers/get_datasets_router.py
@@ -380,10 +380,10 @@ def get_datasets_router() -> APIRouter:
         # Verify user has permission to read dataset
         dataset = await get_authorized_existing_datasets([dataset_id], "read", user)
 
-        if dataset is None:
+        if not dataset:
             return JSONResponse(
                 status_code=404,
-                content=ErrorResponseDTO(f"Dataset ({str(dataset_id)}) not found."),
+                content=ErrorResponseDTO(message=f"Dataset ({str(dataset_id)}) not found.").model_dump(),
             )
 
         dataset_id = dataset[0].id
@@ -691,7 +691,7 @@ def get_datasets_router() -> APIRouter:
         # Verify user has permission to read dataset
         dataset = await get_authorized_existing_datasets([dataset_id], "read", user)
 
-        if dataset is None:
+        if not dataset:
             return JSONResponse(
                 status_code=404, content={"detail": f"Dataset ({dataset_id}) not found."}
             )


ArmagedonFlamer:
Thanks for this @Romil2112, and sorry - #3264 landed the same fix while this was open. It had been in the queue since June and we didn't connect the two, which is on us. Your analysis was correct and the fix was right; closing only because it's now redundant. We welcome your next contribution.

PR #4308, opened 2026-08-01
fix(api): guard unguarded dataset[0] access in export and visualize endpoints

### LINKED ISSUE #4306: GET /activity/export/{id} returns 500, GET /visualize/{id} returns 409 with leaked internal error — same root cause as #4304
## Bug description

Two endpoints mishandle the case where a dataset does not exist or the caller lacks read permission. Both stem from the same root cause as #4304: `get_specific_user_permission_datasets` and `get_authorized_existing_datasets` always return `list[Dataset]`, never `None`, so an unguarded `result[0]` access on an empty list raises `IndexError`.

| Endpoint | Expected | Actual |
|----------|----------|--------|
| `GET /api/v1/activity/export/{dataset_id}` | 403 | **500** — unhandled `IndexError` |
| `GET /api/v1/users/{user_id}/graph` (visualize) | 403 | **409** — `{"error": "list index out of range"}` |

---

## Finding 1 — `GET /activity/export/{dataset_id}` → HTTP 500

`get_activity_router.py`:

```python
dataset_ids = await get_specific_user_permission_datasets(user.id, "read", [dataset_id])
dataset_id = dataset_ids[0].id   # IndexError when dataset_ids == []
```

`get_specific_user_permission_datasets` is typed `-> list[Dataset]`. There is no `if not dataset_ids:` guard and no surrounding `try/except`, so the `IndexError` propagates as an unhandled HTTP 500.

## Finding 2 — `GET /visualize/{dataset_id}` → HTTP 409 + leaked internal detail

`get_visualize_router.py`:

```python
try:
    dataset = await get_authorized_existing_datasets([dataset_id], "read", user)
    html_visualization = await visualize_graph(dataset=dataset[0].id, ...)  # IndexError
    ...
except Exception as error:
    return JSONResponse(status_code=409, content={"error": str(error)})
```

The broad `except Exception` catches the `IndexError`, but the response is **409 Conflict** with `{"error": "list index out of range"}`. This has two problems:

1. **Wrong status code** — 409 Conflict is semantically incorrect for "dataset not found / access denied"; the correct code is 403 or 404.
2. **Information disclosure** — `str(error)` exposes the raw Python exception message (`"list index out of range"`) to the caller, leaking internal implementation details. This is the same secondary issue noted in #4304 (where the Pydantic positional-arg message would leak if that dead-code branch were reachable).

---

## Isolated reproduction (no server required)

```python
import asyncio

async def get_specific_user_permission_datasets(user_id, permission_type, dataset_ids=None):
    return []   # what the real function returns when access is denied

async def get_authorized_existing_datasets(dataset_ids, permission, user):
    return []

# Finding 1
async def export_bug(dataset_id, user=None):
    dataset_ids = await get_specific_user_permission_datasets(user, "read", [dataset_id])
    print(f"returned: {dataset_ids!r}")
    dataset_id = dataset_ids[0].id   # crash

try:
    asyncio.run(export_bug("00000000-0000-0000-0000-000000000000"))
except IndexError as e:
    print(f"IndexError: {e}  →  HTTP 500\n")

# Finding 2
async def visualize_bug(dataset_id, user=None):
    try:
        dataset = await get_authorized_existing_datasets([dataset_id], "read", user)
        print(f"returned: {dataset!r}")
        _ = dataset[0].id
    except Exception as error:
        print(f"Caught: {error!r}")
        print(f"→ HTTP 409  body={{'error': '{error}'}}")
        print("  (wrong status code + raw Python exception exposed to caller)")

asyncio.run(visualize_bug("00000000-0000-0000-0000-000000000000"))
```

**Output:**

```
returned: []
IndexError: list index out of range  →  HTTP 500

returned: []
Caught: IndexError('list index out of range')
→ HTTP 409  body={'error': 'list index out of range'}
  (wrong status code + raw Python exception exposed to caller)
```

## Minimal server repro

```python
import requests

BASE = "http://localhost:8000"
token = requests.post(
    f"{BASE}/api/v1/auth/token",
    data={"username": "default_user@example.com", "password": "default_user_password"},
).json()["access_token"]
headers = {"Authorization": f"Bearer {token}"}

fake_id = "00000000-0000-0000-0000-000000000000"

r1 = requests.get(f"{BASE}/api/v1/activity/export/{fake_id}", headers=headers)
print(r1.status_code, r1.text)   # 500

r2 = requests.get(f"{BASE}/api/v1/users/{fake_id}/graph", headers=headers)
print(r2.status_code, r2.text)   # 409, {"error": "list index out of range"}
```

---

## Fix

The correct pattern is already established in `get_proposals_router.py` and `get_skills_router.py`:

```python
# Both routers use this helper:
async def _authorized_dataset(dataset_id, user, permission):
    datasets = await get_authorized_existing_datasets([dataset_id], permission, user)
    if not datasets:
        raise PermissionDeniedError(message="Not authorized for this dataset")
    return datasets[0]

# Caught downstream as:
except PermissionDeniedError:
    return JSONResponse(status_code=403, content={"error": "Not authorized for this dataset"})
```

Applying the same pattern to both affected endpoints:

```python
# Finding 1 — export endpoint:
dataset_ids = await get_specific_user_permission_datasets(user.id, "read", [dataset_id])
if not dataset_ids:
    return Response(content="Dataset not found or access denied", status_code=403)
dataset_id = dataset_ids[0].id

# Finding 2 — visualize endpoint:
dataset = await get_authorized_existing_datasets([dataset_id], "read", user)
if not dataset:
    return JSONResponse(status_code=403, content={"error": "Not authorized for this dataset"})
# Then catch remaining exceptions with a generic message (not str(error)):
except Exception:
    return JSONResponse(status_code=500, content={"error": "Visualization failed"})
```

I'll open a PR with fixes and regression tests for both endpoints.

Romil2112:
Three endpoints index into the result of `get_authorized_existing_datasets` / `get_specific_user_permission_datasets` without first checking whether the list is empty. Both functions return `list[Dataset]`, never `None`, so when the dataset is absent or the user lacks permission they return `[]` — and `[][0]` raises `IndexError` → HTTP 500 or an unintended exception type.

**`GET /activity/export/{dataset_id}`** (`get_activity_router.py`)
- `dataset_ids[0].id` with no guard → `IndexError` → 500
- Fix: early `return Response(status_code=403)` when list is empty

**`GET /visualize`** (`get_visualize_router.py`)
- `dataset[0].id` inside a bare `except Exception` → 409 with `str(error)` leaking `"list index out of range"` to the caller
- Fix: move auth check outside the try/except so missing-dataset returns 403 before visualization starts; change the catch-all error message to a static string to prevent info disclosure

**`POST /visualize/multi`** (`get_visualize_router.py`)
- `datasets[0]` per-pair with no guard → same `IndexError`
- Fix: guard each pair and return 409 with a clear message

Fixes #4306

### DIFF
--- a/cognee/api/v1/activity/routers/get_activity_router.py
+++ b/cognee/api/v1/activity/routers/get_activity_router.py
@@ -279,6 +279,8 @@ def get_activity_router() -> APIRouter:
         from datetime import datetime, timezone
 
         dataset_ids = await get_specific_user_permission_datasets(user.id, "read", [dataset_id])
+        if not dataset_ids:
+            return Response(content="Dataset not found or access denied", status_code=403)
         dataset_id = dataset_ids[0].id
 
         db_engine = get_relational_engine()
--- a/cognee/api/v1/users/routers/get_visualize_router.py
+++ b/cognee/api/v1/users/routers/get_visualize_router.py
@@ -123,10 +123,15 @@ def get_visualize_router() -> APIRouter:
 
         from cognee.api.v1.visualize import visualize_graph
 
-        try:
-            # Verify user has permission to read dataset
-            dataset = await get_authorized_existing_datasets([dataset_id], "read", user)
+        # Verify user has permission to read dataset
+        dataset = await get_authorized_existing_datasets([dataset_id], "read", user)
+        if not dataset:
+            return JSONResponse(
+                status_code=403,
+                content={"error": "Dataset not found or access denied"},
+            )
 
+        try:
             html_visualization = await visualize_graph(
                 dataset=dataset[0].id,
                 user=user,
@@ -141,7 +146,7 @@ def get_visualize_router() -> APIRouter:
 
         except Exception as error:
             logger.exception("Visualization failed for dataset %s", dataset_id)
-            return JSONResponse(status_code=409, content={"error": str(error)})
+            return JSONResponse(status_code=409, content={"error": "Visualization failed"})
 
     @router.post("/multi", response_model=None)
     async def visualize_multi(
@@ -197,6 +202,11 @@ def get_visualize_router() -> APIRouter:
                 datasets = await get_authorized_existing_datasets(
                     [pair.dataset_id], "read", target_user
                 )
+                if not datasets:
+                    return JSONResponse(
+                        status_code=409,
+                        content={"error": f"Dataset {pair.dataset_id} not found or access denied"},
+                    )
                 user_dataset_pairs.append((target_user, datasets[0]))
 
             html_visualization = await visualize_multi_user_graph(user_dataset_pairs)


Romil2112:
Hi @NMZivkovic — since #4306 was consolidated into #4304, wanted to flag that this PR targets different files and endpoints than #4307.

**What #4307 fixes** (scope of #4304):
- `cognee/api/v1/datasets/routers/get_datasets_router.py`
  - `GET /datasets/{id}/data`
  - `GET /datasets/{id}/data/{node_id}/raw`

**What this PR fixes** (three separate routes in two different files):

1. `cognee/api/v1/activity/routers/get_activity_router.py` — `export_dataset_markdown`
   - `GET /activity/export/{dataset_id}`
   - `get_specific_user_permission_datasets` returns a list, never `None`. Without the guard, `dataset_ids[0].id` throws `IndexError` → 500 on any unrecognized or unauthorized dataset ID.

2. `cognee/api/v1/users/routers/get_visualize_router.py` — `visualize`
   - `GET /visualize/{dataset_id}`
   - The auth check was inside the `try` block, so when `get_authorized_existing_datasets` returned an empty list the bare `dataset[0]` throw was caught by the `except Exception` handler and re-surfaced as a 409 with `{"error": str(error)}` — leaking the internal `IndexError: list index out of range` message verbatim. This PR moves the auth check before the `try`, returns 403 on an empty result, and replaces the catch-all error string with a generic message.

3. `cognee/api/v1/users/routers/get_visualize_router.py` — `visualize_multi`
   - `POST /visualize/multi`
   - Same unguarded `[0]` access inside the per-pair loop; any pair referencing a missing or unauthorized dataset throws `IndexError` → 500 mid-loop.

The root cause is the same pattern as #4304, but none of these three routes are touched by #4307 — they need the guard applied separately. If #4307 merges without this PR, `/activity/export`, `/visualize`, and `/visualize/multi` will still 500 (and leak internal error strings) on a missing or unauthorized dataset.

Happy to rebase onto the latest `dev` or adjust the response codes if the team prefers a different shape.

dexters1:
Thanks for the PR @Romil2112! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #4321, opened 2026-08-03
fix: Budget-overshoot mitigation — fail-fast + concurrency cap (CLO-416/418/420)

LStromann:
Part of **CLO-409** (prod incident: a cognify/`remember` run spent $503 of real inference against a $345 LiteLLM cap, then a ~1.6M-request retry storm). This is the OSS-engine half of the mitigation. Review commit-by-commit.

## Commits (one per sub-ticket)
- **CLO-416 — Fail fast on budget exhaustion.** LiteLLM's `"Budget has been exceeded"` 429 was classified transient, so the shared retry stack hit a spend-capped key ~15× per logical call. Detect budget/payment exhaustion structurally through instructor's `InstructorRetryException` wrapper (`failed_attempts` + `__cause__`/`RetryError` chain) and treat it as terminal by default. One opt-in boolean `llm_budget_retry_enabled` (default off) makes budget errors retryable for a future PAYG path; a bounded attempt/time budget is not implemented yet (see review fixes commit). `raise_if_quota_error` preserves the inner provider wording so the pod's `insufficient_credits` status classification still matches.
- **CLO-418 — Cap concurrent LLM calls at the gateway.** One process-global (per-event-loop) semaphore at `LLMGateway.acreate_structured_output` — the single choke point — bounds the whole engine's concurrent fan-out. `LLM_MAX_CONCURRENT_REQUESTS`, default 0 = unbounded (OSS behavior preserved).
- **CLO-420 — Stop instructor retrying budget errors (prod adapters).** Pass instructor an `AsyncRetrying` using the shared `llm_retry_condition` so it re-asks validation errors but stops immediately on terminal budget/quota/auth. Verified instructor still wraps into `InstructorRetryException`, so adapter fallback/content-policy handling is unchanged. Scoped to `generic_llm_api` (the pod's `custom` provider) + `openai`; other OSS-only adapters stay on int `max_retries`, covered by the outer fail-fast.

## Testing
`test_structured_output_retry.py` + `test_llm_concurrency_limit.py`: budget-429 (structured / instructor-wrapped / string / 402) classified terminal; transient still retries; PAYG opt-in re-enables retry; gateway semaphore caps peak concurrency; instructor factory → 1 attempt on budget, cap on transient. (Local `uv run pytest` can't build `ladybug`; ran via `python3 -m pytest`. The 4 `@pytest.mark.asyncio` failures are a local `pytest-asyncio` gap and reproduce on `dev`.)

## Deploy order ⚠️
This must be live in the pod **before** cloud-backend's key-limits PR (CLO-417), or Phase 0's new admission-429s feed the retry storm. Merge + release this, then `mise run update-cognee <version>` to bump the pod, then merge cloud-backend#(CLO-417).


cognee/infrastructure/llm/config.py
```diff
@@ -133,6 +133,19 @@ class LLMConfig(BaseSettings):
...
+    # preserves OSS default). Bounds the cognify fan-out burst so it cannot overshoot a
+    # LiteLLM key's async budget cap (CLO-409 Phase 0b).
+    llm_max_concurrent_requests: int = 0
+
+    # Budget/payment exhaustion (HTTP 402, or LiteLLM's "Budget has been exceeded" 429)
+    # is terminal by default: retrying a spend-capped key only produces more
+    # proxy-rejected calls (CLO-409). Set > 0 to allow a small bounded retry for
+    # pay-as-you-go tenants where an auto-recharge top-up may lift the cap mid-run.
+    llm_budget_max_retry_attempts: int = 0
+    llm_budget_max_retry_seconds: int = 0
```
LStromann: **Dead / misleading config knobs (medium).** `_budget_retry_enabled()` reads only `llm_budget_max_retry_attempts > 0`; `llm_budget_max_retry_seconds` is never referenced anywhere. And when budget retry *is* enabled, the actual attempt/time budget is governed by the outer adapter's `llm_retry_stop_condition` (`stop_after_attempt(2) & stop_after_delay(240)`), **not** by these values. So `attempts=3, seconds=30` does not yield "3 attempts / 30s" — it just flips a switch and inherits the generic 2-attempt/240s floor.

The docstring and PR body read as if the values are honored. Either wire them into a dedicated budget stop-condition, or collapse to a single `llm_budget_retry_enabled: bool` (and drop `_seconds`) until the bounded policy is actually implemented.

cognee/infrastructure/llm/config.py
```diff
@@ -133,6 +133,19 @@ class LLMConfig(BaseSettings):
     # (with the llm_rate_limit_requests/interval budget) for the rest of the process.
     auto_rate_limit: bool = True
 
+    # Global ceiling on concurrent in-flight structured-output LLM calls across the
+    # process, enforced at the single LLMGateway choke point. 0 = disabled (unbounded,
+    # preserves OSS default). Bounds the cognify fan-out burst so it cannot overshoot a
+    # LiteLLM key's async budget cap (CLO-409 Phase 0b).
+    llm_max_concurrent_requests: int = 0
```
LStromann, line 146: **Tuning caveat worth a comment here (medium/low).** The semaphore in `LLMGateway._limit_concurrency` wraps the adapter's `@retry`, so a slot is held for the *entire* retry lifecycle — up to `LLM_MIN_RETRY_SECONDS = 240s` of backoff on transient errors, not just the in-flight request. With a low cap, a few transiently-failing calls can starve the whole engine for minutes while they sit in backoff holding slots. Defensible (retries also count against the budget you're bounding), but operators should pick the cap accounting for backoff-held slots — please note this here and in the deploy runbook.

cognee/infrastructure/llm/LLMGateway.py
```diff
@@ -64,6 +65,38 @@ async def _fail_fast_on_quota(coro: Coroutine) -> T:
         raise
 
 
+# Per-event-loop concurrency semaphores, keyed by loop id so a process that
+# (re)creates event loops (e.g. tests using asyncio.run) never reuses a semaphore
+# bound to a closed loop. Recreated when the configured limit changes.
+_concurrency_semaphores: dict[int, tuple[int, asyncio.Semaphore]] = {}
+
+
+def _concurrency_semaphore(limit: int) -> asyncio.Semaphore:
```
LStromann, line 80: **`id(loop)` reuse hazard (low, latent).** CPython reuses object addresses, so a new loop can get the same `id` as a closed-and-GC'd loop. If the limit also matches, this returns the *old* `asyncio.Semaphore` whose `_loop` is bound to the dead loop → `RuntimeError: ... bound to a different event loop` on first await. The comment claims keying by loop id makes reuse safe; it only makes it rare. Negligible in prod (one long-lived loop) but a real hazard for the `asyncio.run`-per-call test pattern the comment cites. Consider a `weakref` to the loop as the key, or catching `RuntimeError` and rebuilding.

cognee/infrastructure/llm/retry_config.py
```diff
@@ -67,6 +69,62 @@ def is_quota_or_billing_error(error: BaseException) -> bool:
...
+
+    def _walk(exc: BaseException | None) -> BaseException | None:
+        while exc is not None and id(exc) not in seen:
+            seen.add(id(exc))
+            if is_budget_exhausted_error(exc):
+                return exc
+            # tenacity RetryError wraps the final failed attempt.
+            last_attempt = getattr(exc, "last_attempt", None)
+            if last_attempt is not None:
+                found = _walk(getattr(last_attempt, "_exception", None))
```
LStromann: **Minor:** `last_attempt._exception` reads a private tenacity attribute; `last_attempt.exception()` is the public accessor and won't break if the internal name changes (it won't raise here since the attempt is known-failed).

cognee/infrastructure/llm/retry_config.py
```diff
@@ -77,15 +135,44 @@ def should_retry_llm_exception(error: BaseException) -> bool:
     )
     if isinstance(error, non_retryable):
         return False
+    # Budget/payment exhaustion is terminal: retrying a spend-capped key only produces
+    # more proxy-rejected 429s (the CLO-409 retry storm). Opt-in bounded retry only.
+    if _is_budget_error(error):
+        return _budget_retry_enabled()
```
LStromann, line 151: **Classification asymmetry worth a one-line comment (low).** A *string-only* budget error (e.g. `RuntimeError("Budget has been exceeded!")`) is caught by `is_quota_or_billing_error` via the new pattern, **not** by `_find_budget_error`. Two consequences: (a) it becomes `LLMQuotaExceededError`, not `LLMPaymentRequiredError` (402); (b) it stays terminal *even when PAYG retry is enabled*, since only the structural `_is_budget_error` path consults `_budget_retry_enabled()`. Both are arguably fine (string match is a last-resort heuristic), but a comment would keep the divergence from looking like a bug later.

LStromann (commented):
## Budget-overshoot mitigation review (CLO-416/418/420)

Overall well-targeted and carefully commented. The core fail-fast path is correct: budget error → instructor stops at 1 attempt → `InstructorRetryException` → outer `@retry` sees budget via `failed_attempts` walk → no retry → gateway `raise_if_quota_error` raises `LLMPaymentRequiredError` **preserving the provider wording**, so the pod's `insufficient_credits` classification still matches. Passing `AsyncRetrying` (async client) and *not* setting `reraise=True` are both correct and well-justified in the docstring. Default-off knobs (`0`) preserve OSS behavior.

**Please address before merging into the prod pod:**
1. `llm_budget_max_retry_seconds` is dead config and `llm_budget_max_retry_attempts` is used only as a boolean — the actual attempt/time budget is the generic outer stop condition, not these values. Shipping named-but-inert knobs into the PAYG path is a footgun (inline comment below).
2. Document the semaphore-held-across-backoff behavior given the ⚠️ deploy-order sensitivity (inline comment below).

Non-blocking: `id(loop)` cache reuse hazard, `last_attempt._exception` private access, and the string-only vs structural budget-error classification asymmetry (inline comments below).

LStromann:
Thanks for the review — addressed all five in `cd4142bbb` (rebased onto the latest `dev` merge):

**Blocking**
1. **Dead/misleading knobs** → collapsed `llm_budget_max_retry_attempts`/`_seconds` to a single `llm_budget_retry_enabled: bool`. The comment now states plainly that enabling it just makes budget errors retryable under the generic outer stop condition — no dedicated attempt/time budget (that's future PAYG work). PR body corrected too.
2. **Semaphore held across backoff** → documented on `llm_max_concurrent_requests` that a slot is held for the whole retry lifecycle (up to 240s), so a low cap + transient failures can starve the engine; note to size accordingly (+ runbook).

**Non-blocking**
3. **`id(loop)` reuse hazard** → replaced the `id(loop)` cache with a `WeakKeyDictionary` keyed on the loop object; entries drop when the loop is GC'd, so no stale-loop reuse.
4. **Private attr** → `last_attempt.exception()` instead of `._exception`.
5. **Classification asymmetry** → added a comment explaining that a string-only budget error stays unconditionally terminal / becomes `LLMQuotaExceededError` and never honors the PAYG opt-in (only the structural detector does).

Tests updated for the boolean; sync suite green (the 4 `@pytest.mark.asyncio` failures are a local `pytest-asyncio` gap, green in CI).

LStromann:
Closing per Igor's call in [Slack](https://topoteretes.slack.com/archives/D09FC78D3NY/p1785853818578319): no SDK changes for the CLO-409 budget work — this is handled cloud-side by `cost_guard` (key limits + in-flight reservation ledger) in cloud-backend#283.

Note for whoever picks up CLO-409: closing this leaves the retry amplification on `dev`. A LiteLLM budget-429 wrapped in `InstructorRetryException` isn't detected (`is_budget_exhausted_error` walks only `__cause__`, not `failed_attempts`/`RetryError.last_attempt`), so it's classified transient and retried, and instructor's `max_retries` is still a bare int. Not reachable from the cloud side. Reopen if the retry storm recurs.

Also: cloud-backend#283 wires `LLM_MAX_CONCURRENT_REQUESTS` through both tenant charts — that setting only exists here, so it should be dropped from that PR.

Branch `feat/clo-416-budget-fail-fast` retained.

PR #4324, opened 2026-08-03
feat: Keep cognified data + stop cleanly on budget exhaustion (CLO-421)

LStromann:
**CLO-421** (Fix 3 of the CLO-409 budget-overshoot work). Today a budget-exhausted cognify run marks the whole dataset failed **and rolls back everything already processed** — so paid-for, cognified work is destroyed. This makes budget exhaustion a *clean stop* instead.

## Behavior
- On `is_budget_exhausted_error` mid-run: **cancel in-flight siblings** (stop further billing), **skip the destructive rollback** (all already-cognified documents stay in the graph, queryable), and record a **COMPLETED** run — not errored.
- The run's `run_info` carries `stopped_reason="budget_exhausted"` + `{documents_total, documents_cognified, documents_remaining}`, so callers see how much is left.
- **Resume is automatic**: re-running cognify processes only the remaining documents (incremental loading skips those already marked complete). No re-bill of finished docs.
- Non-budget errors are **unchanged** (rollback + errored + raise).

## Deliberately NOT done
- **No new `PipelineRunStatus`** and **no DB migration** — `run_info` is already a JSON column, and per-document completion already exists as `DATA_ITEM_PROCESSING_COMPLETED`. The split between "errored" and "stopped on budget" is `PipelineRunCompleted` + `run_info.stopped_reason`, not a status enum.

## Also
- Surface the existing completion flag over HTTP as derived **`DataDTO.cognify_completed`** on `GET /v1/datasets/{id}/data`, so clients/UI can show which documents are in the graph vs pending (the Python SDK already exposed it via raw `Data.pipeline_status`).

## OSS impact
None on the happy path or non-budget errors. The graceful stop only triggers on `is_budget_exhausted_error` (HTTP 402 / `litellm.BudgetExceededError` / LiteLLM-proxy `budget_exceeded` 429), which a typical OSS user running against a provider directly won't hit.

## Dependency
Stacks on **#4321** — that PR makes budget errors surface as a detectable `LLMPaymentRequiredError` (402) rather than a wrapped `InstructorRetryException`. Without it the graceful stop simply won't trigger (no regression). Rebase once #4321 lands in `dev`.

## Testing
`test_run_tasks_budget_stop.py`: budget mid-run → COMPLETED (not errored), `stopped_reason` + remaining counts, rollback **not** called. Validated end-to-end locally via an `asyncio.run` harness (the `@pytest.mark.asyncio` test runs in CI; local `uv run pytest` can't build `ladybug`).

### DIFF
--- a/cognee/api/v1/datasets/routers/get_datasets_router.py
+++ b/cognee/api/v1/datasets/routers/get_datasets_router.py
@@ -397,18 +397,13 @@ def get_datasets_router() -> APIRouter:
         if dataset_data is None:
             return []
 
-        from cognee.modules.pipelines.models.DataItemStatus import DataItemStatus
-
-        def _cognify_completed(data) -> bool:
-            return (data.pipeline_status or {}).get("cognify_pipeline", {}).get(
-                str(dataset_id)
-            ) == DataItemStatus.DATA_ITEM_PROCESSING_COMPLETED
+        from cognee.modules.pipelines.utils import is_data_item_cognified
 
         return [
             dict(
                 **jsonable_encoder(data),
                 dataset_id=dataset_id,
-                cognify_completed=_cognify_completed(data),
+                cognify_completed=is_data_item_cognified(data, "cognify_pipeline", dataset_id),
             )
             for data in dataset_data
         ]
--- a/cognee/api/v1/datasets/routers/get_datasets_router.py
+++ b/cognee/api/v1/datasets/routers/get_datasets_router.py
@@ -52,6 +52,10 @@ class DataDTO(OutDTO):
     mime_type: str
     raw_data_location: str
     dataset_id: UUID
+    # Whether this document has been fully cognified into this dataset's graph.
+    # False means it isn't in the graph yet (e.g. a cognify run stopped early on
+    # budget) — re-running cognify processes the still-incomplete documents.
+    cognify_completed: bool = False
 
 
 class DatasetGraphSummaryDTO(OutDTO):
@@ -393,10 +397,18 @@ def get_datasets_router() -> APIRouter:
         if dataset_data is None:
             return []
 
+        from cognee.modules.pipelines.models.DataItemStatus import DataItemStatus
+
+        def _cognify_completed(data) -> bool:
+            return (data.pipeline_status or {}).get("cognify_pipeline", {}).get(
+                str(dataset_id)
+            ) == DataItemStatus.DATA_ITEM_PROCESSING_COMPLETED
+
         return [
             dict(
                 **jsonable_encoder(data),
                 dataset_id=dataset_id,
+                cognify_completed=_cognify_completed(data),
             )
             for data in dataset_data
         ]
--- a/cognee/modules/pipelines/operations/run_tasks.py
+++ b/cognee/modules/pipelines/operations/run_tasks.py
@@ -17,6 +17,9 @@ from cognee.modules.users.methods import get_default_user
 from cognee.modules.pipelines.utils import generate_pipeline_id
 from cognee.modules.pipelines.exceptions import PipelineRunFailedError
 from cognee.tasks.ingestion import resolve_data_directories
+from cognee.infrastructure.llm.exceptions import is_budget_exhausted_error
+from cognee.modules.data.methods import get_dataset_data
+from cognee.modules.pipelines.models.DataItemStatus import DataItemStatus
 from cognee.modules.pipelines.models import PipelineContext
 from cognee.modules.pipelines.models.PipelineRunInfo import (
     PipelineRunCompleted,
@@ -33,6 +36,28 @@ from ..tasks.task import Task
 
 
 logger = get_logger("run_tasks(tasks: [Task], data)")
+
+
+async def _cognify_progress(dataset_id: UUID, pipeline_name: str) -> dict:
+    """Per-document progress for a dataset, derived from the completion markers.
+
+    Reports how many of the dataset's documents are already cognified (carry the
+    ``DATA_ITEM_PROCESSING_COMPLETED`` flag for this pipeline) vs still remaining —
+    used to tell the caller what is left when a run stops early (e.g. budget).
+    """
+    items = await get_dataset_data(dataset_id)
+    total = len(items)
+    cognified = sum(
+        1
+        for d in items
+        if (d.pipeline_status or {}).get(pipeline_name, {}).get(str(dataset_id))
+        == DataItemStatus.DATA_ITEM_PROCESSING_COMPLETED
+    )
+    return {
+        "documents_total": total,
+        "documents_cognified": cognified,
+        "documents_remaining": total - cognified,
+    }
 
 
 def override_run_tasks(new_gen):
@@ -129,9 +154,10 @@ async def run_tasks(
                         data_cache,
                     )
 
-            gathered = await asyncio.gather(
-                *[asyncio.create_task(_run_item(item)) for item in data],
-            )
+            # Keep task references so siblings can be cancelled if the run stops
+            # early (see the budget-exhaustion path in the except below).
+            in_flight = [asyncio.create_task(_run_item(item)) for item in data]
+            gathered = await asyncio.gather(*in_flight)
 
             # Separate successes from unhandled exceptions
             results = []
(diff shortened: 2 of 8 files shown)


cognee/modules/pipelines/operations/run_tasks.py
```diff
@@ -185,6 +211,50 @@ async def _run_item(data_item):
...
+            if in_flight:
+                await asyncio.gather(*in_flight, return_exceptions=True)
+
+            # Budget/quota exhaustion is a resource limit, not a failure. Keep every
+            # document already cognified (skip rollback, so the graph stays queryable)
+            # and record a COMPLETED run that reports what is left — a later cognify
+            # resumes only the remaining documents (incremental loading skips the ones
+            # already marked complete). No new pipeline status is introduced: the
+            # "stopped early" signal lives in run_info.stopped_reason.
+            if is_budget_exhausted_error(error):
```
LStromann: **Detection is shallow where robustness matters most (high).**

`is_budget_exhausted_error` (exceptions.py) only checks the top-level exception's `status_code` / `BudgetExceededError` / 429-response — it does **not** walk `__cause__`. But this `error` is whatever propagates out of `asyncio.gather` after travelling up through `run_tasks_with_telemetry` → the task stack → `run_tasks_data_item`. #4321 needed a dedicated cause-walking detector (`_find_budget_error`, traversing `__cause__` / tenacity `RetryError` / instructor `failed_attempts`) *because* budget errors arrive wrapped in practice. Here the same condition is detected with the shallow helper, and the consequence of a miss is the worst case: `is_budget_exhausted_error` returns False → the code falls through to `rollback_handler` and `PipelineRunErrored` → **the cognified data this PR is trying to preserve gets rolled back**, silently.

Also note the `PipelineRunFailedError` masking path: if `run_tasks_data_item` returns an errored *result* instead of raising (see the `RAISE_INCREMENTAL_LOADING_ERRORS=false` branch, and the results-loop at lines 74–98), `run_tasks` raises `PipelineRunFailedError` — never budget-typed — so detection fails even though budget was the root cause.

Recommend detecting budget exhaustion with the same cause-walking robustness as retry_config (reuse `_find_budget_error`, or extend `is_budget_exhausted_error` to walk `__cause__`), and unwrap `PipelineRunFailedError` to its underlying cause before the check. Otherwise this feature's correctness depends on an unverified invariant about the exact exception shape at this point.

cognee/modules/pipelines/operations/run_tasks.py
```diff
@@ -35,6 +38,28 @@
 logger = get_logger("run_tasks(tasks: [Task], data)")
 
 
+async def _cognify_progress(dataset_id: UUID, pipeline_name: str) -> dict:
```
LStromann, line 40: **Progress counts + resume are only meaningful under incremental loading (medium).** `_cognify_progress` counts docs carrying `DATA_ITEM_PROCESSING_COMPLETED`, but only `run_tasks_data_item_incremental` writes that marker — `run_tasks_data_item_regular` never does. So in a non-incremental run: (a) `documents_cognified` is always 0 regardless of how much actually landed in the graph, and (b) the "resume processes only remaining docs / no re-bill" guarantee doesn't hold — a re-run reprocesses (and re-bills) everything, since there are no markers to skip on. The whole CLO-421 story assumes incremental mode. Please confirm cognify always runs with `incremental_loading=True` (and if so, consider asserting/short-circuiting the graceful-stop path when it isn't, rather than recording misleading counts).

cognee/tests/unit/modules/pipelines/test_run_tasks_budget_stop.py
```diff
@@ -0,0 +1,130 @@
...
+    async def _log_complete(*_args, **kwargs):
+        complete_calls.append(kwargs)
+
+    async def _log_error(*_args, **_kwargs):
+        return None
+
+    monkeypatch.setattr(run_tasks_module, "get_relational_engine", lambda: _FakeEngine(dataset))
+    monkeypatch.setattr(run_tasks_module, "generate_pipeline_id", lambda *_args: uuid4())
+    monkeypatch.setattr(run_tasks_module, "set_database_global_context_variables", _no_op_context)
+    monkeypatch.setattr(run_tasks_module, "run_tasks_data_item", _budget_item)
```
LStromann: **Test bypasses the path most likely to break (medium).** Monkeypatching `run_tasks_data_item` to raise a bare `LLMPaymentRequiredError` means the test never exercises `run_tasks_data_item` → `run_tasks_data_item_incremental` → `run_tasks_with_telemetry`, which is exactly where the budget error could get wrapped (or converted to a `PipelineRunErrored` result → `PipelineRunFailedError`). So this test would keep passing even if, in production, the error reaches the `except` in a shape `is_budget_exhausted_error` can't recognize and the destructive rollback fires. Add a test that drives the real `run_tasks_data_item` with a task that raises a budget error (ideally one wrapped in another exception, and one covering `RAISE_INCREMENTAL_LOADING_ERRORS` both true and false) and asserts the COMPLETED-not-rolled-back outcome. That's the assertion that actually protects the feature.

cognee/modules/pipelines/operations/run_tasks.py
```diff
@@ -185,6 +211,50 @@ async def _run_item(data_item):
             )
 
         except Exception as error:
+            # Stop any siblings still running so a failed or budget-limited run does
+            # not keep issuing (and, on a budget stop, billing) LLM calls after we
+            # have already decided to stop.
+            in_flight = locals().get("in_flight") or []
```
LStromann: **Minor: `locals().get("in_flight")` / `locals().get("results")` is a code smell.** It works around the variables possibly being unbound if an exception is raised before line 70 / before the results loop, but it's fragile and obscures intent. Cleaner to initialize `in_flight: list = []` and `results = None` just before the `try`, then reference them directly in the `except`. Same behavior, no `locals()` introspection.

cognee/api/v1/datasets/routers/get_datasets_router.py
```diff
@@ -393,10 +397,18 @@ async def get_dataset_data(
         if dataset_data is None:
             return []
 
+        from cognee.modules.pipelines.models.DataItemStatus import DataItemStatus
+
+        def _cognify_completed(data) -> bool:
```
LStromann: **Minor: duplicated completion-check + hardcoded pipeline name.** This `_cognify_completed` and `run_tasks._cognify_progress` implement the same `pipeline_status[pipeline][dataset_id] == DATA_ITEM_PROCESSING_COMPLETED` lookup, and this copy hardcodes `"cognify_pipeline"`. Consider a single shared helper (e.g. `is_data_item_cognified(data, pipeline_name, dataset_id)`) so the two can't drift and the pipeline name isn't string-literal'd in the router. Not blocking.

LStromann (commented):
## Graceful stop on budget exhaustion (CLO-421)

The intent is excellent and the design choices are mostly right: keying "stopped early" off `run_info.stopped_reason` instead of a new `PipelineRunStatus` (no migration), skipping rollback so cognified docs stay queryable, cancelling in-flight siblings to stop further billing, and deriving `DataDTO.cognify_completed` from the existing per-doc marker. `DataItemStatus(str, Enum)` with value == name means the stored-JSON comparisons are correct, and `PipelineRunCompleted` does carry `payload`/`data_ingestion_info`, so those wire up fine.

But after reading the execution stack (`run_tasks_data_item` → `run_tasks_data_item_incremental` → telemetry), I think the feature's trigger is **more fragile than the PR/test suggest**, and there's a real chance it silently falls through to the destructive path it's meant to replace. Details inline; the top three:

1. **Shallow detection on a possibly-wrapped error (high).** `is_budget_exhausted_error(error)` only inspects the top-level exception — no `__cause__` walk — yet #4321's own detector (`_find_budget_error`) had to walk `__cause__`/`RetryError`/`failed_attempts` precisely because this error travels wrapped. If any layer between the gateway raise and `asyncio.gather` re-wraps it (`raise X from budget_err`), detection returns False → rollback runs → data destroyed. This is the exact outcome the PR exists to prevent, and it fails silently.
2. **`RAISE_INCREMENTAL_LOADING_ERRORS=false` disables the feature (medium).** In that mode `run_tasks_data_item_incremental` swallows the exception and returns a `PipelineRunErrored` result; `run_tasks` then raises `PipelineRunFailedError` (a wrapper `is_budget_exhausted_error` can't see) → destructive path.
3. **Resume/no-re-bill only holds under incremental loading (medium).** The regular path writes no `DATA_ITEM_PROCESSING_COMPLETED` markers, so `_cognify_progress` would report 0 cognified and a re-run would reprocess and re-bill everything.

The unit test monkeypatches `run_tasks_data_item` to raise a bare `LLMPaymentRequiredError`, so it validates the `except` bookkeeping but **not** that a real budget error actually reaches `_enforce`'s branch unwrapped — the part most likely to break. Recommend an end-to-end-ish test through the real `run_tasks_data_item`.

LStromann:
Thanks — this caught a real fragility. Addressed all of it in `d522d2029`:

**1. Shallow detection (high)** → `is_budget_exhausted_error` now **walks the `__cause__` chain** (like `retry_config.is_quota_or_billing_error`), so a budget error wrapped by an intermediate `raise X from budget_err` is still recognized. A miss here was the worst case (silent destructive rollback), so this is the key fix.

**2. `RAISE_INCREMENTAL_LOADING_ERRORS=false` masking (medium)** → `run_tasks_data_item` now **always re-raises a budget error** (it's terminal — every remaining doc hits the same dead key), so `run_tasks` receives the typed error in *both* modes instead of a `PipelineRunFailedError` it can't classify. Path B collapses into path A.

**3. Incremental-only (medium)** → the graceful stop is now **gated on `incremental_loading`**. Resume-without-re-bill and the counts both rely on the `DATA_ITEM_PROCESSING_COMPLETED` markers that only the incremental path writes; without it we fall through to the normal error path rather than record misleading counts. (cognify runs incremental by default, so the real path is unaffected.)

**Test (your main ask)** → the test now drives a **wrapped** budget error (`outer.__cause__ = LLMPaymentRequiredError`) through `run_tasks`, plus a cause-walk detection test and an incremental-gate test (non-incremental budget → rollback + raise). Validated end-to-end locally via an `asyncio.run` harness: `documents=(3 total, 1 cognified, 2 remaining)`, COMPLETED, rollback not called.

**Minor 4** → dropped `locals().get(...)`; `in_flight`/`results` are initialized before the `try`.
**Minor 5** → extracted `is_data_item_cognified(data, pipeline_name, dataset_id)` in `pipelines/utils`, now used by both `_cognify_progress` and the `DataDTO.cognify_completed` derivation (no more duplicated lookup / hardcoded literal in the router).

Note on **#4321 dependency**: with the `__cause__` walk, detection no longer *requires* #4321 to have already converted the error to a clean 402 — it'll find a budget error wrapped anywhere in the cause chain. #4321 still helps (cleaner typed error), but this is now robust on its own.

LStromann:
Closing per Igor's call in [Slack](https://topoteretes.slack.com/archives/D09FC78D3NY/p1785853818578319): budget handling belongs on the cloud side via `cost_guard`, not in the SDK. The correct fix is preventing cognify from starting without budget, rather than stopping it gracefully mid-run.

Not carried over, worth a small follow-up: `DataDTO.cognify_completed` on `GET /v1/datasets/{id}/data` (~25 lines, no budget dependency) — the per-document processing status Igor suggested keeping on the SDK side.

Branch `feat/clo-421-dataset-pause-resume` retained.

PR #4325, opened 2026-08-03
fix: Return single task result from run_tasks_parallel

### LINKED ISSUE #4319: [Bug]: run_tasks_parallel returns an empty list when given a single task
### Bug Description

`run_tasks_parallel` drops the task result when the task list has exactly one entry. The return guard in `cognee/modules/pipelines/operations/run_parallel.py:11` is `results[len(results) - 1] if len(results) > 1 else []`. With one task, `len(results)` is 1, so the condition is false and the function returns `[]` instead of the result. Nothing is logged, so the next pipeline stage receives an empty list and the run reports success.

### Steps to Reproduce

Pure Python, no DB or LLM needed. This is the body of `parallel_run` copied verbatim:

```python
import asyncio

async def parallel_run(coros):
    parallel_tasks = [asyncio.create_task(c) for c in coros]
    results = await asyncio.gather(*parallel_tasks)
    return results[len(results) - 1] if len(results) > 1 else []

async def work(v):
    return v

async def main():
    print("one task :", await parallel_run([work("important-value")]))
    print("two tasks:", await parallel_run([work("a"), work("b")]))

asyncio.run(main())
```

### Expected Behavior

One task returns that task's result, `important-value`.

### Actual Behavior

```
one task : []
two tasks: b
```

### Environment

- OS: macOS 15
- Python version: 3.11
- Cognee version: main @ 459069a
- LLM Provider: n/a, no LLM needed to reproduce
- Database: n/a, no DB needed to reproduce

### Additional Context

`run_tasks_parallel` is a public export (`cognee/pipelines/__init__.py:45`), so a caller building a task list programmatically hits this whenever the list happens to have one element. `results[-1] if results else []` covers all three cases.

Happy to open the PR if you want to assign this to me.

### Pre-submission Checklist

chiruu12:
Fixes #4319.

`run_tasks_parallel` returned `[]` when the task list had exactly one entry.
The guard was `results[len(results) - 1] if len(results) > 1 else []`, so a
single result fell to the else branch and the task's output was dropped
silently. The next pipeline stage received an empty list and the run still
reported success.

Changed to `results[-1] if results else []`, which keeps the existing
"last result wins" contract for multiple tasks and the empty list for no
tasks.

### Test plan

- Added `cognee/tests/unit/pipelines/test_run_tasks_parallel.py` covering one,
  many, and zero tasks. The single-task case fails on dev and passes with this
  change.
- `pytest cognee/tests/unit/pipelines/`: 45 passed.

### DIFF
--- a/cognee/modules/pipelines/operations/run_parallel.py
+++ b/cognee/modules/pipelines/operations/run_parallel.py
@@ -8,6 +8,6 @@ def run_tasks_parallel(tasks: List[Task]) -> Callable[[Any], Generator[Any, Any,
         parallel_tasks = [asyncio.create_task(task.run(*args, **kwargs)) for task in tasks]
 
         results = await asyncio.gather(*parallel_tasks)
-        return results[len(results) - 1] if len(results) > 1 else []
+        return results[-1] if results else []
 
     return Task(parallel_run)
--- a/cognee/tests/unit/pipelines/test_run_tasks_parallel.py
+++ b/cognee/tests/unit/pipelines/test_run_tasks_parallel.py
@@ -1,0 +1,32 @@
+"""Tests for run_tasks_parallel's return value."""
+
+import pytest
+
+from cognee.modules.pipelines.operations.run_parallel import run_tasks_parallel
+from cognee.modules.pipelines.tasks.task import Task
+
+
+async def echo(value):
+    return value
+
+
+@pytest.mark.asyncio
+async def test_single_task_returns_its_result():
+    """A one-task list used to return [] because the guard tested len(results) > 1."""
+    parallel = run_tasks_parallel([Task(echo)])
+
+    assert await parallel.run("important-value") == "important-value"
+
+
+@pytest.mark.asyncio
+async def test_multiple_tasks_return_last_result():
+    parallel = run_tasks_parallel([Task(echo), Task(echo)])
+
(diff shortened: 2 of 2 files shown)


chiruu12:
Bump on this one. CI red here is the usual fork-secrets pattern rather than anything in the diff. The change is one function in run_tasks_parallel plus its test. Let me know if you would rather it target a different branch.

chiruu12:
Rebased onto current dev at 5aef529c. The branch was 596 commits behind and the diff still applies unchanged.

The red checks are not coming from this change. The unit and OS matrix jobs in `basic_tests.yml` and `test_different_operating_systems.yml` read `secrets.LLM_API_KEY` and friends, and the vector and relational DB jobs start service containers, so on a fork PR they fail during setup before any test runs. The jobs that need no secrets all pass: CLI Unit Tests (Mocked, No Secrets), Telemetry Test (Mocked, No Secrets), Code Quality, basic checks, check-dco, and the Docker backend image build.

The change is one line in `run_parallel.py` plus a unit test, so it should be cheap to confirm on a run that has the secrets.

dexters1:
Thanks for the PR @chiruu12! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #4330, opened 2026-08-04
fix(storage): isolate file health check probes

### LINKED ISSUE #4329: [Bug]: File-storage health check can delete an existing object
### Bug Description

`HealthChecker.check_file_storage()` always probes the fixed key `health_check_test` and then removes it. If that name already belongs to real data, a health request deletes the data it is supposed to be checking.

The two current paths fail slightly differently:

- Local storage overwrites the existing file with `test`, then removes it.
- S3 calls `store()` with its default `overwrite=False`. If the object already exists, the write is skipped, but the following `remove()` still deletes the existing object.

Concurrent health probes also share the same key, so one worker can remove another worker's probe and make the endpoint intermittently report unhealthy.

### Steps to Reproduce

Run this with an isolated local data directory on the current `dev` branch:

```python
import asyncio
import tempfile
from pathlib import Path

import cognee
from cognee.api.v1.health.health import HealthChecker

with tempfile.TemporaryDirectory() as root:
    cognee.config.data_root_directory(root)
    existing = Path(root) / "health_check_test"
    existing.write_text("keep me")

    result = asyncio.run(HealthChecker().check_file_storage())
    print(result.status, existing.exists())
```

Output on `dev`:

```text
HealthStatus.HEALTHY False
```

### Expected Behavior

A health check should only remove the probe object created by that invocation. Existing files or objects must be left untouched, and concurrent probes should not collide.

### Actual Behavior

The check returns `healthy` after deleting the pre-existing object.

### Environment

- Branch: `dev` at `108f57b0e`
- Python: 3.11
- Reproduced with local storage
- The S3 behavior follows from the same fixed key plus `store(..., overwrite=False)`

### Suggested Scope

A small fix would generate a unique key for each invocation, use the storage abstraction for both local and S3, and remove that key in `finally`. I have a patch and focused tests ready for the existing-file, concurrent-key, and partial-write cases.

I searched the existing issues and PRs and did not find this report. #4190 also changes the health endpoint, but its current diff leaves `check_file_storage()` unchanged. Happy to open the storage fix separately if that scope makes sense.

JiataiWang:
## What changed

`check_file_storage()` used the same `health_check_test` key for every probe. If that key already existed, a local probe overwrote and removed it. With S3, `store(overwrite=False)` could leave the existing object untouched and cleanup would still remove it.

This change gives each probe its own hidden UUID-based key and cleans it up in `finally`, including after a partial write failure. Local and S3-backed checks now use the same `StorageManager` path.

I added focused tests for:

- preserving an existing `health_check_test` object
- using a fresh key for each probe
- cleaning up after a store failure

Fixes #4329.

## Local checks

```text
uv run pytest cognee/tests/unit/api/v1/health/test_file_storage_health.py -q
3 passed, 10 warnings in 0.01s

uv run ruff check cognee/api/v1/health/health.py cognee/tests/unit/api/v1/health/test_file_storage_health.py
All checks passed!

uv run ruff format --check cognee/api/v1/health/health.py cognee/tests/unit/api/v1/health/test_file_storage_health.py
2 files already formatted
```

## Coordination

#4190 also edits `health.py`, but it covers endpoint routing and timeouts. This patch is limited to the storage probe lifecycle and can be rebased if that lands first.

## Checklist

- [x] Bug fix with the smallest practical scope
- [x] Regression tests added
- [x] Existing PRs checked for overlap
- [x] Commit includes DCO sign-off

## DCO Affirmation

### DIFF
--- a/cognee/api/v1/health/health.py
+++ b/cognee/api/v1/health/health.py
@@ -7,6 +7,7 @@ from sqlalchemy import text
 from datetime import datetime, timezone
 from typing import Dict
 from enum import Enum
+from uuid import uuid4
 from pydantic import BaseModel
 
 from cognee.version import get_cognee_version
@@ -148,7 +149,6 @@ class HealthChecker:
         """Check file storage health."""
         start_time = time.time()
         try:
-            import os
             from cognee.infrastructure.files.storage.get_file_storage import get_file_storage
             from cognee.base_config import get_base_config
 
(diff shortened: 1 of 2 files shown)


JiataiWang:
Closing this as superseded by #4528, which introduced unique per-probe paths and completed #4329, resolving the original overwrite/concurrency risk. This branch's remaining differences (unified StorageManager use and cleanup after a partial store failure) no longer justify rebasing the now-conflicting branch; those can be proposed separately if a concrete remaining failure is observed.

PR #4332, opened 2026-08-04
feat(telemetry): Add tenant-local sink, fix event identity (COG-6062)

LStromann:
Makes the existing SDK telemetry usable as the cloud activity log, replacing the bespoke per-operation activity log from #4291 (closed). Paired with cloud-backend#287, which enables it for v2 tenants.

Linear: [COG-6062](https://linear.app/cognee/issue/COG-6062/enable-enriched-sdk-telemetry-for-cloud-tenants)

**Open-source behaviour is unchanged** — `TELEMETRY_SINK` defaults to `http`, so nothing new is written or stored unless a deployment opts in.

## 1. Identity — fixes a live OSS bug

`send_telemetry` does `str(user_id)`, but `remember.py`, `improve.py` and `forget.py` pass the **`User` model object**, which defines no `__str__`. Those events have been recording `<cognee.modules.users.models.User.User object at 0x...>` as the user id. From the warehouse, last 3 days:

| `tracking_event` | events | with `tenant_id` | `user_id` = object repr |
|---|---:|---:|---:|
| `Pipeline Run Started` | 1,043,396 | 1,043,253 | 0 |
| `cognee.improve` | 289,828 | **0** | **288,511** |
| `cognee.remember` | 64,313 | **0** | **35,387** |
| `cognee.recall` | 147,365 | **0** | 355 |
| `cognee.forget` | 4,978 | **0** | **4,173** |
| `* API Endpoint Invoked` | ~2.0M | **0** | 0 |

~2.6M corrupted events per 30 days. `run_tasks_with_telemetry` already did this correctly, which is why `Pipeline Run *` is clean — this PR generalises that pattern.

`send_telemetry` now resolves its caller itself (`User` / `UUID` / `"sdk"`) and records `tenant_id`. Fixing it **centrally** rather than at four call sites corrects every emitter at once and avoids threading `tenant_id` through ~40 sites. Sites that pre-resolved `user.id` now pass the model, so their events gain a tenant too — done with an AST rewrite restricted to `send_telemetry`'s second positional argument, not a regex.

## 2. Sinks are a list, not a switch

`TELEMETRY_SINK` is comma-separated. A deployment can run `postgres` (nothing leaves its boundary) and later `postgres,http` to *also* feed the shared analytics pipeline. Adding a destination is configuration, not code.

The `postgres` sink buffers events and writes them to the deployment's own relational DB as a rolling window (`TELEMETRY_RETENTION_DAYS`, default 30), read back via `GET /v1/activity/events`. It is bounded, batched, flushed on shutdown, and lossy under pressure by design — telemetry must never slow down or fail the operation it observes.

Per-task events (`f"{task_type} Task Started"` etc.) are excluded from local sinks: 54% of all volume and meaningless to an end user. The HTTP sink still receives them, so historical analytics are unaffected.

The table mirrors `analytics.main.pipeline_events` columns so warehouse queries port over without a reshape.

## 3. Columns promoted out of `properties`

So an activity view can filter on them without a JSON scan:

| Column | Source | Null when |
|---|---|---|
| `dataset_id` | normalised from the three spellings call sites use (`dataset_id`, `dataset_ids`, `dataset`) | no dataset, or the event spans several — the list stays in `properties` |
| `pipeline_run_id` | `current_pipeline_run_id` ContextVar, published by `run_tasks_with_telemetry` from `ctx.pipeline_run_id` | event raised outside a pipeline |
| `origin` | `telemetry_origin` ContextVar, set per request by the API layer from `X-Cognee-Client`; falls back to `TELEMETRY_ORIGIN`, then `"sdk"` | never |

`pipeline_run_id` is what makes `/v1/activity/events` joinable to `/v1/activity/pipeline-runs`. `origin` distinguishes dashboard traffic from MCP/coding-agent traffic. Both travel on ContextVars rather than arguments, because the emitters worth labelling sit deep inside pipelines that have no reason to know how they were invoked.

`improve()`'s ambiguous `dataset` property is also split into explicit `dataset_id` / `dataset_name` — it accepts either a name or an id, so the column was previously only populated when the caller happened to pass an id. `dataset` is unchanged for the warehouse queries that read it.

## Also

- `TELEMETRY_ENDPOINT` overrides the hardcoded proxy URL.
- The deprecated `user_id=` keyword still works, for out-of-tree callers.
- New env vars documented in `.env.template`.

## Testing

- 30 unit tests in `cognee/tests/unit/shared/test_telemetry_identity_and_sinks.py`.
- Round-trip verified against a real SQLite DB: real UUIDs persist, internal events excluded, `http`-only writes nothing locally, `postgres,http` delivers to both, `TELEMETRY_DISABLED` beats every sink, and `dataset_id`/`pipeline_run_id`/`origin` populate as documented.
- Nullability matrix confirmed: `id`, `event_name`, `origin`, `anonymous_id`, `properties`, `created_at` are never null; `user_id`/`tenant_id` are null only without a real `User` (the OSS default user or the `"sdk"` sentinel).
- `ruff` v0.15.11 (the pre-commit pin) clean and formatted.
- Pipeline suite compared against clean `dev` at identical scope — 6 failed / 61 passed on both, identical failure sets. `run_tasks_with_telemetry` gained a `finally`, so this was checked deliberately.

### Red CI is pre-existing on `dev`

`Code Quality`, `basic checks / Lockfile and Pre-commit Hooks`, `Community Tests Status`, `Unit Tests (Mocked, No Secrets) (3.11.x)` and `Test Completion Status` fail identically on unrelated PRs (e.g. #4331). On a clean `origin/dev` worktree with the pinned hook version: `ruff format --check` wants 8 files (`cognee/api/v1/search/search.py`, four `cognee/modules/observability/*`, …), `ruff check` reports 1 error, and `test_memory_semconv.py` raises `ImportError: OpenTelemetry packages are required for tracing`. These look to originate in `107fdd857`. I deliberately did not fold an unrelated 8-file reformat plus a lint fix into a telemetry PR.

## Note for the cloud side

Tenant pods run cloud-backend's Alembic chain (`COGNEE_ALEMBIC_PATH`), not this one, so the companion PR adds `telemetry_events` there as well. The migration here is for OSS/self-hosted deployments.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -395,6 +395,30 @@ LITELLM_LOG="ERROR"
 #DEFAULT_USER_EMAIL=""
 #DEFAULT_USER_PASSWORD=""
 
+# -- Product Telemetry ---------------------------------------------------------
+# Where telemetry events go. Comma-separated, so a deployment can write to more
+# than one place:
+#   http     (default) POST to the analytics proxy
+#   postgres store in this deployment's own relational DB as a rolling window,
+#            readable via GET /v1/activity/events. Nothing leaves the machine.
+# e.g. "postgres" for local-only, "postgres,http" for both.
+#TELEMETRY_SINK="http"
+# Point the http sink at your own collector instead of the default proxy.
+#TELEMETRY_ENDPOINT="https://telemetry.example.com"
+# Labels which interface a call came through ("api", "mcp", "cli", …). The API
+# sets this per request from the X-Cognee-Client header; this is the fallback.
+#TELEMETRY_ORIGIN="sdk"
+# How long the postgres sink keeps events before pruning them.
+#TELEMETRY_RETENTION_DAYS=30
+# Postgres-sink buffering. Events are batched rather than written one at a time;
+# the buffer is bounded and drops the oldest events rather than ever blocking.
+#TELEMETRY_BUFFER_MAX=5000
+#TELEMETRY_FLUSH_INTERVAL=5
+#TELEMETRY_FLUSH_BATCH=200
+# Overrides the machine-level analytics id. Set this when the filesystem is
+# ephemeral (containers), or the id is regenerated on every restart.
+#TRACKING_ID=""
+
 # -- Cognee Logging -----------------------------------------------------------
 # Console log level: DEBUG, INFO, WARNING, ERROR, CRITICAL (default: INFO)
 #LOG_LEVEL="INFO"
--- a/cognee/alembic/env.py
+++ b/cognee/alembic/env.py
@@ -9,6 +9,7 @@ from sqlalchemy.ext.asyncio import async_engine_from_config
 from cognee.infrastructure.databases.relational import get_relational_engine, Base
 import cognee.modules.session_lifecycle.models  # noqa: F401
 import cognee.modules.migrations.models  # noqa: F401
+import cognee.modules.telemetry.models  # noqa: F401
 
 # this is the Alembic Config object, which provides
 # access to the values within the .ini file in use.
--- a/cognee/alembic/versions/a3f7c1b9d2e4_add_telemetry_events.py
+++ b/cognee/alembic/versions/a3f7c1b9d2e4_add_telemetry_events.py
@@ -31,6 +31,8 @@ def upgrade() -> None:
         "telemetry_events",
         sa.Column("id", sa.UUID(), primary_key=True),
         sa.Column("event_name", sa.String(), nullable=True),
+        sa.Column("operation", sa.String(), nullable=True),
+        sa.Column("event_kind", sa.String(), nullable=True),
         sa.Column("user_id", sa.UUID(), nullable=True),
         sa.Column("tenant_id", sa.UUID(), nullable=True),
         sa.Column("dataset_id", sa.UUID(), nullable=True),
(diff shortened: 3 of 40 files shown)


LStromann:
**Heads-up on the two red checks — they are inherited from `dev`, not from this PR.**

`Code Quality` and `basic checks / Lockfile and Pre-commit Hooks` both fail on the `pre-commit` step, which runs `--all-files`. The file it wants to reformat (`cognee/tests/unit/modules/observability/test_memory_semconv.py`) is untouched by this branch.

Reproduced on a clean `origin/dev` worktree with the pinned hook version (`ruff` v0.15.11):

```
$ ruff format --check .
8 files would be reformatted, 2005 files already formatted
  cognee/api/v1/search/search.py
  cognee/modules/observability/logs.py
  cognee/modules/observability/metrics.py
  cognee/modules/observability/tracing.py
  cognee/tests/unit/modules/observability/test_memory_semconv.py
  ...

$ ruff check .
Found 1 error.   # unused variable `provider`
```

These appear to originate in `107fdd857` (`fix(observability): fix Dynatrace OTLP export`), so **every** PR against `dev` is currently red on this check. I deliberately did not fold an unrelated 8-file reformat plus a lint fix into a telemetry PR — that belongs in a small fix straight to `dev`.

Everything in this branch is clean under the same pinned version:

```
36 files already formatted
All checks passed!
```

`linear-issue-check` (the required check) passes.

LStromann:
Superseded by #4344, which keeps only the identity bug fix.

The local `postgres` sink is dropped. Checked the warehouse: tenant pods already deliver telemetry to Segment — 316 distinct tenants reporting `Pipeline Run Started` with `tenant_id`, ingesting today — so Igor's question about whether pod→Segment egress is possible is already answered in the affirmative, and a second delivery path wasn't buying anything.

What survives in #4344 is the part that's needed regardless of destination: without it, `Remember Entry API Endpoint Invoked` (378,393 events) and `Improve API Endpoint Invoked` (255,346) carry no `tenant_id` at all, so no per-tenant view can be built from the Segment data either.

Also dropped: the `origin` / `pipeline_run_id` ContextVars, the event-name taxonomy, and the `TELEMETRY_*` settings — about 1,000 lines across 12 files. The two ContextVars are cheap and independently useful for warehouse joins; they can come back on their own if wanted.

Branch `feat/cog-6062-telemetry-tenant-sink` retained — the sink is the natural starting point for on-pod live event access, which is being investigated separately.

PR #4336, opened 2026-08-04
feat(permissions): let role members see their own roles and co-members (COG-6064)

LStromann:
## Summary

Listing roles and role members required tenant-wide user management permission, which in practice means **only the tenant owner** — `has_user_management_permission` allows the owner, or a member of a role literally named `admin`, and nothing in the codebase ever creates that role (only the owner can, via `create_role`). Ordinary tenant members could not see who shares their roles.

Role membership is now sufficient authorization for the caller's own roles.

| Endpoint | Owner / admin | Plain member |
| --- | --- | --- |
| `GET /permissions/tenants/{tenant_id}/roles` | all roles in tenant | only roles they belong to |
| `GET /permissions/tenants/{tenant_id}/roles/{role_id}/users` | any role in tenant | roles they belong to (403 otherwise) |

`GET /tenants/{tenant_id}/users` (the full tenant directory) is unchanged and stays owner/admin only. No new tables and no new permission types were needed.

## Commits

- **`4926b88`** `fix(permissions)` — `get_users_in_role` selected members by `role_id` alone, without scoping to the tenant the caller was authorized for, so a caller with user management permission in one tenant could pass a role id from another tenant and read its members' emails. The role is now looked up by `(id, tenant_id)` and returns 404 when it does not belong to the tenant.
- **`e8af58f`** `feat(permissions)` — the visibility change above. `get_tenant_roles` catches `PermissionDeniedError` to decide whether to join `user_roles` as a filter; `get_users_in_role` checks membership first and only falls back to `has_user_management_permission` for non-members, leaving the existing 403 path intact.
- **`b388995`** `chore` — see *Unrelated lint commit* below.

## Behavior change to note

`GET /tenants/{tenant_id}/roles` no longer returns 403 for non-privileged callers — it returns a filtered, possibly empty list. A caller passing some other tenant's id gets `[]`. No data crosses tenants. If a 403 is preferred there, that needs an extra `UserTenant` membership check.

## Unrelated lint commit

`b388995` applies `ruff format` to 8 files and drops an unused `provider` binding in `test_memory_semconv.py`. None of those files are touched by this feature — the repo-wide `Code Quality` check is **already failing on `dev` for every PR** because of them (e.g. #4334 fails identically). Included so this PR's required check can pass; happy to split it into its own PR instead.

## Testing

- 6 new tests in `cognee/tests/unit/users/tenants/test_role_visibility.py`: member-can-read, non-member-denied, admin-still-allowed, cross-tenant-404, and both branches of the roles filter.
- `cognee/tests/unit/users` — 16 passed.
- `ruff check .` and `ruff format --check .` — clean repo-wide.

Note on how tests were run locally: `uv run` cannot build the env in this checkout (`ladybug` 0.18.1 fails its native build, unrelated to these changes), so pytest ran from a sibling venv resolving `cognee` from this worktree. CI is the real check.

Closes COG-6064

### DIFF
--- a/cognee/modules/users/tenants/methods/get_tenant_roles.py
+++ b/cognee/modules/users/tenants/methods/get_tenant_roles.py
@@ -3,23 +3,36 @@ from sqlalchemy.orm import selectinload
 from uuid import UUID
 
 from cognee.infrastructure.databases.relational import get_relational_engine
+from cognee.modules.users.exceptions import PermissionDeniedError
+from cognee.modules.users.models import Role, UserRole
 from cognee.modules.users.permissions.methods.has_user_management_permission import (
     has_user_management_permission,
 )
 
 
 async def get_tenant_roles(tenant_id: UUID, user):
-    # Check permissions - only tenant owner or users with specific roles (e.g. admin) can list roles in this tenant
-    # TODO: Consider if regular users should be able to see the list of roles with info
-    await has_user_management_permission(user.id, tenant_id)
+    """
+    List roles in a tenant.
+
+    Callers with user management permission (owner or admin) see every role in
+    the tenant. Everyone else sees only the roles they are a member of.
+    """
+    try:
+        await has_user_management_permission(user.id, tenant_id)
+        can_manage_users = True
+    except PermissionDeniedError:
+        can_manage_users = False
 
     db_engine = get_relational_engine()
     async with db_engine.get_async_session() as session:
-        from cognee.modules.users.models import Role
+        query = select(Role).options(selectinload(Role.users)).where(Role.tenant_id == tenant_id)
 
-        roles_result = await session.execute(
-            select(Role).options(selectinload(Role.users)).where(Role.tenant_id == tenant_id)
-        )
+        if not can_manage_users:
+            query = query.join(UserRole, Role.id == UserRole.role_id).where(
+                UserRole.user_id == user.id
+            )
+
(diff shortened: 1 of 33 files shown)


cognee/modules/users/tenants/methods/get_tenant_roles.py
```diff
@@ -3,23 +3,36 @@
...
+    except PermissionDeniedError:
+        can_manage_users = False
 
     db_engine = get_relational_engine()
     async with db_engine.get_async_session() as session:
-        from cognee.modules.users.models import Role
+        query = select(Role).options(selectinload(Role.users)).where(Role.tenant_id == tenant_id)
+
+        if not can_manage_users:
+            query = query.join(UserRole, Role.id == UserRole.role_id).where(
```
dexters1: We need to make sure also that the Role is part of the tenant as well

LStromann: Good catch — reworked this in a184bd22: instead of joining `UserRole` onto the query, the membership check is now a `Role.id.in_(select(UserRole.role_id).where(UserRole.user_id == user.id))` subquery filter. That way `Role.tenant_id == tenant_id` stays a standalone top-level predicate that always applies, rather than being combined with the join — so it's unambiguous that a role must belong to the tenant regardless of the membership check. Verified against the existing test suite (16 passed).

PR #4342, opened 2026-08-05
CLO-434 feat: Expose per-agent session usage

goran-radonic:
Cloud UI needs to show which client (Claude Code, Codex, Slack, ...) and which user is driving token/cost spend. session_records and the agent-connections registry tracked this separately with no way to correlate them, and token counts were a chars/4 estimate everywhere.

Add two read endpoints that join on session_id:
- GET /sessions/with-agent-info: per-session detail merged with agent-connection metadata (type, source, session name, origin).
- GET /sessions/cost-by-user-agent: totals grouped by (user, agent type), sorted by spend, chart-ready. Role-scoped: a tenant owner/admin sees every member, a regular member falls back to their own sessions instead of being denied outright.

Join/aggregation logic lives in modules/session_lifecycle/agent_usage.py so it's unit-testable independent of FastAPI; the router stays a thin HTTP layer.

AgentConnectionType moves from a closed Literal to a free string so a new client (Cursor, Windsurf, ...) can self-declare its own type at registration without a backend code change; derive_connection_type() is now only a best-effort fallback for sessions that never registered. Also fixes a dead cc_ prefix check (the leading space in the joined text meant it could never match).

Token counts are now exact, not estimated, for the default instructor-based LLM path: LLMGateway reads the real prompt/completion counts instructor already attaches to every parsed response as _raw_response, instead of guessing from character length. This is what feeds every endpoint above, so "who spends the most" is now a real number for that path (litellm_native/BAML still estimate).

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/api/v1/sessions/routers/get_sessions_router.py
+++ b/cognee/api/v1/sessions/routers/get_sessions_router.py
@@ -407,15 +407,24 @@ def get_sessions_router() -> APIRouter:
         """Cost + token totals grouped by (user, agent type) — feeds a
         "who spends the most, with which agent" chart (CLO-434 follow-up).
 
-        Role-scoped, not all-or-nothing: a tenant owner/admin (same
-        check ``GET /tenants/{id}/users`` uses) sees every member's
-        spend. A regular member — or anyone with no tenant, i.e.
-        single-user/local mode — falls back to just their own sessions
-        rather than being denied outright.
+        Visibility matches every other endpoint in this router: the
+        caller, their child agents, and dataset-shared sessions. On top
+        of that base scope, a tenant owner/admin (same check
+        ``GET /tenants/{id}/users`` uses) additionally sees every
+        member's spend. A regular member — or anyone with no tenant,
+        i.e. single-user/local mode — just keeps the base scope rather
+        than being denied outright.
         """
         since = _range_since(range)
         try:
-            result = await compute_cost_by_user_agent(user=user, since=since)
+            permitted = await _permitted_dataset_ids_for(user)
+            visible_ids = await _visible_user_ids(user)
+            result = await compute_cost_by_user_agent(
+                user=user,
+                visible_user_ids=visible_ids,
+                permitted_dataset_ids=permitted,
+                since=since,
+            )
             return jsonable_encoder(result)
         except Exception as exc:
             logger.error("cost_by_user_agent failed: %s", exc, exc_info=True)
--- a/cognee/api/v1/sessions/routers/get_sessions_router.py
+++ b/cognee/api/v1/sessions/routers/get_sessions_router.py
@@ -7,17 +7,19 @@ abandonment-by-idle rule so no sweeper is needed.
 
 from datetime import datetime, timedelta, timezone
 from typing import Literal, Optional
-from uuid import UUID as UUIDType
 
 from fastapi import APIRouter, Depends, HTTPException, Path, Query
 from fastapi.encoders import jsonable_encoder
 from fastapi.responses import JSONResponse
 from sqlalchemy import and_, func, or_, select
 
+from cognee.exceptions import CogneeApiError
 from cognee.infrastructure.databases.relational import get_relational_engine
 from cognee.modules.session_lifecycle.agent_usage import (
-    build_sessions_with_agent_info_page,
-    compute_cost_by_user_agent,
+    _permitted_dataset_ids_for,
+    _visible_user_ids,
+    get_cost_by_user_agent,
+    get_sessions_with_agent_info,
 )
 from cognee.modules.session_lifecycle.metrics import (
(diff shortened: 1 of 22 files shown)


cognee/api/v1/sessions/routers/get_sessions_router.py
```diff
@@ -338,6 +342,94 @@ async def cost_by_model(
...
+                visible_user_ids=visible_ids,
+                since=since,
+                status_filter=status,
+                limit=limit,
+                offset=offset,
+                order_by=order_by,
+                descending=descending,
+            )
+            return jsonable_encoder(result)
+        except Exception as exc:
```
dexters1, line 362: We need to stop doing this all encompassing exception handling, CogneeApiError type errors should not be caught by these mechanisms as they already handle all responses and are desinged to give the user the necessary info on why an error happened

I cleaned this already in several places

cognee/api/v1/sessions/routers/get_sessions_router.py
```diff
@@ -338,6 +342,94 @@ async def cost_by_model(
...
+            permitted = await _permitted_dataset_ids_for(user)
+            visible_ids = await _visible_user_ids(user)
+            result = await compute_cost_by_user_agent(
+                user=user,
+                visible_user_ids=visible_ids,
+                permitted_dataset_ids=permitted,
+                since=since,
+            )
+            return jsonable_encoder(result)
+        except Exception as exc:
```
dexters1, line 395: We need to stop doing this all encompassing exception handling, CogneeApiError type errors should not be caught by these mechanisms as they already handle all responses and are desinged to give the user the necessary info on why an error happened

I cleaned this already in several places

cognee/api/v1/sessions/routers/get_sessions_router.py
```diff
@@ -338,6 +342,94 @@ async def cost_by_model(
...
+
+        Visibility matches every other endpoint in this router: the
+        caller, their child agents, and dataset-shared sessions. On top
+        of that base scope, a tenant owner/admin (same check
+        ``GET /tenants/{id}/users`` uses) additionally sees every
+        member's spend. A regular member — or anyone with no tenant,
+        i.e. single-user/local mode — just keeps the base scope rather
+        than being denied outright.
+        """
+        since = _range_since(range)
```
dexters1, line 386: Can we move all buisness logic out of the routers? And make a method for these so pure SDK users can use these as well

cognee/modules/session_lifecycle/agent_usage.py
```diff
@@ -0,0 +1,345 @@
...
+# full in-memory list, then slices. This is the slice size we ask for
+# when we need "effectively all of them" to build the session->agent
+# map below. If a scope ever has more connections than this, the excess
+# silently falls back to the prefix heuristic (see the has_more check
+# below) rather than erroring, so it's raised here as a single visible
+# constant instead of a bare literal at the call site.
+_AGENT_CONNECTIONS_PAGE_CAP = 10_000
+
+
+async def _permitted_dataset_ids_for(user: User) -> list[UUIDType]:
```
dexters1: Can you check if we have a function like this already? This seems that it will be called by more than one place and should be put in a better location

Let's also remove except Exception:

We should almost never use such exception handling as it masks real errors and cause issues. We should only catch errors we expect and are okay with like PermissionDeniedError

I would also personally prefer the error PermissionDeniedError itself would be raised instead of caught so we have more transparency with the user and he knows the operation was not succesful and not silently return 200 with an empty list

cognee/modules/session_lifecycle/agent_usage.py
```diff
@@ -0,0 +1,345 @@
...
+    try:
+        datasets = await get_specific_user_permission_datasets(user.id, "read", None)
+        return [ds.id for ds in datasets] if datasets else []
+    except PermissionDeniedError:
+        return []
+    except Exception:
+        return []
+
+
+async def _child_agent_user_ids(user_id: UUIDType) -> list[UUIDType]:
```
dexters1: This does not seem like it should be a private function and that it should be stored at this location

as it could be used outside of the session_lifecycle system

cognee/modules/session_lifecycle/agent_usage.py
```diff
@@ -0,0 +1,345 @@
...
+
+    engine = get_relational_engine()
+    async with engine.get_async_session() as session:
+        rows = (
+            await session.execute(select(UserModel.id).where(UserModel.parent_user_id == user_id))
+        ).all()
+        return [row.id for row in rows]
+
+
+async def _visible_user_ids(user: User) -> list[UUIDType]:
```
dexters1: Same comment as for the function above

PR #4343, opened 2026-08-05
feat(score): Add memory accuracy score endpoint (COG-6067)

LStromann:
## Goal

Give customers a measured **memory accuracy score** for a dataset, so they have a concrete reason to upload more data or define a graph schema.

Pipeline: sample the customer's own data → cluster into topics → generate synthetic questions grounded in real chunks → answer them via cognee recall → judge with LLM-as-a-judge → aggregate per topic and overall.

Linear: [COG-6067](https://linear.app/cognee/issue/COG-6067/memory-accuracy-score-endpoint-synthetic-real-questions-llm-as-judge)

## Endpoints

* `POST /api/v1/score` — start a run. Body: `dataset_id` (**required**), `run_in_background` (default true), `synthetic_target` (default 100, max 500), `real_question_limit` (default 20, max 200). Returns `{run_id}`, or the full document when blocking.
* `GET /api/v1/score/{run_id}` — full run document
* `GET /api/v1/score/latest` — most recent run the caller may read

Errors: 400 no tenant, **403 no read permission on the dataset**, 404 dataset or run not found, 409 run already active, 400 spend parameter out of range (`client.py` maps every `RequestValidationError` to 400 app-wide).

## Two KPIs, deliberately not blended

**`overall_accuracy`** — of the questions the memory attempted, how many were right. From **synthetic questions only**, which have a golden answer derived from their source chunk.

**`coverage`** — of the questions actually asked by users, how many the memory could answer at all. From **real questions only**.

Both are needed because they move in opposite directions on a thin dataset: a 5-chunk graph answers questions drawn from those same 5 chunks almost perfectly, so accuracy alone *flatters* exactly the customer who needs to upload more. Coverage is what actually drops. Blended into one number, 0.6 could mean either wrong answers or absent data — opposite remedies. Kept apart the pair maps straight onto the CTA:

| accuracy | coverage | reading |
| -- | -- | -- |
| high | low | the data is missing → upload more |
| low | high | the data is there, retrieval is failing |
| high | high | memory is working |

Synthetic questions can never reveal a coverage gap — they are generated from chunks that exist by construction. A question the tenant actually asked is the only input that can.

**Two distinct real-question failures, two lists.** `unanswerable_real_questions` is the "questions your users asked that could not be answered" list (coverage gap). `ungrounded_real_questions` is hallucination — the memory answered, and the answer was not supported by the retrieved context. They must not be conflated: a coverage gap is `answered=false, grounded=true`, a hallucination is the reverse.

## Other design decisions

**1. Access control.** Sharing a tenant is not permission to score a dataset — a run returns `expected_answer` values lifted verbatim from its chunk text plus full recall output over it, so scoring is a read of its contents. `resolve_memory_score_dataset` checks `get_specific_user_permission_datasets(user_id, "read", [dataset_id])`. Tenancy is checked *first*, so a cross-tenant id stays a 404 rather than a 403 that confirms it exists. This cannot be delegated to the DB layer: the run enters context as `dataset.owner_id`, so per-user isolation never sees the caller. Reads are gated the same way and answer 404, and `/latest` means the latest run *the caller may read*.

**2. `dataset_id` is required, never inferred.** Both tenant charts set `ENABLE_BACKEND_ACCESS_CONTROL: true`, so graph/vector engines resolve per user+dataset. There is no single "tenant graph" and no defensible way to guess. Callers state it: `default_dataset` on Cloud, `main_dataset` in OSS. The run enters that context via `set_database_global_context_variables(dataset.id, dataset.owner_id)` — `owner_id`, so a scheduled run with no acting user works.

**3. Real questions are read per user, never tenant-wide.** `queries` rows are one member's search history and their verbatim text is returned in the run document, so a tenant-wide read would show every member what colleagues had been searching for. The run replays only the questions of the user it is attributed to (or `dataset.owner_id` for a scheduled run). Cost: real-traffic topic weighting reflects one member's usage.

**4. Spend is capped in two places.** `synthetic_target` ≤ 500, `real_question_limit` ≤ 200. The API refuses out-of-range so the caller knows it was refused; the run *clamps* whatever it is handed, so a scheduler or SDK caller bypassing FastAPI validation is bounded too. One constant each, imported into the request model, so the advertised limit cannot drift from the enforced one. `synthetic_target` is a strict ceiling — when the budget cannot cover one question per topic, the highest-real-traffic topics keep theirs and the rest are skipped.

**5. Retriever and `top_k` pinned** (`cognee_graph_completion`, `top_k=5`) rather than inherited from tenant search defaults, so a tenant flipping its default cannot move its score without its memory changing, and two tenants' scores stay comparable.

**6. Topics capped at 5.** `compute_clusters` is borrowed from the visualization pipeline, whose `default_k` goes to 12 — there more clusters means a better picture and nothing is measured. Here `_allocate_counts` floors every topic at one question, so the topic count is a direct divisor of the budget: 12 topics left 1–2 questions each, and a per-topic accuracy from two samples can only read 0/50/100%. `k` is now `min(default_k, 5)` — the `min` matters, since forcing 5 on a thin graph would create clusters below `min_nodes_per_topic` and gate the run for a reason that is an artifact of the ceiling.

**7. Question generation batches.** One call over `MAX_CHUNKS_PER_PROMPT` (3) chunks can ground about 15 pairs, and the prompt returns fewer rather than inventing. A topic allocated 39 questions in one call therefore returned ~15 and the rest were silently lost — `synthetic_target` was never reachable. Topics needing more than one call's worth now get more calls over *different* chunks, with per-call context still bounded and the allocation still a strict ceiling.

**8. Data floor gate, before any LLM spend.** ≥50 chunks AND clustering yields ≥3 topics of ≥5 nodes each. Below it: persist `SKIPPED_INSUFFICIENT_DATA` + `floor_reason` and return. *Caveat:* the topic half costs one embedding batch. Zero LLM completion tokens. Now that `coverage` exists, the floor is arguably less load-bearing — a 5-chunk dataset would honestly report ~95% accuracy / ~10% coverage instead of a misleading 95% — so relaxing it is a reasonable follow-up.

**9. One run per tenant**, with a 2h staleness escape so a dead process cannot lock a tenant out. Bounded concurrency (5) on answer+judge.

**10. No call-to-action in the API.** Raw signals only — `coverage`, `below_data_floor`, `floor_reason`, `schema_defined`, per-topic accuracy, the two failure lists. Thresholds and copy are the UI's job.

**11. Not built on `run_tasks`.** It models ingestion — `DATASET_PROCESSING_*` statuses, `pipeline_runs` rows — and per-question rows can't be queried out of `PipelineRun.run_info` JSON. Porting later to inherit the progress queue and activity cost attribution is a reasonable follow-up.

## Reused rather than rebuilt

* `eval_framework/answer_generation/answer_generation_executor.py` — recall
* `eval_framework/evaluation/direct_llm_eval_adapter.py` — correctness judge
* `modules/visualization/semantic_clusters.py::compute_clusters` — topics, pure numpy, no LLM
* `modules/search/operations/get_queries.py` — real question history
* `api/v1/visualize/get_schema_inventory.py` — schema signal

The measured path is the same one a real user query takes.

## Verification

**46 unit tests** in `cognee/tests/unit/api/v1/score/`, run against the **real app** from `cognee.api.client` so asserted status codes are the ones callers receive. They use `asyncio.run` rather than `pytest.mark.asyncio`, so no `pytest_asyncio` plugin is needed.

**Live, against a running server.** 19-case authorisation and validation matrix: 400 no tenant · 404 absent/cross-tenant dataset · **403 tenant peer without read permission** · 400 caps/negative/missing `dataset_id` · at-cap accepted · `SKIPPED_INSUFFICIENT_DATA` + `floor_reason` on an empty graph · 200 owner reads · **404 peer reading owner's run** · **404 peer `/latest`** · 409 concurrent · 401 unauthenticated.

**Full end-to-end**, 45k chars of docs → 179 DocumentChunks → `synthetic_target=100`, 5 replayed real questions (3 answerable, 2 not):

```
status: COMPLETED
ACCURACY  0.899    synthetic asked/generated/judged: 100 / 95 / 95
COVERAGE  0.6      real answered/measured/replayed:    3 /  5 /  5
unanswerable_real_questions: ["Who won the 1966 world cup final?", ...]
histogram: {0.0: 7, 0.2: 1, 0.33: 1, 0.4: 1, 0.7: 1, 0.8: 1, 1.0: 83}
```

Live testing found three bugs the unit tests had missed, all fixed: `get_queries` imported from the package bound the submodule so every run with real question history died `ERRORED`; `synthetic_target` was unreachable (59/100); and half the generated questions asked about their own source. Single alembic head `d4f6a8b0c2e3`; the migration was edited in place (it is unmerged) and re-verified from an empty database.

## Known limitations

**`overall_accuracy` is not deterministic.** 0.932 and 0.899 are the same code on the same graph — only the sampled questions differ, because generation is an LLM call. Roughly ±3pp of run-to-run noise at 100 questions, more at smaller targets. This PR pins the retriever and `top_k` precisely so the score cannot move without the memory changing; question generation is the remaining unpinned input, and it moves the number more than most real memory improvements would. **The UI should not render small deltas as movement.** Seeding generation deterministically, or reporting an interval, is not in this PR. `coverage` does not have this problem — it is computed over a fixed replayed question set.

**The generation prompt is part of the score's definition.** Editing `memory_score_question_generation_system.txt` re-baselines every tenant's score; tightening it moved the number 13 points in testing without touching a graph node. It lives in a shared prompt directory where it does not look load-bearing.

**Coverage needs search history.** It is null when a tenant has none — including a brand-new customer, who is exactly who the CTA targets — and it is bounded by `real_question_limit` (default 20). The UI needs a "not enough usage yet" state distinct from "coverage is bad". `coverage` is also null if every real question was left unmeasured by a recall or judge failure; `real_question_count` vs `measured_real_question_count` distinguishes the two cases.

**Real questions carry no dataset attribution.** `queries` records `text`, `query_type`, `user_id`, `created_at` — no dataset. A user active in several datasets can have questions replayed against one they were never asked against. Proper fix is `dataset_id` on `Query` at log time. Mitigation: the response always names the scored `dataset_id`.

**`synthetic_target` is a ceiling, not a guarantee.** 95/100 in the run above; a topic that cannot ground its allocation logs the shortfall, and `judged_synthetic_question_count` reports the true denominator.

**`schema_defined` is derived, not stored**, and returns `True` after a plain cognify in practice. Reviewed and accepted as-is.

**An `ERRORED` run returns HTTP 200 with no diagnostic** — `MemoryScoreRun.error` is populated but not surfaced. Accepted; to be handled separately.

**Topic labels are weak** — `default_label` picks top entities by degree, giving labels like `"none, ui, nodeset"`. Fine as a graph tooltip, poor as a UI section heading.

**Chunk-derived questions are single-hop.** 0.899 on a corpus 3.5× the data floor. Questions drawn from one chunk and answered by retrieving that same chunk are close to tautological, so accuracy may never get low enough to motivate an upload on its own — which is much of why `coverage` exists.

## Follow-ups

- [ ] Frontend: score display, topic breakdown, coverage + CTA thresholds, "not enough usage yet" state (**cognee-saas**)
- [ ] Auto-trigger when the last run is >1 week old. **Prefer a scheduled sweep over tenant warm-up firing** — warm-up means a deploy restarting every tenant pod launches N simultaneous runs into the shared LiteLLM gateway while each pod is cold. Filter to tenants active in the last 7 days. Pattern exists: `analytics_etl.py`. Lives in **cloud-backend**.
- [ ] `dataset_id` on `Query` at log time
- [ ] Surface `MemoryScoreRun.error` in the run document
- [ ] Seed question generation, or report a confidence interval, so the score is comparable over time
- [ ] Reconsider the 50-chunk floor now that coverage exists
- [ ] Persist questions incrementally so a partial failure keeps its spend
- [ ] Decouple topic selection from the visualization sampling cap (`SEMANTIC_NODE_CAP = 2000`, sample shifts as the graph grows)
- [ ] Distinct status for "completed but generated nothing"
- [ ] Consider porting onto `run_tasks` for progress queue + activity cost attribution (needs cognee#4291)


LStromann:
# Review: `feat(score): Add memory accuracy score endpoint (COG-6067)`

**+2383 / −0 across 22 files** — new `/api/v1/score` endpoint, `memory_score` module, 2 relational tables, 4 prompt files, 1 alembic revision.

## Overview

Samples a dataset's graph → clusters into topics (numpy k-means, no LLM) → generates synthetic Q/expected-answer pairs grounded in real chunks → answers via cognee recall → judges (correctness for synthetic, groundedness for real) → aggregates. A data-floor gate short-circuits before LLM spend.

## What's good

- **Reuse over reimplementation** is real, not claimed: `AnswerGeneratorExecutor`, `DirectLLMEvalAdapter`, `compute_clusters`, `get_schema_inventory`. I verified every contract against `dev` — `retriever_options["cognee_graph_completion"]` accepts `top_k`, `question_answering_non_parallel` returns `retrieval_context`, `evaluate_correctness` returns `{score, reason}`, cluster dicts carry `label`/`node_ids`/`size`, and `set_database_global_context_variables` is a dual-mode async CM. All correct.
- **Synthetic/real separation enforced structurally**, not just by convention — separate columns, separate aggregates, `overall_accuracy` synthetic-only. This is the right call and it's hard to accidentally violate later.
- **`_claim_initiated_run`** is careful: conditional UPDATE so racing runners can't both claim, plus retiring orphan `INITIATED` rows that would otherwise lock the tenant out for 2h.
- **Alembic verified**: `dev` has exactly one head (`c3d5e7f9a1b2`); `d4f6a8b0c2e3 <- c3d5e7f9a1b2` keeps it single. Migration is idempotent-guarded and drops the Postgres enum on downgrade.
- Docstrings are unusually good — the "why not `run_tasks`", the cost accounting, the honest known-limitations section.

---

## High

**1. No dataset ACL check — cross-dataset disclosure inside a tenant.** `resolve_memory_score_dataset` (`run_memory_score.py`) validates only `dataset.tenant_id != tenant_id`. The repo convention is a permission check: `get_specific_user_permission_datasets(user.id, "read", [dataset_id])` (used in `get_activity_router.py`, the sync router, the sessions router) or `resolve_authorized_user_datasets`. Consequence: any authenticated tenant member can score **any** dataset in the tenant and read back `expected_answer` values — verbatim facts the LLM extracted from that dataset's chunk text — plus recall answers over it. And because the run enters context as `dataset.owner_id`, DB-layer isolation is bypassed too, so ACL is never consulted at either level. Suggest adding the read-permission check in `resolve_memory_score_dataset`.

**2. Unbounded cost inputs.** `synthetic_target` and `real_question_limit` are `ge=0` with no `le=`. The PR body prices a run at ~230k tokens / $0.58 for 100 questions; nothing rejects `synthetic_target=1000000`. Suggest adding `le=` caps and clamping server-side. Related: `_allocate_counts` returns `[1] * topic_count` when `target_count <= topic_count`, so `synthetic_target` isn't a strict ceiling either — e.g. `synthetic_target=2` with 5 topics generates 5.

**3. Tenant-wide query replay exposes users' search history to each other.** `get_tenant_queries` joins `queries.user_id → users.id` on `users.tenant_id`, pulling every tenant member's question text — and those texts come back in `questions[].text` and `ungrounded_real_questions` to whichever member calls `GET /score/{run_id}`. The existing `cognee.modules.search.operations.get_queries` is per-user deliberately. Suggest scoping to the triggering user, or gating the read behind an admin permission.

## Medium

**4. The accuracy denominator drops failures, inflating the headline number.** `_score_synthetic_question` leaves `score=None` when recall returns an empty answer, and `_synthetic_scores` filters `None` out of the mean. But an empty recall answer for a question grounded in an *actually ingested* chunk is a memory failure and arguably should score 0, not vanish from the average. Note the real-question path does the opposite (empty context → `grounded=False`, explicitly reasoned as a verdict). This asymmetry could bias `overall_accuracy` upward precisely on the thin/degraded datasets the feature exists to flag. Suggest scoring 0 for empty-answer (reserving `None` for judge/infra errors), and exposing a `judged_question_count` in the document — `synthetic_question_count` currently counts unjudged rows too, so no consumer can see the true denominator.

**5. All-or-nothing persistence discards a full run's spend on partial failure.** Nothing is written until both `asyncio.gather` calls complete and `_persist_questions` commits. Any failure in between — including cancellation on shutdown or a commit error — sends the run to `ERRORED` and discards every completed answer and judgement (~$0.58 worth). The gathers also don't pass `return_exceptions=True`, so one escaped exception aborts both while the sibling's in-flight tasks keep burning tokens with nowhere to report the result. Suggest persisting incrementally, or at minimum passing `return_exceptions=True`.

**6. Two full-graph reads per run, with all node properties.** `_count_chunk_ids` calls `get_filtered_graph_data([{"type": ["DocumentChunk"]}])`; on the ladybug adapter this merges each node's full `properties` JSON (including chunk `text`) into the returned dict, and the caller keeps only the ids. `_embedded_nodes` then does a second full `get_graph_data()` over the whole graph. The docstring calls the chunk-count half "free" — it's free of LLM cost, not of I/O/memory, and on a large corpus this materializes the whole graph twice. Worth either an id-only count path or softening the "free" framing.

**7. Topics come from a capped, sampled subset (`SEMANTIC_NODE_CAP = 2000`, k ≤ 12) borrowed from the visualization pipeline.** Reusing that pipeline for *measurement* rather than *display* means topic coverage silently stops tracking corpus size past 2000 nodes, and since the over-cap sample is seeded on `len(ordered)`, the sampled set shifts as the graph grows — undercutting the run-to-run comparability that pinning the retriever/top_k was meant to protect. Worth calling out explicitly in the docs given the score's value is largely about being comparable over time.

**8. Zero generated questions still reports `COMPLETED`.** `generate_questions` skips topics best-effort; if every topic fails, the run finishes `COMPLETED` with `overall_accuracy=None`, `synthetic_question_count=0`, and no `floor_reason` — indistinguishable from a real "we measured and there's nothing to report" result. A distinct status or explicit reason would make this legible to the UI.

**9. No FK constraints.** `memory_score_questions.run_id` and `memory_score_runs.{dataset_id,tenant_id,triggered_by_user_id}` are bare UUID columns with no cascade — question rows orphan on run/dataset deletion. Worth aligning with neighboring models' conventions if there's an established one.

## Low

- `send_telemetry` fires before the tenant check, so 400s get logged as invocations.
- `request = payload` in `start_score_run` is a dead alias. The deferred `from ...methods import (...)` block is duplicated verbatim across all three handlers — could be hoisted to module scope if the deferral isn't there specifically to dodge a circular import.
- `POST /score` uses `response_model=None` and returns two different shapes depending on `run_in_background`, so its 200 response has no OpenAPI schema.
- In the "claim" path, `_update_run(run_id, dataset_id=dataset_id)` is a redundant round-trip — `create_memory_score_run` already wrote that value in the background-registration path.
- `find_active_memory_score_run(tenant_id: UUID)` and `run_memory_score(tenant_id: UUID)` are typed non-optional, but `_tenant_filter` and both docstrings explicitly support and rely on `None` (OSS/no-tenant case). Should be `UUID | None`.
- `groundedness_eval_system.txt` uses `•` bullets and looks pasted from a doc; other prompts in the repo use `-`. Minor style inconsistency, not a bug.
- `ScoredQuestion.created_at` is a per-row `datetime.now()`, so ties within a fast batch are likely and "oldest first" ordering isn't fully deterministic.

## Tests

Zero tests for 2383 lines. The "local env can't build ladybug" rationale doesn't cover the pure functions, which is also where the subtlest logic lives: `_allocate_counts` (sums to target, floor behavior, real-traffic weighting), `_aggregate_topics` (label-collision merge, `None` vs `0.0` semantics), `_build_context` (char budget), `_chunks_first`, `build_memory_score_document` (shape stability, ungrounded filter), `_tenant_filter`. Given findings 1–3 are security/cost-shaped, a couple of router-level tests (403/404 on an unpermitted dataset, cap enforcement on `synthetic_target`) would be the highest-value additions before merge.

---

**Overall:** the design reasoning is strong and the reuse is genuine — this isn't a case of reinventing wheels. Finding 1 (missing dataset permission check) is the one I'd treat as a merge-blocker, since it's a deviation from an established repo convention with real data-disclosure consequences, and the fix is small. Findings 2–3 are likewise small changes with outsized blast radius. Finding 4 is worth resolving before any UI surfaces `overall_accuracy`, since it affects the integrity of the one number this feature exists to produce.

LStromann:
## Addressed: findings 1–4 + tests (`5484c7c`)

Took the four you flagged as pre-merge. Findings 5–9 and the Low list are untouched and answered at the bottom.

### 1. Dataset ACL — fixed

`resolve_memory_score_dataset` now takes `requesting_user_id` and calls `get_specific_user_permission_datasets(user_id, "read", [dataset_id])`. Ordering matters, so: **tenancy first (404), permission second (403)** — checking permission first would turn a cross-tenant id into a 403 and confirm it exists.

Your note that "ACL is never consulted at either level" is the reason this can't be pushed down to the DB layer, and I've written that into the docstring so nobody later assumes isolation covers it.

**Extended beyond the POST**: fixing only the write path leaves the disclosure open, because `GET /score/{run_id}` and `/score/latest` serve the same `expected_answer` values for runs a colleague started. Both are gated now via a non-raising `readable_dataset_ids(user_id)`:

- `GET /{run_id}` → **404**, not 403, consistent with the handler's existing documented stance of never confirming an id exists.
- `GET /latest` → "latest" now means *the latest run the caller may read*, not the tenant's latest. Otherwise a colleague's newer run over a dataset this caller has no permission on lands at that URL.

Scheduled runs (`triggered_by_user_id=None`) skip the check — there's no caller to authorise and the trigger is the trusted party.

### 2. Unbounded cost inputs — fixed, both halves

`MAX_SYNTHETIC_TARGET = 500` (~1.15M tokens, ~$2.90) and `MAX_REAL_QUESTION_LIMIT = 200`, enforced in **both** places deliberately:
- API: `le=`, so out-of-range gets a **422** and the caller learns it was refused rather than silently clamped.
- Run: clamps whatever it's handed, so the cloud-backend scheduler and SDK callers that never pass through FastAPI validation are bounded too.

One constant each, imported into the request model, so the advertised limit can't drift from the enforced one — asserted by a test.

On `_allocate_counts` not being a strict ceiling: right, and your `synthetic_target=2` / 5 topics → 5 case is exactly it. A short budget now goes to the highest-real-traffic topics and the rest get 0 (skipped before their graph reads). Measuring fewer topics than asked is legible; generating more than was authorised isn't.

### 3. Tenant-wide query replay — fixed by deletion

Took your first option and scoped to the triggering user. That made `get_tenant_queries` redundant against the existing per-user `get_queries`, so the file is **deleted** rather than narrowed — −53 lines. Scheduled runs fall back to `dataset.owner_id`.

**The tradeoff, explicitly**: real-traffic topic weighting now reflects one member's usage rather than the tenant's, so two members scoring the same dataset can get slightly different topic splits. `overall_accuracy` stays comparable — it's synthetic-only with retriever and `top_k` pinned. Shout if you'd rather have the admin-gated tenant-wide read; I judged leaking search history the worse failure.

### 4. Accuracy denominator — fixed

Split the two cases that were collapsed into one branch:

- recall/judge **failed** → `score` stays NULL. Unmeasured; an outage isn't evidence about the memory.
- recall **succeeded** with an empty answer → **`score = 0.0`**. The question was generated from a chunk that's definitely in the graph, so this is the failure the score exists to measure.

Plus `judged_synthetic_question_count` in the document — I named it more explicitly than your `judged_question_count` since real questions are judged too, and it's unambiguously `overall_accuracy`'s denominator. A consumer can now tell "82% over 100" from "82% over 9".

### Tests — 36, and the "can't run pytest" claim was wrong

The local env **does** run pytest fine (`python3 -m pytest`, 3.14.5) — the ladybug blocker doesn't apply to these. My mistake in the PR body; corrected.

`cognee/tests/unit/api/v1/score/`:
- `test_memory_score_units.py` — `_allocate_counts` (sums to target, strict ceiling, real-traffic weighting, degenerate inputs), the measured/unmeasured distinction incl. *no judge call paid for on an empty answer*, `_aggregate_topics` label-collision merge and None-vs-0.0, document shape and ungrounded filter.
- `test_memory_score_router.py` — **403 on an unpermitted dataset with an assertion the run never starts**, caller identity actually reaching the resolver, 422 on each out-of-range param, advertised-vs-enforced caps, 404 on an unreadable run, `/latest` scoping.
- `test_memory_score_run_scoping.py` — per-user query read, owner fallback, in-run clamping of both params, dataset context entered as owner, gated run spending nothing.

They use `asyncio.run` rather than `pytest.mark.asyncio`, so they don't need the `pytest_asyncio` plugin (which is missing locally and breaks `test_session_visibility.py` on `dev` too — pre-existing, not mine).

Verification: `36 passed`. Full `cognee/tests/unit/api/` is 118 failed / 200 passed vs a **118 failed / 164 passed baseline on the same env with my changes stashed** — same 118 pre-existing env failures, +36 new passes, no regressions. `ruff 0.15.15` clean, `py_compile` clean, app OpenAPI builds, `POST /score` now documents 403.

### Deliberately not addressed

- **5 (all-or-nothing persistence)** — agreed, and worth noting `return_exceptions=True` isn't the one-liner it looks like: the row lists then contain exceptions that `_persist_questions` and `_aggregate_topics` would choke on, so it needs filtering plus a decision about incremental persistence. Medium-tier work, kept out of a security fix.
- **6, 7, 9** — accepted as written. 7 in particular (`SEMANTIC_NODE_CAP` sampling shifting as the graph grows) undercuts run-to-run comparability, which is most of the score's value, so it deserves its own change rather than a docstring caveat buried here.
- **8 (zero questions still COMPLETED)** — real legibility gap; wants a distinct status, which is a migration.
- **Low list** — took the three that were free and adjacent: telemetry moved after the tenant check on all three handlers, the dead `request = payload` alias gone, `UUID | None` on the three signatures that documented `None` support. Left the deferred-import duplication (measured: the methods package costs 0.00s once the router is imported, so the deferral buys nothing — but that's cosmetic churn on a +2383 diff), `response_model=None` on POST, the redundant `_update_run`, `•` bullets, and `created_at` tie-ordering.

Happy to fold 5–9 into this PR if you'd rather not split.

LStromann:
## Correction + a real bug, both found by running this against a live server (`c679502`)

I claimed 422 for out-of-range spend parameters in the comment above and in the PR body. **That was wrong — it's 400.** `client.py:173` maps every `RequestValidationError` to 400 app-wide. My router tests "passed" on 422 only because they built a bare `FastAPI()` and mounted the router on it, which has none of the app-wide handlers. Corrected, and the tests now run against the real app so they can't certify a status no caller sees.

### The bug that mattered

`from cognee.modules.search.operations import get_queries` binds the **submodule** — that package doesn't re-export the function. Every run over a dataset with any real question history died with `TypeError: 'module' object is not callable` and was persisted as `ERRORED`. Because the crash happens *before* the floor gate writes its verdict, even an empty dataset reported `ERRORED` instead of `SKIPPED_INSUFFICIENT_DATA`.

My 36 unit tests missed it because `test_memory_score_run_scoping.py` monkeypatches `get_queries` to observe which user it's called for — the stub replaced the broken reference with a working callable. A test now asserts every patched boundary is a real callable *before* any patching.

Fixed and re-verified live:

```json
{"status": "SKIPPED_INSUFFICIENT_DATA", "below_data_floor": true,
 "floor_reason": "Graph has 0 document chunk(s); 50 required.", ...}
```

### Live matrix — 19 cases, all as intended

| | Result |
|---|---|
| No tenant | 400 |
| Dataset absent | 404 |
| Cross-tenant dataset | 404 (not 403 — doesn't confirm existence) |
| **Tenant peer without read permission** | **403** ← finding 1, live |
| Caps exceeded / negative / missing `dataset_id` | 400 |
| At-cap values (500 / 200) | accepted |
| Empty graph | `SKIPPED_INSUFFICIENT_DATA` + `floor_reason` |
| Owner reads own run / `/latest` | 200 |
| **Peer reads owner's run** | **404** ← read-path gating |
| **Peer `/latest`** | **404** ← scoped to readable datasets |
| Concurrent run | 409, names the active run |
| Unauthenticated | 401 |

### Two things I found but did not change

**1. The endpoint is unreachable in stock OSS.** The default user has `tenant_id = None` (only `create_tenant` / `add_user_to_tenant` ever set it), so `POST /api/v1/score` always 400s on a fresh OSS install. Verified directly. This contradicts the module's own docs, which advertise OSS support — `_tenant_filter` handles the NULL-tenant case deliberately and `run_memory_score` accepts `tenant_id=None`, but the router rejects it before that code is reachable. Either the router should allow a null tenant or the docs should stop promising OSS.

**2. An `ERRORED` run tells the caller nothing.** `MemoryScoreRun.error` is populated but `build_memory_score_document` never returns it, so a client gets HTTP 200, `status: "ERRORED"`, and no diagnostic. Had the document surfaced `error`, bug 1 above would have been obvious from the first response instead of needing the server log.

Both are small. Say the word and I'll fold them in.

### Still untested

The happy path — generation → recall → judge → aggregate — needs an LLM key and a corpus above the 50-chunk floor (~$1 in spend). Everything above is the paths that don't spend tokens.

LStromann:
## Full end-to-end run against a live server — works, but two findings worth deciding on

Setup: 45k chars of cognee docs → `chunk_size=128` → **179 DocumentChunks** (floor is 50), 1456 nodes / 3975 edges, 5 seeded real questions (3 answerable from the corpus, 2 deliberately not). Blocking run, `synthetic_target=20`, `real_question_limit=5`. **2m57s.**

```
status: COMPLETED          overall_accuracy: 0.956
synthetic_question_count: 20   judged_synthetic_question_count: 20
real_question_count: 5     topics: 12     schema_defined: True
```

**What's confirmed working:** generation → recall → judge → aggregate end to end. Topics cluster and 4 of 12 picked up real traffic, so real-traffic weighting is live. Per-topic accuracies populate. `sum(synthetic_count) == 20` exactly against a target of 20 across 12 topics — the strict-ceiling fix holds where the old floor would have produced more. Judged count equals asked count. Question quality is reasonable (`"What is the Kuzu Dataset Database Handler described in the source text?"` → expected/actual match, 1.0). Score spread: 19 × 1.0 and one 0.12.

---

### 1. `ungrounded_real_questions` cannot detect the failure the PR says it's for

The PR justifies real questions as surfacing *named failures* — "6 questions your users actually asked couldn't be answered". I seeded two questions the corpus cannot possibly answer. Both came back **`grounded: true`**, and `ungrounded_real_questions` is `[]`.

```
q: "Who won the 1966 world cup final?"
answer:  "I can't answer from the provided context — it contains no information about the 1966 World Cup final."
grounded: true
reason:   "True — the provided answer accurately states that the retrieved context contains no information..."
```

Nothing is malfunctioning. The retriever honestly declined, and the groundedness judge correctly ruled that *the refusal itself* is supported by the context. But that means the metric measures **hallucination** (answer unsupported by context), while the feature is sold on **coverage gaps** (memory has nothing to say). Those are opposite failure modes, and `ungrounded_real_questions` will stay empty for exactly the customer the CTA targets — the one whose memory is too thin to answer.

A refusal-detection signal (did the answer decline / was retrieved context empty) is a different check from groundedness and would need to be its own field.

### 2. `schema_defined` is `True` after a plain cognify

The PR documents "Plain cognify → `False`" and calls the signal "deliberately conservative". This run was a plain cognify with no custom graph model and got `True`.

`_detect_schema_defined` excludes `_INGESTION_TYPES` plus the 187 `EntityType` names. Two inventory entries survive both filters:

```
TRIGGERING schema_defined: ['built-in group container', 'token-based lexical chunk search']
```

Neither is an `EntityType` name (no exact or case-insensitive match among the 187), so the exclusion set never had a chance at them. The signal is therefore a false-positive generator rather than a conservative one, and it's data-dependent — whether a given graph trips it depends on what the inventory happens to surface, so it will be flaky run to run and corpus to corpus. Any UI branching on "define a schema" will tell users who have no schema that they already have one.

### 3. Observation, not a defect: 0.956 on a 179-chunk corpus

Well above the data floor and still scoring 96%. The PR's own argument for the floor is that chunk-derived questions are trivially answerable from those same chunks — that effect clearly doesn't stop at 50 chunks. Worth a sanity check on whether this number will ever be low enough to motivate the upload it's meant to motivate, or whether the floor needs to be much higher / the questions harder (multi-hop, cross-chunk).

---

Findings 1 and 2 are both about signals the UI is meant to branch on, so they seem worth settling before the frontend work starts. Happy to take either on — 2 is a bug with a contained fix, 1 needs a product call on whether to add a separate refusal signal.

LStromann:
## Dispositions on the findings from local testing

Two of the three findings I raised above are **closed by author decision**, recorded here so review doesn't re-open them:

- **`schema_defined` returning `True` after a plain cognify** — not an issue. No change.
- **`ERRORED` runs returning HTTP 200 with no diagnostic** (`MemoryScoreRun.error` populated but never surfaced) — accepted, to be handled in separate work.

**Still open, and it needs a product decision rather than a fix:**

`ungrounded_real_questions` does not detect coverage gaps. Verified live — two questions the corpus cannot possibly answer both returned `grounded: true`, because the retriever honestly declined and the judge correctly ruled the refusal supported by context:

```
q:        "Who won the 1966 world cup final?"
answer:   "I can't answer from the provided context — it contains no information about..."
grounded: true
```

The metric therefore measures hallucination, while the PR justifies real questions as surfacing "6 questions your users actually asked couldn't be answered". Those are opposite failure modes, so that list stays empty for exactly the thin-memory customer the CTA targets. Detecting a refusal (or empty retrieved context) is a separate signal from groundedness and would need its own field. **The frontend should not build against `ungrounded_real_questions` as a coverage-gap indicator until this is settled.**

LStromann:
## Before/after on the same 179-chunk corpus, `synthetic_target=100` both runs

|  | before (`64f5dfd`) | after (`fc22870`) |
|---|---|---|
| questions generated | **59** / 100 | **90** / 100 |
| judged | 59 | 90 |
| `overall_accuracy` | 0.7954 | **0.9317** |
| source-referencing questions | 30 / 59 (51%) | **4 / 90 (4%)** |
| score histogram | `{0.0: 10, 0.33: 1, 0.5: 1, 0.6: 2, 0.9: 1, 1.0: 44}` | `{0.0: 3, 0.15: 1, 0.2: 1, 0.5: 1, 0.6: 1, 0.7: 2, 1.0: 81}` |

Per-topic allocation vs delivery:

```
                                    before          after
datasets, dataset_ids           20 ->  7        20 -> 20
cognee, cognee cloud            39 -> 15        39 -> 34
llm_api_key, start_ui()         11 -> 11        11 -> 11
none, ui, nodeset               20 -> 16        20 -> 15
graph_completion, knowledge…    10 -> 10        10 -> 10
```

### The accuracy went UP, and that needs saying plainly

0.795 → 0.932 is not the fix "improving the memory" — nothing about the graph changed. Zeros fell from 10/59 to 3/90, and the questions that disappeared are the source-referencing ones, which averaged 0.77 against 0.82 for self-contained ones. The old baseline was partly measuring **bad question generation** rather than bad recall.

Which surfaces something worth deciding on: **the generation prompt is now effectively part of the score's definition**, exactly like the pinned retriever and `top_k`. This commit moved the headline number by 13 points without touching a single graph node. Editing that prompt file later re-baselines every tenant's score and breaks comparability over time — the property the pinning in `run_memory_score.py` exists to protect. That constraint currently isn't written down anywhere, and the prompt lives in a shared directory where it doesn't look load-bearing.

### Still 90, not 100

The remaining shortfall is the honest case: two topics could not ground their full allocation from the chunks they have (34/39, 15/20). That is now logged per topic, and `judged_synthetic_question_count` reports the true denominator, so the gap is visible rather than silent. Worth a reviewer's eye on whether `synthetic_target` should be described as a target or a ceiling in the API docs.

Separately, and unchanged by any of this: **0.93 on a corpus 3.5x the data floor.** Single-hop questions drawn from one chunk and answered by retrieving that same chunk are close to tautological. If this number is meant to motivate uploading more data, it may need harder questions (multi-hop, cross-chunk) rather than a higher floor.

LStromann:
## Coverage added as a second KPI (`690de48`)

The gap I reported earlier is now measurable. Verified live on the same 179-chunk corpus, replaying 5 real questions — 3 answerable from the corpus, 2 not:

```
overall_accuracy: 0.945      <- unchanged, not blended
coverage:         0.6        <- 3 of 5 answered
unanswerable_real_questions: ["Who won the 1966 world cup final?",
                              "What is the tensile strength of galvanised steel cable?"]
ungrounded_real_questions:   []

  answered=False grounded=True   Who won the 1966 world cup final?
  answered=False grounded=True   What is the tensile strength of galvanised steel cable?
  answered=True  grounded=True   How does the MCP integration work?
  answered=True  grounded=True   What search types does cognee support?
  answered=True  grounded=True   How does cognee handle dataset permissions?
```

### Why one boolean couldn't carry it

`groundedness_eval_system.txt` **already** instructed the judge to mark refusals as not grounded. The judge overrode it, because "grounded means every claim traces to the context" is the more salient rule and a refusal's claim genuinely is traceable. The prompt was asking one boolean to mean two independent things.

They're now separate, from the same judge call — so the new signal costs one extra field, not extra calls:

- **`answered`** — coverage. Did the memory supply what was asked, or decline?
- **`grounded`** — hallucination. Was what it asserted supported?

A coverage gap is `answered=false, grounded=true`. A hallucination is the reverse. Collapsing them lost the former entirely.

### Why coverage is not folded into `overall_accuracy`

A thin memory scores **high** on accuracy and **low** on coverage — which is the inversion the data-floor gate exists to paper over, so this addresses the root cause rather than the symptom. It may be worth revisiting whether the 50-chunk floor still needs to be as blunt now that a 5-chunk dataset would honestly report ~95% accuracy / ~10% coverage instead of a misleading 95%.

Blended, 0.6 could mean either wrong answers or absent data — opposite remedies. Kept apart the pair maps straight onto the CTA: low coverage → upload more data; high coverage with low accuracy → the data is there and retrieval is failing.

### New response fields

`coverage`, `measured_real_question_count` (denominator), `answered_real_question_count` (numerator), `unanswerable_real_questions`, and `answered` on each question row. `ungrounded_real_questions` keeps its narrower meaning and should not be read as the coverage list.

### Notes for review

- The migration was **edited in place** rather than adding a new revision, since `d4f6a8b0c2e3` is unmerged. Re-verified from an empty database; single head. Anyone who already ran this branch locally needs the column added or the tables recreated — my own local DB hit exactly that and it is the only reason a run 500'd mid-testing.
- **Coverage is null when the tenant has no search history** — including a brand-new customer, who is precisely who the CTA targets. Coverage only becomes meaningful once there is traffic, so the UI needs a "not enough usage yet" state distinct from "coverage is bad".
- 46 unit tests, including that a perfect-accuracy/thin-coverage run reports 1.0 and 0.25 rather than one blended figure.

LStromann:
## Full run with everything in, `synthetic_target=100` (`690de48`)

179-chunk corpus, 5 replayed real questions, blocking run.

```
status: COMPLETED

ACCURACY  0.899      synthetic asked/generated/judged: 100 / 95 / 95
COVERAGE  0.6        real answered/measured/replayed:     3 /  5 /  5

unanswerable_real_questions: ["Who won the 1966 world cup final?",
                              "What is the tensile strength of galvanised steel cable?"]
ungrounded_real_questions:   []

score histogram: {0.0: 7, 0.2: 1, 0.33: 1, 0.4: 1, 0.7: 1, 0.8: 1, 1.0: 83}
source-referencing questions: 7 / 95
```

Per-topic allocation vs delivery — four of five topics now fill their allocation exactly:

```
topic                                  asked  got     acc  real
datasets, dataset, dataset_ids            20   20   0.795     1
cognee, cognee cloud, cognee.search       39   39   0.891     3
llm_api_key, start_ui(), session_id       11   11   1.000     0
none, ui, nodeset                         20   15   0.987     1
graph_completion, knowledge graph…        10   10   0.900     0
                                         100   95
```

### Progression across the three runs

| | before fixes | + batching/prompt | + coverage |
|---|---|---|---|
| questions generated | 59 / 100 | 90 / 100 | **95 / 100** |
| source-referencing | 30/59 (51%) | 4/90 (4%) | 7/95 (7%) |
| `overall_accuracy` | 0.795 | 0.932 | 0.899 |
| `coverage` | — | — | **0.6** |

### The number is not deterministic, and that matters for this feature

0.932 and 0.899 are the **same code on the same graph** — only the sampled questions differ, because generation is an LLM call. That is roughly ±3pp of run-to-run noise, and the 20-question run earlier swung between 0.956 and 0.983.

This PR pins the retriever and `top_k` specifically so a score cannot move without the memory changing. Question generation is the remaining unpinned input, and it moves the number by more than most real memory improvements would. Two consequences worth settling before the UI ships:

- **Don't render small deltas as movement.** A drop from 0.93 to 0.90 is indistinguishable from noise at this sample size.
- **Consider seeding generation deterministically**, or reporting a confidence interval alongside the point estimate. Neither is in this PR.

`coverage` has the opposite property — it is computed from a fixed replayed question set, so it *is* stable run to run. It is also the number that actually moves when a customer uploads more data, which makes it the better CTA trigger of the two.

Issue #4362, opened 2026-08-07
[Bug]: PGVectorAdapter never creates an ANN index — every similarity search is a full sequential scan


mckernanin:
### Bug Description

On pgvector, every collection cognee creates has **only the btree primary key**.
`create_vector_index` is a table-create, not an index-create:

```python
# cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py (v1.4.1)
async def create_vector_index(self, index_name: str, index_property_name: str):
    """Create the underlying index collection (table) for the given name/property pair."""
    await self.create_collection(f"{index_name}_{index_property_name}")
```

Nothing in the adapter ever issues `CREATE INDEX ... USING hnsw` (or ivfflat),
so `search()`'s cosine-distance query scans the whole table on every call.
Search latency grows linearly with graph size, and the constant is large:
ingestion makes the store bigger, which makes every subsequent search slower,
with no ceiling. (LanceDB gets ANN behavior from its own defaults — pgvector is
the backend where cognee has to opt in explicitly, and currently never does.)

### Steps to Reproduce

1. Run cognee v1.4.1 with `VECTOR_DB_PROVIDER=pgvector` and ingest any corpus.
2. Inspect the indexes on the created collections:

   ```sql
   SELECT t.relname, i.relname, am.amname
   FROM pg_index x
   JOIN pg_class t ON t.oid = x.indrelid
   JOIN pg_class i ON i.oid = x.indexrelid
   JOIN pg_am am ON am.oid = i.relam
   WHERE t.relname IN ('DocumentChunk_text','EdgeType_relationship_name', ...);
   ```

3. Observe every row is `<collection>_pkey | btree` — no hnsw or ivfflat index
   anywhere.
4. `EXPLAIN ANALYZE` any `search()` query and observe a sequential scan.

### Expected Behavior

Similarity search against a vector store should be approximate-nearest-neighbor,
not O(n): an HNSW index (`vector_cosine_ops`, matching the `1 -
cosine_distance` scoring in `search()`) created when a collection is created,
or exposed as an option.

**Caveat — just adding the index today breaks retrieval** (measured, details
under Additional Context): the `limit=None` full-score fetches in both
retrieval modes must become bounded first (a real `wide_search_limit` default,
or top-k + rescore). Also: HNSW caps at 2,000 dims, so dimension needs a check;
and building on an existing large table should use `CREATE INDEX CONCURRENTLY`
to avoid blocking writers.

### Actual Behavior

[...]

PR #4375, opened 2026-08-07
Fix: PGVectorAdapter never creates an ANN index

Ajaysingh-2003:
PGVector never actually indexes collections — `create_vector_index()` just creates a table, so every similarity search is a full sequential scan that gets slower as the graph grows (issue #4362). This adds a real HNSW index on new collections, and fixes the reason a naive index would have silently broken correctness: some retrieval code paths ask for `limit=None` expecting every row back, and an HNSW index would otherwise truncate that to ~40 results with no error. Existing collections get the index backfilled via the migration path.

## Acceptance Criteria

- New pgvector collections get an HNSW index automatically.
- `search(limit=None)` still returns every row, even with the index present.
- Existing collections (created before this change) get the index backfilled via `run_migrations()`.
- No other vector backend or retrieval code path is affected.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots
image (1)
image (2)
image

PGVector never actually indexes collections — `create_vector_index()` just creates a table, so every similarity search is a full sequential scan that gets slower as the graph grows (issue #4362). This adds a real HNSW index on new collections, and fixes the reason a naive index would have silently broken correctness: some retrieval code paths ask for `limit=None` expecting every row back, and an HNSW index would otherwise truncate that to ~40 results with no error. Existing collections get the index backfilled via the migration path.

## Acceptance Criteria

- New pgvector collections get an HNSW index automatically.
- `search(limit=None)` still returns every row, even with the index present.
- Existing collections (created before this change) get the index backfilled via `run_migrations()`.
- No other vector backend or retrieval code path is affected.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

## Pre-submission Checklist

## DCO Affirmation


cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py
```diff
@@ -506,6 +541,13 @@ async def search(
         )
         # Use async session to connect to the database
         async with self.get_async_session() as session:
+            if fetch_all:
```
linhongyu510, line 544: **P1: Preserve explicit limits larger than `hnsw.ef_search`**

`fetch_all` is only true for `limit=None`. With the new index, `search(..., limit=60)` still executes the HNSW path at its default `hnsw.ef_search=40`, so it can return about 40 rows even when at least 60 exist. That violates the same result-cardinality contract this block fixes for `None`; callers are allowed to pass arbitrary explicit limits (`wide_search_top_k` and `top_k_candidates + 1` flow through here).

The new 60-row test can expose this directly by adding a `limit=point_count` search before/after migration and asserting 60 results. Please either force an exact scan when the requested limit exceeds the effective ANN breadth, or set a transaction-local `hnsw.ef_search` high enough for the requested limit (while accounting for post-index filtering). Otherwise the PR replaces the current exact `LIMIT n` behavior with silent partial recall for explicit `n > 40`.

linhongyu510 (commented):
One remaining result-cardinality issue: the exact-scan fallback only covers `limit=None`, while the adapter accepts arbitrary explicit limits and the added test itself demonstrates that default HNSW search breadth is 40.

cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py
```diff
@@ -260,12 +264,41 @@ def __init__(self, id, payload, vector):
...
+    async def _create_hnsw_index(self, connection, collection_name: str, vector_size: int) -> None:
+        """Build the HNSW ANN index for a table just created in this same transaction.
+
+        The table is still empty here, so a plain (non-CONCURRENTLY) build is instant
+        and holds its lock for a negligible amount of time.
+        """
+        if not self._hnsw_index_enabled:
+            return
+
+        if vector_size > _HNSW_MAX_DIMENSIONS:
```
NMZivkovic, line 283: With the default embedding config this branch always runs. `EmbeddingConfig` defaults to `openai/text-embedding-3-large`, which resolves to 3072 dimensions (and `.env.template` suggests `EMBEDDING_DIMENSIONS=3072`). So someone on a stock setup who hits #4362 still gets a sequential scan, plus this warning every time a collection is created. It's also why the new test passes in CI, where `EMBEDDING_DIMENSIONS=300`, but fails at the first `hnsw_index_exists()` assert on the default config.

pgvector 0.7+ can index up to 4000 dimensions through `halfvec`, using an expression index:

```sql
CREATE INDEX ... ON "<collection>" USING hnsw ((vector::halfvec(3072)) halfvec_cosine_ops);
```

The catch is that `search()` has to order by the same expression (`vector::halfvec(3072) <=> $1::halfvec(3072)`), or the planner won't use the index. That's fine as a follow-up if you'd rather keep this PR smaller. In that case, please skip the index asserts in the test when the dimension is over 2000, and mention the limit in the PR description so people on the default model know what to expect.

cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py
```diff
@@ -506,6 +541,13 @@ async def search(
...
         # Use async session to connect to the database
         async with self.get_async_session() as session:
+            if fetch_all:
+                # An HNSW/ivfflat index turns "ORDER BY ... LIMIT <full count>" into an
+                # approximate, silently-truncated scan. Force an exact scan for this
+                # query so a caller that asked for every row actually gets every row.
+                await session.execute(text("SET LOCAL enable_indexscan = off"))
+                await session.execute(text("SET LOCAL enable_bitmapscan = off"))
+
             if node_name:
```
NMZivkovic, line 551: Related to @linhongyu510's comment, but this one hits small limits too. When the index is used, Postgres applies the `belongs_to_set` filter to the rows the HNSW scan returns, and the scan returns at most `hnsw.ef_search` (40) candidates. So if a NodeSet covers 10% of a collection, a `limit=15` search comes back with about 4 rows, and a small NodeSet can come back empty. The pgvector README covers this in its filtering section.

The graph-completion wide search (`brute_force_triplet_search`) already passes `limit=None` when `node_name` is set, so your `fetch_all` branch covers it. But the CHUNKS, RAG-completion, TRIPLET_COMPLETION and hybrid chunk retrievers pass `node_name` together with a `top_k`-style limit. So does `search_session_qa_ids` in `session_embeddings.py`, which scopes every session-history lookup with a per-session tag. That one picks a single session's rows out of every session stored in the collection, so it's the most likely to come back empty, and an empty result there just looks like there's no relevant history.

Two ways to handle it:
- Force the exact scan whenever `node_name` is set (the same `SET LOCAL` pair you use for `fetch_all`). Filtered searches then return what they return today.
- On pgvector 0.8+, `SET LOCAL hnsw.iterative_scan = strict_order` makes the index keep scanning until it has enough matches (up to `hnsw.max_scan_tuples`, 20,000 by default). That would also cover explicit limits above 40.

Raising `hnsw.ef_search` alone won't cover large limits, since it maxes out at 1000.

cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py
```diff
@@ -756,5 +798,82 @@ async def prune(self):
         await self.delete_database()
 
     async def run_migrations(self):
```
NMZivkovic, line 800: How this gets called on `dev` matters for the backfill. The migration runner only calls the adapter's `run_migrations()` when the Cognee version changes. It records the new version first, then syncs each dataset's vector database one at a time under the global migration lock, and the API lifespan waits for all of that before the server starts serving.

Two consequences:
- The first start after an upgrade builds HNSW indexes for every existing collection before the API comes up. On big collections that can take a while.
- If that start is interrupted, the next start sees the version as already recorded and skips the sync. Collections it hadn't reached stay unindexed until the next upgrade. A build that was cut off leaves an invalid index behind, which Postgres ignores for queries but still updates on every write. `cognee-cli upgrade` goes through the same version check, so it can't re-trigger this either.

This doesn't need a redesign in this PR, but please make it recoverable. Log each table's build at info (start and finish), log failures at warning with the exact `CREATE INDEX CONCURRENTLY` statement so an operator can run it by hand, and mention in the PR description that the first start after upgrading builds these indexes.

cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py
```diff
@@ -756,5 +798,82 @@ async def prune(self):
...
+                        text(
+                            f"CREATE INDEX CONCURRENTLY IF NOT EXISTS {quoted_index} "
+                            f"ON {quoted_table} USING hnsw (vector vector_cosine_ops)"
+                        )
+                    )
+                    indexed_tables.append(table_name)
+                finally:
+                    await connection.close()
+            except exc.SQLAlchemyError as e:
+                logger.debug("HNSW index migration skipped '%s': %s", table_name, e)
```
NMZivkovic, line 855: This is the failure path from my `run_migrations` comment. The runner logs the returned dict at info, so the table name does show up under `skipped_tables`, but the reason only goes to `debug`. Since nothing retries it until the next upgrade, I'd log the error at `warning` so an operator can see why it failed.

cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py
```diff
@@ -756,5 +798,82 @@ async def prune(self):
...
+
+        Without this, ``CREATE INDEX CONCURRENTLY IF NOT EXISTS`` sees the invalid
+        index as "already there" and never retries it, leaving the collection
+        permanently unindexed after a crash mid-build.
+        """
+        result = await connection.execute(
+            text(
+                "SELECT 1 FROM pg_index i "
+                "JOIN pg_class c ON c.oid = i.indexrelid "
+                "WHERE c.relname = :index_name AND NOT i.indisvalid"
```
NMZivkovic, line 874: Small one for the rebase. `dev` now has a schema-per-dataset mode, where datasets share one database and each gets its own schema, and index names only have to be unique within a schema. `relname` alone can match an invalid index in another dataset's schema. The `DROP INDEX` below then drops this schema's valid index, and it gets rebuilt for nothing. Resolving the name through the search path avoids that:

```sql
SELECT 1 FROM pg_index WHERE indexrelid = to_regclass(:index_name) AND NOT indisvalid
```

(pass the name double-quoted, since collection names are mixed case).

NMZivkovic (commented):
Thanks for taking this on, @Ajaysingh-2003. #4362 is a real problem, and this goes further than a naive index would: the `limit=None` exact scan, building concurrently for existing tables, and cleaning up invalid indexes after an interrupted build are details people usually miss.

Before it can go in:

1. **Rebase on `dev`.** The branch has conflicts. The adapter has changed since you branched (schema-per-dataset mode, `score_by_ids` from #5038, typing and lint rewrites), but `search()` and `run_migrations()` still have the same shape, so it should be mostly mechanical.
2. **The default embedding model doesn't get an index.** Out of the box it's 3072 dimensions, over the 2000 limit (inline).
3. **Result counts.** +1 to @linhongyu510's comment on explicit limits. It's the default path too: `wide_search_top_k` defaults to 100 in the graph-completion retrievers, so once the index exists those searches get at most 40 rows per collection. NodeSet-filtered searches have a related problem (inline).
4. **Backfill recovery.** A backfill that fails or gets interrupted isn't retried until the next Cognee upgrade (inline on `run_migrations`).

On the red checks: those workflow runs never started any jobs, so they aren't about your change. The pgvector suite also needs API secrets, so it can't run from a fork. Once it's rebased we can run it on our side.

dexters1:
Thanks for the PR @Ajaysingh-2003! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

Ajaysingh-2003:
Heyy @dexters1 . Thanks for the clear heads up and informing me about this. Take your time, if any changes are required, i will be avaliable. Thanks!

PR #4391, opened 2026-08-08
fix: Share Docker storage as one non-root user

dexters1:
## Description

Fixes the storage-ownership failures reported on Discord: the main image ran as **root** while `cognee-mcp` runs as **uid 1000**, and both bind-mounted `./cognee`, whose `.cognee_system` doubled as the persistence location.

The ownership arithmetic made every mixed-uid deployment converge to broken, regardless of start order:
- **Main first**: storage created root-owned → the MCP server (uid 1000) cannot write the SQLite file or create anything in the databases dir.
- **MCP first**: works initially (root ignores ownership) — but every *new* filesystem object the root container creates (per-dataset database dirs, SQLite `-wal`/`-shm` sidecars) lands root-owned, and MCP progressively loses write access to exactly the newest data.
- Bonus failure mode: persistence inside the source bind mount pollutes the host checkout with root-owned files and is host-uid-sensitive (the reporter's host uid was 1002).

## Change

- Main image now runs as the same `cognee` uid/gid-1000 user the MCP image already defines (`COPY --chown` in one layer, writable `HOME=/app`, no more root-by-default).
- Default storage moves **outside the source tree** to `/cognee-storage/{system,data}`, baked into both images cognee-owned — so a fresh named volume mounted there initializes with the right ownership.
- docker-compose mounts shared named volumes (`cognee_system`, `cognee_data`) on **both** services: shared memory between the API and the MCP server is the point of running both, now by design instead of by accident of the bind mount. The `./cognee` bind mount goes back to being dev-reload only.
- `.env` mounts are `:ro` (Discord bug #2), and the main service gets the same `DB_*` passthrough the MCP service already had, with a comment on wiring `--profile postgres` (`DB_HOST=postgres`).

## Migration note for existing deployments

Data previously written under `./cognee/.cognee_system` stays on the host; copy it into the named volume or override `SYSTEM_ROOT_DIRECTORY`/`DATA_ROOT_DIRECTORY` to keep the old location.

## Related

Complements #4390 (Docker/MCP migration-path fixes) — disjoint files, either merge order works.

## Testing

Docker build + boot verification in both start orders and cross-container write checks — results being posted as a follow-up comment.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -81,10 +81,15 @@ WORKDIR /app
 # separate `chown -R /app` would copy the whole tree up into a second layer.
 # /cognee-storage is baked into the image cognee-owned so a fresh named
 # volume mounted there initializes with the right ownership.
+# ``chown cognee /app`` (the directory inode only): WORKDIR created /app as
+# root, and ``COPY --chown`` sets ownership on the copied content, not the
+# pre-existing target dir — without this the non-root user cannot create
+# ``$HOME/.lbdb`` and the build-time extension pre-install silently fails.
 RUN groupadd --system --gid 1000 cognee \
     && useradd --system --uid 1000 --gid cognee --no-create-home --shell /usr/sbin/nologin cognee \
     && mkdir -p /cognee-storage/system /cognee-storage/data \
-    && chown -R cognee:cognee /cognee-storage
+    && chown -R cognee:cognee /cognee-storage \
+    && chown cognee:cognee /app
 
 COPY --from=uv --chown=cognee:cognee /app /app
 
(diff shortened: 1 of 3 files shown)


dexters1:
**Docker verification results** — both images built from this branch merged with #4390, full matrix:

```
=== TEST 1: fresh boot, main service first ===
PASS: server healthy
PASS: runs as cognee (not root)
PASS: storage owned by uid 1000
PASS: alembic stamped at head on fresh DB
PASS: fresh-stamp path taken (no history replay)
PASS: no .cognee_system polluting the source tree
=== TEST 2: MCP joins the shared volume ===
PASS: mcp runs as cognee
PASS: mcp can WRITE the shared databases dir
PASS: mcp-side migrations run clean (existing path)
=== TEST 3: order flip — MCP first on fresh volumes ===
PASS: MCP-first creates+stamps fresh DB
PASS: main healthy on MCP-created storage
PASS: main can write MCP-created databases dir
PASS: main took existing path (no double stamp)
=== RESULT: 13 passed, 0 failed ===
```

The verification surfaced two additional real bugs, fixed in `ac2c91912`:
1. `WORKDIR` creates `/app` as root and `COPY --chown` does not change the pre-existing target dir — the non-root user couldn't create `$HOME/.lbdb`. **In the MCP image this had been silently breaking the build-time JSON-extension pre-install since it was added** (the `||` fallback ate the error).
2. The main image never pre-baked the graph JSON extension (as root it re-installed at runtime every boot); non-root runtime installs race between graph workers and fail health checks. It now uses the same build-time pre-install as the MCP image.

PR #4413, opened 2026-08-09
fix(security): scope visualization events by dataset

### LINKED ISSUE #4411: [Security]: Visualization live-events can leak QA content across dataset permissions
﻿## Summary

The new GET /v1/visualize/live-events endpoint can expose session questions and answers from sessions unrelated to the dataset used for authorization.

cognee/api/v1/visualize/visualize.py:391-395 verifies read access to the requested dataset_id, but then calls collect_session_events(user=user) without passing a dataset/session scope. collect_session_events() lists recent sessions for the user (cognee/modules/visualization/session_events.py:122-133) and maps each QA entry?including its question and answer at :40-54?into the response.

## Reproduction

1. Create two datasets, A and B, under a tenant/user setup where the caller can read A but cannot read B.
2. Use a session associated with B to create a search/QA entry containing distinctive private text, for example private-question-B and private-answer-B.
3. Request GET /v1/visualize/live-events?dataset_id=<A>&since=... as the caller who can read A but not B.
4. Observe that the response can contain the B session question, answer, and graph element IDs, provided that session belongs to the caller and is among the recent sessions scanned.

## Expected behavior

The endpoint should return only events that are authorized for the requested dataset (or explicitly document and enforce a broader authorization scope that covers every session included).

## Actual behavior

Dataset A acts only as an access gate; the returned event list is user-wide. The endpoint documentation says the events are "the caller's own", but ownership of a session is not equivalent to read access to every dataset whose content may appear in that session.

## Why it matters

Questions, generated answers, feedback text, and graph IDs can contain sensitive tenant or dataset content. A user with access to one dataset may receive private content from another dataset through a visualization polling endpoint. Filter events by authorized dataset/session membership before serialization, or avoid returning content that cannot be tied to the requested dataset, and add a multi-dataset authorization regression test.

## Related audit context

The current unit tests cover cursor behavior with mocked collect_session_events, but do not exercise dataset-scoped filtering or cross-dataset authorization.

abhinavkr26104:
﻿## Summary

Fixes #4411.

Scope visualization session events to the authorized dataset instead of returning every recent session event for the user.

## Changes

- Add dataset filtering to session lifecycle session lookup.
- Propagate SessionManager.dataset_id into lifecycle activity records.
- Pass the authorized dataset into visualization JSON/HTML event collection and live-events polling.
- Exclude unscoped/legacy session records from dataset-scoped event responses.
- Add route-level and collector-level regression coverage, including filtering of explicitly supplied out-of-scope session IDs.

## Security impact

A caller with access to dataset A can no longer receive QA events collected from a session recorded for dataset B through the visualization live-events endpoint.

## Verification

- python -m compileall -q on all changed Python files
- git diff --check
- Focused pytest could not run because pytest is not installed in the local environment.

### DIFF
--- a/cognee/api/v1/visualize/visualize.py
+++ b/cognee/api/v1/visualize/visualize.py
@@ -87,7 +87,11 @@ async def fetch_visualization_data(
     if include_session_events:
         from cognee.modules.visualization.session_events import collect_session_events
 
-        search_events = await collect_session_events(user=user, session_ids=session_ids)
+        search_events = await collect_session_events(
+            user=user,
+            session_ids=session_ids,
+            dataset_id=getattr(dataset, "id", None),
+        )
 
     return graph_data, search_events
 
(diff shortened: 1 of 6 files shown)


ArmagedonFlamer:
Thanks @abhinavkr26104 for picking up #4411 and for the regression tests around the live-events endpoint. The same fix has since landed on dev in #4843, which scopes session events for /visualize/json and the live-events endpoint to the authorized dataset. It also matches sessions recorded before dataset attribution existed. Since your branch now conflicts with that work in every file it touches, we're closing this one. Thanks again, and we welcome more PRs from you.

PR #4416, opened 2026-08-09
fix(validate): reject unauthorized dataset fallback

### LINKED ISSUE #4414: [Security]: validate endpoint falls back to unscoped graph for unauthorized or missing datasets
## Summary
`/api/v1/validate` can run against an unscoped/default graph when the requested dataset does not exist or the caller has no read access.

## Reproduction
1. Enable backend access control and create dataset A owned by user A.
2. Authenticate as user B, who has no read permission for A.
3. Request `GET /api/v1/validate?dataset=A` (or pass a nonexistent dataset name).
4. Mock or inspect the graph/vector adapters and observe that validation still proceeds instead of rejecting the dataset.

The fallback is in `cognee/api/v1/validate/validate.py:221-229`: `resolved_dataset` remains `None` when `get_authorized_existing_datasets()` returns no match, but the function still enters `set_database_global_context_variables(None, None)` and calls `get_graph_engine()` / `get_graph_data()`.

## Expected behavior
A missing or unauthorized dataset should return the same authorization/not-found error used by other dataset-scoped endpoints, and no graph/vector reads should occur.

## Actual behavior
The request returns a validation report for whatever backend context is selected with `None` (typically the default/unscoped graph). The response can expose node counts, type distributions, and integrity issue details from data outside the requested dataset. With multiple requested names, inaccessible names are silently ignored if another name is authorized.

## Why it matters
This bypasses the dataset access check for a diagnostic endpoint and can disclose cross-tenant metadata or integrity details. It also makes `validate(dataset=missing)` appear successful while validating the wrong store.

## Suggested fix
Require exactly the requested dataset(s) to resolve to authorized dataset objects before opening the database context. Raise the standard permission/not-found error when the result is empty (and define explicit behavior for mixed authorized/unauthorized lists), with an API regression test asserting no adapter calls on denial.

abhinavkr26104:
## Summary

Fixes #4414.

The validator now fails closed when any explicitly requested dataset is missing or not readable. It raises the standard `DatasetNotFoundError` before opening graph/vector adapters, preventing fallback to an unscoped or default graph.

## Changes

- Require every requested dataset to resolve to an authorized dataset.
- Preserve first-dataset selection when all requested datasets are authorized.
- Add regression tests for fully unauthorized and mixed authorized/unauthorized requests.

## Verification

- `python -m compileall -q cognee/api/v1/validate/validate.py cognee/tests/unit/api/v1/validate/test_validate.py`
- `git diff --check`
- `python -m pytest cognee/tests/unit/api/v1/validate/test_validate.py -q` - 18 passed, 1 skipped.

### DIFF
--- a/cognee/api/v1/validate/validate.py
+++ b/cognee/api/v1/validate/validate.py
@@ -42,6 +42,7 @@ from cognee.infrastructure.databases.vector.get_vector_engine import get_vector_
 from cognee.infrastructure.engine import DataPoint
 from cognee.modules.chunking.models import DocumentChunk
 from cognee.modules.data.constants import DEFAULT_DATASET_NAME
+from cognee.modules.data.exceptions import DatasetNotFoundError
 from cognee.modules.data.methods import get_authorized_existing_datasets
 from cognee.modules.engine.models import Entity, EntityType
 from cognee.modules.users.methods import get_default_user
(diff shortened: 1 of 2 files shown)


dexters1:
Thank you for your contribution! Your PR was merged through this branch:

https://github.com/topoteretes/cognee/pull/4455

We needed to run CI for the change and make some formatting changes

PR #4419, opened 2026-08-09
fix(recall): reject empty schema unions

### LINKED ISSUE #4418: [Bug]: Empty response-schema unions raise TypeError instead of HTTP 422
﻿## Summary

The HTTP recall `response_schema` converter does not reject empty unions. A schema containing `anyOf: []` reaches `typing.Union[tuple()]`, which raises a raw `TypeError` instead of the documented `CogneeValidationError`/HTTP 422 response.

## Reproduction

Send a recall request with a structurally valid object root and an empty `anyOf` property:

```json
{
  "query": "hello",
  "response_schema": {
    "type": "object",
    "properties": {
      "value": {"anyOf": []}
    },
    "required": ["value"]
  }
}
```

The failure is in `cognee/modules/recall/methods/model_from_json_schema.py:94-101`, where the empty member list is passed to `Union[tuple(members)]`. The same issue applies to a JSON Schema `type: []` list at lines 103-107.

## Expected behavior

Invalid or unsupported schemas should raise `CogneeValidationError`, which the recall API maps to a 422 validation response, without starting the recall pipeline or invoking an LLM.

## Actual behavior

Python raises `TypeError: Cannot take a Union of no types.` This bypasses the converter's validation-error contract and can surface as an internal server error/500.

## Why it matters

A malformed client request can produce an incorrect server error and may trigger generic error handling, confusing API clients and obscuring the actual input problem. It also violates the documented supported-schema boundary.

## Suggested fix

Explicitly reject empty `anyOf` and empty type lists with `_fail(...)`, and add converter/router regression tests asserting a 422-style `CogneeValidationError`.

abhinavkr26104:
## Summary

Fixes #4418.

Reject empty JSON Schema unions as client validation errors instead of allowing Python to raise `TypeError: Cannot take a Union of no types.`

## Changes

- Add a shared union-construction helper that rejects empty member lists with `CogneeValidationError`.
- Apply it to both `anyOf` and JSON Schema `type: []` forms.
- Add regression tests for both malformed schemas.

## Verification

- `python -m pytest cognee/tests/unit/modules/recall/test_model_from_json_schema.py -q` — 15 passed.
- `python -m compileall -q cognee/modules/recall/methods/model_from_json_schema.py cognee/tests/unit/modules/recall/test_model_from_json_schema.py` — passed.
- `git diff --check` — passed.

The targeted Ruff run reports existing UP007/UP045 modernization findings in this module/test; the change introduces no new import or formatting errors.

### DIFF
--- a/cognee/modules/recall/methods/model_from_json_schema.py
+++ b/cognee/modules/recall/methods/model_from_json_schema.py
@@ -39,6 +39,15 @@ def _fail(message: str) -> CogneeValidationError:
     return CogneeValidationError(message=f"response_schema: {message}")
 
 
+def _union_from_members(members: list[Any], source: str) -> Any:
+    """Build a type union while preserving schema errors as API validation errors."""
+    if not members:
+        raise _fail(f"{source} must contain at least one schema")
+    if len(members) == 1:
+        return members[0]
+    return Union[tuple(members)]
+
+
 class _Budget:
     """Caps total property count across the whole schema."""
 
@@ -97,7 +106,7 @@ def _type_from(
             _type_from(member, defs, depth + 1, in_flight_refs, budget)
             for member in schema["anyOf"]
         ]
-        return Union[tuple(members)]
+        return _union_from_members(members, "anyOf")
 
     schema_type = schema.get("type")
     if isinstance(schema_type, list):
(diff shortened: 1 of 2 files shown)


dexters1:
Thanks for the PR @abhinavkr26104! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #4433, opened 2026-08-11
feat: Add TELEMETRY_ORIGIN to telemetry events

dexters1:
## What

Every telemetry event now carries a `telemetry_origin` property, read from the
`TELEMETRY_ORIGIN` env var and defaulting to `"sdk"` when unset.

```python
telemetry_origin = os.getenv("TELEMETRY_ORIGIN", "sdk")
# added to the event `properties` payload in send_telemetry()
```

## Why

So telemetry can be segmented by where it originates. Managed deployments (the
cloud) set `TELEMETRY_ORIGIN=cloud`; everything else (local SDK usage) reports
the default `"sdk"`. Makes tracking telemetry origin straightforward.

## Notes
- Read directly via `os.getenv`, matching how `send_telemetry` already reads
  `TELEMETRY_DISABLED` / `ENV`. No config-schema change, no new required setup.
- Placed before `**additional_properties`, so a per-call override remains possible.
- Test added: defaults to `"sdk"` when unset, honored as `"cloud"` when set.

## Verification
ruff clean; `cognee/tests/unit/shared/test_telemetry_tracking.py` passes (5 passed).

### DIFF
--- a/cognee/shared/utils.py
+++ b/cognee/shared/utils.py
@@ -266,6 +266,10 @@ def send_telemetry(event_name: str, user_id: str | UUID, additional_properties: 
     anonymous_id = str(get_anonymous_id())
     persistent_id = str(get_persistent_id())
     api_key_tracking_id = _get_api_key_tracking_id()
+    # Where this telemetry event originates. Defaults to "sdk"; deployments such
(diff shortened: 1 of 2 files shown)


NMZivkovic (dismissed):
Add a test, but looks good.

PR #4435, opened 2026-08-11
refactor(data): dataset-scoped Data rows — dedup as lookup, stable data ids

dexters1:
## What

`Data.id` was `uuid5(content_hash, user)` — identity WAS the content, so the same bytes in every dataset shared one row, in-place updates bled across datasets, and changed content always meant a changed id. This PR completes the full dataset-scoping of the Data table:

**1. Dataset-scoped rows, dedup as lookup** (`4635b5e7f`): `Data.dataset_id` names the owning dataset; ids are random `uuid4`, stable forever; dedup is a lookup on `(dataset_id, owner, tenant, content_hash)`. Same content in two datasets → two documents with independent futures. Re-adds stay idempotent.

**2. Owner/tenant scoping** (`3f187edfc`): dedup probes filter by owner and tenant, matching the old identity's scope and the tenant-scoped file storage — in shared multi-writer datasets, another user's identical bytes never resolve to (or steal ownership of) your row.

**3. Backfill + retire `DatasetData`** (`05279b9cb`): membership IS `Data.dataset_id`. The startup-run migration backfills legacy rows: **sole-membership rows keep their original `data_id`** (only stamped — user-held id mappings stay valid); multi-membership rows keep the original id for the *oldest* membership while other datasets get fresh rows with `legacy_id` provenance and repointed ledger rows; `dataset_data` is dropped. All membership-join consumers rewritten; dataset deletion removes its rows with `raw_data_location`-refcounted file cleanup.

**4. E2E contract alignment** (`4ea332cf0`): deduplication/legacy/label suites updated to the scoped semantics.

**5. Every id ever issued keeps resolving** (`68d18b625`, `170d8b445`): the backwards-compatibility layer for fork rows (the multi-membership splits from step 3):

- `legacy_id` (indexed) records the pre-fork original — only the backfill split ever sets it, so alias chains cannot form. Ingest resolves pinned `DataItem` ids through the same exact-then-legacy lookup.
- **`update()` keeps the document's `data_id`**: the replacement is re-ingested pinned to the resolved id (exact, or the pre-fork `legacy_id`), and fork lineage is restored onto the recreated row — externally held id mappings never move. The old design churned the id on every update because identity was the content hash; that is gone. Multi-item lists are refused (update replaces exactly one document).
- `resolve_data_id(dataset_id, data_id)`: exact id first, then `legacy_id` within the dataset — deterministic, used by delete/update/ingest.
- `get_data` without dataset context refuses a forked id with `AmbiguousDataIdError` (HTTP 409) listing every candidate `(dataset_id, dataset_name, data_id)` instead of silently returning another dataset's document; an id with a single surviving fork resolves directly.
- **dlt ids are dataset-scoped**: dlt derived pin ids from (row identifier, user) only, so the same source loaded into two datasets collided on the foreign-pin guard. The derivation now includes the target dataset (independent stable id families per dataset), with a batched adoption probe that keeps pre-scoping rows' ids stable in their own dataset — existing dlt deployments upgrade without re-minting or orphan-cleanup churn.
- Chain migration `rekey_fork_document_ids` (auto-runs per database pair at startup) moves fork documents' identity to the canonical id in every store: relational ledger slug/edge-endpoint references, chunk `document_id` properties in the graph **and in the vector index payloads** (re-upserted through `index_data_points`, cognify's own write path, so the row shape matches production on every vector backend — chunk point ids never change), the document node itself with full property + provenance preservation, and — on graph-provenance backends, which never write the ledger and delete through source-ref keys embedding the data id — every node/edge source ref, attach-then-remove so an interrupted run converges on re-run. `downgrade()` fully reverses all of it. On shared stores (access control off), fork pairs whose pre-fork id is still owned by a live keeper row are provenance-only — the shared node, ledger, chunks, and payloads stay with the keeper, and refcounted deletion handles the rest; multi-fork legacy ids get one deterministic claimant. After it runs, fork documents are reached by canonical id with no legacy awareness in the hot paths.

## Compatibility contract

- **Every `data_id` ever issued keeps resolving** — for sole-membership rows the id is unchanged; for fork rows the pre-fork id resolves via `legacy_id` in every dataset-scoped API (delete, update, ingest pinning, `get_data` with dataset context, raw download). And the id is now **stable through `update()`** — the old system churned ids on every update, so mapped-id users strictly gain.
- Context-free `get_data` of a forked id is the one call that cannot be answered safely — it now fails loudly (409 with candidates) rather than guessing.
- Client-side id *derivation* (computing `uuid5(content…)`) no longer works for new adds — deliberately: derivable ids are the aliasing bug. Store the `data_id` returned by `add()`; it now survives updates.
- `ENABLE_AUTO_MIGRATIONS=false` deployments must run `cognee-cli upgrade` before this version touches data. There is deliberately **no runtime fallback** for un-migrated stores: deleting fork data there fails fast with `DocumentSubgraphNotFoundError` instead of mutating stores the migration has not touched.
- The raw-data download endpoint now resolves pre-fork ids like every other dataset-scoped API (it previously matched exact ids only).
- Ships in **v1.5.0** (chain migration version-stamped accordingly).

## Testing

- **The backfill is set-based and scale-tested**: one `UPDATE … FROM` stamps all sole-membership rows (a 50k-row + 1,200-fork run completes in ~1.2s; the original per-row loop was effectively non-terminating on SQLite due to a UUID type-affinity trap that degraded correlated subqueries to per-row table scans), fork rows are copied server-side, the migration opens its own transaction on SQLite (alembic's context there is non-transactional), and boot logs carry pre-flight/completion counts. Verified end to end against a genuine locally-seeded v1.2.0 database through the full compat flow — including a large-scale ROUND TRIP: 1,082 docs / 65 shared rows seeded by real v1.2.0 (two datasets cognified with real LLM), upgraded (76 forks, exact accounting, 141 pre-fork-id resolutions), then downgraded (`cognee-cli downgrade base --alembic <v1.2.0 head>`) and driven by real v1.2.0 again: search, cognify of a new document, and delete all work on the downgraded store.
- `test_dataset_scoped_backfill.py`: builds a genuine pre-migration DB (legacy NULL rows, hand-created `dataset_data`, ledger entries) and runs the real `upgrade()` via alembic `MigrationContext` — pins sole-id preservation, keeper selection, split provenance, ledger repoint, table drop. A downgrade test pins the reverse: `dataset_data` recreated with one membership per scoped row (data-preserving, never id-merging), `legacy_id` dropped — with the documented consequence that fork lineage cannot survive a downgrade (the old schema has nowhere to keep it).
- `test_dataset_scoped_data.py` (~3s, no LLM): per-dataset identities; the bleed regression (rewrite alpha in place → beta byte-identical); idempotent re-add; owner-scoped dedup; foreign-pin refusal; resolver exact/legacy checks; the `get_data` guard matrix (unforked, single-fork, scoped, ambiguous-409 with candidate set); pinned pre-fork ids binding to the canonical row; `update()` id stability (fork row updated by pre-fork id keeps its canonical id with lineage restored; normal row keeps its id; multi-item list refused); delete by pre-fork id through the real API.
- `test_rekey_fork_migration.py` (~8s, no LLM): cognifies a real document, simulates the backfill fork relationally, runs the real chain migration — asserts the graph node re-key with properties intact, chunk `document_id` follow-up in graph and vector payloads, idempotency, and end-to-end deletion by the **pre-fork** id leaving zero stranded graph nodes (with a probe asserting the delete enters the graph by the canonical id only); dataset-scoped vs global ledger-reference updates for ledger backends; the full downgrade round trip — migrate → downgrade restores the exact pre-migration store state (graph node, chunk properties, vector payloads, provenance refs), downgrade is idempotent, re-migrate lands canonical again and still deletes cleanly by the pre-fork id; and the shared-store keeper scenario — a subgraph carrying two datasets' provenance, migrate twice, keeper's node/chunks/payloads untouched with only the fork's provenance keys moved, then refcounted deletion dataset by dataset down to zero stranded nodes.
- `test_update_id_stability.py` (~5s, no LLM): the id-stable `update()` contract end to end — same `data_id` with the full stack (graph doc node, chunk `document_id` in graph and vector payloads) rebuilt under it and old chunks gone; repeated + unchanged-content updates; duplicate-content update does not collapse documents; unknown id creates under that id; list handling; fork lineage survives updates by either id; teardown by pre-fork id with zero stranded nodes.
- **Backwards-compat CI now exercises the fork split on real legacy data**: Phase 1 (run under cognee v1.2.0) adds identical content to two datasets — one genuinely shared legacy row; Phase 2 (this branch) verifies the backfill split it into keeper + fork, the pre-fork id resolves in both datasets via dataset-scoped `get_data`, context-free lookup raises `AmbiguousDataIdError` with both candidates, and the delete sweep removes fork rows **through their pre-fork id**. Runs on both compat stacks (SQLite+LanceDB+Kuzu, PostgreSQL+PGVector+Kuzu), covering the chain migration's dataset-scoped mode on real legacy data.
- Delete e2e scripts pass end-to-end through the real startup migration chain (which caught a multiple-heads mistake during development — the only local test layer that exercises it).

## Notes

- Incremental-update integration (shared-row precondition + update-path bleed tests) follows on #4241 separately.
- Draft until a Linear key is assigned. Known-inherited unit reds fail identically on dev.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


cognee/tests/unit/modules/data/test_dataset_scoped_backfill.py
dexters1, line 1: This is a test for the one-time migration, it should not be part of the final PR and merge.

I kept it here if anyone is interested in how the migration was verified

cognee/tests/integration/tasks/test_rekey_fork_migration.py
dexters1, line 1: This is a test for the one-time migration, it should not be part of the final PR and merge.

I kept it here if anyone is interested in how the migration was verified

PR #4436, opened 2026-08-11
feat(recall): Log recall queries to search history (SDK-401)

LStromann:
Recall answered questions without recording them.

`log_query` is called from exactly one place in the codebase — `cognee/modules/search/methods/search.py:94` — and recall calls `authorized_search()` directly, skipping the logging that `search()` wraps around that call. Only `/v1/search` traffic reached the `queries` table.

Agents recall through `/v1/recall` (every cloud integration card hands users `curl -X POST {{BASE_URL}}/api/v1/recall`, and cloud's own `recallKnowledge` deliberately uses `/v1/recall` over `/v1/search`), so agent questions were absent from recall history entirely. That leaves nothing for SDK-393's `dataset_id` to attribute, and blocks the agent-scoped recall coverage analysis in COG-6067.

## What changed

`cognee/api/v1/recall/recall.py` — one `log_query` call in the graph path, before `authorized_search()`:

* Records the **resolved** `local_query_type.value`, so a router-picked type is logged as itself and history stays consistent with `/v1/search`.
* Reuses `search()`'s `_single_dataset_id()` rather than reimplementing it, so the two endpoints can never disagree about when a question belongs to exactly one dataset.
* Wrapped in try/except with a warning. A failed history insert must not cost the caller their answer.

No `log_result` — recall's per-dataset answers are a separate concern, since `Result.value` is currently a JSON array of unlabelled completions.

## Tests

`cognee/tests/unit/api/recall/test_recall_logs_query_history.py` covers the three behaviours: a single-dataset recall logs question + type + user + dataset; a fan-out logs `dataset_id = NULL`; a failing insert still returns results.

`57 passed` across `cognee/tests/unit/api/recall` and `cognee/tests/unit/modules/search`.

Fixes SDK-401

### DIFF
--- a/cognee/api/v1/recall/recall.py
+++ b/cognee/api/v1/recall/recall.py
@@ -670,7 +670,9 @@ async def recall(
                     local_query_type.value,
                     user.id,
                     graph_results,
-                    _single_dataset_id(search_dataset_ids),
+                    # Each row takes its dataset from its own payload; this is
+                    # only used for payloads that carry none.
+                    fallback_dataset_id=_single_dataset_id(search_dataset_ids),
                 )
             except Exception as log_error:
                 logger.warning(f"Failed to log recall to search history: {log_error}")
--- a/cognee/api/v1/recall/recall.py
+++ b/cognee/api/v1/recall/recall.py
@@ -579,13 +579,7 @@ async def recall(
                 normalize_search_payload,
             )
 
-            # _single_dataset_id is imported rather than reimplemented so recall
-            # and search can never disagree about which dataset a question is
-            # attributed to.
-            from cognee.modules.search.methods.search import (
-                authorized_search,
-                _single_dataset_id,
-            )
+            from cognee.modules.search.methods.search import authorized_search
             from cognee.modules.search.operations import log_search_history
 
             if user is None:
@@ -662,20 +656,7 @@ async def recall(
             # because it calls authorized_search() directly and skips the
             # logging that search() wraps around it. Agents recall through this
             # endpoint, so their questions were absent from history entirely.
-            # Never let logging break a search — a failed insert must not cost
-            # the caller their answer.
-            try:
-                await log_search_history(
-                    query_text,
-                    local_query_type.value,
-                    user.id,
-                    graph_results,
-                    # Each row takes its dataset from its own payload; this is
-                    # only used for payloads that carry none.
-                    fallback_dataset_id=_single_dataset_id(search_dataset_ids),
-                )
-            except Exception as log_error:
-                logger.warning(f"Failed to log recall to search history: {log_error}")
+            await log_search_history(query_text, local_query_type.value, user.id, graph_results)
 
             tagged = []
             for r in graph_results:
--- a/cognee/api/v1/recall/recall.py
+++ b/cognee/api/v1/recall/recall.py
@@ -586,7 +586,7 @@ async def recall(
                 authorized_search,
                 _single_dataset_id,
             )
-            from cognee.modules.search.operations import log_query
+            from cognee.modules.search.operations import log_search_history
 
             if user is None:
                 try:
@@ -635,23 +635,6 @@ async def recall(
                 if not search_dataset_ids:
                     raise DatasetNotFoundError(message="No datasets found.")
 
-            # /v1/search records every question it answers; recall never did,
-            # because it calls authorized_search() directly and skips the
-            # logging that search() wraps around it. Agents recall through this
-            # endpoint, so their questions were absent from recall history
-            # entirely. Reuses search()'s dataset attribution rule so both
-            # endpoints record the same thing. Never let logging break a
-            # search — a failed insert must not cost the caller their answer.
-            try:
-                await log_query(
-                    query_text,
-                    local_query_type.value,
-                    user.id,
-                    _single_dataset_id(search_dataset_ids),
-                )
-            except Exception as log_error:
-                logger.warning(f"Failed to log recall query to search history: {log_error}")
-
             graph_results = await authorized_search(
                 query_text=query_text,
                 query_type=local_query_type,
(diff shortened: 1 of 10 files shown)


LStromann:
@iigor Both points addressed in `fc504e085`.

**1. One history row per dataset searched.** This was a bug in `search()` too, as you said — `_single_dataset_id()` recorded `NULL` whenever more than one dataset was searched, and `log_result` wrote a single row holding `json.dumps([...])`, a JSON array of unlabelled completions with no way to tell which dataset produced which answer.

Both endpoints now go through `log_search_history`, which writes **one query and one result per payload**, taking the dataset from `SearchResultPayload.dataset_id` — the results already carry it, so nothing new had to be threaded through.

The logging moved to *after* the search, because only the results say which datasets were actually searched: `dataset_ids=None` means "every dataset the user can read", which is unknowable beforehand without duplicating the authorization step.

**2. Result logging in recall.** Same helper, so recall now records answers exactly as search does.

**Fallback:** payloads carry no `dataset_id` when access control is disabled, so the request's own dataset is used there. A search that returns nothing still logs the question — a question nothing could answer is worth recording.

**Net effect on `search()`: -12 lines.** The per-dataset loop replaced the collapse-into-JSON logic rather than adding to it.

Tests cover the fan-out (two datasets → two query rows and two result rows, each result tied to its own dataset's query), the access-control-off fallback, the empty-result case, and that a failed insert still returns the caller's answer. `58 passed` across `cognee/tests/unit/api/recall` and `cognee/tests/unit/modules/search`; the two existing search tests moved their patch seam to follow the code.

cognee/api/v1/recall/recall.py
```diff
@@ -650,6 +658,23 @@ async def _run_graph() -> list[RecallResponse]:
...
+            # endpoint, so their questions were absent from history entirely.
+            # Never let logging break a search — a failed insert must not cost
+            # the caller their answer.
+            try:
+                await log_search_history(
+                    query_text,
+                    local_query_type.value,
+                    user.id,
+                    graph_results,
+                    _single_dataset_id(search_dataset_ids),
```
dexters1: Won't this reduce the list of UUIDs to just one dataset_id?

We want to make a search/result entry/row per dataset in the database so we can filter out the events based on dataset_ids

cognee/modules/search/methods/search.py
```diff
@@ -146,21 +144,12 @@ async def search(
...
-    await log_result(
-        query.id,
-        json.dumps(completions) if completions else "[]",
-        user.id,
-        logged_dataset_id,
+    # Logged after the search because only the results say which datasets were
+    # actually searched — dataset_ids=None means "every dataset the user can
+    # read". Only the completion text is stored, never the raw result_objects,
+    # which run 50-100 KB each and would grow the DB without bound.
+    await log_search_history(
```
dexters1: I am not 100% sure, but just logging the result in the for loop above should also work? And provide a good result with proper dataset_id info to the DB without a need for a dedicated function for it

cognee/api/v1/recall/recall.py
LStromann: No — that argument is only the *fallback*. Each row takes its dataset from its own payload (`SearchResultPayload.dataset_id`), so a fan-out over 3 datasets writes 3 query rows and 3 result rows, each with its own id. `_single_dataset_id()` is used only for payloads that carry no dataset, which is what happens with access control disabled.

It read as the source rather than the fallback, so I've made it a named argument at both call sites with a one-line comment (`ddc9d7ba0`). Test `test_recall_logs_one_query_and_result_per_dataset` asserts the per-dataset rows.

cognee/modules/search/methods/search.py
LStromann: Inlining the loop here would work for `search()`, but recall calls `authorized_search()` directly and never reaches this function — that is exactly why recall was logging nothing. It would need the same loop copied, and the two would drift again.

One shared function is also what let the per-dataset fix land in both endpoints at once. It cost 41 lines and removed 12 from here, so `search()` is net shorter than before.

Happy to inline it if you would rather have zero indirection — recall would then get its own copy.

cognee/api/v1/recall/recall.py
```diff
@@ -650,6 +658,25 @@ async def _run_graph() -> list[RecallResponse]:
...
+                await log_search_history(
+                    query_text,
+                    local_query_type.value,
+                    user.id,
+                    graph_results,
+                    # Each row takes its dataset from its own payload; this is
+                    # only used for payloads that carry none.
+                    fallback_dataset_id=_single_dataset_id(search_dataset_ids),
+                )
+            except Exception as log_error:
```
dexters1: Please remove the try/except 

if an error happens it should break the system and not silently not work otherwise CI can never detect issues with it

cognee/modules/search/operations/log_search_history.py
```diff
@@ -0,0 +1,42 @@
...
+    was searched, and collapsing the answers into one blob loses which dataset
+    answered — so each payload gets its own query and result row.
+
+    ``fallback_dataset_id`` covers payloads that carry no dataset of their own,
+    which is what happens with access control disabled.
+    """
+    payloads = [item[0] if isinstance(item, tuple) else item for item in search_results] or [None]
+
+    for payload in payloads:
+        dataset_id = getattr(payload, "dataset_id", None) or fallback_dataset_id
```
dexters1: Lets just keep the dataset_id as None if none is provided and remove fallback_dataset_id as a concept from the function in case when backend access control is turned off so the function is more clear, readable and reliable

cognee/api/v1/recall/recall.py
LStromann: Removed in `b280c3fe8`. `search()` never guarded its logging either, so the two paths now behave the same way.

Swapped the test that asserted the swallow for one asserting it raises, so nobody quietly reintroduces it.

cognee/modules/search/operations/log_search_history.py
LStromann: Done in `b280c3fe8` — the helper takes only `(query_text, query_type, user_id, search_results)` and each row records whatever its payload carries, `None` included. One rule instead of two.

Knock-on worth flagging: `_single_dataset_id()` in `search.py` is now unused by production code — its only remaining callers are its own two tests from #4426. I left it rather than delete someone else's helper in this PR, but happy to drop it and those tests if you want it gone.

dexters1 (dismissed):

dexters1 (approved):

PR #4438, opened 2026-08-12
feat(permissions): read which principals have access to a dataset (COG-6140)

LStromann:
## Summary

Dataset permissions can be written but never read. `POST` and `DELETE /permissions/datasets/{principal_id}` both exist, and `get_principal_datasets` answers *"which datasets can this principal reach"* — but nothing answers the reverse, so a client that shares a dataset has no way to show who it is already shared with.

This adds the missing direction:

```
GET /v1/permissions/datasets/{dataset_id}/principals
```

```json
[
  {"principal_id": "…", "kind": "role",   "name": "engineering",        "permissions": ["read", "write"]},
  {"principal_id": "…", "kind": "user",   "name": "member@example.com", "permissions": ["read"]},
  {"principal_id": "…", "kind": "tenant", "name": "acme",               "permissions": ["read", "write"]}
]
```

No new tables and no new permission types. `acls` is already a plain `(principal_id, permission_id, dataset_id)` join table, so the inverse is one query — rows are collapsed per principal, since the table stores a row per permission and would otherwise repeat the principal once per grant.

## Design notes

- **Authorization mirrors the grant side.** The caller must hold `share` on the dataset (`get_specific_user_permission_datasets`, which raises `PermissionDeniedError`), so someone who cannot share a dataset cannot enumerate who it is shared with either.
- **`kind` comes from the polymorphic discriminator** on `principals`, not an `isinstance` ladder — so a dataset shared workspace-wide reads as a single `tenant` entry rather than one row per member, which is what a "who has access" list wants to show.
- **Dangling rows are skipped**, not fatal: an ACL whose principal or permission has gone is dropped from the listing.

## Testing

5 unit tests in `cognee/tests/unit/users/permissions/test_get_dataset_principals.py`: all three principal kinds are labelled correctly, permissions collapse per principal, the unauthorized caller is refused, a dangling ACL row is skipped, and a dataset nobody shares returns `[]`.

- `cognee/tests/unit/users` — 21 passed
- `ruff check .` and `ruff format --check .` — clean repo-wide

## Open question

`share` is the strict reading, and it matches the grant side. It does mean a plain member holding only `read` on a brain cannot see which teams it is shared with — so a UI listing teams-with-access would come back empty for them.

If this should instead follow the philosophy of #4336 (you can see what you are part of), `read` is the one-line alternative. Happy to switch before this leaves draft.

## Context

Needed by the cloud dashboard's brains page, where the Teams column currently lists every visible team rather than the ones with access, and the share modal forgets existing shares on reopen. Both track the missing read back to COG-6140.

### DIFF
--- a/cognee/api/v1/permissions/routers/get_permissions_router.py
+++ b/cognee/api/v1/permissions/routers/get_permissions_router.py
@@ -140,6 +140,48 @@ def get_permissions_router() -> APIRouter:
             status_code=200, content={"message": "Permission revoked from principal"}
         )
 
+    @permissions_router.get("/datasets/{dataset_id}/principals")
+    async def get_principals_with_dataset_permission(
+        dataset_id: UUID,
+        user: User = Depends(get_authenticated_user),
+    ):
+        """
+        List the principals (users, roles, tenants) with permission on a dataset.
+
+        The inverse of the grant/revoke endpoints above: those write access, this
+        one reads back who holds it. The authenticated user must have "share"
+        permission on the dataset — a caller who cannot share it cannot
+        enumerate who it is shared with either.
+
+        ## Path Parameters
+        - **dataset_id** (UUID): The UUID of the dataset (list yours via GET /api/v1/datasets)
+
+        ## Response
+        Returns a JSON list: [{"principal_id", "kind", "name", "permissions"}],
+        where kind is "user", "role" or "tenant", name is the user's email or the
+        role/tenant name, and permissions lists what that principal holds
+        (e.g. ["read", "write"]). A dataset shared with the whole workspace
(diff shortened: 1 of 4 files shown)


LStromann:
Closing unmerged — the product decision this was built on changed.

Team membership now implies dataset visibility: if you are in a team, you see all of the team's brains. Permissions stay manageable for **read/write**, but not for general visibility, so there is no per-dataset visibility list for a client to read back. The Teams column listing every visible team — the thing this endpoint was meant to correct — is the intended behaviour under that model, not a workaround.

COG-6140 is canceled with the same rationale. The branch `feat/dataset-principals-endpoint` is left in place in case the reverse lookup is wanted later for read/write specifically.

PR #4447, opened 2026-08-12
feat(permissions): list the datasets granted to a principal (COG-6158)

LStromann:
COG-6158

## Summary

A team is a role, and a role is a principal — so a Teams-first Brain page asks
"which brains does this team have?" on every render. `get_principal_datasets`
has always answered that, but only as a method: there is no route over it, so
no client can ask.

This adds one:

```
GET /v1/permissions/principals/{principal_id}/datasets?permission_name=read
```

No new concepts and no new model code — it exposes the existing lookup.

## Not the endpoint from #4438

#4438 added the *reverse* direction (dataset → principals) and was closed
unmerged: team membership now implies dataset visibility, so there is no
per-dataset visibility list worth reading back. This is the other direction,
which that decision makes *more* load-bearing, not less — the Brain page reads
its middle column from it.

## Authorization

Reuses `get_users_in_role`, the check already behind the members endpoint: it
scopes the role to the caller's tenant, then admits role members and tenant
user managers. So this route can see exactly the teams its caller can already
see, and a role id from another tenant resolves to nothing.

## Test plan

- [x] `pytest cognee/tests/unit/api/test_permissions_router.py` — route is registered at the expected path/method
- [x] `ruff check` / `ruff format`


cognee/api/v1/permissions/routers/get_permissions_router.py
```diff
@@ -140,6 +146,68 @@ async def revoke_datasets_permission_from_principal(
...
+
+        db_engine = get_relational_engine()
+        async with db_engine.get_async_session() as session:
+            role = (
+                await session.execute(
+                    select(Role).where(Role.id == principal_id, Role.tenant_id == user.tenant_id)
+                )
+            ).scalar_one()
+
+        datasets = await method_get_principal_datasets(role, permission_name)
```
dexters1: I do not understand the PR

a principal is either a user, a role or the whole tenant

why are only datasets for a role retrieved here? The name is misleading

get_principal_datasets implies it can retrieve the datasets related to either a user, a role or the whole tenant

You need to create a new SDK method for this called authorized_get_principal_datasets

which checks permissions of the caller and if he can get access to the data he requests and that data is filtered by the current tenant_id of the user (so if access is requested for information about a different tenant it wont be given)

Additionally:

1) if the requester requests to get datasets of a specific user (that is not him) he can only do so if the function has_user_management_permission allows him

2) If the requester requests to get datasets of a specific role he can only do so if has_user_management_permission allows him to do it or he is a member of the role (And the role is part of the current tenant)

3) Requester requests to get datasets of a specific tenant if he is currently in that tenant

LStromann:
Thanks Igor — you were right, the name promised a principal and the code only handled roles. Fixed in 32656d4, along with both bot majors, which the same change subsumes.

Added `authorized_get_principal_datasets` next to `authorized_give_/revoke_permission_on_datasets`, and the router now just calls it:

| Principal | Allowed when |
| -- | -- |
| user | themselves, or anyone if `has_user_management_permission` |
| role | a role of this tenant they belong to, or any of its roles if `has_user_management_permission` |
| tenant | only the tenant they are currently in |

Results are filtered to the caller's `tenant_id`, so a principal spanning tenants cannot disclose another tenant's datasets.

Two decisions worth flagging:

- **A role from another tenant raises `RoleNotFoundError` (404), not `PermissionDeniedError` (403).** Otherwise the status code confirms that a role id exists in some tenant you cannot see.
- **Membership reuses `get_user_role_names_in_tenant` rather than a new query.** Role names are unique per tenant (`uq_roles_tenant_id_name`) and the role is scoped to the tenant before the check, so matching on name is a sound membership test — and it kept the diff to no new SQL.

The duplicate role query is gone: resolving the principal happens once, inside the method. The router lost more lines than it gained.

11 unit tests in `cognee/tests/unit/users/permissions/test_authorized_get_principal_datasets.py` cover the matrix above, both cross-tenant refusals, the dataset filtering, and an unsupported principal type. `pytest cognee/tests/unit/users` — 28 passed; `ruff check`/`format` clean.

cognee/modules/users/permissions/methods/authorized_get_principal_datasets.py
```diff
@@ -0,0 +1,73 @@
...
+from cognee.modules.users.permissions.methods.get_user_role_names_in_tenant import (
+    get_user_role_names_in_tenant,
+)
+from cognee.modules.users.permissions.methods.has_user_management_permission import (
+    has_user_management_permission,
+)
+
+
+async def authorized_get_principal_datasets(
+    principal_id: UUID, permission_name: str, requester_id: UUID, tenant_id: UUID
```
dexters1: tenant_id should not be a parameter here, a User object should be found based on requester_id.

And from that user object the tenant_id needs to be specified, it specifies what is the current tenant the requester is in

LStromann: Agreed — done in 60b6562.

`tenant_id` is gone from the signature; the method now resolves the requester with `get_user(requester_id)` and reads `tenant_id` off them. Your framing is the right one: the current tenant is a property of who is asking, not an argument to the question. As written before, a call site could have named a tenant the requester was not in — the router happened to pass `user.tenant_id`, but nothing in the signature required it.

```python
async def authorized_get_principal_datasets(
    principal_id: UUID, permission_name: str, requester_id: UUID
) -> List[Dataset]:
    requester = await get_user(requester_id)
    tenant_id = requester.tenant_id
```

`get_specific_user_permission_datasets` already imports `get_user` the same way, so no new import cycle.

Added a test that pins the property directly (`test_tenant_comes_from_the_requester_not_the_caller`): a role in a foreign tenant is refused even when the requester is a "member" of it and can manage users — the requester's own tenant is the only thing that decides scope. 29 passed in `cognee/tests/unit/users`.

dexters1 (approved):

PR #4451, opened 2026-08-12
fix(search): Missing collection is empty results, not NoDataError

dexters1:
## Summary

A dataset can be legitimately cognified **without** carrying every standard vector collection — a pipeline route that skips summarization never creates `TextSummary_text`, and an added-but-never-cognified dataset has no collections at all. With per-dataset databases (access control on, the default), `CHUNKS` / `SUMMARIES` / `RAG_COMPLETION` / hybrid searches iterate every readable dataset — and **one** such dataset raised `NoDataError` (*"No data found in the system, please add data first"*), aborting the entire merged search and hiding every healthy dataset's results behind a message that is simply false.

Caught live by the backwards-compatibility CI on the DLT branch: a DLT dataset (which legitimately has no summaries) broke the SUMMARIES smoke search over the document dataset sitting right next to it.

## The fix

The inference *"missing collection ⇒ empty system"* predates routes that skip pipeline stages; it no longer holds. Retrievers now treat a missing collection exactly like the graph-completion vector sweep already does (`node_edge_vector_search.py` catches per collection and continues): **that scope contributes zero results and the search goes on.**

- `chunks_retriever`, `summaries_retriever`, `completion_retriever` (RAG): `CollectionNotFoundError` → `[]`
- hybrid `search_collection`: the `required=True` escape hatch removed — a missing collection is always an empty channel
- `triplet_retriever`: **deliberately unchanged** where it matters — the `has_collection` guard still raises its actionable *"first use the create_triplet_embeddings memify pipeline"* guidance for the opt-in feature; only the race-window catch behind it returns empty now

Nothing in production catches `NoDataError` (verified — only tests), so no handler/API contract changes; empty-result semantics also match the documented permission-filtering behavior ("returns empty list rather than error").

## Testing

- **New integration test** pins the demonstrated scenario: one healthy cognified dataset + one collectionless dataset; `CHUNKS`/`SUMMARIES`/`RAG_COMPLETION` spanning both return the healthy dataset's results, and a search scoped to the bare dataset returns an empty `search_result` instead of erroring. Offline (LLM + embeddings mocked).
- Five existing unit tests flipped from asserting `NoDataError` to asserting empty results; the triplet guidance test unchanged.
- Full retrieval unit suite green (400 tests), flipped RAG integration test green.

## Note

No Linear key yet — happy to add `COG-xxx` to the title once a ticket exists (the `Require Linear issue` check will flag until then).

### DIFF
--- a/cognee/modules/retrieval/chunks_retriever.py
+++ b/cognee/modules/retrieval/chunks_retriever.py
@@ -2,7 +2,6 @@ from typing import Any, Optional, List, Union
 from cognee.shared.logging_utils import get_logger
 from cognee.infrastructure.databases.unified import get_unified_engine
 from cognee.modules.retrieval.base_retriever import BaseRetriever
-from cognee.modules.retrieval.exceptions.exceptions import NoDataError
 from cognee.infrastructure.databases.vector.exceptions.exceptions import CollectionNotFoundError
 
 logger = get_logger("ChunksRetriever")
@@ -97,7 +96,7 @@ class ChunksRetriever(BaseRetriever):
         """
         Retrieves document chunks context based on the query.
         Searches for document chunks relevant to the specified query using a vector engine.
-        Raises a NoDataError if no data is found in the system.
+        A missing chunk collection yields an empty result list.
         Parameters:
         -----------
             - query (str): The query string to search for relevant document chunks.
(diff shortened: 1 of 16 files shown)


dexters1:
Closing this one, changed my mind. Users should properly specify their datasets when searching, no reason to safe guard so much

PR #4458, opened 2026-08-13
feat(dlt): CSV auto-detection for S3 and fsspec URLs

dexters1:
## Summary

CSV auto-detection routed every path through `os.path.abspath`, which treats `s3://bucket/data.csv` as a *relative local path* and mangles it into `<cwd>/s3:/bucket/data.csv` — dlt's filesystem reader then pointed at a nonexistent local directory (the stray `s3:/` directories some of us have in repo roots are this bug's droppings). So a CSV living on S3 could never take the DLT route.

## The fix

One branch in `create_dlt_source_from_csv`: inputs matching an fsspec URL scheme (`s3://`, `gs://`, `az://`, `file://`, …) are split into parent-URL + filename and handed to dlt's filesystem source **untouched**; plain local paths keep the abspath resolution. `is_csv_path` already accepted URL forms (it only excludes http/https), so detection needed no change.

Requirements for remote protocols, documented in the docstring: the protocol's fsspec backend must be installed (s3 ships with the `aws` extra) and credentials come from the standard environment (`AWS_ACCESS_KEY_ID` / `AWS_SECRET_ACCESS_KEY` / region).

## Testing — the full ladder

1. **Unit** (`test_create_dlt_source_csv.py`): pins the URL parsing for s3/gs/file/memory forms — including the exact abspath-mangling regression — plus local absolute/relative behavior and `is_csv_path` semantics.
2. **Hermetic integration** (`test_dlt_csv_url_ingestion.py`): drives a `file://` URL through plain `add()`/`cognify()` end to end — same code path as s3, different fsspec backend — asserting the manifest stamp (`row_count: 3`) and graph census (`DltRow: 3, SchemaTable: 1`). Offline, embeddings mocked, zero LLM calls.
3. **Real S3 e2e**: the existing **S3 Bucket Test** job now also self-provisions a run-scoped CSV in the shared `github-runner-cognee-tests` bucket (uuid key, removed in `finally`), ingests it via the bare `s3://…/instruments.csv` string through plain add/cognify, and asserts manifest + `DltRow` census under the dataset context. The job gains the `dlt` extra and the dlt-telemetry exit-hang guard (see the backwards-compat jobs for its history).

Local verification: could not run real S3 from this machine (no s3fs/credentials in the dev venv) — the CI job is the live proof; the hermetic ladder below it passed locally (8 unit + 1 integration, plus the full ingestion unit suite).

## Notes

- `source_name` for auto-detected CSVs is still dlt's reader name (`_read_csv`) rather than the filename — pre-existing cosmetic wart, untouched here to avoid changing manifest identity.
- No Linear key yet; happy to retitle with `COG-xxx` when a ticket exists.

### DIFF
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -286,11 +286,16 @@ jobs:
           uses: ./.github/actions/cognee_setup
           with:
             python-version: '3.11.x'
-            extra-dependencies: "aws"
+            # dlt: the test also covers CSV-from-S3 auto-detection into the
+            # DLT ingestion route, which is gated on the dlt extra.
+            extra-dependencies: "aws dlt"
 
         - name: Run S3 Bucket Test
           env:
             ENV: 'dev'
+            # dlt's telemetry thread is non-daemon and a wedged call blocks
+            # interpreter exit (see the backwards-compat jobs).
+            RUNTIME__DLTHUB_TELEMETRY: "false"
             LLM_MODEL: ${{ secrets.LLM_MODEL }}
             LLM_ENDPOINT: ${{ secrets.LLM_ENDPOINT }}
             LLM_API_KEY: ${{ secrets.LLM_API_KEY }}
--- a/cognee/tasks/ingestion/create_dlt_source.py
+++ b/cognee/tasks/ingestion/create_dlt_source.py
@@ -56,34 +56,57 @@ def create_dlt_source_from_connection_string(
     return source
 
 
-# fsspec-style URL scheme (s3://, gs://, az://, file://, memory://, ...).
-# http(s) is already excluded by is_csv_path. A URL must NOT go through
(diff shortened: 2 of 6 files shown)


dexters1:
Superseded: dlt-pipeline (#4278, merged to dev) absorbed S3-CSV handling via the storage layer (#4466's spool machinery), and PR #4470 (dlt-loader-engine) now moves all CSV ingestion into the loader engine — where s3:// localization rides the loader path's built-in pull_from_s3 through cognee's S3Config credentials. This branch's one remaining unique asset — the live-S3 CI e2e (dlt_csv_from_s3 + the aws dlt extras and telemetry guard on the S3 Bucket Test job) — was ported onto #4470 as a42916f7f, upgraded with a dlt_csv_loader ownership assertion. Nothing else here survives: the production diff was fully superseded.

PR #4461, opened 2026-08-13
fix(vector): report LanceDB dimension mismatches

### LINKED ISSUE #4313: LanceDB merge_insert 'Spill has sent an error' with 1024-dim custom embeddings — only inside uvicorn server; same pipeline succeeds out-of-process
## Summary

With a **1024-dim custom (OpenAI-compatible) embedding provider**, every cognify run executed **inside the uvicorn server process** fails during `add_data_points` with:

```
RuntimeError: lance error: LanceError(IO): Execution error: Spill has sent an error,
/rustc/.../library/core/src/task/poll.rs:291:44
```

The identical pipeline, config and data succeed when `cognee.cognify()` is run in a plain Python process. 512-dim fastembed embeddings work fine in the server context. This blocks using any 1024-dim OpenAI-compatible embedding endpoint with the served API.

## Environment

- cognee 1.4.0, lancedb 0.36.0, pylance 0.36.0, pyarrow 25.0.0
- Python 3.12.13, macOS arm64 (Darwin 25.3)
- `EMBEDDING_PROVIDER=custom` (OpenAI-compatible endpoint, qwen text-embedding), `EMBEDDING_DIMENSIONS=1024`, `LITELLM_DROP_PARAMS=true`, `EMBEDDING_BATCH_SIZE=10`
- LLM provider: custom OpenAI-compatible (works fine; graph extraction completes)

## Failure point

```
File ".../cognee/infrastructure/databases/vector/lancedb/LanceDBAdapter.py", line 521, in create_data_points
File ".../cognee/infrastructure/databases/vector/lancedb/LanceDBAdapter.py", line 1296, in index_data_points
File ".../lancedb/table.py", line 5470, in _do_merge
    return await self._inner.execute_merge_insert(
RuntimeError: lance error: LanceError(IO): Execution error: Spill has sent an error, ...
```

Reproduces 100% of the time (5/5 clean-slate attempts) whether `VECTOR_DB_SUBPROCESS_ENABLED` is true (failure inside `cognee_db_workers/lancedb_worker.py`) **or** false (failure directly in the server process). The dataset is tiny (one small text file), so this is not a real memory-pressure spill.

## What does NOT reproduce it (isolation matrix)

All of the following succeed with the exact same venv, config and 1024-dim data:

| Repro attempt | Result |
|---|---|
| raw lancedb, sync API, create_table 30×1024-dim rows | OK |
| raw lancedb, async API, merge_insert 10/100/500 rows | OK |
| raw lancedb, 12 concurrent merge_inserts, same table | OK |
| raw lancedb, async merge_insert with `pa.Table` input (as `_records_for_write` produces) | OK |
| cognee `LanceDBAdapter.create_data_points` called in isolation (real custom 1024-dim embeddings) | OK |
| same + 5 collections × 3 concurrent batches | OK |
| same + graph engine initialized in the same process | OK |
| **full `cognee.add()` + `cognee.cognify()` in a plain Python process** | **OK — graph built, search works** |
| any cognify routed through the uvicorn server (`POST /api/v1/cognify` or memify-triggered) | **FAILS** |
| same server context but 512-dim fastembed embeddings | OK |

So the trigger seems to require: uvicorn server process context × 1024-dim custom embedding path. "Spill has sent an error" is the receiver-side report of the input stream erroring, which suggests the record-batch producer feeding `execute_merge_insert` dies mid-stream in that context (event-loop interaction?).

## Workaround we're using

Run graph builds out-of-process (`cognee.cognify()` via a helper script against the same storage) and keep the server for reads/ingestion — or stay on 512-dim fastembed.

Happy to provide full logs, the isolation scripts, or run instrumented builds if that helps.

EthanReedLabs:
## Description

LanceDB can surface a fixed-size vector mismatch during `merge_insert` as a
cryptic write error. This change diagnoses the failure only after the LanceDB
write fails. It then reads the existing collection schema and raises
`VectorDimensionMismatchError` only when the stored and incoming dimensions are
confirmed to differ.

Successful writes do not perform a schema check. Same-dimension failures and
inconclusive diagnosis preserve the existing exception and schema-migration
handling.

The subprocess-backed adapter exposes a schema-only table operation, so failure
diagnosis does not materialize collection rows.

Closes #4313

## Acceptance Criteria

- [x] Successful writes do not perform vector-dimension schema checks.
- [x] Dimension diagnosis runs only after a LanceDB write failure.
- [x] Schema-confirmed mismatches raise a dedicated `CogneeApiError` subtype
      with HTTP 422.
- [x] The error reports both dimensions and remediation guidance.
- [x] The original LanceDB failure is preserved as the exception cause.
- [x] Same-dimension and inconclusive failures retain existing handling.
- [x] Local and subprocess diagnosis reads only table schema.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots

Local test results:

```text
cognee/tests/unit/infrastructure/databases/vector/test_lancedb_vector_dimensions.py: 6 passed
cognee/tests/unit/infrastructure/databases/vector: 48 passed, 14 skipped
pre-commit: all hooks passed
git diff --check: passed
```

## Pre-submission Checklist

- [x] All new and existing relevant tests pass

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/databases/vector/exceptions/__init__.py
+++ b/cognee/infrastructure/databases/vector/exceptions/__init__.py
@@ -1,1 +1,1 @@
-from .exceptions import CollectionNotFoundError
+from .exceptions import CollectionNotFoundError, VectorDimensionMismatchError
--- a/cognee/infrastructure/databases/vector/exceptions/exceptions.py
+++ b/cognee/infrastructure/databases/vector/exceptions/exceptions.py
@@ -1,5 +1,5 @@
 from fastapi import status
-from cognee.exceptions import CogneeValidationError
+from cognee.exceptions import CogneeApiError, CogneeValidationError
 
 
 class CollectionNotFoundError(CogneeValidationError):
@@ -18,5 +18,19 @@ class CollectionNotFoundError(CogneeValidationError):
         status_code: int = status.HTTP_422_UNPROCESSABLE_CONTENT,
         log=True,
         log_level="DEBUG",
+    ):
+        super().__init__(message, name, status_code, log, log_level)
+
+
+class VectorDimensionMismatchError(CogneeApiError):
+    """Raised when a collection and its embedding engine use different vector sizes."""
+
+    def __init__(
+        self,
+        message: str = "A vector dimension mismatch occurred.",
+        name: str = "VectorDimensionMismatchError",
+        status_code: int = status.HTTP_422_UNPROCESSABLE_CONTENT,
+        log=True,
+        log_level="ERROR",
     ):
         super().__init__(message, name, status_code, log, log_level)
--- a/cognee/infrastructure/databases/vector/lancedb/LanceDBAdapter.py
+++ b/cognee/infrastructure/databases/vector/lancedb/LanceDBAdapter.py
@@ -17,7 +17,10 @@ from cognee.infrastructure.engine import DataPoint
 from cognee.infrastructure.engine.utils import parse_id
 from cognee.infrastructure.files.storage import get_file_storage
 from cognee.modules.storage.utils import copy_model
-from cognee.infrastructure.databases.vector.exceptions import CollectionNotFoundError
+from cognee.infrastructure.databases.vector.exceptions import (
+    CollectionNotFoundError,
+    VectorDimensionMismatchError,
+)
 from cognee.infrastructure.databases.vector.pgvector.serialize_data import serialize_data
 from cognee.shared.logging_utils import get_logger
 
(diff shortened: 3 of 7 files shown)


dexters1:
@EthanReedLabs can we change to only check for the dimensions in the case the LanceDB exception happens? So we don't continuously check for dimension size? 

In case of the LanceDB exception we try and verify if it caused by a dimension mismatch and then raise the issue

EthanReedLabs:
Updated in 553aefd1a. The successful write path no longer reads the table schema. Dimension verification now runs only after a LanceDB write error. A schema-confirmed mismatch raises `VectorDimensionMismatchError` from the original exception, while same-dimension or inconclusive failures preserve the existing handling.

The write flow is now:

1. Open or create the collection and get the incoming vector size.
2. Generate embeddings and build the `LanceDataPoint` records.
3. Attempt `merge_insert` without reading the table schema.
4. If the write succeeds, return without dimension diagnosis.
5. If the LanceDB write fails, read the collection schema through the schema-only path.
6. If the stored and incoming dimensions are confirmed to differ, raise `VectorDimensionMismatchError` from the original write exception.
7. Otherwise, continue the existing schema-migration handling or re-raise the original exception.

The subprocess path uses the schema-only RPC and does not materialize collection rows.

EthanReedLabs:
I have an idea for a possible follow-up: in addition to the current API-based LLM providers, Cognee could expose an optional entry point that uses a locally available Claude Code or Codex runtime to execute supported LLM tasks.

Has the team explored an approach like this, or is anything similar being considered for the project roadmap? I would be interested in the team's view on whether this fits Cognee's current LLM and provider architecture and overall direction.

EthanReedLabs:
@dexters1, I would like to confirm the team’s preferred direction for #4313. Would the team prefer to keep this at the design/discussion stage for now, or is it appropriate to continue addressing it through this implementation PR? I have updated #4461 based on your feedback in 553aefd1a. If the PR route is appropriate, please let me know whether the team would like any further changes before review.

PR #4467, opened 2026-08-13
SDK-100 feat: Bundle Ladybug JSON extension with cognee

dexters1:
Linear: [SDK-100](https://linear.app/cognee/issue/SDK-100/ladybug-hosting-issue-json-extension)

## Description

Removes cognee's runtime dependency on `extension.ladybugdb.com` by bundling the official Ladybug JSON extension binaries with cognee and loading them by absolute path. **The ladybug constraint in `pyproject.toml` is the single source of truth for what gets bundled.**

### Why

Ladybug's **Linux and Windows wheels dynamic-load the JSON extension**: `INSTALL JSON` downloads `libjson.lbug_extension` from `extension.ladybugdb.com` at runtime (macOS wheels compile it in statically). That server's origin has been **down since at least Aug 13** (522 from Cloudflare; last provably alive Aug 10 ~18:55 UTC per edge-cache `age` headers). Deployments currently survive only where a Cloudflare edge still has the file cached — CI included — and fail on any cache-cold edge, air-gapped network, or once caches evict. JSON-dependent features (recall, temporal search) then break on fresh databases. The latest ladybug (0.19.1) has the same gap with thinner cache coverage.

### How

- **Loader ladder** in `cognee_db_workers/_kuzu_helpers.py::load_json_extension`: by-name `LOAD` (static builds / preinstalled) → bundled binary via `LOAD EXTENSION '<path>'` (reads the file directly, never contacts the repo) → classic remote `INSTALL` as last resort. Shared by the local-mode adapter, the warm-up (skipped entirely when a bundle exists), and the subprocess worker. Wrong-binary loads are impossible: lookup is keyed by the *installed* ladybug's version, misses degrade to the old behavior, and ladybug's own load-time version check backstops everything.
- **Source of truth**: `scripts/ladybug_extension_versions.py` (stdlib-only) intersects the pyproject ladybug constraint with the empirically verified mapping `_EXTENSION_REPO_VERSIONS` (0.16.0→v0.16.0, 0.17.x→v0.17.0, 0.18.0→v0.18.0, 0.18.1/0.18.2→v0.18.1 — each entry read from offline `INSTALL` error URLs). Release workflows call the fetch script argument-free, so **bumping the ladybug constraint automatically changes what ships**.
- **Guard rails**: unit tests fail when the uv.lock-pinned version or any constraint bound lacks a mapping entry (with instructions in the assertion message), and cross-check the resolver against `packaging`'s PEP 440 semantics. Release workflows assert **post-build** that every supported version's binary is inside the wheel — a missing bundle cannot ship silently.
- **Binaries** live in `cognee_db_workers/ladybug_extensions/` — gitignored, fetched at build time by `scripts/fetch_ladybug_json_extension.sh` from `ghcr.io/ladybugdb/extension-repo` (the public origin image behind the extension server — same official artifacts, available during the outage). The hatch `artifacts` entry lets them into wheels.
- **Dockerfile** collects every published Linux binary pair in a build stage — version-agnostic, never needs touching on ladybug bumps.
- **Version bump** to 1.4.3.dev1 + lock regen.

### Size impact

Wheel with the full supported range (four extension versions × five platforms): 7.15 MB → **29 MB**. The four `win_amd64` blobs dominate (~4.6 MB compressed each — Windows statically links the lbug core). Levers if that's too much: bundle Windows only for the newest version, or drop Windows entirely (falls back to remote) — both are one-line changes in the workflow fetch call. Docker image: +~15 MB. sdist: unchanged.

### Verification (all local / local Docker — no release workflow touched)

- Offline (`--network=none`) end-to-end on **linux amd64 + arm64**: ladybug 0.18.1, 0.18.2, and **0.16.0 each select their own version's binary** and `to_json()` works.
- **Built wheel** installed into a bare Linux container, offline: bundle resolves from site-packages and loads.
- **Docker image** built from this branch, run offline: binaries baked in, **local-mode adapter** loads JSON + executes a JSON Cypher query, **subprocess-mode adapter** (production default) loads it in the worker.
- macOS: static linking verified on 0.16.0/0.17.1/0.18.2 — by-name load, no bundle needed (osx binaries included as insurance against a repeat of the 0.18 Linux static-linking regression).
- 12 unit tests (ladder, resolution, source-of-truth guards); worker import-hygiene tests pass; `pre-commit` clean.

### Caveats

- `win_amd64` binaries bundled but not runtime-tested (no Windows host); path handling normalizes backslashes.
- musllinux (Alpine) wheels report the same platform token but can't dlopen glibc binaries — the loader falls back to the remote path there.
- Upstream asks worth filing separately: restart the extension origin (it's the GHCR nginx image; the deployment is down), and restore Linux/Windows static linking (existed in 0.17.x) — which would eventually make this bundle deletable.

---

## Update (Sep 9): merged dev, adapted to the split ladybug constraint, re-verified offline

Dev moved substantially since this PR was written (ladybug pin bumped to 0.19.0, the constraint split across platform markers, ruff bumped with a no-`noqa` blind-except gate). Changes on top of the original:

- **Merged `dev`** (4 conflicts). In `adapter.py`, dev had added its own INSTALL-and-retry in the except handler; it is a strict subset of this PR's `load_json_extension` ladder (whose last rung runs the identical `INSTALL JSON; LOAD EXTENSION JSON;` on the same connection), so the ladder stands alone.
- **Resolver handles the split constraint**: pyproject now carries two ladybug lines (`>=0.17.0,<0.18` for legacy macOS, `==0.19.0` elsewhere). `ladybug_requirements()` returns every line and `bundled_extension_dirs()` bundles the **union** of their ranges — currently `v0.17.0` + `v0.19.0`, matching both uv.lock pins. The guard tests cover all lines and all locked versions.
- **Test fixes**: dev's worker sequence test now expects the ladder's version probe between LOAD and INSTALL; the bundled-extension sequence asserts mirror production's forward-slash quoting so they hold on Windows.
- **Supply-chain**: the extension-repo image is now pulled **by digest** in both the fetch script and the Dockerfile bundle stage (a compromised `:latest` tag can no longer inject binaries into wheels or the image); a unit test keeps the two pins present and identical, and the pinned digest is the exact content the offline e2e validated. The refresh command is documented in the fetch script for ladybug bumps.
- **Lint conformance**: the two deliberate blind catches now log with `exc_info=True` (Code Quality's `--ignore-noqa` BLE001 pass rejects suppression comments); the bundled-load fallback logs through the module logger instead of a stderr print.

### Re-verification (post-merge, real binaries, offline)

- `scripts/fetch_ladybug_json_extension.sh` (argument-free) against the live GHCR image: resolver selected exactly `v0.17.0` + `v0.19.0`, all 5 platforms fetched (10 binaries).
- Linux (arm64) containers with `--network=none`:
  - **ladybug 0.19.0** → probe announced `v0.19.0/linux_arm64`, bundled binary loaded, `to_json([1,2,3])` returned correctly. (0.19.0 postdates the original verification — first proof on the new pin.)
  - **ladybug 0.17.1** → served by the **trailing dir** `v0.17.0`, loaded, `to_json` works — exercising the below-floor dir logic end-to-end.
- Full unit suite locally: 5886 passed; pre-commit and both CI ruff passes clean.
- **The offline proof is now a permanent CI job**: `Bundled JSON Extension Offline e2e` (`cognee/tests/e2e/bundled_extension/`, wired into `e2e_tests.yml`) reruns the fetch → probe → `--network=none` load → JSON query chain for every ladybug version pinned in uv.lock on each PR — the canary for the probe's error-format dependency, which mocked unit tests cannot see.


cognee_db_workers/_kuzu_helpers.py
```diff
@@ -30,6 +31,107 @@ def _safe_close(obj) -> None:
...
+# directory is populated.
+
+# Maps the installed ladybug package version to the extension-repo version
+# directory its binary requests (baked into the native lib; it can trail the
+# package version). Discover a new entry by running ``INSTALL JSON;`` offline —
+# the error message prints the exact URL, e.g.
+# ``.../v0.18.1/linux_arm64/json/libjson.lbug_extension``. Every entry below
+# was verified that way. Covers cognee's full supported range
+# (ladybug>=0.16.0,<=0.18.2 in pyproject.toml).
+_EXTENSION_REPO_VERSIONS = {
```
dexters1: This seems like it could be handled better, having to manage a variable mapping ourselves all the time seems like a nightmare

cognee_db_workers/_kuzu_helpers.py
```diff
@@ -30,6 +31,107 @@ def _safe_close(obj) -> None:
...
+    "0.17.1": "v0.17.0",
+    "0.18.0": "v0.18.0",
+    "0.18.1": "v0.18.1",
+    "0.18.2": "v0.18.1",
+}
+
+_BUNDLED_EXTENSIONS_DIR = os.path.join(os.path.dirname(__file__), "ladybug_extensions")
+
+
+def _ladybug_platform() -> Optional[str]:
```
dexters1: There are claims that the MacOS version is still bundled with the JSON extension.

1) I am not sure that is true it needs verifying
2) Even if it is true it seems likely they might exclude it from the MacOS version at some point as they have for other OSs.

Would be good to have also the MacOS binaries (if possible) so we do not need to make changes to support MacOS maintenance this way in the future

cognee_db_workers/_kuzu_helpers.py
dexters1: Fixed in aab4b6ffa — the mapping is gone entirely. The engine now announces its own requirement: the loader runs `INSTALL JSON FROM '<invalid local path>'`, which fails instantly offline (~0.01s, no network — the path is treated as an unreachable URL) with the exact `<version>/<platform>` in the error message. Only that announced file is loaded. At fetch time, the extension repo's published dirs are filtered by the pyproject ladybug constraint, so bumping the constraint changes what ships with no other edit anywhere.

Worth knowing why it's a probe and not "try each bundled binary until one loads": I tested that alternative and loading a real-but-wrong-version extension binary **segfaults the process** (exit 139), so guessing was never an option — the probe is what makes selection safe with zero maintained state.

cognee_db_workers/_kuzu_helpers.py
dexters1: Both points addressed:

**1) Verified** — empirically, on macOS arm64 with ladybug 0.16.0, 0.17.1 and 0.18.2 (fresh venv each): `LOAD EXTENSION JSON` succeeds on a brand-new database with zero extension files anywhere on disk (checked ~/.kuzu, ~/.lbdb, ~/.ladybug and the DB dir before/after) and with the extension server dead, so no download was possible even in principle. The macOS wheel contains only the single native library — the JSON code can only be inside it. One-command repro (works with Wi-Fi off): `python3 -m venv /tmp/v && /tmp/v/bin/pip install ladybug && /tmp/v/bin/python -c "import ladybug,tempfile,os; c=ladybug.Connection(ladybug.Database(os.path.join(tempfile.mkdtemp(),'p.db'))); c.execute('LOAD EXTENSION JSON;'); print(c.execute('RETURN to_json({ok:1});').get_next())"`

**2) Already bundled** — `osx_amd64` and `osx_arm64` binaries ship in the wheel for every supported version, precisely as insurance against them dropping static linking later (they already did exactly that on Linux in 0.18). The ladder makes the transition automatic: today macOS succeeds at the by-name rung and never touches the bundle; the day a macOS wheel stops embedding JSON, the probe rung picks up the bundled osx binary with zero code changes. Docstring wording softened to 'currently compile it in' in aab4b6ffa.

NMZivkovic (dismissed):

NMZivkovic (approved):

PR #4477, opened 2026-08-14
fix: Plumb LLM_TEMPERATURE into calls, add LLM_SEED (COG-6174)

Vasilije1990:
## Description

`LLM_TEMPERATURE` was dead config, surfaced while investigating customer feedback about nondeterministic extraction hurting reproducibility:

- `llm_temperature` is defined in `LLMConfig` (default `0.0`) and documented in `docs/ollama_models.md` and `.env.template` as "Cognee's default" for deterministic output — but it was never plumbed to any adapter on the instructor or litellm-native paths. No adapter takes a temperature argument, so the provider default (OpenAI = 1.0) always applied. The only working path was the `LLM_ARGS='{"temperature": 0}'` escape hatch. (The BAML path wires its own `baml_llm_temperature` correctly.)
- There was also no seed support on the extraction path.

## Fix

One `model_validator` on `LLMConfig` folds `llm_temperature` and a new `llm_seed` field into `llm_args` — the dict every adapter already merges into each completion call (`{**self.llm_args, **kwargs}`). Both structured-output paths pick the values up with no adapter changes, and the instructor client cache key already includes `llm_args`, so differing temperatures cannot collide in the adapter cache.

Two deliberate semantics:
- **Temperature is folded only when explicitly set** (env var or kwarg). The default model `openai/gpt-5-mini` rejects any temperature other than 1, so silently sending the field's `0.0` default would break every default install.
- **Explicit `LLM_ARGS` keys win** over the dedicated fields, preserving existing workaround behavior.

Docs updated to stop claiming an unplumbed default; `.env.template` documents both knobs including the gpt-5 caveat.

## Verification

- 5 new unit tests in `test_llm_config.py` (unset not folded, kwarg/env folded, `LLM_ARGS` precedence, seed folding preserves keys).
- Verified end-to-end that with `LLM_TEMPERATURE=0.0 LLM_SEED=7`, both the instructor `OpenAIAdapter` and `NativeLiteLLMAdapter` carry `{"temperature": 0.0, "seed": 7}`.
- The two failures in `test_llm_config_rate_limit_defaults.py` are pre-existing on clean dev (verified via stash).

Fixes COG-6174

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -537,7 +537,11 @@ WEB_SCRAPER_MAX_DELAY=10.0
 ################################################################################
 
 # -- LLM tuning ---------------------------------------------------------------
+# Sampling temperature, sent with every LLM call when set. Leave unset to use
+# the provider's default — note gpt-5 models reject any value other than 1.
 #LLM_TEMPERATURE=0.0
+# Sampling seed for reproducible outputs, sent when set (provider support varies).
+#LLM_SEED=42
 #LLM_STREAMING=false
 # Optional fallback model used when the primary completion fails.
 #FALLBACK_MODEL=""
--- a/cognee/infrastructure/llm/config.py
+++ b/cognee/infrastructure/llm/config.py
@@ -109,6 +109,7 @@ class LLMConfig(BaseSettings):
     llm_query_api_version: str | None = None
 
     llm_temperature: float = 0.0
+    llm_seed: int | None = None
     llm_streaming: bool = False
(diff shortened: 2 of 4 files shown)


dexters1:
@Vasilije1990 Maybe we should only support these types of fine tune LLM argument setting through the LLM_ARGS environment variable in system, to make it simpler to maintain on our side

Temperature, max tokens and the rest of the various params can be set through it directly

Issue #2717, opened 2026-04-24
[Bug] SQLite deadlock during cognify() — still reproducible in v1.0.2


pocklabsdotcom:
**Reported by:** @Tim-Rambo

## Summary

`cognee.cognify()` fails with `sqlite3.OperationalError: database is locked` when processing datasets containing more than a handful of documents. This blocks any production use of Cognee with non-trivial data volumes on the default SQLite backend.

The root cause is a **fundamental mismatch between Cognee's task engine and SQLite's concurrency model**: Cognee's pipeline spawns parallel greenlets that all attempt to write to the same SQLite relational database simultaneously. SQLite supports only one concurrent writer. The greenlets deadlock.

## Updated Context (2026-04-24): We Already Tried the API Client

**Cognee provides a documented workaround:** `cognee/cli/api_client.py` and the `--api-url` flag route all CLI commands through a single FastAPI server process, serializing all database access through one connection owner.

From the official codebase (`cognee/cli/api_client.py`, lines 1-11):

> """Thin HTTP client that delegates CLI commands to a running Cognee API server.
>
> When ``--api-url`` is supplied, every CLI command is forwarded as an HTTP request
> to the server instead of being executed in-process. **This avoids file-based
> database locking issues** (SQLite, KuzuDB, LanceDB) that arise when multiple CLI
> processes try to access the same files concurrently.
>
> **The single API server process owns all database connections and serialises
> writes, which is the correct model for file-based backends.**"""

### What We Tried

We followed this recommendation and ran the FastAPI server (`cognee.api.client`, port 8000). We then used the HTTP client (`httpx` via `api_client.py`) with `--api-url http://localhost:8000`.

**Result:** The lock still occurs. The FastAPI server itself is single-process, but **internally, `cognify()` spawns parallel greenlets** that all attempt concurrent `UPDATE data` statements within the same process. The single-process API server does NOT serialize the internal greenlet writes — it only serializes access from *external* CLI processes.

### The Real Bug

The `api_client.py` workaround solves the **multi-process** file-locking problem. Our bug is the **intra-process** greenlet-locking problem within `cognify()` itself.

So this is not a documentation issue — it's a genuine concurrency bug in the `cognify()` pipeline that even the API server cannot avoid.

## Impact

- **Production blocker**: Any dataset >50 documents reliably triggers the error.
- **Silent corruption risk**: Some greenlets partially commit before the lock error, leaving the database in an inconsistent state between relational and vector stores.
[...]

PR #4482, opened 2026-08-14
fix: Lower add data_per_batch default to 100

dexters1:
## Description

`add()`'s `data_per_batch` default of 2000 effectively runs every item of a large add fully concurrently. Each in-flight item writes its `Data` row in parallel, and on the default SQLite backend hundreds of simultaneous read-then-write transactions hit WAL **snapshot-upgrade conflicts**: `database is locked` raised immediately, bypassing the existing 120s `busy_timeout`/WAL pragmas (issue #2717's mitigations don't cover this mode — the busy handler is not invoked for `SQLITE_BUSY_SNAPSHOT`).

Observed live: a 448-item directory add on SQLite failed 34 seconds in with a plain `INSERT INTO data` lock error, despite WAL + 120s timeouts being active. The same add at bounded concurrency completes cleanly.

This lowers the default to 100 — the single default all add surfaces inherit (SDK `add()`, the HTTP router, and `remember()` all flow through it). Callers can still pass a higher value explicitly. Postgres deployments are unaffected either way.

Note: deliberately **not** raising the SQLite timeouts — they are already 120s and the observed failure was not a timeout expiry; a bump would be a placebo for this failure mode.

## Testing
- Signature default verified (100), ruff clean.
- Live A/B on cognee's own repo directory (448 items incl. images through vision): 2000 → `database is locked` at 34s; bounded concurrency → clean completion. A confirmation run of this exact PR state on SQLite is in the PR conversation.

No Linear key attached yet — `linear-issue-check` will fail until one is added.

### DIFF
--- a/cognee/api/v1/add/add.py
+++ b/cognee/api/v1/add/add.py
@@ -50,7 +50,12 @@ async def add(
     dataset_id: Optional[UUID] = None,
     preferred_loaders: Optional[List[Union[str, dict[str, dict[str, Any]]]]] = None,
     incremental_loading: bool = True,
-    data_per_batch: Optional[int] = 2000,
+    # 100, not unbounded: every in-flight item writes its Data row concurrently,
+    # and on the default SQLite backend hundreds of parallel read-then-write
+    # transactions hit WAL snapshot-upgrade conflicts ("database is locked"
+    # immediately, bypassing busy_timeout) — observed live at 448 concurrent
+    # items. 100 keeps ingestion fast while staying under that pressure.
+    data_per_batch: Optional[int] = 100,
     importance_weight: Optional[float] = 0.5,
     run_in_background: bool = False,
     llm_config: Optional[LLMConfig] = None,


dexters1:
Confirmation run of this PR state, same machine, same input (the cognee repo directory, 448 items incl. 52 images through vision transcription):

| Backend | `data_per_batch` | Result |
|---|---|---|
| SQLite | 2000 (old default) | `database is locked` **34s in**, aborted after 147 records — WAL + 120s busy_timeout active and bypassed (snapshot-upgrade conflict) |
| SQLite | **100 (this PR)** | **clean completion in 61s** — 423 records (1 repo manifest + 416 documents + 6 DLT manifests), zero lock errors |
| Postgres | 2000 | clean in 45s (multi-writer backend unaffected either way) |

The 100 default costs ~16s on this workload vs unbounded-on-Postgres while making the default-install (SQLite) path survive large directory adds.

PR #4484, opened 2026-08-14
fix: Raise on pinned-id re-add with changed content

dexters1:
## Description

Items ingested with a **pinned, stable `data_id`** (DLT source manifests; code-repo manifests once #4481 lands) were silently dropped on re-add even when their content had changed: the incremental pre-check in `run_tasks_data_item_incremental` short-circuited on *pipeline already completed* **before any content comparison**, so the stored record — and every graph artifact derived from it — went stale with no signal.

**Observed live** (2026-08-14, cognee repo ingestion): a repository manifest re-added after the tree gained 19 files kept its old `file_count=2453` while detection logged 2,472 covered files; cognify then skipped it, and the graph kept describing the pre-change repository.

## The change

The completed short-circuit now computes the incoming item's content hash first (the same save+classify formula that produced the stored `Data.content_hash`, so the comparison is exact):

- **Unchanged content** → the cheap idempotent skip stays (re-running an unchanged sync is quiet and free).
- **Changed content** → new `DataContentConflictError` (HTTP 409) naming all three ways out: `update()` for an in-place update, a different `data_id` to ingest as a separate item, or `add(..., incremental_loading=False, data_cache=False)` to force re-ingestion in place.
- **Unpinned items are untouched** by construction — their identity derives from content, so changed content is a new id and no conflict can occur.

This narrows DLT's documented idempotency contract (was: *re-add keeps the completed record, changed or not*): re-syncing **changed** data now fails loudly instead of no-opping. The contract comment in `resolve_dlt_sources.py` and its invariant tests are updated accordingly.

## Testing

- New `test_pinned_content_conflict.py`: same hash → `PipelineRunAlreadyCompleted` skip; different hash → `DataContentConflictError` with all three remedies in the message; undeterminable hash → legacy skip (no false conflicts); helper returns None for unpinned items.
- `test_dlt_stable_manifest_id.py` updated to the narrowed contract (unchanged → fast skip; changed → raise).
- Live end-to-end through the real `add()` pipeline: v1 add → unchanged re-add silently skipped (hash proven identical) → changed re-add raised 409 with the record verifiably untouched → `incremental_loading=False, data_cache=False` refreshed the record in place (content_hash changed, same single record) → two unpinned adds with different content created two records, no conflict.
- Pipelines + tasks + data unit suites: 448 passed.

Found while validating #4481's repo manifests; fixes the mechanism for both manifest families. One caveat surfaced during testing and encoded in the message: `incremental_loading=False` alone does **not** bypass the pre-check — `data_cache` also selects it, so the escape hatch is both flags (matching DLT's existing documentation).

No Linear key attached yet — `linear-issue-check` will fail until one is added.

### DIFF
--- a/cognee/modules/ingestion/exceptions/__init__.py
+++ b/cognee/modules/ingestion/exceptions/__init__.py
@@ -5,5 +5,6 @@ This module defines a set of exceptions for handling various ingestion errors
 """
 
 from .exceptions import (
+    DataContentConflictError,
     IngestionError,
 )
--- a/cognee/modules/ingestion/exceptions/exceptions.py
+++ b/cognee/modules/ingestion/exceptions/exceptions.py
@@ -10,3 +10,20 @@ class IngestionError(CogneeValidationError):
         status_code=status.HTTP_415_UNSUPPORTED_MEDIA_TYPE,
     ):
         super().__init__(message, name, status_code)
+
+
+class DataContentConflictError(CogneeValidationError):
+    """A pinned data_id already holds different content in the dataset.
+
+    Raised by the incremental add pre-check instead of silently keeping the
+    stale record: the caller must choose between updating the existing
+    document, ingesting under a different identity, or forcing re-ingestion.
+    """
+
+    def __init__(
+        self,
+        message: str = "Data id already exists with different content.",
+        name: str = "DataContentConflictError",
+        status_code=status.HTTP_409_CONFLICT,
+    ):
+        super().__init__(message, name, status_code)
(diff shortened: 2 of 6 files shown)


ArmagedonFlamer:
Thanks @dexters1 for catching the stale repo manifest during the #4481 validation and for the thorough live end-to-end write-up. We're closing this because #5017 settled the area with a different design, released in v1.6.0. add() now refuses changed content once per request with DocumentUpdateRequiredError, and pinned items are exempt because they go through update(), which rebuilds DLT manifests. 
Thanks again, looking forward to what you send next.

PR #4502, opened 2026-08-14
fix(api): serve GET /datasets/graph-summary the dashboard calls (#4431)

### LINKED ISSUE #4431: v1.4.2 dashboard calls graph-summary route/method not served by backend
### Summary

In a stock/self-hosted Cognee `v1.4.2` deployment, the `v1.4.2` dashboard frontend calls `GET /api/v1/datasets/graph-summary`, but the official `cognee/cognee:1.4.2` backend responds `405 Method Not Allowed` on that concrete request.

This causes visible browser console errors in the web interface on a stock deployment.

### Environment

- Cognee backend image: `cognee/cognee:1.4.2`
- Cognee MCP image: `cognee/cognee-mcp:1.4.2`
- Frontend: built from unchanged upstream tag `v1.4.2`
- Upstream source commit used for frontend build: `b0ea53e95caa7bfebf373ffabfb44f9fff922`
- Self-hosted Docker deployment

### Measured evidence

The `v1.4.2` frontend references the graph summary call in:

- `src/modules/datasets/getDatasetGraphSummary.ts`
- `dashboard/hooks/useGraphSummary.ts` (mentions upstream `COG-5726`)

The live stock backend responds:

```text
GET /api/v1/datasets/graph-summary -> HTTP 405 {"detail":"Method Not Allowed"}
```

### Expected behavior

The released frontend and backend should agree on the available API surface for the same stable release tag/image. Either:

1. the backend should expose the route/method the frontend calls, or
2. the frontend should guard/use the route/method actually available in `v1.4.2`.

### Actual behavior

The UI calls a graph-summary route/method that the `v1.4.2` backend does not serve as called, producing console errors in the dashboard.

### Slack note

An earlier local note also mentioned `POST /api/v1/integrations/slack/authorize`. Rechecking the live `v1.4.2` OpenAPI/runtime shows that Slack authorize **does exist** through the generic path `/api/v1/integrations/{provider}/authorize`. On an unconfigured self-hosted server it returns `503` with `"slack integration is not configured on this server."`, which is expected until the documented `SLACK_*` environment variables are set. Please treat Slack as a config-state/UX note, not as the confirmed missing-route bug here.

### Notes

This was measured on 2026-08-11 while rebuilding a self-hosted Cognee deployment back to manufacturer/stock release behavior. We intentionally did not patch locally, because this looks like a release packaging/API-surface mismatch rather than a deployment-specific defect.

Anai-Guo:
## What & why

Fixes #4431. On a stock `v1.4.2` deployment the dashboard calls `GET /api/v1/datasets/graph-summary`, but that route doesn't exist. The path matches the dynamic `DELETE /{dataset_id}` route, so a `GET` returns **405 Method Not Allowed** and floods the browser console on every dashboard load.

## Fix

Adds the missing `GET /api/v1/datasets/graph-summary` route. It returns the precomputed node/edge counts the frontend already expects (`getDatasetGraphSummary.ts` / `useGraphSummary.ts`, COG-5726) instead of the frontend having to walk the full graph via `/{dataset_id}/graph`.

Everything it needs already exists on `main`:

- **Counts** come from the cached `GraphMetrics` rows a cognify run writes in `get_pipeline_run_metrics.py`, keyed by `GraphMetrics.id == pipeline_run_id`.
- **Latest-per-dataset** selection reuses the same `row_number() OVER (partition_by dataset_id ORDER BY created_at DESC)` pattern as `get_pipeline_status.py`. Because `pipeline_run_id` is not unique in `pipeline_runs` (one row per status transition), the run→dataset map is collapsed with `DISTINCT` before the metrics join to avoid a fan-out.
- **Authorization** reuses `get_authorized_existing_datasets(dataset_ids or None, "read", user)`: omit `dataset_ids` → every dataset the caller can read; foreign/unknown ids are silently dropped — exactly the behavior the frontend TS documents.
- **Serialization** uses an `OutDTO`, so `dataset_id/pipeline_run_id/num_nodes/num_edges/computed_at` are emitted as `datasetId/pipelineRunId/numNodes/numEdges/computedAt`, matching the `DatasetGraphSummary` interface.

Datasets with no metrics computed yet report `numNodes/numEdges = 0` with a `null` `pipelineRunId`/`computedAt`.

## Note for maintainers

The response contract is inferred from the frontend `DatasetGraphSummary` interface (the current consumer only sums `numNodes`/`numEdges`). If COG-5726 intended different degraded-case semantics for `computedAt` when a run exists but its metrics row is missing, happy to adjust — but the core fix (serving the route so the `405` stops) stands regardless. This is a single-file, read-only endpoint with no new dependencies.

### DIFF
--- a/cognee/api/v1/datasets/routers/get_datasets_router.py
+++ b/cognee/api/v1/datasets/routers/get_datasets_router.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from pydantic import BaseModel, Field
 from typing import Any, Dict, List, Optional, Union
 from typing_extensions import Annotated
+from sqlalchemy import select, func
 from fastapi import status
 from fastapi import APIRouter
 from fastapi.encoders import jsonable_encoder
@@ -24,7 +25,8 @@ from cognee.modules.users.models import User
 from cognee.modules.users.methods import get_authenticated_user
 from cognee.modules.users.permissions.methods import get_all_user_permission_datasets
 from cognee.modules.graph.methods import get_formatted_graph_data
-from cognee.modules.pipelines.models import PipelineRunStatus
+from cognee.modules.data.models import GraphMetrics
+from cognee.modules.pipelines.models import PipelineRun, PipelineRunStatus
 from cognee.shared.utils import send_telemetry
 from cognee import __version__ as cognee_version
 
@@ -70,6 +72,17 @@ class GraphEdgeDTO(OutDTO):
 class GraphDTO(OutDTO):
     nodes: List[GraphNodeDTO]
     edges: List[GraphEdgeDTO]
+
+
+class GraphSummaryDTO(OutDTO):
+    dataset_id: UUID
+    pipeline_run_id: Optional[UUID] = None
+    num_nodes: int = 0
+    num_edges: int = 0
+    # ``None`` (while ``pipeline_run_id`` is null) means no metrics have been
+    # computed for the dataset yet. Serialized as ``computedAt`` for the
+    # dashboard's precomputed node/edge counts (frontend COG-5726).
+    computed_at: Optional[datetime] = None
 
 
 class DatasetCreationPayload(InDTO):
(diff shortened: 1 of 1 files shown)


dexters1:
Thanks for the PR @Anai-Guo! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

ArmagedonFlamer:
Thanks @Anai-Guo for tracking down the dashboard's 405 and for the careful write-up, especially the note on why pipeline_run_id needs de-duplicating before the join. GET /api/v1/datasets/graph-summary was already implemented on dev in #4179, just not yet released when #4431 was reported, and it has shipped since v1.5.0. With the route on both dev and main now, there's nothing left for this PR to add, so we're closing it rather than asking you to retarget and rebase. Thanks again, and we'd welcome more PRs from you.

PR #4506, opened 2026-08-15
fix(skills): pass explicit limit to avoid silent 200-skill truncation

### LINKED ISSUE #4279: Skills page silently truncates at 200 skills — frontend sends no limit/offset while `GET /v1/skills` defaults to `limit=200`

asthashukla233:
## Description
The Skills page permanently shows "200 skills" even though the dataset contains 378. There is no truncation indicator and no pagination control, so it is impossible to tell from the UI that ~half the skills are missing from the list.
Fixes #4279

## Acceptance Criteria
1. getSkills.ts now sends an explicit limit param instead of relying on
   the backend's default
2. Datasets with more than 200 skills (up to 1000) now display fully
   on the Skills page instead of silently cutting off at 200

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots
N/A — frontend query-param change, no visual UI change to screenshot.

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee-frontend/src/modules/skills/getSkills.ts
+++ b/cognee-frontend/src/modules/skills/getSkills.ts
@@ -12,6 +12,7 @@ export default function getSkills(
 ): Promise<Skill[]> {
   const params = new URLSearchParams({ dataset_id: datasetId });
   if (includeInactive) params.set("include_inactive", "true");
+  params.set("limit", "1000");
 
   return instance
     .fetch(`/v1/skills/?${params.toString()}`, {


dexters1:
Thanks for the PR @asthashukla233! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #4508, opened 2026-08-15
fix(storage): isolate metadata for multi-field indexing

### LINKED ISSUE #4491: index_data_points: pydantic model_copy() shallow-copies metadata dict, corrupting multi-field index buckets
## Bug
`cognee/tasks/storage/index_data_points.py` uses `data_point.model_copy()` (pydantic **shallow copy**) then mutates `indexed_data_point.metadata["index_fields"] = [field_name]`.

Because `model_copy()` shallow-copies the `metadata` dict (the dict itself is **shared** between the copy and the original), the mutation leaks back into the original data point. When a DataPoint has **more than one** `index_fields` entry (e.g. `Entity` with `index_fields=["name", "description"]`), every field bucket gets the **last** field name written:

- The `Entity_name` vector table ends up storing the **description** text instead of the name
- Result: name-based retrieval is silently corrupted (all `Entity_name` rows contain description content)

## Repro
```python
e = Entity(name="A", description="B", id=uuid4())
await index_data_points([e])
# Entity_name.payload.text == "B"  (should be "A")
# Entity_description.payload.text == "B"  (correct)
```

## Root cause
```python
indexed_data_point = data_point.model_copy()          # metadata dict SHARED
```
`data_point.model_copy()` → shallow copy → `metadata` is the same dict object → later loop iterations overwrite `index_fields` for ALL previously-queued copies.

## Fix (what we patched locally)
Deep-copy metadata per bucket:
```python
indexed_data_point.metadata = {**data_point.metadata, "index_fields": [field_name]}
```

This is invisible with single-field `index_fields` (the common default), which is why it ships undetected. We hit it after registering a second index field on `Entity`.

mikemikimike:
## Summary

Fixes #4491.

Issue: https://github.com/topoteretes/cognee/issues/4491

`index_data_points` creates one model copy per indexed field. The previous implementation mutated a shared `metadata` dictionary, so later fields overwrote earlier copies and corrupted the field-specific vector payload.

## Changes

- Assign a fresh metadata dictionary to each indexed copy while preserving all other metadata.
- Add a regression test covering multiple index fields, field-specific embeddable payloads, source metadata immutability, and metadata object isolation.

## Compatibility

Single-field indexing is unchanged. Multi-field indexing now isolates the metadata used by each vector bucket; no public API changes.

## Testing

- `uv run pytest cognee/tests/unit/infrastructure/databases/test_index_data_points.py -v` — 5 passed.
- `uv run ruff check cognee/tasks/storage/index_data_points.py cognee/tests/unit/infrastructure/databases/test_index_data_points.py` — passed.
- `uv run ruff format --check cognee/tasks/storage/index_data_points.py cognee/tests/unit/infrastructure/databases/test_index_data_points.py` — passed.
- `uv lock --check` — passed.
- `uv run --with pre-commit pre-commit run --all-files` — passed.

The full unit and integration suites were also run. Remaining failures are outside this change, including existing recall/code-graph behavior, LLM-key-gated integration paths, session/retrieval paths, URL crawler tests, and Windows file-lock cleanup.

## DCO Affirmation


mikemikimike:
I rechecked this PR against issue #4491. The PR targets the shallow-copy metadata corruption described in the issue and includes the corresponding fix. The remaining red jobs span unrelated E2E, database, LLM, notebook, example, migration, and integration suites, with no human changes-requested review present. This pattern looks like repository/environment-wide CI failure rather than a failure introduced by this focused change. Could a maintainer please review the fix and advise which required checks are meaningful or need to be rerun?

Vasilije1990:
@mikemikimike this has conflicts + does't do anything to index data point file

mikemikimike:
Thanks for pointing this out. I updated the branch with the current `dev`, and the PR now has no merge conflicts.

The production change is in `cognee/tasks/storage/index_data_points.py`: each indexed data-point copy keeps the shallow `model_copy()` behavior but receives a fresh metadata dictionary with only the current `index_fields` entry. This prevents later fields from overwriting earlier indexed batches and lets `DataPoint.get_embeddable_data()` select the correct field payload.

I also added regression coverage in `cognee/tests/unit/infrastructure/databases/test_index_data_points.py` for independent metadata, source immutability, and field-specific embeddable payloads.

Local verification:
- `uv run --frozen pytest cognee/tests/unit/infrastructure/databases/test_index_data_points.py -v` — 5 passed.
- Docker CI-compatible Python 3.11.14 gate — focused pytest 5 passed; Ruff E/F checks, format check, and `git diff --check` passed.

The PR description has been updated with these results. Remote GitHub Actions may still need maintainer approval for this fork PR.

dexters1:
Thanks for the PR @mikemikimike! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

mikemikimike:
Updated this PR in commit 7cfd575d4e2b9588ab56e4d61a9d7f620860552a. The production fix is in cognee/tasks/storage/index_data_points.py: each indexed copy now receives a fresh metadata dictionary, so multi-field indexing keeps field-specific payloads and does not mutate the source DataPoint. Added regression coverage for metadata isolation, source immutability, and embeddable payload selection. Focused tests: 5 passed. Ruff, format, lock, pre-commit, and type checks passed. The full unit and integration suites were also run; remaining failures are outside this change in existing recall/code-graph, LLM-key-gated integration, session/retrieval, URL crawler, and Windows file-lock paths.

PR #4514, opened 2026-08-15
perf: Set-based fork re-key on Ladybug (COG-6185)

dexters1:
## Description

Migrating a 100k-node fork on Ladybug took **54 minutes**, and two phases were nearly all of it. Both are replaced by statements the engine answers in **one scan**. Everything lives in the migration scripts — **no adapter changes** — and both shortcuts fall back to today's behaviour on backends that cannot answer them.

Stacked on #4512 (ladybug 0.19.0 pin + single-clause edge MATCH), so review that first.

### 1. Provenance move: ~400k point updates → 2 statements

The generic path lists every artifact carrying the old source ref, reads its provenance, attaches the new key and removes the old one — **3 reads + 2 writes** over the whole subgraph, each a per-row primary-key probe.

Ladybug stores provenance as delimiter-wrapped strings, and a re-key only changes the **data-id half** of the key, so the entire move is a substring rewrite:

```cypher
MATCH (n:Node)
WHERE n.source_ref_keys CONTAINS $old_key AND NOT n.source_ref_keys CONTAINS $new_key
SET n.source_ref_keys = replace(n.source_ref_keys, $old_key, $new_key),
    n.source_run_refs = replace(n.source_run_refs, $old_key, $new_key)
```

`source_run_refs` embeds the key verbatim, so the same replace fixes it. `source_dataset_ids` and `source_run_ids` contain no data id and stay correct untouched.

### 2. Edge re-assert: 102,340 blind upserts → only what actually vanished

Step 4 of `_migrate_graph` re-asserts every survivor edge incident to a rewritten neighbourhood, because a detach-delete *could* drop them. On this fixture that is 102,340 writes costing ~22 minutes — and the instrumented run reports:

```
namespace migration: 0 of 102340 at-risk edge(s) were dropped and need re-asserting
```

**Zero.** The whole phase was repairing nothing. Listing the edge table once answers the question directly, and only genuinely dropped edges are rewritten — so the safety net still works if a backend ever drops them again.

## Measured

Same 100k fixture, same real `text-embedding-3-large` embeddings, both runs restored from one pre-migration snapshot so they are directly comparable:

| | baseline | this PR |
|---|---|---|
| `run_migrations` | **3263.4s** (54.4 min) | **1255.0s** (20.9 min) |
| provenance move | ~9 min | two statements |
| edge re-assert | ~22 min (102,340 writes) | 0 writes — one scan |

**2.6× end-to-end.** What remains is dominated by the real-embedding vector sync and the verification suite, not graph writes.

## Output is identical

Compared against **both** the pre-migration seed and the unoptimized baseline run:

- relationship-name distribution identical on both datasets (all ~1,800 relationship types at their original counts)
- node-type distribution identical
- 100,722 nodes / 295,290 edges on both datasets
- **0 identity-less rows, 0 empty-endpoint edges**, distinct ids == node count
- migration chain at head with no errors, fork split verified, CHUNKS search executes, post-migration ingest OK

## Safety

- Both shortcuts are gated on the graph provider (`ladybug` / `kuzu`) and wrapped so an engine that rejects the statement falls back to the existing path — behaviour off Ladybug is unchanged.
- Artifacts carrying **both** the old and the new key (a run interrupted mid-move) are deliberately excluded from the rewrite, since a blind `replace` would duplicate the key. They fall through to the generic path, which dedupes — **convergence on re-run is preserved**.
- 8 new unit tests cover: only-dropped-edges narrowing, no-op when nothing dropped, fallback when edge listing is unsupported, the two-column rewrite shape, fallback when `replace` is unavailable, skipping the generic sweep once the fast move clears the key, still sweeping the both-keys residue, and non-Ladybug backends keeping the generic path.
- 584 tests pass across `cognee/tests/unit/infrastructure/databases/` and `cognee/tests/unit/modules/migrations/`.

Part of COG-6185.


cognee/tests/unit/modules/migrations/test_ladybug_migration_fast_paths.py
dexters1, line 1: Should be deleted before merge

PR #4534, opened 2026-08-17
fix: rollback must not clear cognify markers of already-completed data

plexus77:
## The problem

`cognify_rollback_handler` clears the cognify marker of records the failed run never processed. One unprocessable document costs the **whole dataset** its markers.

The path, on current `dev`:

1. **`run_tasks_data_item.py`** — an item already extracted by an earlier run is skipped, and yields `{"run_info": PipelineRunAlreadyCompleted(...), "data_id": ...}`. Note it still carries a `data_id`.
2. **`run_tasks.py`** — appends that to `results`, raises `PipelineRunFailedError` if *any* item errored, then hands the rollback `data_ingestion_info=results` — the whole list, errored and skipped alike.
3. **`rollback.py`** — `_extract_data_ids` walks every entry and takes `data_id` **without inspecting `run_info`**. `_reset_pipeline_status` then deletes `pipeline_status["cognify_pipeline"][dataset_id]` for all of them.

## Why it's a bug rather than a deliberate choice

The graph and vector deletion in the same handler **is** correctly scoped by `pipeline_run_id`, so it only removes the failed run's own nodes and edges. The marker reset is not scoped at all.

The two halves therefore disagree. Records extracted by an earlier run keep their graph artifacts but lose their markers: the marker says "not extracted", the graph says otherwise, and the next `cognify` re-extracts them at full LLM cost. Nothing detects it, because the run reports a rollback rather than corruption.

We hit this on a 400-record dataset. A single unprocessable document cleared markers for all 400; 260 were still present in the graph and were re-extracted on the following run.

## The fix

Skip `PipelineRunAlreadyCompleted` entries in `_extract_data_ids`, restoring the invariant that a rollback clears markers only for data the run actually touched. `PipelineRunCompleted` and `PipelineRunErrored` entries are still collected, so a genuine partial failure still rolls back fully.

The graph-provenance branch gets the same benefit, since it unions this helper's result with the ids read from provenance.

Three lines of behaviour change; the rest of the diff is the explanatory comment and tests.

## Tests

Two cases added to `cognee/tests/unit/modules/cognify/test_rollback.py`:

- `test_extract_data_ids_skips_already_completed_items` — the helper directly.
- `test_rollback_preserves_markers_of_previously_extracted_data` — through `cognify_rollback_handler`, with no nodes or edges from the failed run so `data_ingestion_info` is the only source of target ids.

Both **fail on current `dev` and pass with the fix**. The three existing rollback tests pass either way.

```
5 passed
```

## Notes

- Verified against `dev` (`3aba120`); `rollback.py` is byte-identical in `v1.4.2` and `v1.5.0`, so this affects both released versions.
- Related to #4233, which fixed a different problem in the same area (the memory-fragment projection). The rollback trap itself was never addressed — #4233 removed a common *trigger*, not the trap.

### DIFF
--- a/cognee/modules/cognify/rollback.py
+++ b/cognee/modules/cognify/rollback.py
@@ -14,6 +14,7 @@ from cognee.modules.graph.legacy.has_edges_in_legacy_ledger import has_edges_in_
 from cognee.modules.graph.legacy.has_nodes_in_legacy_ledger import has_nodes_in_legacy_ledger
 from cognee.modules.graph.methods.delete_from_graph_and_vector import delete_from_graph_and_vector
 from cognee.modules.graph.models import Edge, Node
+from cognee.modules.pipelines.models.PipelineRunInfo import PipelineRunAlreadyCompleted
 from cognee.shared.logging_utils import get_logger
 
 logger = get_logger("cognify.rollback")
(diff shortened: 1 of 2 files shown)


siillee (approved):
Hey @plexus77, thanks for opening this PR! 

This was a nice catch, and your solution looks good. I verified the PR against our CI via this PR #4536, and all is green, so this is good to go!

PR #4538, opened 2026-08-17
feat(llm): Stream answer tokens through a side-channel, off by default [CLO-556]

NMZivkovic:
First of two SDK PRs for LLM streaming on cloud recall (CLO-556). **This one is the engine — nothing is exposed over HTTP yet, and with the flag off it is inert.** PR 2 adds the SSE transport; cloud consumes both after a release.

## Why a side-channel and not a generator

Every hop from an HTTP handler down to litellm is single-value:

```
recall()                    → list[RecallResponse]      (materialised)
  authorized_search         → SearchResultPayload        (pydantic)
    get_completion_from_context → List[Any]
      generate_completion   → Any                        (one str)
        LLMGateway          → T
          adapter           → response.choices[0].message.content
```

`SearchResultPayload.completion` has a `field_serializer` that stringifies anything unrecognised, and `commit_turn` needs the finished string to write session memory. Threading a generator through means changing six return types and breaking the session write, usage accounting and serialisation on the way.

So tokens travel sideways instead: **the adapter pushes each delta into a sink while still returning the complete string.** Every existing signature is untouched, so retrieval, the session lock, `add_qa` and usage accounting run byte-for-byte as before. If nothing is listening — or the request takes a path that cannot stream — the caller gets the identical payload with zero events.

## Two ContextVars, both load-bearing

`requested_token_sink` is set once at request entry so every task inherits it. `active_token_sink` is promoted **only** around the answer call, in `complete_turn`.

That split is the whole design. The live cloud path runs `asyncio.gather(analyze_turn, _retrieve_and_answer)` — **two concurrent LLM calls** — and `gather` snapshots context per task, so a variable set inside the answer task is structurally invisible to its sibling. A single request-scoped flag would interleave `SessionTurnAnalysis` output into the user's visible answer.

`try_claim` handles the other fan-out: `search_in_datasets_context` runs one answer call per authorised dataset. First claim wins; the rest run normally and still contribute to the final payload.

## Adapter details worth reviewing

- Skips chunks whose `choices` list is empty. That is the final `include_usage` chunk, and `chunk.choices[0]` on it is the single most common way to break a streaming integration.
- Skips role-only / finish chunks whose `delta.content` is `None`.
- The `async for` stays **inside** `llm_rate_limiter_context_manager()`. `acompletion(stream=True)` returns as soon as the connection opens, so releasing there would leave generations effectively unlimited.
- Exceptions **propagate** so the existing tenacity retry still fires. `begin_attempt()` emits a `reset` on re-entry, so a retried call does not leave the consumer holding two copies of the answer.

## Not `llm_streaming`

That flag already exists and is broken: unread by OpenAI and Azure, absent from every other provider, and on Bedrock it injects `stream=True` into the *instructor* path where nothing consumes a stream. It is also part of the adapter LRU cache key, so flipping it changes adapter identity. This adds `llm_answer_streaming` (env `LLM_ANSWER_STREAMING`) instead and leaves the old one alone for separate deprecation.

## Testing

**814 existing tests pass** across `infrastructure/llm`, `infrastructure/session` and `modules/retrieval` — the check that matters, since the claim is "inert with the flag off".

17 new tests. The ones worth reading:

- `test_only_the_answer_task_can_stream` — the isolation invariant, run through a real `asyncio.gather` with forced interleaving. If ContextVar-per-task scoping regresses, this fails.
- `test_second_concurrent_answer_call_does_not_stream` — multi-dataset fan-out claims once.
- `test_usage_chunk_with_empty_choices_does_not_raise` — regression guard for the `choices[0]` bug.
- `test_retry_emits_reset_so_the_answer_is_not_duplicated`
- `test_flag_off_ignores_an_active_sink` and `test_no_sink_takes_the_original_non_streaming_path` — the inertness claims, asserted rather than assumed.
- `test_mid_stream_failure_propagates_so_retry_still_fires`

## Next

PR 2 adds `recall_stream.py`, SSE encoding and `Accept: text/event-stream` negotiation inside the existing recall handler (a separate route would bypass the pod's credit guard). Then an SDK release, then cloud enables the env flag.

**One thing to settle before cloud turns this on:** `disable_spend_logs: true` on the LiteLLM gateway means budget enforcement reads the aggregate spend column. If streamed calls do not move it, streaming is silently free and the cloud credit guard goes blind. `stream_options={"include_usage": True}` is set here as the mitigation, but it needs verifying against `/key/info` on a real tenant.


NMZivkovic:
Heads-up for anyone mid-review: the diff moved. `de9c392` reworks how the streaming decision is made, in response to review feedback.

**Before:** `complete_turn` held `async with stream_answer_tokens(...)`, so `infrastructure/session/` imported `infrastructure/llm/streaming/` and the session turn decided which LLM call in the whole request a client may watch — two unrelated jobs in one function.

**After:** the decision has a name and lives where answers are generated. `generate_answer` sits beside `generate_completion` in `completion.py` and is the only completion a listening client may watch; choosing it *is* the decision to stream. `complete_turn` calls it and imports nothing from streaming. Turn analysis, summarisation, subquery answers and agentic steps keep calling `generate_completion`, so they can never take the stream — without any of them knowing streaming exists.

`stream_answer_tokens` is now `answer_scope`. A scope that usually promotes nothing is ordinary; a verb that usually does nothing is not, and the old name read like a promise its internal checks routinely decline. It stays in `token_sink.py` — it reads the feature switch and the adapter capability, which is streaming policy.

**Behaviour is unchanged.** Promotion still happens inside the answer task, so `asyncio.gather`'s per-task context copy still keeps `SessionTurnAnalysis` out of the user's stream.

The point of the change is the third one, though: the rule had no name, so nothing could pin it. Deleting the `async with` used to leave SSE silent with the whole suite green. Two new tests now drive `complete_turn` with a requested sink and assert the sink is active inside the real completion (not a mocked `generate_answer` — a mock would hide exactly what is being asserted), plus the negative case. Reverting `complete_turn` to `generate_completion` now fails four tests; before this commit it failed none — verified by actually making that mutation.

One consequence worth calling out for scope: sessionless and sequential recall stream nothing today, and this change makes enabling them a one-line change each (point them at `generate_answer`) rather than copying a context manager into another module. Whether CLO-556 wants that is still open.

lxobr (dismissed):

PR #4561, opened 2026-08-18
feat: Extend pipeline_runs to all cognee operations (SDK-399)

Vasilije1990:
## Description

`pipeline_runs` becomes the durable, queryable record of **every** cognee operation — answering "what did this user run, did it work, how long did it take, what did it cost in tokens" with one indexed SQL query, no JSON parsing.

Design decisions (this PR description is the written decision record for SDK-399, to be linked from CLO-395 / CLO-451):
- **System of record**: OSS `pipeline_runs` is the durable record of all operations; cloud activity_log (CLO-395/CLO-451) and tenant_logs (COG-6073) are downstream projections.
- **Row model**: append-only stays; the terminal row is self-contained (start/end/user/tokens/outcome on one record). Non-pipeline operations write exactly one row.
- **Status vs outcome**: `PipelineRunStatus` is frozen (public API + native PG enum); operation result lives in a plain-String `outcome` + `error_class`.

### What's new

- **Coverage**: `forget`, `recall`, `search`, `remember`, `delete`, `prune` now write one self-contained row each via a new `record_operation()` context manager (`cognee/modules/operations/`); pipeline terminal rows carry the same fields. The recorder is fail-open: it can never break the operation it records.
- **Schema**: 14 new nullable columns — `user_id`, `tenant_id`, `operation_name`, `started_at`/`ended_at`, `outcome`, `error_class`, `error_message` (PII-scrubbed, ≤512 chars), `tokens_in`/`tokens_out`, `origin`, `session_id`, `parent_operation_id`, `background`. One idempotent Alembic migration (`a7f3c9e1b5d2`), old rows stay NULL, no backfill. Append-only pattern, status enum, and `run_info` untouched; NULL-status operation rows are invisible to all legacy status readers.
- **Exact token accounting**: a session-independent accumulator captures tokens even outside session scope; counts are provider-billed `response.usage` (includes hidden reasoning tokens), cross-checked live against OpenAI-reported usage. Nesting is an explicit `parent_operation_id` tree whose edges mirror the token chain, so children of any row sum to ≤ the row itself.
- **Origin attribution**: `sdk` / `api` / `cli` / `mcp` / `background` stamped per surface.
- **Structured output default → `litellm_native`**: no instructor in the default call path (instructor/BAML stay as opt-ins); `get_max_chunk_tokens` no longer instantiates an LLM client. Remaining instructor touchpoints ticketed as SDK-420.

### Contains PR #4344 (COG-6062)

Also merged: the telemetry identity fix (`send_telemetry` resolves User/UUID/str itself and carries `tenant_id`; ~20 router call sites stop stringifying User objects). One conflict resolved in `recall.py` (kept both the response_model folding and the pass-the-User-through telemetry line); its test suite adapted to dev's newer `_TELEMETRY_TASKS` tracking. When this merges, #4344 can be closed. cc @LStromann

### Contains PR #4342 (CLO-434)

This branch merges `feature/clo-434-...` (per-agent session usage + the canonical `_exact_usage_from_result` exact-capture helper) rather than duplicating its LLMGateway change. When this merges, #4342 can be closed. cc @goran-radonic

### Drive-by fix

sqlite `get_table()` reflected the on-disk schema into the shared declarative `Base.metadata`, breaking later model imports (reproduced on a clean tree via `test_app_registration`). Now declarative-first with a throwaway MetaData for unknown tables.

### Testing

41 new unit tests (`cognee/tests/unit/modules/operations/`): non-pipeline op end-to-end, failed op with error class, entry-point wiring, token capture exact/fallback, parent-chain nesting, PII scrubbing. Full unit regression green (2251+ tests); migration exercised up/down/re-up against a pre-upgrade sqlite. Live verification: remember→recall flows recording exact billed tokens across sdk/cli/background origins.

Recording overhead measured at ~2.5 ms per operation (one INSERT, off the hot path); token accumulation ~0.1 µs per LLM call; zero added LLM/network calls.


cognee/modules/agents/models.py
```diff
@@ -7,7 +7,26 @@
...
-AgentConnectionType = Literal["sdk", "api", "mcp", "claude_code", "opencode", "workflow", "unknown"]
+# Free-form on purpose: the set of client types keeps growing (Claude Code,
+# Codex, Slack, OpenCode, Cursor, Windsurf, ...) and gating it behind a
+# closed Literal meant every new integration needed a backend code change
+# just to be recognized. Callers should self-declare their type via
+# ``RegisterAgentRequest.type`` at registration; ``derive_connection_type()``
+# is only a best-effort fallback for sessions that never registered. This
+# list is documentation, not an enum member set — pick a value from it when
+# your client matches one, otherwise just use your own lowercase name.
+KNOWN_AGENT_CONNECTION_TYPES = (
```
NMZivkovic, line 21: there is value "unknown" in the "KNOWN_..." variable.

cognee/modules/pipelines/operations/run_tasks.py
```diff
@@ -66,8 +67,13 @@ async def run_tasks(
         dataset = await session.get(Dataset, dataset_id)
 
     pipeline_id = generate_pipeline_id(user.id, dataset.id, pipeline_name)
-    pipeline_run = await log_pipeline_run_start(pipeline_id, pipeline_name, dataset.id, data)
+    pipeline_run = await log_pipeline_run_start(
+        pipeline_id, pipeline_name, dataset.id, data, user=user
+    )
     pipeline_run_id = pipeline_run.pipeline_run_id
+    run_started_at = getattr(pipeline_run, "started_at", None) or getattr(
+        pipeline_run, "created_at", None
```
NMZivkovic: created_at is dead code

cognee/modules/operations/scrub_error.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+def scrub_error_message(message: Optional[object]) -> Optional[str]:
+    """Redact PII/secret-shaped substrings and truncate to the column bound."""
+    if message is None:
+        return None
+    text = str(message)
+    if not text:
+        return None
+    for pattern, replacement in _SCRUB_PATTERNS:
```
NMZivkovic, line 51: Small eficiancy note: All regexes run on the full str(exception), that can be improved.

cognee/modules/pipelines/operations/log_pipeline_run_error.py
```diff
@@ -26,6 +35,20 @@ async def log_pipeline_run_error(
             "data": data_info,
             "error": str(e),
```
NMZivkovic: The whole error is still written here, so error_message=scrub_error_message(e) looses the point.

cognee/modules/pipelines/operations/run_tasks.py
```diff
@@ -78,148 +84,171 @@ async def run_tasks(
...
-            for i, result in enumerate(gathered):
-                if isinstance(result, BaseException):
-                    logger.error(f"Item {i} failed: {result}", exc_info=result)
-                    results.append(
-                        {
-                            "run_info": PipelineRunErrored(
+    # parent_run_scope makes nested runs (a pipeline started by one of our
+    # tasks, or a recorded operation called mid-pipeline) parent to THIS run,
+    # mirroring how their tokens chain into run_usage.
+    with operation_usage_scope() as run_usage, parent_run_scope(pipeline_run_id):
```
NMZivkovic, line 90: We could have a bug here, it is more of a hunch.
What happens when tokens are reset from another context?

cognee/alembic/versions/a7f3c9e1b5d2_extend_pipeline_runs_operation_record.py
```diff
@@ -0,0 +1,125 @@
+"""Extend pipeline_runs into the operation record (SDK-399)
```
NMZivkovic, line 1: This migration is ok, I regenerated and got the same output.

NMZivkovic (changes requested):
The only agreement we need to have is regarding storing the whole error message.

Other comments are minor.

cognee/modules/agents/models.py
Vasilije1990: Fixed in 9762c00f6 — removed `"unknown"` from the tuple with a note that it is the `derive_connection_type()` fallback, not a declarable type.

cognee/modules/pipelines/operations/run_tasks.py
Vasilije1990: Fixed in 9762c00f6 — dropped the `created_at` fallback (start rows always carry `started_at`; the `getattr` stays only because unit tests stub the writer with plain namespaces).

cognee/modules/operations/scrub_error.py
Vasilije1990: Fixed in 9762c00f6 — regexes now run over a bounded 2KB window (`_SCRUB_WINDOW = 4 × ERROR_MESSAGE_MAX_LENGTH`) before the final 512-char truncation, so exceptions embedding huge payloads cannot make scrubbing expensive. The window is larger than the final bound so redaction shrinkage cannot starve the output.

cognee/modules/pipelines/operations/log_pipeline_run_error.py
Vasilije1990: Agreed — fixed in 9762c00f6: `run_info["error"]` is now `scrub_error_message(e)` too, so raw exception text is never persisted anywhere. Nothing in-repo reads `run_info["error"]` (only this writer sets it), so I kept the key for any external consumers but scrubbed its value; this also keeps run_info bounded (COG-5359).

cognee/modules/pipelines/operations/run_tasks.py
Vasilije1990: Good hunch — there is a real (narrow) case. The scope is opened inside the `run_tasks` async generator, and generators have no context of their own: if the generator is finalized from a different context than the one that last resumed it (executor teardown / GC), `ContextVar.reset(token)` raises `ValueError` inside `finally` and would mask the pipeline's real error. Token *totals* are safe either way — `OperationUsage` is a shared object mutated in place, so counts cannot be lost or double-reset; only the var cleanup can misfire. Fixed in 9762c00f6: both scope cleanups tolerate the cross-context reset (try/except ValueError with a comment explaining why).

cognee/alembic/versions/a7f3c9e1b5d2_extend_pipeline_runs_operation_record.py
Vasilije1990: Thanks for regenerating and confirming — no action taken here.

Vasilije1990:
All six review comments addressed in 9762c00f6:

- **The main agreement item** — raw error text is now gone everywhere: `run_info["error"]` gets the same PII scrub as the `error_message` column (nothing in-repo reads that key; kept it, scrubbed, for external consumers).
- Scrubber regexes bounded to a 2KB window before the 512-char truncation.
- Dead `created_at` fallback removed from `run_tasks`.
- `"unknown"` removed from `KNOWN_AGENT_CONNECTION_TYPES`.
- The cross-context token-reset hunch was a real narrow case (async-generator finalization from another context) — scope cleanups are now tolerant; details in the inline reply.

Suites re-run green: operations/pipelines/agents/session_lifecycle (136) + unit/api (416).

NMZivkovic (approved):

PR #4568, opened 2026-08-18
feat(retrieval): add Personalized-PageRank graph-completion retriever

RajdeepKushwaha5:
## What

Adds a new search type `GRAPH_COMPLETION_PPR` and `GraphCompletionPPRRetriever`, a graph-structural alternative to the default graph-completion ranking. It is fully opt-in via the new search type; no existing path changes.

Opened at @Vasilije1990's suggestion to add this as a new retriever.

## Why

The default graph-completion path ranks every triplet purely by the vector distance of its own elements (`CogneeGraph._calculate_query_top_triplet_importances`). So the graph decides *which* facts are candidates and expands the neighborhood, but the *ranking* is effectively vector-only: `importance_weight`, the one field in that scorer that could carry graph signal, is a constant `0.5` in the default pipeline (and is not projected onto edges), so `(2 - importance_weight)` is a uniform multiplier with no effect on relative order.

The consequence: a fact that sits between several query-relevant entities (a "bridge") is scored the same as an isolated fact of equal text similarity. That is exactly the multi-hop case a vector-only ranking misses.

## How

`ppr_triplet_search` seeds Personalized PageRank on the vector hits and lets relevance spread through the projected fragment, so bridge facts accrue activation and surface. This is the mechanism HippoRAG (Gutierrez et al., NeurIPS 2024) uses for single-step multi-hop retrieval, applied to the entity graph cognee already builds.

- The retriever subclasses `GraphCompletionRetriever` and overrides only `get_triplets`; all context-building and completion behavior is inherited.
- The vector step, graph projection, and distance mapping are reused verbatim from the brute-force path; only the final ranking differs.
- The final score blends normalized PageRank with direct vector similarity via a tunable `ppr_weight` (higher = more structural). `ppr_alpha` (damping) and a one-hop neighborhood expansion are configurable through `retriever_specific_config`.
- Everything is "higher is better" and non-matches contribute 0, so no penalty term can dominate.
- Robust by design: PageRank failures degrade to similarity-only ranking; an adapter without neighborhood support falls back to an ID-filtered projection; no vector hits returns `[]`.

## Cost / scale

- No extra LLM calls (PageRank is graph math).
- Query-time PPR runs on the already-projected local fragment, not the whole graph, so its cost is bounded by the fragment size. The natural next step for very large graphs is a local-push / approximate PPR (the technique that makes HippoRAG scale); happy to add it if wanted.

## Tests

New offline unit tests (`test_ppr_triplet_search.py`): the headline test proves a bridge triplet outranks a leaf triplet that has no direct similarity (a vector-only ranking would tie them), plus empty/edge-case handling, `top_k`, PageRank-failure fallback, input validation, and dispatch wiring. Existing dispatch and result-normalization tests were updated. `ruff format` and `ruff check` are clean; the search + retrieval unit suites pass.

## Suggested follow-up (not in this PR)

To quantify the win, run a multi-hop QA benchmark (MuSiQue / 2WikiMultiHopQA) comparing `GRAPH_COMPLETION` vs `GRAPH_COMPLETION_PPR`, sweeping `ppr_weight`. Prediction: little change on single-hop, a lift on multi-hop that widens as the graph grows. Happy to wire that up.


dexters1:
Thanks for the PR @RajdeepKushwaha5! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #4572, opened 2026-08-19
fix(graph): delete unowned edges before nodes in source-ref removal

yarrrly:
## Description

`execute_source_ref_removal` in `provenance_delete_planner.py` hard-deletes unowned nodes before it deletes unowned edges. Backends implement `delete_nodes` as a detaching delete: Neo4j issues `DETACH DELETE`, Kuzu/Ladybug detaches as well. Deleting an unowned endpoint therefore also destroys relationships that other source refs still own, including relationships the removed refs never matched.

I hit this on a dataset built from a Markdown repository where the same entities are referenced from many files. Content-hash identity makes every edited file a delete plus a re-add, so `delete_by_source_ref` is a hot path rather than a rare operation. On a fixture exported from that graph (7,615 nodes, 16,818 logical relationships), removing 10 source refs through the current planner drops 167 protected nodes and 354 protected edge rows. Reversing the iteration order changes the node count to 212 and leaves the same 354 edge rows missing, so the exact damage depends on traversal order but the class does not.

The planner now deletes the unowned edges first. It then asks `GraphDBInterface.get_edges(node_id)` for each node that looked unowned. A node with any remaining incident relationship is retained: it loses the targeted refs and keeps its vectors and its foreign relationships. Only isolated nodes are hard-deleted, so detach semantics cannot reach anything foreign-owned. Because the check runs after the unowned edges are already gone, an edge that is still incident is by definition a surviving one, and the check does not need to compare edge direction.

Unowned artifacts keep their source refs until their own hard delete. That preserves the Part 0 retry invariant: a failure at any step leaves them discoverable by source ref, and a retry converges.

No new adapter method was needed. The retention check goes through the existing public `get_edges` and costs one indexed lookup per candidate node rather than a scan.

The ordering dates back to #3299, which introduced the planner. The file is identical on `v1.5.0` and on current `dev`.

## Acceptance Criteria

The planner unit suite passes with the fix and fails without it:

```
uv sync --group dev
uv run pytest cognee/tests/unit/infrastructure/databases/provenance/test_unified_graph_provenance_delete.py -q
```

With the fix: `19 passed`.
With the planner reverted to its previous ordering: `5 failed, 14 passed`.

The suite could not catch this class before, because its fake graph engine deleted nodes without detaching. The fake now models detaching `delete_nodes` like the real backends and implements `get_edges`. The new regressions cover an endpoint carrying a foreign edge outside the matched set, a shared detached edge, dataset-scoped multi-ref retention, an idempotent repeat, order dependence, a crash between the edge delete and the node delete, and a retained NodeSet keeping its tag.

Wider suites on the same tree, Python 3.13.12, macOS arm64:

```
uv run pytest cognee/tests/unit/infrastructure/databases -q
uv run pytest cognee/tests/integration/infrastructure/graph -q
```

```
563 passed, 16 skipped
87 passed, 39 skipped, 4 xfailed, 2 xpassed
```

The two xpassed cases are long-standing `KuzuAdapter.get_graph_metrics` quirks and are unrelated to this change.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots

```
$ uv run pytest cognee/tests/unit/infrastructure/databases/provenance/test_unified_graph_provenance_delete.py -q
19 passed, 10 warnings in 0.05s
```

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/databases/unified/provenance_delete_planner.py
+++ b/cognee/infrastructure/databases/unified/provenance_delete_planner.py
@@ -6,13 +6,22 @@ source ref remains -> hard delete) versus which merely *survive* (some ref
 remains -> detach the targeted refs only). It then performs the removal in a
 retry-safe order:
 
-  1. delete vectors for unowned artifacts (from snapshots only),
-  2. ``remove_*_source_refs`` for the targeted refs on ALL matched artifacts
-     (idempotent),
-  3. ``delete_nodes`` / ``delete_edge_triples`` for the unowned artifacts.
+  1. delete per-edge triplet vectors for unowned edges (from snapshots only),
+  2. ``remove_*_source_refs`` for the targeted refs on SURVIVING matched
+     artifacts (idempotent),
+  3. ``delete_edge_triples`` for the unowned edges (edge-first),
+  4. retention check: an unowned-candidate node that is still an endpoint of any
+     remaining relationship is NOT hard-deleted — backends implement node
+     deletion as a detaching delete, which would destroy foreign-owned incident
+     relationships. Such nodes only lose the targeted refs (idempotent) and keep
+     their vectors,
+  5. delete vectors for the remaining truly-unowned nodes, then ``delete_nodes``
+     for them (they have no incident relationships at this point, so the
+     backend's detach semantics cannot remove anything foreign-owned).
 
-Vectors are deleted first so that a failure leaves graph provenance intact and a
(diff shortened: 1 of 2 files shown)


dexters1:
Thanks for the PR @yarrrly! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #4591, opened 2026-08-20
Fix metadata sharing when indexing multiple fields

### LINKED ISSUE #4491: index_data_points: pydantic model_copy() shallow-copies metadata dict, corrupting multi-field index buckets

aravmittal123:
## Description

Fixes #4491.

When a DataPoint contains multiple `index_fields`, `model_copy()` creates a shallow copy and the `metadata` dictionary remains shared between indexed copies.

As a result, processing a later field can overwrite the `index_fields` metadata of previously created indexed DataPoints.

This change creates a new metadata dictionary for each indexed field while preserving the original DataPoint metadata.

## Changes

- Create independent metadata for each indexed DataPoint.
- Preserve the original `index_fields` metadata.
- Add a regression test covering multiple indexed fields.

## Testing

Ran:

`pytest cognee/tests/unit/infrastructure/databases/test_index_data_points.py -v`

Result:

`4 passed`


Vasilije1990:
@aravmittal123 can you check the conflict and post screenshot of multiple datapoint indexing so i can see how this works

aravmittal123:
Screenshot 2026-08-20 162732


Checked the conflict and resolved it. I also tested multiple-field indexing locally. The screenshot below shows that `name` and `description` get independent `index_fields` metadata, while the original DataPoint remains unchanged. The regression test passes as well.

dexters1:
Thanks for the PR @aravmittal123! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

ArmagedonFlamer:
Thanks @aravmittal123 for the clear write-up of the shallow model_copy() problem, and for testing multi-field indexing with the screenshot. The same fix landed on dev in #4487, which switched index_data_points to model_copy(deep=True) and added a multi-field regression test. After the conflict merge, this branch no longer carries that change and instead picks up unrelated Dockerfile and docker-compose changes, so there is nothing left here to merge. We're closing it as obsolete. We're looking forward to your next contribution.

PR #4592, opened 2026-08-20
feat: day-one seeding — auto-ingest what already exists on install [COG-6253]

Vasilije1990:
## Goal

On a fresh install the first `recall` returns nothing. Retention analysis showed the retained accounts all had automatic ingestion wired **before** value was judged — this makes that the default instead of an accident. On install, cognee immediately ingests what already exists: agent memory files (`MEMORY.md`, `SOUL.md`, `AGENTS.md`, `CLAUDE.md`, …), the workspace README, recent coding-agent session logs, and the codebase itself. **First recall then has something to return.**

Plan + design rationale: `docs/day-one-seeding.md`.

### `cognee/modules/seeding/` (exported as `cognee.seed()`)

- **`discovery.py`** — pure discovery returning a `SeedPlan`:
  - agent memory files at the workspace root + `memory/*.md` + allowlisted `.claude/CLAUDE.md` + Claude Code auto-memory for this workspace (`~/.claude/projects/<slug>/memory/*.md`),
  - `README.*`,
  - the newest N (default 3, ≤5 MB each) Claude Code transcripts for this workspace (slug derivation verified against real `~/.claude/projects` layout),
  - the workspace as a **code project** — reuses the existing repo-manifest resolution (`resolve_code_repository`), so "index any codebase automatically" costs one directory path.
  - Hidden files are never picked up beyond the explicit allowlist → `.env` can never enter the plan. Everything is size-capped; every skip is recorded with a reason.
- **`seed.py`** — staged runner, **fast-first-value order**: memory files + README (seconds) → session logs → codebase. Each stage runs `add()` (LLM-free) then `cognify()`, isolated: no LLM key ⇒ data ingested, cognify deferred and reported — never a failed seed. Sources outside the local-file allowlist are **staged** into `{data_root}/seed_staging/` instead of punching holes in `resolve_local_path`. Lands in one dataset (default `workspace`), which a default `recall()` finds with no arguments.

### Getting-started wiring

- **`cognee-cli seed`** — `--dry-run` prints the full plan without ingesting; `--force`, `--no-code`, `--no-session-logs`, `-w`, `-d`. Refuses to re-seed an existing seed dataset without `--force`; content-hash dedup makes `--force` cheap.
- **Empty recall now hints at `cognee-cli seed`** instead of only `remember`.
- **MCP server auto-seed**: on startup in local mode, when the user owns **zero datasets** (genuinely fresh — never re-seeds behind your back), seed in the background. Startup never blocks. Opt out with `COGNEE_AUTO_SEED=false`. Import is guarded so cognee-mcp keeps booting against PyPI cognee releases that predate the module.

## Verification

- 15 unit tests (`cognee/tests/unit/modules/seeding/`): allowlists & secret-exclusion, size caps, newest-N transcript selection, env/flag toggles, workspace-root walking, slug derivation, stage ordering, out-of-root staging, cognify deferral, idempotency, dry-run, empty-plan short-circuit.
- `cognee-cli seed --dry-run` run against a real workspace correctly discovered its `AGENTS.md`/`CLAUDE.md`, both READMEs, the live session transcript under `~/.claude/projects/<slug>/`, and resolved the repo as a code project.

## Follow-ups (deferred)

`/api/v1/seed` endpoint; cognee-memory plugin install hook calling `cognee-cli seed`; re-seed on change detection (per-source hashes); more agent adapters (Codex, Cursor) beyond the `COGNEE_SEED_SESSION_LOG_GLOB` escape hatch.

### DIFF
--- a/cognee-mcp/src/server.py
+++ b/cognee-mcp/src/server.py
@@ -1934,6 +1934,46 @@ def load_class(model_file, model_name):
     return model_class
 
 
+async def _auto_seed_if_fresh() -> None:
+    """Day-one seeding: on a fresh local install, ingest what already exists.
+
+    Memory files (MEMORY.md, SOUL.md, ...), the workspace README, recent
+    session transcripts, and the codebase are ingested automatically so the
+    first recall has something to return — the default, not an accident.
+
+    Runs only when COGNEE_AUTO_SEED is not false and the user owns zero
+    datasets, which makes it once-per-fresh-install by construction. The
+    import is guarded because cognee-mcp depends on cognee from PyPI, which
+    may predate cognee.modules.seeding.
+    """
+    if os.getenv("COGNEE_AUTO_SEED", "true").strip().lower() in {"0", "false", "no", "off"}:
+        logger.info("Day-one auto-seed disabled via COGNEE_AUTO_SEED")
+        return
+
+    try:
+        from cognee.modules.seeding import seed
+    except ImportError:
+        logger.info("Installed cognee has no seeding module; skipping day-one auto-seed.")
+        return
+
+    try:
+        existing = await cognee_client.list_datasets()
+        if existing:
+            return  # not a fresh install — never re-seed behind the user's back
+
+        with redirect_stdout(sys.stderr):
+            result = await seed()
+
+        if result.plan.is_empty:
+            logger.info(
+                "Day-one auto-seed found nothing to ingest around %s", result.plan.workspace
+            )
+        else:
+            logger.info("Day-one auto-seed:\n%s", result.summary())
+    except Exception as error:
+        logger.warning(f"Day-one auto-seed failed (non-fatal): {error}")
+
+
 async def main():
     global cognee_client
 
@@ -2087,6 +2127,13 @@ async def main():
         logger.info("Database migrations done.")
     elif not is_remote:
         logger.info("Skipping DB migrations")
+
+    # Day-one seeding (local mode only): a fresh install immediately ingests
+    # what already exists around the workspace so the first recall has
+    # something to return. Backgrounded so startup never blocks on it; opt
+    # out with COGNEE_AUTO_SEED=false.
+    if not is_remote:
+        _track_background(_auto_seed_if_fresh())
 
     try:
         match args.transport.lower():
--- a/cognee-mcp/src/server.py
+++ b/cognee-mcp/src/server.py
@@ -1960,6 +1960,10 @@ async def _auto_seed_if_fresh() -> None:
         existing = await cognee_client.list_datasets()
         if existing:
             return  # not a fresh install — never re-seed behind the user's back
+        # This check races with concurrent server instances, but the race is
+        # benign: seed() re-checks dataset existence itself, every instance
+        # seeds the same dataset name, and add()'s content-hash dedup makes
+        # concurrent seeds converge to one copy.
 
         with redirect_stdout(sys.stderr):
             result = await seed()
--- a/cognee/__init__.py
+++ b/cognee/__init__.py
@@ -59,6 +59,7 @@ from cognee.modules.visualization.cognee_network_visualization import (
 )
 from .api.v1.ui import start_ui
 from .api.v1.session import session
+from .modules.seeding import seed
 
 # Pipelines
 from .modules import pipelines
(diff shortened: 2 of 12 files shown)


Vasilije1990:
Addressed the review in db87f1a60:

- **Blocker (env glob)**: `COGNEE_SEED_SESSION_LOG_GLOB` is removed entirely — the sharper version of the finding is that cognee loads `.env` from cwd, so a hostile repo could set the glob and have auto-seed ingest arbitrary files. Extra agents' transcript locations will land as explicit discovery adapters (tracked in COG-6253 follow-ups).
- **Staging dedup**: now `filecmp.cmp(shallow=False)` (chunked reads) instead of `read_bytes()` comparison.
- **TOCTOU on fresh-install check**: left as-is, documented in place — `dataset_lock` is in-process and needs an existing dataset UUID, so it can't guard a cross-process race; the race is benign because `seed()` re-checks dataset existence, every instance targets the same dataset name, and `add()`'s content-hash dedup makes concurrent seeds converge to one copy.
- **CLI**: `--workspace` is validated as a directory up front.

Vasilije1990:
Extended session-log discovery from Claude-Code-only to per-agent adapters (1704bef99): **Codex CLI** (`~/.codex/sessions/YYYY/MM/DD/rollout-*.jsonl`, matched to the workspace via the first-line session meta's `payload.cwd` — verified against a real `~/.codex` tree), **Gemini CLI** (`~/.gemini/tmp/<sha256(project_root)>/` — hash derivation confirmed from gemini-cli source), **pi** (`~/.pi/agent/sessions/--<slug>--/`, observed layout), and **Aider** (`.aider.chat.history.md` at the workspace root). Cursor/opencode/Amp keep sessions in SQLite or server-synced stores and are documented as not-yet-supported. Adapter table added to `docs/day-one-seeding.md`; 4 new tests (19 total in the discovery/seed suite).

NMZivkovic:
The codebase stage doesn't survive the documented invocation. At `seed.py:200` it runs `plan.codebase` through `resolve_local_path`, whose allowed roots are cwd, tempdir and cognee's own data dirs — but `find_workspace_root` walks *up*, so the workspace is typically an ancestor of cwd, not inside it. Run `cognee-cli seed -w /path/to/workspace -d my_dataset` from anywhere else (verbatim from `docs/day-one-seeding.md`), or `cd repo/backend && cognee-cli seed` in a monorepo, and the repo is dropped. The user still gets exit 0 and "Day-one seed complete."; the only trace is a `(skipped)` line in the summary body, and the memory/README/session stages all succeed, so it reads as success. Treating `plan.workspace` as an allowed root for the seed run would fix it, and `plan.skipped` should go through `fmt.warning`. Worth noting `test_seed_stages_in_fast_first_value_order` passes only because pytest's `tmp_path` sits under `tempfile.gettempdir()`, so the suite structurally can't see this.

The other one I'd push on is the MCP auto-seed. `server.py:2139` backgrounds `_auto_seed_if_fresh()` in local mode by default, and the freshness test is `list_datasets()` returning empty — which is also true after a prune, a dataset delete, or a `data_root_directory` change. When it fires it pulls the newest Claude Code transcripts from `~/.claude/projects/<slug>/*.jsonl` at up to 5 MB each, copies them permanently into `seed_staging/session_logs/`, and cognifies them against the user's LLM key. Those JSONL files record tool output verbatim, so a session where the agent ran `cat .env` or read `~/.aws/credentials` carries those bytes. The docstring's "secrets in `.env` or other hidden files can never enter the plan" is true of the file and false of a transcript quoting it. I'd default `COGNEE_AUTO_SEED` off for MCP, or at minimum gate on a persisted "already seeded" marker under `system_root_directory` and default `include_session_logs=False` when nobody typed the command.

Two smaller things. `find_workspace_root` will happily select `$HOME` if there's a dotfiles repo or a stray `package.json` there, and the codebase stage is the only one with no size or file-count cap, so `partition_repo_files` rglobs the whole home tree — Documents, Downloads and Library included, since only dot-prefixed components are skipped. Refusing `Path.home()` and `/` plus a ceiling that records a `plan.skipped` reason would match the caps the other stages already have. And in `_prepare_paths` the `while target.exists() and not filecmp.cmp(...)` loop means `--force` accretes `session-1-1.jsonl`, `session-1-2.jsonl` and so on for exactly the category that always changes between runs, with nothing cleaning `seed_staging/` up. Keying the staged filename on a hash of the source path would let a changed source overwrite its own copy.

The discovery tests are solid otherwise — the allowlist and secret-exclusion coverage in particular.

PR #4601, opened 2026-08-21
docs: note direct mode requirement for list_dataset_data_json and create_dataset_json (#4599)

### LINKED ISSUE #4599: [Docs]: MCP tools page doesn't mention that some workspace tools only work in direct mode
### Documentation Type

Missing documentation

### Documentation Location

https://docs.cognee.ai/cognee-mcp/mcp-tools

### Issue Description

`list_dataset_data_json` and `create_dataset_json` return an error when the MCP server is running in API mode. They only work in direct mode.

The restriction looks intentional, but the tools page doesn't mention it. So if you're running the MCP server against a cognee API, docker compose for example, you get an error with no way to know it's expected. A user ran into this and assumed the tool was broken.

### Suggested Improvement

Add a note on those two tools saying they only work in direct mode and will error under API mode.

KHUSHI423:
Closes #4599

Added a callout note specifying that list_dataset_data_json and create_dataset_json only operate in Direct Mode and will raise an error under API Mode.

### DIFF
--- a/cognee-mcp/README.md
+++ b/cognee-mcp/README.md
@@ -483,6 +483,7 @@ The MCP server intentionally exposes only the memory API: `remember`, `recall`, 
 In API mode these tools call the Cognee API server endpoints directly. Operational helpers such as
 `cognify`, `search`, `list_data`, `delete`, `prune`, `improve`, and document retrieval helpers are
 kept internal and are not exposed as MCP tools.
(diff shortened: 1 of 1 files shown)


ArmagedonFlamer:
Thanks for drafting a quick fix. These methods were removed in PR #4575 so I'll close this PR

PR #4607, opened 2026-08-21
fix: decode each token id in decode_token_list

### LINKED ISSUE #4594: TikToken decode_token_list passes a bare int to tiktoken and raises
`decode_token_list` in the TikToken adapter decodes each id by calling `self.tokenizer.decode(i)` with a bare int. tiktoken's `Encoding.decode` expects a sequence of ids, so this raises `'int' object is not an instance of 'Sequence'`.

`cognee/infrastructure/llm/tokenizer/TikToken/adapter.py`

```python
def decode_token_list(self, tokens: list[Any]) -> list[Any]:
    if not isinstance(tokens, list):
        tokens = [tokens]
    return [self.tokenizer.decode(i) for i in tokens]   # decode(i) gets an int
```

Repro:

```python
from cognee.infrastructure.llm.tokenizer.TikToken.adapter import TikTokenTokenizer
t = TikTokenTokenizer(model=None)
t.decode_token_list(t.extract_tokens("hello world"))   # raises
```

Fix is to decode each id as a one-element list: `self.tokenizer.decode([i])`.

One honest note: this method is not part of `TokenizerInterface` and I could not find a caller in the repo, so if you would rather drop the method than fix it, that works too. I am happy to send either the fix or the removal. Could this be assigned to me?

chiruu12:
Fixes #4594.

`decode_token_list` passes each id to `Encoding.decode` as a bare int. tiktoken expects a sequence there, so the call raises `'int' object is not an instance of 'Sequence'` and the method cannot return for any non-empty input.

```python
return [self.tokenizer.decode(i) for i in tokens]
```

Wrapping the id makes it decode per token, which is what the list comprehension was already trying to do.

Added `cognee/tests/unit/infrastructure/llm/test_tiktoken_adapter.py`, which round trips `extract_tokens` into `decode_token_list` and checks the pieces rejoin into the original string.

### Validation

```
.venv/bin/python -m pytest -q cognee/tests/unit/infrastructure/llm/test_tiktoken_adapter.py -p no:randomly
1 passed
```

With `adapter.py` reverted to `dev` the same test fails on the `Sequence` error, so it pins the behavior.

One note carried over from the issue. This method is not on `TokenizerInterface` and I could not find a caller for it in the repo, so if you would rather delete it than fix it, say so and I will swap this for the removal.

### DIFF
--- a/cognee/infrastructure/llm/tokenizer/TikToken/adapter.py
+++ b/cognee/infrastructure/llm/tokenizer/TikToken/adapter.py
@@ -59,7 +59,9 @@ class TikTokenTokenizer(TokenizerInterface):
         """
         if not isinstance(tokens, list):
             tokens = [tokens]
-        return [self.tokenizer.decode(i) for i in tokens]
+        # tiktoken's decode expects a sequence of ids, so wrap each id: passing a
+        # bare int raises "'int' object is not an instance of 'Sequence'".
+        return [self.tokenizer.decode([i]) for i in tokens]
 
     def decode_single_token(self, token: int) -> str:
         """
--- a/cognee/tests/unit/infrastructure/llm/test_tiktoken_adapter.py
+++ b/cognee/tests/unit/infrastructure/llm/test_tiktoken_adapter.py
@@ -1,0 +1,16 @@
+from cognee.infrastructure.llm.tokenizer.TikToken.adapter import TikTokenTokenizer
+
+
+def test_decode_token_list_returns_text_for_each_token():
+    """decode_token_list must decode each id, not pass a bare int to tiktoken.
+
+    tiktoken's decode expects a sequence of ids, so decoding a list previously
+    raised "'int' object is not an instance of 'Sequence'".
+    """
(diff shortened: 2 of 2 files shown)


chiruu12:
The red on this one is the fork secrets model, not the change. `adapter_caching_tests.yml`, `integration_tests.yml` and `e2e_tests.yml` all read `secrets.LLM_API_KEY` and `secrets.EMBEDDING_API_KEY`, and a pull request from a fork gets those empty, so the jobs die at setup before any test runs. 27 of the 42 workflows here inject a secret.

The diff is one character inside a list comprehension and a unit test that needs no network:

```
.venv/bin/python -m pytest -q cognee/tests/unit/infrastructure/llm/test_tiktoken_adapter.py -p no:randomly
1 passed
```

Reverting `adapter.py` to `dev` makes it fail on `'int' object is not an instance of 'Sequence'`, so the test pins the fix rather than describing it. If someone runs the workflows with secrets it should come back clean.

Vasilije1990 (approved):

PR #4624, opened 2026-08-22
feat: Loud first-build failures + status-aware warm-up marker (COG-6276)

Vasilije1990:
## Description

Implements the top two activation fixes from the onboarding forensics (Veljko §6): 57.4% of SDK accounts' **first cognify run errors**, 854/854 accounts with a day-1 cognify error **never issued a single search**, and the recorded error text is **null everywhere**. One commit per fix.

> **Scope note:** an earlier revision of this PR also shipped an error-taxonomy classifier (auth/quota/schema/… categories + canned remedies) and a `build_in_progress` marker state. Those were cut after review discussion as over-engineering — categories can be derived at analysis time from `error_class`, and the remedy text belonged in docs. What remains is the minimal surface that closes the kill-switch.

### 1. Make a failed first build loud and typed — COG-6276

The mechanics of "error unknowable", traced: `run_tasks` wraps per-item failures in a generic `PipelineRunFailedError` ("Pipeline run failed. Data item could not be processed."), `cognify()` returns a silently-errored run info, and `remember()` folds it into `result.status = "errored"` on an object no quickstart inspects.

Now:
- **Root-cause propagation**: `run_tasks` keeps the first per-item exception on `PipelineRunFailedError.first_error`, so the run record's `error_class`/`error_message` columns (added by SDK-399 but never fed the root cause) and the yielded `PipelineRunErrored` (new `error_class`/`error_message` fields, PII-scrubbed) say what actually broke.
- **`CognifyFailedError` raised by default** from foreground `cognify()`/`remember()`:
  ```
  CognifyFailedError: Cognify failed for dataset 'main_dataset':
  AuthenticationError: Incorrect API key provided: sk-inval****pose
  | Pass raise_on_error=False to get the errored run info instead of this exception.
  ```
- **`raise_on_error=False`** keeps today's swallow behavior for batch users; the `/cognify` and `/remember` routers and `update()` pass it, so HTTP contracts (run-info bodies, the 409 response) are unchanged. Run-level crashes that bypass the errored-run-info path (e.g. an `AuthenticationError` escaping `summarize_text`) wrap into the same typed surface.

### 2. Make the warm-up marker build-status aware — COG-6277

The COG-6254 probe treated *any* pipeline run as warm — **including an ERRORED one** — so exactly the users whose first build failed got empty search results with no explanation. The probe now distinguishes:

| state | meaning | marker |
|---|---|---|
| `warm` | a graph-writing pipeline completed (or staging-only rows: custom pipelines keep their fail-safe fall-through) | normal search |
| **`build_failed`** | no graph-writing run ever completed, an attempt ERRORED | "Memory build failed: the last ingestion ended in an error (AuthenticationError: Incorrect API key provided…). Fix the cause and re-run remember() or cognify()." + `error_class`/`error_message` fields |
| `never_built` | no runs at all | today's warming-up marker |

All COG-6254 review guarantees preserved: populated graphs can never read cold, **only warm verdicts are cached** (a fix is visible on the very next recall), explicit dataset ids remain authorization-filtered before probing, probe failures fail open, multi-source recalls contribute `[]` (tools `on_empty` intact), `only_context` skips the guard, config surface unchanged.

Together: the measured kill-switch (write → silent error → empty read → gone in 24–48 h) becomes: the write **raises with the root cause**, and any read that happens anyway **explains the failure**.

## Testing

- New `test_cognify_failed_error.py`: `first_error` propagation, exception message shape, `raise_if_cognify_errored` (completed passes / errored raises / legacy payload-only run infos), run-level wrapping with no double-wrap.
- `test_recall_warmup.py`: all pre-existing tests migrated to the status probe, plus `build_failed` marker carries the cause and cold-not-cached/warm-cached sequencing.
- Full affected suites green: 648 passed (api, pipelines, recall, search); ruff/pre-commit clean.
- Verified end-to-end with a genuinely invalid key: `remember()` raises `CognifyFailedError` quoting the `AuthenticationError`, and `recall()` returns the `build_failed` marker with the same cause.

Fixes COG-6276
Fixes COG-6277


NMZivkovic:
Defaulting `raise_on_error=True` at `cognify.py:119` changes the contract for every existing SDK caller. `test_cognify_rollback_recovery.py:273` does exactly what users do — `cognify_result = await cognee.cognify(...)` then `assert run_info.status == "PipelineRunErrored"` — and it's now red in "Run Integration Tests / Integration - Documents & Tasks (2/2)" with `CognifyFailedError`. That file is unmodified by this PR, which is the clearest evidence of the blast radius. In-tree it also changes behaviour at `cognee/tasks/memify/cognify_session.py:69` and `cognify_agent_trace_feedback.py:52`, where a per-window cognify failure used to be swallowed so the run continued and the watermark advanced; now it escapes into `except Exception -> raise CogneeSystemError` and kills the whole memify run. Either default to `False` and let callers opt in, or land the flip together with the caller updates and a changelog entry.

Separately, `test_run_tasks_error_propagation.py` is an integration test living in the unit tree. `community_tests.yml:96` runs `cognee/tests/unit/modules/pipelines/` without secrets and the test calls `cognee.add(...)`, so "Unit Tests (Mocked, No Secrets) (3.11.x)" dies with `LLMAPIKeyNotSetError`. It also runs `prune_data()` and `prune_system(metadata=True)` against whatever the ambient config points at, with no tmp root and without the `runner_plumbing` fixture every sibling file uses, so running the unit suite locally wipes real data. Move it to `cognee/tests/integration/` and use `monkeypatch.setenv` instead of writing `os.environ` directly.

Two more. The `except Exception` at `cognify.py:384` wraps everything the pipeline can raise, including `PermissionDeniedError` and `DatasetNotFoundError` from `resolve_authorized_user_datasets`, into a 422 `CognifyFailedError` — so `except PermissionDeniedError` around an SDK cognify stops matching, and the appended "Pass raise_on_error=False to get the errored run info" is false on that path since it re-raises the original. Narrow the wrap to pipeline/build failures and let `CogneeApiError` subclasses through. And `remember(run_in_background=True)` loses its run metadata on failure: the inner foreground cognify raises, `_remember_background` routes to `_fail()`, and `pipeline_run_id`/`dataset_id`/`raw_result` stay None, so the caller can't correlate the run it just launched. Passing `raise_on_error=raise_on_error and not run_in_background` there mirrors what cognify already does for itself.

Smaller: the `build_failed` branch in `graph_warmup.py:138` is evaluated before the `any_run` fail-safe and its query doesn't exclude `add_pipeline`, so one errored `add()` of a corrupt file makes every `recall()` on a populated graph return zero results plus "Fix the cause and re-run". The custom-pipeline fall-through the module docstring describes only works now when no errored row exists anywhere.

Vasilije1990:
Addressed all five points in 6c5415a78 — the loud default stays, with the callers updated deliberately:

**1. `raise_on_error=True` blast radius** — kept the flip (it's the point of COG-6276) and landed the caller updates:
- `test_cognify_rollback_recovery.py` now passes `raise_on_error=False` where it inspects the errored run info.
- `cognify_session` and `cognify_agent_trace_feedback` opt out of raising and inspect the returned run info instead: one failed window/trace session logs its root cause and the memify run continues. As a deliberate improvement, `cognify_session` no longer advances the watermark past a failed window — which its docstring always promised ("on failure the watermark stays put") but the old silent-errored path violated; the window is re-extracted on the next `improve()`.
- Swept the remaining in-tree foreground callers: `update()` and both routers already passed `False`; `sync.py` and the CLI already catch; `_distill_sessions` is per-session try/except (fail-open), and the responses-API `handle_cognify` previously returned "successfully converted" over an errored run — raising there is the intended fix.

**2. `test_run_tasks_error_propagation.py`** — moved to `cognee/tests/integration/tasks/` (it drives the real add pipeline), rewritten as pytest with a per-test tmp-root fixture and `monkeypatch.setenv`; it no longer prunes ambient data and the no-secrets unit job no longer collects it.

**3. Over-broad wrap at `cognify.py:384`** — `CogneeApiError` subclasses now pass through unwrapped, so `except PermissionDeniedError` / `DatasetNotFoundError` around an SDK cognify keeps matching (pinned by a new integration test). The wrapped-crash message also got an accurate hint: on that path `raise_on_error=False` re-raises the original exception, and the text now says so instead of promising errored run info.

**4. `remember(run_in_background=True)` losing run metadata** — the inner cognify now gets `raise_on_error=raise_on_error and not run_in_background`, so `_resolve()` runs and the errored result keeps `pipeline_run_id`/`dataset_id`/`raw_result`.

**5. `build_failed` vs an errored `add()`** — the errored-run query now excludes `add_pipeline` rows (same guard the warm query already had), restoring the staging-only warm fail-safe: an errored `add()` of a corrupt file can no longer flip a populated graph to "Memory build failed". Added SQL-level probe tests (`test_graph_warmup_probe.py`) since the unit suite stubs the probe wholesale.

Affected suites locally: 700 unit (pipelines/api/memify) + 10 integration (rollback recovery, error propagation, warm-up probe) all green; ruff/pre-commit clean.

PR #4628, opened 2026-08-23
feat: Add authority and traceability opt-ins (COG-6279)

Vasilije1990:
## Description

An external audit (right-to-reply issue on the article *"your memory has no authority model"*, audited commit `fd5045f6`) validated four gaps in cognee's authority model. This PR is the code response for two of them — the third (`ONTOLOGY_MODE=strict`) was split out to #4848 and should merge first, with the review findings on it fixed there.

The standard pipelines behave identically unless opted in, with one deliberate exception called out in §2: `assertion_source` stamping is **default-on** (it rides the existing `COGNEE_PROVENANCE_MODE` gate, whose default is `lightweight`), so every node/edge gains the new property on defaults. It is write-only on defaults — ranking, dedup, and entity merging ignore it, and untagged edges resolve byte-identically — but it is a new stored property, not a no-op.

### 1. `COPY_LOCAL_FILES` — original-bytes snapshot (default `false`)
Local file paths have always been ingested **by reference**: cognee records the path and `content_hash`, which can verify the original but not recover it once the file moves or is deleted. With the flag on, local-path originals are snapshotted content-addressed into cognee-managed storage at add time via the existing `save_data_to_file_detailed` machinery, so `original_data_location` stays resolvable and hash-verifiable. `s3://` inputs remain pass-through by design.

### 2. Assertion authority + superseded-edge filtering
- `DataPoint.assertion_source` (`user_stated` / `document_extracted` / `llm_inferred`) is stamped whenever provenance is not disabled — i.e. **on by default** under `COGNEE_PROVENANCE_MODE=lightweight`: `user_stated` for session-bridged facts, `document_extracted` otherwise; `cross_connect_entities` and `detect_contradictions` write `llm_inferred`. Copied onto edge properties so the conflict resolver can see it.
- The `functional_relationships` temporal conflict resolver now ranks **authority first, then recency** — an older user-stated fact beats a newer LLM inference. Untagged/pre-existing edges rank as `document_extracted` and resolve **byte-identically to today** (locked by a regression test).
- Graph retrieval skips edges tagged `superseded=True` at both choke points (`CogneeGraph` projection and the hybrid entity lane). Safe by construction: the tag only ever exists for callers who opted into `functional_relationships`. `contradicts` edges stay visible; precomputed triplet embeddings and raw chunk text are unfiltered (documented).

## Known review follow-ups (to address before merge)
- **Stale `superseded` tag on re-assertion**: nothing clears the tag, and Neo4j's `SET rel += edge.properties` merges rather than replaces, so a re-asserted edge that later wins can keep `superseded=True` from an earlier round — with the retrieval filter on, both edges in a group can end up hidden. Fix: `tag_superseded_edges` should write the winner back with `superseded=False`.
- **`supersession_reason` wording**: hardcoded "superseded by a more recent ... assertion", but with authority-first ranking the winner is often the *older* edge — the audit string can contradict the timestamps on the same edges.

## Test plan
- [x] Unit tests across the touched areas (ingestion, resolver, provenance stamping, memify, CogneeGraph, hybrid entities) — all pass on this branch
- [x] 12 ingestion integration tests, including the new snapshot test: copy survives `os.remove(source)` with md5 == `content_hash`
- [x] Byte-identity regression tests at every hook site (untagged edges, flag-off ingestion)
- [x] `pre-commit run --all-files` clean

Linear: COG-6279

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -543,6 +543,7 @@ Configuration:
 ONTOLOGY_RESOLVER=rdflib  # Default: uses rdflib and OWL files
 MATCHING_STRATEGY=fuzzy   # Default: fuzzy matching with 80% similarity
 ONTOLOGY_FILE_PATH=/path/to/your/ontology.owl  # Full path to ontology file
+ONTOLOGY_MODE=annotate    # Default: enrich only. strict drops entities with no ontology grounding
 ```
 
 Implementation: `cognee/modules/ontology/`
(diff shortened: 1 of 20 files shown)


NMZivkovic:
`ontology_env_config.py:40` breaks ontology-via-env outright. The dict returned by `to_dict()` exists only to be splatted into `get_ontology_resolver_from_env(ontology_resolver, matching_strategy, ontology_file_path)`, which takes exactly three parameters and no `**kwargs`, so adding the fourth key means anyone with `ONTOLOGY_FILE_PATH` set gets `TypeError: got an unexpected keyword argument 'ontology_mode'` out of `cognee.cognify()` — before any ingestion work. That's the setup documented in `.env.template`, `CLAUDE.md` and `catalog/entries/use-cases/custom-ontology.yaml`, and it's what `examples/demos/comprehensive_example` does at line 34. Drop the key from `to_dict()` (the mode is already read separately via `get_ontology_env_config().ontology_mode`), or pass the three fields explicitly at the call site.

CI didn't catch it because `test_none_config_uses_env_when_configured` patches `get_ontology_env_config` to a hand-written 3-key `to_dict` *and* mocks `get_ontology_resolver_from_env`, so neither half of the mismatch can surface. Worth a test that builds the real config and stubs only the resolver's body.

Two other things. `ontology_mode: Literal["annotate", "strict"]` with no fallback means `ONTOLOGY_MODE=Strict` raises a pydantic ValidationError that kills every cognify run, including runs with no ontology at all, since the config object is constructed before the file-path guard. `get_provenance_config()` already has the pattern for this — plain `str`, lowercase, warn and fall back. And the stale-`superseded` case: nothing ever clears the tag, and Neo4j's `SET rel += edge.properties` merges rather than replaces, so an edge that gets re-asserted and later wins keeps `superseded=True` from the earlier round. Now that `CogneeGraph.py:198` skips superseded edges, both edges in a group can end up hidden and the current, correct fact disappears from GRAPH_COMPLETION while sitting right there in the graph. Having `tag_superseded_edges` write the winner back with `superseded=False` would close it.

Minor: `supersession_reason` is hardcoded to "superseded by a more recent ... assertion", but with authority-first ranking the winner is often the older edge, so `test_older_user_stated_edge_beats_newer_llm_inferred` is producing an audit string that contradicts the timestamps on the same edges.

Also worth correcting the PR description — `assertion_source` stamping is default-on, not opt-in. `_stamp_provenance` runs whenever provenance isn't disabled, and the default mode is `lightweight`, so every user on defaults gets the new property on every node and edge. Storage-wise it's safe and the ranking is byte-identical for untagged edges, so it isn't a bug, but "each default-off or default-byte-identical" isn't true of that write path.

PR #4630, opened 2026-08-23
feat: Presort folder pre-organization + JSON graph-model DSL (COG-6281)

Vasilije1990:
Presort adds an explicit pre-ingestion scan for messy folders, plus a JSON graph-model DSL for defining entity and relationship schemas. `remember(folder, dry_run="presort")` returns a persisted report of junk, duplicates, version candidates, redacted personal-data findings, existing ingestion status, and proposed dataset groups. `remember(report)` applies the selected groups; `auto_apply=True` combines the two phases.

Ordinary `remember(folder)` calls retain their existing dataset and `RememberResult` behavior. Automatic presort requires `PRESORT_FOLDERS_ENABLED=true` (default: false). The CLI keeps boolean `--dry-run` for estimates and uses `--presort` for folder analysis, with `--apply`, `--apply-graph`, and `--from-report` for ingestion.

Apply preserves caller node sets, database configurations, loaders, batching, chunking, and error-handling options. Empty reports return no ingestion results when no groups were explicitly requested. Without an LLM, deterministic analysis still works and apply stages files through `add()` only. Report reads and folder probes require permitted roots (configured allowlist, or cwd, temporary directory, and Cognee storage by default); `--allow-root` extends those roots explicitly; raw input is no longer logged by automatic routing.

The DSL validates schemas before code generation, accepts frontend-compatible aliases, and supports identity fields for node merging. Presort uses it for duplicate, version, group, and PII relationships. Examples are in `examples/guides/presort_downloads.py` and `examples/guides/graph_model_from_json.py`. There are no MCP or frontend changes.

Merged current `dev`, resolved the `remember()` signature conflict, and applied current lint rules. Python 3.10 tests patch module objects instead of shadowed import paths; filename tests use `Path.name` for Windows compatibility.

Validation:

- Python 3.10 and 3.12: 494 passed, 1 skipped each across presort, graph-model DSL, all CLI unit tests, adjacent remember/forget/estimator/path-policy tests, and real local ingestion regression tests.
- `uv run ruff check .`
- `uv run ruff format --check .`
- `uv lock --check`
- `git diff --check`

Fixes COG-6281.


NMZivkovic:
The part I'd want changed before this merges is that presort isn't opt-in. `remember.py:979` reroutes any local directory into the presort path whenever `dataset_name` is still the default — and `main_dataset` is also the parameter's default, so passing `dataset_name="main_dataset"` explicitly gets rerouted too:

```python
if dry_run is False and _should_auto_presort(data, dataset_name, dataset_id, session_id, kwargs):
    dry_run = "presort"
    kwargs.setdefault("auto_apply", True)
```

Today `remember("./docs")` rglobs every file into `main_dataset` and returns a `RememberResult`. After this it returns a `PresortReport`, writes into several group-derived datasets with `node_set=["presort", <group>]`, and drops duplicate copies, symlinks, hidden files, zero-byte files and junk extensions that used to be ingested. Anything reading `result.dataset_id` or `result.pipeline_run_id` now raises AttributeError, and `PRESORT_FOLDERS_ENABLED=false` is the only way out. For this release I'd require an explicit `dry_run="presort"` and keep the auto-route behind an env flag defaulting to off.

Second thing: the new tests fail deterministically on 3.10 — 12 failures in "OS and Python Tests Ubuntu / Unit tests 3.10.x", and 3.10 is a supported version per `requires-python`. It isn't flakiness or import order. `mock.patch("cognee.modules.presort.run_presort._report_destination")` resolves through `_dot_lookup` up to 3.10 and `pkgutil.resolve_name` from 3.11 on, so the getattr walk lands on the re-exported *function* from `__init__.py` instead of the module and raises `AttributeError: <function run_presort ...> does not have the attribute '_report_destination'`. Same shadowing on `cognee.tasks.presort.detect_pii.LLMGateway`, `cognee.api.v1.add.add.add` and `cognee.api.v1.remember.remember._remember_inner`. Fix is to patch the module object: `import cognee.modules.presort.run_presort as m; patch.object(m, "_report_destination", ...)`. Two more fail on Windows because the tests do `path.rsplit("/", 1)[-1]`, which never splits a backslash — use `Path(...).name`.

Three smaller ones while you're in there. A folder of only dotfiles or `.DS_Store` now hits `apply_presort.py:112` and raises `ValueError: No groups to apply` on a path the caller never opted into, where dev ingested those files fine; return the empty report instead when the groups weren't explicitly requested. When a folder auto-presorts, every kwarg except `run_in_background`/`self_improvement`/`user` is dropped, including `node_set`, `vector_db_config` and `graph_db_config` — so data lands in the default databases with no error. And `--dry-run` as `nargs="?"` next to the `nargs="*"` positional breaks `cognee-cli remember --dry-run notes.txt`; that now parses as `data=[], dry_run='/tmp/notes.txt'` and aborts with "Unsupported --dry-run value". Keeping `store_true` and adding a separate `--presort` avoids that.

The DSL and the test volume are good — my objection is entirely about the default and the 3.10 gate.

PR #4649, opened 2026-08-24
fix(workers): close the Linux parent-death spawn-window race that orphans DB workers holding locks

GrowDev1:
## Summary

On Linux, a DB worker can outlive its parent forever and keep holding an exclusive Kuzu/LanceDB file lock. The next run then blocks on a lock owned by a process whose parent is long gone.

This closes the window between `Process.start()` and the moment the child arms its own parent-death detection.

## This reproduces on 1.5.3 today

I raised this before in #4249, which was closed as `Handled`. I want to be careful here: it is entirely possible something related was fixed, and I am not asserting otherwise. What I can say is what I measured on the released `1.5.3` this morning, three times, on a fresh venv:

```
interpreter : /mnt/bulk/cognee-153-sandbox/bin/python
cognee      : 1.5.3
scenario B  : parent dies INSIDE the spawn window, before arming

  run 1:  FAIL  worker 48058 was STILL ALIVE 20.0s after its parent died
  run 2:  FAIL  worker 48201 was STILL ALIVE 20.0s after its parent died
  run 3:  FAIL  worker 48963 was STILL ALIVE 20.0s after its parent died
```

Three different PIDs, three orphans. `grep -c 'parent_process()' cognee_db_workers/harness.py` returns `0` on 1.5.3.

With this branch applied to the same venv, same harness, same timeout:

```
  run 1:  PASS  worker exited 0.0s after
  run 2:  PASS  worker exited 0.0s after
  run 3:  PASS  worker exited 0.0s after
```

If I am testing the wrong thing, I would genuinely like to know which scenario you consider covered — the harness is below so the disagreement can be settled by running it rather than by reading.

## Why the obvious fix is wrong, which is the part worth reviewing

The intuitive approach is to compare `os.getppid()` against the parent PID captured at launch, mirroring the Windows logic. **That would kill every worker at startup on modern Linux.**

Measured across all three start methods:

```
spawn        parent_process().pid=199682  getppid()=199682
fork         parent_process().pid=199682  getppid()=199682
forkserver   parent_process().pid=199682  getppid()=199706   <-- differ, parent perfectly healthy
```

Under `forkserver` the child is forked by the *fork server*, not by the original parent, so those two values differ **by design**. `forkserver` is Python 3.14's default start method on Linux. A `getppid()`-based check would `os._exit(0)` every worker immediately.

So this uses `multiprocessing.parent_process().is_alive()` instead — a kernel-held sentinel, start-method agnostic, and immune to PID reuse.

## The window itself

`prctl(PR_SET_PDEATHSIG)` is the right primitive but it is **not retroactive**: if the parent already died while the child was still re-exec'ing and importing, the signal never fires. Worse, `prctl` *succeeding* suppresses the polling fallback, and the fallback compares `getppid()` to a baseline it reads **after** reparenting has already happened — so it can never detect a death that already occurred.

The fix adds an explicit `parent_already_exited()` check placed **after** arming, which is the only ordering that closes the window without introducing a new one.

## Scope

Two files:

- `cognee_db_workers/harness.py`
- `cognee_db_workers/tests/test_linux_parent_death_race.py` (new)

Both commits cherry-pick cleanly onto `25200a548` (the 1.5.3 bump) with no conflicts.

## Reproducing it yourself

The harness spawns a worker through the real `run_worker_loop`, records its PID, then has the parent `SIGKILL` **itself** — no `atexit`, no `finally`, no signal handler can tidy up on the worker's behalf, which isolates the mechanism under test.

It runs two scenarios, and only one is interesting:

- **A** — parent dies *after* the worker armed its detection. `prctl` covers this. Every version passes. A is the control: without it, a pass on B proves nothing, because a worker that never started also never survives.
- **B** — parent dies *during* the spawn window, before the worker armed anything. This is the race.

```
python3 verify_parent_death.py --python /path/to/venv/bin/python --scenario both --timeout 20
```

Exit 0 all passed · 1 a worker outlived its parent · 2 the harness could not run (never reported as a pass). Happy to attach it to this PR as a script if that is useful — say the word and I will add it under `cognee_db_workers/tests/`.

## What I am not claiming

I have not measured this on macOS, and I have not measured whether the orphaned worker is reachable through every code path that constructs a worker — only through `run_worker_loop`, which is the production entry point rather than a replica of it.


cognee_db_workers/tests/test_linux_parent_death_race.py
```diff
@@ -0,0 +1,541 @@
+"""Tests for the POSIX parent-death ARM-WINDOW race in `run_worker_loop`.
```
NMZivkovic, line 1: ruff-format rejects the new test file.

cognee_db_workers/tests/test_linux_parent_death_race.py
```diff
@@ -0,0 +1,541 @@
...
+        multiprocessing.parent_process = lambda: _BadFdParent()
+        try:
+            got = harness._parent_sentinel_alive()
+        finally:
+            multiprocessing.parent_process = original
+
+        assert got is None, (
+            f"invalid fd yielded {got!r}; False would be read as 'parent dead' "
+            "and kill a worker whose parent is fine"
+        )
```
NMZivkovic, line 541: there is no bare pytest, and no testpaths in pyproject.toml. 
cognee_db_workers/tests/ appears in no workflow at all. 
The only evidence for its own fix never runs.

cognee_db_workers/harness.py
```diff
@@ -257,12 +257,249 @@ def set_pdeathsig() -> bool:
...
+        if parent.is_alive():
+            return True
+
+        # is_alive() said "dead". Do NOT act on that alone -- an invalid fd is
+        # reported as ready (POLLNVAL) and is indistinguishable from a real EOF
+        # at this layer. Corroborate the fd is still valid before confirming.
+        # POSIX only: on Windows _sentinel is a process HANDLE, not an fd.
+        if sys.platform != "win32":
+            try:
+                os.fstat(sentinel)
```
NMZivkovic, line 424: The os.fstat(sentinel) corroboration proves the fd is valid, not that it is still the parent's sentinel. A closed-and-reused fd passes it, yields “parent confirmed dead”, and fires os._exit(0) against a perfectly healthy parent.

NMZivkovic (changes requested):

GrowDev1:
@Vasilije1990 Thank you for moving the suite and for addressing the fd-reuse point. I added one lint-only commit (`f2ee46d50`) to fix the PR-file Ruff and strict exception diagnostics; Code Quality and basic-checks pre-commit are now green on that head. This does not claim the full PR checks are green.

The Test Suites reusable-workflow calls at `.github/workflows/test_suites.yml:105` fail workflow evaluation because the required LLM/embedding secrets are unavailable to fork PR jobs. Could the maintainers judge a secure way to run those checks?

@NMZivkovic, could you re-review the updated PR?

PR #4654, opened 2026-08-24
feat: only_context returns session information

siillee:
## Description

`only_context=True` returns the bare retrieval context — no session
guidance, no conversation history, no rendered prompt — so callers
feeding it to their own LLM get less than cognee's own completion does.

This PR adds an opt-in `context_format="prompt"` that returns the full envelope
instead: question, context, session_context, user_prompt, system_prompt.
The session layer is rebuilt read-only (no LLM call, no session write),
and both the preview and the real completion now render their prompts
through one shared `build_completion_prompts`, so they cannot drift.

The default stays byte-identical: changing it would silently turn a
string into a dict on a public API, add an embedding round-trip per call,
and feed JSON to `@agent_memory(memory_only_context=True)`.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation


cognee/modules/recall/methods/normalize_search_payload.py
```diff
@@ -151,6 +157,20 @@ def normalize_search_payload(payload: SearchResultPayload) -> list[SearchResultI
     """Normalize one dataset's retriever payload into SearchResultItems."""
     kind = _KIND_BY_SEARCH_TYPE.get(payload.search_type, SearchResultKind.UNKNOWN)
 
+    if payload.only_context and payload.context_format == CONTEXT_FORMAT_PROMPT:
```
NMZivkovic: **The prompt branch always returns exactly one item, so an empty retrieval now looks like a hit — and that silently disables recall's `on_empty` tools fallback.**

The default path returns `_flatten(payload.context)`, which is `[]` when the context is `None`/empty. This branch returns a one-element list unconditionally.

Concretely, `recall(q, scope=['graph','tools'], tools_trigger='on_empty', only_context=True, context_format='prompt')`:

- `recall.py:506` reorders `tools` last precisely so it observes every other source
- `recall.py:938`: `if src == 'tools' and tools_trigger == 'on_empty' and merged: continue`

With prompt format the graph lane always contributes one envelope item even when the graph returned nothing, so `merged` is always truthy and the tools/text-to-SQL fallback can never fire. The same applies to `auto_fallthrough` at `recall.py:938`. With the default format the identical call yields zero items and tools runs.

Suggest returning `[]` when the underlying context is empty, so the item count still carries "did the retrieval find anything".

cognee/modules/recall/methods/normalize_search_payload.py
```diff
@@ -151,6 +157,20 @@ def normalize_search_payload(payload: SearchResultPayload) -> list[SearchResultI
...
+        # One item, not one per context entry: the caller asked for the prompt a
+        # completion would receive, and that is a single artifact. The parts stay on
+        # `raw` so a consumer can still take just the context or just the history.
+        envelope = payload.prompt_envelope
+        return [
+            _build_item(
+                envelope,
+                payload,
+                kind,
+                text_override=payload.user_prompt or _text_from_dict(envelope),
```
NMZivkovic: **When `user_prompt` is `None` this falls through to `json.dumps(envelope)`, so `text` becomes an escaped JSON blob.**

`_text_from_dict` probes `text`/`completion`/`summary`/`name`/`content`/`answer` — the envelope has none of them — and lands on the `json.dumps` fallback at line 66.

`recall(..., query_type=SearchType.CHUNKS, only_context=True, context_format='prompt')`: CHUNKS has no prompt templates so `user_prompt is None`, and `SearchResultItem.text` — documented as "always populated and always renderable" — comes back as:

```
{"question": "...", "context": [...], "session_context": "## Active session guidance\n- ...", "user_prompt": null, "system_prompt": null}
```

That pushes the guidance block and the conversation history into the display field. Same fires for any generative retriever whenever `build_context_preview` fails open (`context_preview.py:161`).

A `session_context`-or-`context` fallback would keep `text` readable.

cognee/modules/retrieval/context_preview.py
```diff
@@ -0,0 +1,169 @@
...
+            load_preference_lines_safe,
+            select_session_history,
+        )
+
+        user_uuid = getattr(session_user.get(), "id", None)
+        if not user_uuid:
+            return ""
+
+        session_manager = get_session_manager()
+        if not session_manager.is_session_available_for_completion(user_uuid):
```
NMZivkovic, line 134: **Sessionless callers get a preview that is missing the preference block the real completion actually sends.**

With `CACHING=false`, `is_session_available_for_completion` returns `False` and this returns `""`. But the real path takes `_use_session_cache() == False` and lands in `_generate_completion_without_session`, which does:

```python
preference_text = await load_preference_text()
completion = await generate_completion(query=query, conversation_history=preference_text, **kwargs)
```

(`graph_completion_retriever.py:338`, `completion_retriever.py:216`, `hybrid_retriever.py:261`)

So the real `system_prompt` is `PREFERENCES...\nTASK:<template>`, while the preview reports the bare template. A user with a stored "Answer in Portuguese" preference is told cognee sends no preference instruction at all — which is exactly the discrepancy this PR exists to close.

The sessionless branch needs its own `load_preference_text()` call.

cognee/modules/retrieval/context_preview.py
```diff
@@ -0,0 +1,169 @@
...
+            return ""
+
+        user_id = str(user_uuid)
+        requested_session_id = (
+            session_id if session_id is not None else getattr(retriever, "session_id", None)
+        )
+        session_id = session_manager.resolve_session_id(requested_session_id)
+
+        history = await select_session_history(
```
NMZivkovic: **The module docstring's "No LLM call" invariant isn't true — this path issues a billed embedding call, once per dataset.**

`select_session_history` -> `search_session_qa_ids` -> `vector_engine.search(SESSION_QA_VECTOR_COLLECTION, query_text=query_text, query_vector=None, ...)` (`session_embeddings.py:133`). `query_vector=None` means the engine embeds `query_text` — a network round-trip to the embedding provider.

And it runs per dataset: `search_in_datasets_context` gathers one `get_retriever_output` per authorized dataset, each calling `build_context_preview`. Ten datasets = ten embedding calls and ten `build_active_context_block` reads for a session layer that is dataset-independent and byte-identical in all ten payloads.

The PR description says the default was kept specifically to avoid "an embedding round-trip per call" — worth saying plainly that the opt-in path pays exactly that, since `only_context` callers often choose it to avoid model spend (cf. `@agent_memory(memory_only_context=True)`).

`test_preview_never_calls_an_llm` can't catch this: `patched_session` replaces `select_session_history` with an `AsyncMock` before the assertion runs.

Hoisting the preview above the dataset fan-out would fix both the cost and the duplication.

cognee/modules/retrieval/context_preview.py
```diff
@@ -0,0 +1,169 @@
...
+        return ""
+
+
+async def build_context_preview(
+    retriever, *, query: str, context: Any, session_id: Optional[str] = None
+) -> ContextPreview:
+    """Assemble the session layer and the rendered prompts for one ``only_context`` call."""
+    session_context = await load_read_only_session_prompt(retriever, query, session_id=session_id)
+
+    user_prompt_path = getattr(retriever, "user_prompt_path", None)
```
NMZivkovic, line 182: **`AgenticRetriever` answers with different templates than the ones read here.**

Nothing rejects `search(query_type=SearchType.AGENTIC_COMPLETION, only_context=True, context_format='prompt')`, and `AgenticRetriever` inherits `user_prompt_path`/`system_prompt_path`, so the `getattr` succeeds and the envelope reports `graph_context_for_question.txt` + `answer_simple_question.txt`.

The real loop uses the agentic pair (`agentic_retriever.py:487`):

```python
step: AgentStep = await generate_completion(
    user_prompt_path=self.agentic_user_prompt_path,
    system_prompt_path=self.agentic_system_prompt_path,
    response_model=AgentStep,
)
```

The inherited pair is only the budget-exhausted fallback at line 527. Its session layer is assembled differently too (`_get_session_history()` + `_maybe_active_context_block`, not `select_session_history` + `compose_session_prompt`).

Either special-case the agentic retriever or have retrievers expose the pair they will actually use.

cognee/modules/retrieval/context_preview.py
```diff
@@ -0,0 +1,169 @@
...
+    system_prompt_path = getattr(retriever, "system_prompt_path", None)
+    if not user_prompt_path or not system_prompt_path:
```
NMZivkovic: **Attribute presence isn't the same as "is generative" — `CypherSearchRetriever` passes this guard but never sends a prompt.**

`cypher_search_retriever.py:24-30` sets both `user_prompt_path` and `system_prompt_path`, but `get_context_from_objects` returns `None` (line 79).

So `search(query_type=SearchType.CYPHER, only_context=True, context_format='prompt')` renders `context_for_question.txt` with `context=None`. `render_prompt` sets no Jinja `finalize`, so the caller gets:

```
And here is the context: `None`
```

A complete, plausible prompt pair for a search type cognee never sends a prompt for — which is precisely what the `ContextPreview` docstring at line 50 says it avoids ("inventing one would misrepresent a search type whose contract is explicitly non-generative").

NMZivkovic (commented):
Reviewed the whole diff. The `build_completion_prompts` extraction is byte-faithful to the old `generate_completion` — nice. My concerns are all in the new preview layer and the shapes it produces.

The recurring theme: `context_preview.py` is a hand-copy of `generate_session_answer` rather than a read-only mode of it, and it has already drifted on day one (sessionless preference block, agentic templates, cypher, `answer_query`). The deeper fix for several of the comments below is to extract the session assembly into a shared `build_session_prompt(..., stamp_served=True)` in `session_turn.py` — the same move this PR made for the prompt layer with `build_completion_prompts` — instead of a parallel implementation.

Two of the new tests also can't fail; flagged inline.

cognee/modules/retrieval/context_preview.py
```diff
@@ -0,0 +1,169 @@
...
+        return ContextPreview(session_context=session_context)
+
+    try:
+        user_prompt, system_prompt = build_completion_prompts(
+            query=query,
```
NMZivkovic: **The real sequential turn renders `{{question}}` with the rewritten query, not the raw one — and the same rewrite drives history selection and guidance ranking.**

`session_manager.py:423`:

```python
answer_query = (turn_preparation.effective_query or '').strip() or (effective_query or '').strip() or query
```

`generate_session_answer` then passes `answer_query` to `select_session_history(query_text=)`, `build_active_context_block(query=)`, and through `generate_session_completion_with_optional_summary` into `render_prompt(..., {'question': answer_query})`.

So with `AUTO_FEEDBACK=true`, a follow-up like "and in French?" is rewritten into a standalone query, and a real turn's `user_prompt` carries the rewritten text with a differently-ranked guidance block and a different set of vector-recalled turns. The preview reports the literal follow-up for all three.

The module docstring concedes only the retrieval-context gap — worth widening it to the question slot and the session-layer ranking, or noting it in the CLAUDE.md caveat paragraph too.

cognee/modules/retrieval/utils/completion.py
```diff
@@ -7,6 +7,32 @@
...
+) -> Tuple[str, str]:
+    """Assemble the exact ``(user_prompt, system_prompt)`` pair a completion sends.
+
+    Pure and side-effect free. ``generate_completion`` builds its prompts here, and so
+    does the ``only_context`` preview, so a caller asking what the LLM *would* receive
+    gets the real strings instead of a reconstruction that drifts as templates change.
+    """
+    user_prompt = render_prompt(user_prompt_path, {"question": query, "context": context})
+    resolved_system_prompt = (
+        system_prompt if system_prompt else read_query_prompt(system_prompt_path)
```
NMZivkovic, line 27: **`read_query_prompt` returns `None` on a missing template, and the new signature promises `Tuple[str, str]`.**

`read_query_prompt.py:40` logs and returns `None` on `FileNotFoundError`. Two outcomes here:

- no `conversation_history` -> returns `(user_prompt, None)`
- with `conversation_history` -> `TypeError: can only concatenate str (not "NoneType") to str` on line 31

On the preview path both collapse into the same output: `build_context_preview`'s `except Exception` (line 161) returns `ContextPreview(user_prompt=None, system_prompt=None)` — indistinguishable from the documented "this retriever is non-generative" sentinel.

So `search(system_prompt_path='typo.txt', only_context=True, context_format='prompt')` on GRAPH_COMPLETION tells the caller GRAPH_COMPLETION has no prompt template, and they never learn the path is wrong. (This was latent in `generate_completion` before, but the new pure helper with a `Tuple[str, str]` annotation is the right place to raise instead.)

cognee/modules/search/methods/search.py
```diff
@@ -409,6 +419,21 @@ async def _search_without_context() -> SearchResultPayload:
     return await asyncio.gather(*tasks)
 
 
+def _prompt_preview_fields(search_result) -> dict:
+    """Prompt-preview keys for verbose output, present only when a preview was built.
+
+    Verbose callers parse a fixed set of keys, so an ordinary search must keep the shape
+    it always had; these appear only for only_context calls that asked for the prompt.
+    """
+    if search_result.user_prompt is None and search_result.session_context is None:
```
NMZivkovic: **This gates on value-nullness rather than on `context_format`, so an explicitly requested prompt format can return a response byte-identical to one that never sent the field.**

The docstring says "Verbose callers parse a fixed set of keys" — but the returned key set now depends on runtime session state, which is the opposite.

`search(query_type=CHUNKS, only_context=True, context_format='prompt', verbose=True)` with no active session: `user_prompt` is `None` (non-generative by design) and `get_retriever_output.py:104` collapses the empty session layer to `None` via `preview.session_context or None`, so this returns `{}`. A client doing `d['user_prompt_result']` gets a `KeyError` not on bad input but on a session that happened to be empty — passes in dev, raises in prod. They also can't distinguish "nothing to show" from "an older server dropped my field".

Gating on `search_result.context_format == CONTEXT_FORMAT_PROMPT` and always emitting the three keys (possibly `None`) keeps the shape stable.

cognee/modules/search/methods/get_retriever_output.py
```diff
@@ -72,11 +78,31 @@ async def get_retriever_output(
         only_context=only_context,
         search_type_for_spans=effective_query_type,
     )
+
+    # An unknown format degrades to the historical shape rather than failing a retrieval
+    # that already succeeded; search()/recall() reject bad values at their boundary.
+    context_format = kwargs.get("context_format") or CONTEXT_FORMAT_CONTEXT
```
NMZivkovic: **One rule, three behaviours — and the `or` only normalises falsy values, so a truthy-invalid one is stored and echoed back.**

- `api/v1/search/search.py:79` raises with `name='InvalidContextFormat'`
- `api/v1/recall/recall.py:488` raises with `name='InvalidContextFormatError'` and different wording
- here: silently degrades

The two error names and messages have already drifted in the first commit, so a client branching on `name` has to handle both spellings for one condition.

And `context_format='Prompt'` (case mismatch) reaching `modules.search.methods.search.search`/`authorized_search` directly — neither validates — survives the `or`, gets written to `SearchResultPayload.context_format` on line 103 (typed plain `str`, so pydantic catches nothing), and is echoed to the caller while `result` returns the bare-context shape.

`Literal["context", "prompt"]` on the two DTOs would delete all three sites and publish the valid values in the OpenAPI schema.

cognee/modules/search/models/SearchResultPayload.py
```diff
@@ -2,6 +2,10 @@
 from typing import Optional, Any, List, Union
 from pydantic import BaseModel, ConfigDict, field_serializer
 from pydantic.alias_generators import to_camel
+from cognee.modules.retrieval.context_preview import (
```
NMZivkovic: **Layering inversion: a leaf result model now imports its consumer layer at module scope, just to read two string literals.**

Static import-closure check: this module's cognee closure goes from 2 modules to ~220+. `context_preview` pulls in `cognee.context_global_variables` (users methods, relational/graph/vector DB config, `get_or_create_dataset_database`) and `retrieval.utils.completion` -> `LLMGateway` -> `retry_config`, which does `import litellm` at module scope (`retry_config.py:21`).

No cycle today, but any module already inside that closure that later needs `SearchResultPayload` becomes a partially-initialized-module `ImportError`. The PR already feels this pressure — `load_read_only_session_prompt` does its `cognee.infrastructure.session.*` imports lazily inside the function body for the same reason.

The repo convention for a closed string knob is a `(str, Enum)` next to `SearchType` — compare `ContextProfile` + `VALID_PROFILES` for the sibling `context_profile` knob. Moving `CONTEXT_FORMAT_*` into `cognee/modules/search/types/` fixes the weight and the eight duplicated imports at once.

cognee/api/v1/search/routers/get_search_router.py
```diff
@@ -64,6 +65,16 @@ class SearchPayloadDTO(InDTO):
     )
     top_k: Optional[int] = Field(default=15)
     only_context: bool = Field(default=False)
+    context_format: str = Field(
```
NMZivkovic: **This description promises "session guidance, conversation history", but `SearchPayloadDTO` has no `session_id` field and the router never forwards one.**

`POST /api/v1/search` with `onlyContext: true, contextFormat: 'prompt'` reaches `build_context_preview(session_id=None)`, and the retrievers this router builds carry no `session_id` either, so `resolve_session_id(None)` picks the default session.

Every HTTP and cloud-client caller therefore gets the *default* session's guidance block and history, reported as "the envelope a completion would have received" — regardless of which conversation they're actually in. Only `/api/v1/recall` carries a session id.

Either add `session_id` to this DTO, or note in the description that the session layer on this endpoint is always the default session.

CLAUDE.md
```diff
@@ -400,6 +400,34 @@ Concurrent mode applies only to `GraphCompletionRetriever`,
...
+
+`only_context=True` returns the retrieval context instead of an LLM completion. By
+default that is the bare context string and nothing else — no session guidance, no
+conversation history, no rendered prompt — which is less than a real completion
+receives. Pass `context_format="prompt"` to get the full envelope instead:
+
+```python
+result = await cognee.recall(
+    "why did the migration stall?",
+    session_id="s1",
```
NMZivkovic: **This example won't produce the envelope it documents.**

A bare `session_id` with no `datasets` and no `query_type` is exactly the shape that hits `recall.py:468`:

```python
if session_id and not has_dataset_scope and query_type is None:
    sources = ['session', 'graph']
    auto_fallthrough = True  # session hit short-circuits graph
```

and then `recall.py:938`: `if auto_fallthrough and src == 'graph' and merged: break`.

The session lane never applies `context_format`, so as soon as session memory returns anything the graph lane is skipped and the reader gets session entries — not `question`/`context`/`session_context`/`user_prompt`/`system_prompt`. This doc already states that rule at line 110.

Adding a `datasets=[...]` or an explicit `query_type=` to the example makes it do what the surrounding text says.

cognee/tests/unit/modules/retrieval/test_context_preview.py
```diff
@@ -0,0 +1,298 @@
...
+
+@pytest.mark.asyncio
+async def test_session_prompt_performs_no_session_writes(as_user):
+    """The guidance block must be built with stamp_served=False."""
+    manager = _FakeSessionManager()
+    with patched_session(manager, block="## Active session guidance\n- x") as mocks:
+        await load_read_only_session_prompt(_PromptRetriever(), "why?")
+
+    assert mocks.build_block.await_args.kwargs["stamp_served"] is False
+    assert manager.updated_entries == []
```
NMZivkovic, line 215: **These two assertions can't fail.**

The only write on this path lives *inside* `build_active_context_block`, which `patched_session` replaces wholesale with an `AsyncMock`. So `manager.updated_entries` and `manager.added_qas` are unreachable under any implementation — including a broken one.

If someone later flips `stamp_served` back to the default, or adds a direct `add_qa`/`update_session_context_entry` call to `load_read_only_session_prompt`, this test still passes and `only_context` previews start stamping `last_served_at` on session-context entries — mutating guidance-ranking state on a path the module docstring guarantees is read-only.

Only the `stamp_served is False` kwarg check is real, and it asserts the argument rather than that the argument suppresses the write. Letting the real `build_active_context_block` run against `_FakeSessionManager` would make the other two assertions load-bearing.

cognee/tests/unit/modules/retrieval/test_context_preview.py
```diff
@@ -0,0 +1,298 @@
...
+    with patch.object(
+        get_session_manager_module,
+        "get_session_manager",
+        side_effect=RuntimeError("cache down"),
+    ):
+        assert await load_read_only_session_prompt(_PromptRetriever(), "why?") == ""
+
+
+@pytest.mark.asyncio
+async def test_preview_matches_what_generate_completion_would_build(as_user):
```
NMZivkovic: **"The whole point" test is self-referential, so it can't catch the one thing the preview adds on top of the shared assembly.**

Two issues:

1. `expected_user`/`expected_system` are computed with the same `build_completion_prompts` the implementation calls, and fed the implementation's own output back in as input (`conversation_history=preview.session_context`). That reduces to `f(x) == f(x)` for everything the shared helper already guarantees.
2. `context="node1 -- rel -- node2"` is a plain string, so `render_context_for_prompt` is the identity — the only transformation `build_context_preview` layers on top is never exercised.

If `render_context_for_prompt` starts joining with the wrong separator, drops entries, or `str()`-reprs a list context, this test still passes. A case with `context=[...]` asserting the `\n---\n` join actually appears in `user_prompt` would close that.

NMZivkovic (approved):

PR #4659, opened 2026-08-25
fix(retrieval): thread live node schemas into NL Cypher-generation prompt

### LINKED ISSUE #4655: [Bug]: NATURAL_LANGUAGE never sees node labels — node schemas are computed then dropped before the prompt
`NaturalLanguageRetriever` computes node schemas and then passes only the edge schemas into the Cypher-generation prompt, so the model is never told which node labels exist. It guesses, and every generated query fails to bind.

## Where

`cognee/modules/retrieval/natural_language_retriever.py`, unchanged between v1.4.2 and v1.5.2.

`_get_graph_schema` queries and returns both:

```python
node_schemas = await graph_engine.query(...)
edge_schemas = await graph_engine.query(...)
node_schemas = [row for row in node_schemas or [] if not _is_internal_schema_row(row)]
return node_schemas, edge_schemas
```

The caller unpacks both:

```python
node_schemas, edge_schemas = await self._get_graph_schema(graph_engine)
```

and then passes only one:

```python
cypher_query = await self._generate_cypher_query(
    query, edge_schemas, previous_attempts
)
```

`_generate_cypher_query(self, query, edge_schemas, previous_attempts=None)` builds its prompt context from `{"edge_schemas": edge_schemas, ...}`. `node_schemas` is computed, filtered, returned, unpacked, and then dropped.

## Effect

The generated Cypher references labels the model inferred rather than labels that exist. On our store it consistently produced `DocumentChunk`, `Entity` and `TextSummary` as if they were tables, and every query failed to bind.

We ran `NATURAL_LANGUAGE` against a 6,879-document store as one arm of a retrieval comparison. It returned an empty result for all 29 questions, scoring 0/29 where other search types on the same store scored 20/29 to 28/29. The failure is silent from the caller's side: empty results, no error.

## Suggested fix

Pass the node schemas into the prompt alongside the edges, which is what the docstring of `_get_graph_schema` implies they are collected for. Happy to send a PR if you would prefer it in that form.

## Environment

cognee 1.4.2, kuzu/ladybug graph adapter, LanceDB vector adapter. Code inspected at v1.5.2 as well; the behaviour is unchanged there.

sainikhiljuluri:
## Description

`NaturalLanguageRetriever` translates a natural-language question into Cypher using an LLM, and to do that it shows the LLM the graph's schema. It fetches **both** the node schema and the edge schema, but only the edge schema actually reached the model:

- `_execute_cypher_query` computes `node_schemas, edge_schemas = await self._get_graph_schema(...)`, then calls `_generate_cypher_query(query, edge_schemas, previous_attempts)` — `node_schemas` is dropped on the floor.
- `_generate_cypher_query` only put `edge_schemas` into the prompt context.
- The prompt template compounded this: it rendered edges live via `{{edge_schemas}}`, but the node section was a **hardcoded static block** listing five fixed labels (`EntityType`, `Entity`, `TextDocument`, `DocumentChunk`, `TextSummary`) with fixed properties.

So the LLM never saw the store's *real* node labels. On any graph whose labels differ from that static set, it generated Cypher against guessed labels that bind nothing, and `NATURAL_LANGUAGE` search silently returned empty results for every query — no error, just nothing. (The issue reporter measured 0/29 answered for `NATURAL_LANGUAGE` versus 20–28/29 for the other search types on the same store.)

The fix threads the already-computed `node_schemas` through to the prompt using the exact same mechanism `edge_schemas` already uses, and replaces the static block with `{{node_schemas}}`. No new graph query is issued — `node_schemas` was already being fetched and simply discarded. Internal schema rows are still filtered upstream in `_get_graph_schema` (`_is_internal_schema_row`), so nothing internal is newly exposed.

## Acceptance Criteria

- The live node labels of the store appear in the Cypher-generation system prompt (they did not before).
- Existing internal-node filtering behaviour is unchanged (internal schema rows stay out of the prompt).
- No regression in the retriever test suite.

Proof: a new regression test renders the **real** template through `render_prompt` and asserts a distinctive live node label reaches the generation prompt. It fails on the pre-fix code (`AssertionError: live node labels were not threaded into the Cypher-generation prompt`) and passes with the fix. Retriever tests: 16 passed, 1 skipped.

## Type of Change
- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots

Non-UI change. Test evidence (local): the regression test passes with the fix and fails without it (`AssertionError: live node labels were not threaded into the Cypher-generation prompt`).

## Pre-submission Checklist
- [x] My code follows the project's coding standards and style guidelines (ruff format + check clean)
- [ ] I have added necessary documentation (if applicable) — n/a

## DCO Affirmation

Fixes #4655

### DIFF
--- a/cognee/infrastructure/llm/prompts/natural_language_retriever_system.txt
+++ b/cognee/infrastructure/llm/prompts/natural_language_retriever_system.txt
@@ -33,26 +33,8 @@ ERROR PREVENTION:
 4. Use pattern variables consistently throughout the query
 5. If previous attempts failed, analyze the failures and adjust your approach
 
-Node schemas:
-- EntityType
-Properties: description, ontology_valid, name, created_at, type, version, topological_rank, updated_at, metadata, id
-Purpose: Represents the categories or classifications for entities in the database.
-
-- Entity
-Properties: description, ontology_valid, name, created_at, type, version, topological_rank, updated_at, metadata, id
-Purpose: Represents individual entities that belong to a specific type or classification.
-
-- TextDocument
-Properties: raw_data_location, name, mime_type, external_metadata, created_at, type, version, topological_rank, updated_at, metadata, id
-Purpose: Represents documents containing text data, along with metadata about their storage and format.
-
-- DocumentChunk
-Properties: version, created_at, type, topological_rank, cut_type, text, metadata, chunk_index, chunk_size, updated_at, id
-Purpose: Represents segmented portions of larger documents, useful for processing or analysis at a more granular level.
-
-- TextSummary
-Properties: topological_rank, metadata, id, type, updated_at, created_at, text, version
-Purpose: Represents summarized content generated from larger text documents, retaining essential information and metadata.
+Node schemas (node labels and their properties):
+`{{node_schemas}}`
 
 Edge schema (relationship properties):
 `{{edge_schemas}}`
--- a/cognee/modules/retrieval/natural_language_retriever.py
+++ b/cognee/modules/retrieval/natural_language_retriever.py
@@ -101,11 +101,14 @@ class NaturalLanguageRetriever(BaseRetriever):
         node_schemas = [row for row in node_schemas or [] if not _is_internal_schema_row(row)]
         return node_schemas, edge_schemas
 
-    async def _generate_cypher_query(self, query: str, edge_schemas, previous_attempts=None) -> str:
+    async def _generate_cypher_query(
+        self, query: str, node_schemas, edge_schemas, previous_attempts=None
+    ) -> str:
         """Generate a Cypher query using LLM based on natural language query and schema information."""
         system_prompt = render_prompt(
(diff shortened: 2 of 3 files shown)


ptimizeroracle:
@sainikhiljuluri checking in: this pr has been quiet for about 25 days. are you still planning to carry it forward? if life got busy, no problem at all: i am happy to take #4655 over from here, starting from your branch with credit, or fresh if you prefer.

ptimizeroracle:
@sainikhiljuluri taking this forward as offered in my check-in above (no response after three days, and this pr has been quiet for a month). your fork branch is gone so i recovered the diff from this pr itself and rebuilt it on current dev: the template placeholder, the retriever threading, and your regression test all applied cleanly, and your test fails on dev without the fix and passes with it. full credit to you in the replacement pr on #4655; chime in anytime and i will fold in whatever you were still planning.

PR #4660, opened 2026-08-25
fix(graph): dedup skeleton neighbours by id set, not O(D^2) list scan

### LINKED ISSUE #4648: Graph projection is quadratic in node degree (`add_skeleton_neighbor` scans a list)
Projecting a graph gets very slow when one node has a lot of neighbours. The time grows with the **square** of that node's degree.

Adding neighbours to a single node, median of 5 runs:

| Neighbours | Time |
|---|---|
| 1,000 | 0.090s |
| 2,000 | 0.379s |
| 4,000 | 1.338s |

Double the neighbours, roughly quadruple the time.

## Why

`Node.add_skeleton_neighbor` looks for the neighbour in a list before adding it:

```python
def add_skeleton_neighbor(self, neighbor: "Node") -> None:
    if neighbor not in self.skeleton_neighbours:
        self.skeleton_neighbours.append(neighbor)
```

`CogneeGraph.add_edge` calls `add_skeleton_edge` on both ends of every edge, and that calls this method. Each call scans a list as long as the node's current degree. At degree 4,000 that is about 8 million comparisons for one node.

## To reproduce

```python
import time
from cognee.modules.graph.cognee_graph.CogneeGraphElements import Node, Edge

for degree in (1000, 2000, 4000):
    hub = Node("hub", {}, dimension=1)
    others = [Node(str(i), {}, dimension=1) for i in range(degree)]
    start = time.perf_counter()
    for other in others:
        hub.add_skeleton_edge(Edge(hub, other, attributes={}, dimension=1))
    print(f"degree={degree:<6} {time.perf_counter() - start:7.3f}s")
```

Python 3.11, pydantic 2.13.4. The loop also builds the `Edge` objects, which is about 1% of the time before any change.

## A fix that works

`Node` already hashes and compares by id, so a set of ids answers the same question in O(1). With one, the numbers above become 0.006 / 0.012 / 0.024s.

Two things surprised us while testing it.

**Do not declare the set as a `PrivateAttr` field.** That slows down building *every* `Node`, even ones that never get a neighbour. We measured 20,000 nodes going from 0.11s to 0.60s. On a graph with many nodes and normal degree, that cancels out most of the speedup. Creating the set on first use instead keeps node building at 0.12s and keeps the whole speedup.

**A plain set drifts out of sync with the list.** `skeleton_neighbours` is public and can be changed directly, and pydantic can build a `Node` without calling `__init__`. When that happens you get duplicate neighbours, or a `ValueError` from `list.remove`. Two guards fixed every case we found:

- rebuild the set when its size does not match the list — covers `model_construct()`, unpickling an older `Node`, and any change that alters the length
- reset it in `model_copy` — covers `model_copy(update={"skeleton_neighbours": ...})`

On a test graph of 20,000 nodes and 83,000 edges, the whole projection goes from 1.83s to 0.96s. When one node holds 25,000 of those edges, it goes from 38.6s to 0.83s.

## Two cases this does not fix

The set can still disagree with the list without the length changing:

1. **Replacing an item in the list in place** — `neighbours[0] = other`, or `pop()` then `append()`.
2. **Changing a neighbour's `.id` after adding it.** Today the code compares ids at the moment of the call, so it follows the change. A stored id cannot.

## The question

**Are `Node.id` and `skeleton_neighbours` meant to be changed directly after a node is built?**

If **no**, then freezing `id` and only changing the list through the two methods makes a set safe, and both cases above become impossible. (`Field(frozen=True)` makes changing `id` raise an error, which is a visible API change — that is why we are asking rather than assuming.)

If **yes**, a set is the wrong tool here and the fix needs to be something else. An index where neighbours notify their owner when their id changes would work, but it is a lot more machinery.

That is your call, which is why this is an issue and not a PR.

Happy to send a PR either way — just say which one you want, and we will include tests for the cases above.

sainikhiljuluri:
## Description

`Node.add_skeleton_neighbor` deduplicated neighbours with a list membership scan:

    def add_skeleton_neighbor(self, neighbor):
        if neighbor not in self.skeleton_neighbours:   # O(degree) scan every call
            self.skeleton_neighbours.append(neighbor)

`CogneeGraph.add_edge` calls `add_skeleton_edge` (→ `add_skeleton_neighbor`) on **both** endpoints of every edge, so building a node of degree `D` costs `O(D²)`. This shows up as slow graph projection for high-degree nodes (the behaviour in #4648).

Measured on this machine (build a node with N neighbours):

| N | before | after |
|---|---|---|
| 1000 | 20 ms | 0.8 ms |
| 2000 | 80 ms | 1.8 ms |
| 4000 | 314 ms | 3.5 ms |
| 8000 | 1269 ms | 6.9 ms |

Before, time quadruples as degree doubles (O(D²)); after, it scales linearly (~180× faster at 8k).

## Fix

`Node` equality is by id (`__eq__`/`__hash__` use `self.id`), so list-membership dedup is exactly id dedup. Keep a `PrivateAttr` set of neighbour ids for O(1) membership in `add_skeleton_neighbor` / `remove_skeleton_neighbor`, synced via `model_post_init`. `skeleton_neighbours` remains the same serialized `List[Node]`, so ordering, the `field_serializer`, and the public interface are unchanged.

## Acceptance Criteria

- Neighbour dedup behaviour is unchanged (a repeated add, incl. a distinct object with the same id, is still a no-op).
- A neighbour removed and then re-added is added back (the id index stays in sync).
- Building a high-degree node is linear, not quadratic (table above).

Regression tests added in `cognee_graph_elements_test.py` (`test_add_skeleton_neighbor_deduplicates_by_id`, `test_skeleton_neighbor_remove_then_readd`). Full graph-elements suite: 25 passed. `ruff format`/`ruff check` clean.

## Type of Change
- [ ] New feature
- [ ] Code refactoring
- [ ] Other

## Screenshots

Non-UI change; before/after timing table above (in-process microbenchmark, no network).

## Pre-submission Checklist
- [x] This PR contains minimal changes necessary to address the issue
- [ ] I have added necessary documentation (if applicable) — n/a

## DCO Affirmation

Fixes #4648

### DIFF
--- a/cognee/modules/graph/cognee_graph/CogneeGraphElements.py
+++ b/cognee/modules/graph/cognee_graph/CogneeGraphElements.py
@@ -1,6 +1,6 @@
 import numpy as np
 from typing import List, Dict, Optional, Any, Union
-from pydantic import BaseModel, ConfigDict, field_serializer
+from pydantic import BaseModel, ConfigDict, PrivateAttr, field_serializer
 from cognee.modules.graph.exceptions import InvalidDimensionsError, DimensionOutOfRangeError
 
 
@@ -20,6 +20,15 @@ class Node(BaseModel):
     skeleton_edges: List["Edge"]
     status: np.ndarray
     model_config = ConfigDict(arbitrary_types_allowed=True)
+
+    # O(1) membership index for skeleton_neighbours, kept in sync with the list.
+    # Node equality is by id (see __eq__/__hash__), so id membership == Node
+    # membership. Without this, add_skeleton_neighbor scanned the list on every
+    # call, making a degree-D node cost O(D^2) to build.
+    _neighbour_ids: set = PrivateAttr(default_factory=set)
+
+    def model_post_init(self, __context: Any) -> None:
+        self._neighbour_ids = {neighbor.id for neighbor in self.skeleton_neighbours}
 
     def __init__(
         self,
@@ -75,11 +84,13 @@ class Node(BaseModel):
         distances[query_index] = score
 
     def add_skeleton_neighbor(self, neighbor: "Node") -> None:
-        if neighbor not in self.skeleton_neighbours:
+        if neighbor.id not in self._neighbour_ids:
+            self._neighbour_ids.add(neighbor.id)
             self.skeleton_neighbours.append(neighbor)
 
     def remove_skeleton_neighbor(self, neighbor: "Node") -> None:
-        if neighbor in self.skeleton_neighbours:
+        if neighbor.id in self._neighbour_ids:
+            self._neighbour_ids.discard(neighbor.id)
             self.skeleton_neighbours.remove(neighbor)
 
     def add_skeleton_edge(self, edge: "Edge") -> None:
--- a/cognee/tests/unit/modules/graph/cognee_graph_elements_test.py
+++ b/cognee/tests/unit/modules/graph/cognee_graph_elements_test.py
@@ -38,6 +38,27 @@ def test_remove_skeleton_neighbor():
     assert node2 not in node1.skeleton_neighbours
 
 
+def test_add_skeleton_neighbor_deduplicates_by_id():
+    """Neighbours are deduplicated by id (a repeated add is a no-op)."""
+    node1 = Node("node1")
+    node2 = Node("node2")
+    node1.add_skeleton_neighbor(node2)
+    node1.add_skeleton_neighbor(node2)  # same object again
+    node1.add_skeleton_neighbor(Node("node2"))  # distinct object, same id
+    assert node1.skeleton_neighbours == [node2]
+
+
+def test_skeleton_neighbor_remove_then_readd():
+    """After removal a neighbour can be added again (id index stays in sync)."""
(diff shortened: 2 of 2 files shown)


abhinandan6123 (changes requested):
The O(1) index can become stale because skeleton_neighbours remains publicly mutable. For example, after adding a neighbour, calling node.skeleton_neighbours.clear() leaves _neighbour_ids unchanged, so adding the same neighbour again incorrectly becomes a no-op. It can also cause remove_skeleton_neighbor() to raise ValueError when the id exists in the index but the neighbour is no longer in the list.

Since the PR keeps skeleton_neighbours as a mutable public List[Node], the private set and list can diverge. Please either ensure mutations to skeleton_neighbours cannot bypass synchronization, or use a design that keeps the membership index reliably synchronized with the public list.

Image

dexters1:
Thanks for the PR @sainikhiljuluri! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #4665, opened 2026-08-25
docs: remove stray colon in README translations section

roarzx:
## What
Removed a lone `:` line in the "This README is also available in:" block that
rendered as a stray colon above the language links.

## Why
Leftover from editing — the sentence already ends with a colon on the previous
line.

## How verified
Confirmed the surrounding block renders cleanly with no duplicate punctuation.


Vasilije1990:
Don't waste my time

PR #4677, opened 2026-08-26
Fix/4632 graph model config

### LINKED ISSUE #4632: [Bug]: cognee.config.set_graph_model() is inert — graph extraction ignores the configured model
## Bug Description

`cognee.config.set_graph_model(MyModel)` is documented as setting the graph
extraction model, but it has no effect on extraction. The setter writes
`graph_config.graph_model`, and no extraction path ever reads that field. As a
result, calling `set_graph_model(...)` silently does nothing — extraction always
uses the hardcoded default (`KnowledgeGraph`) unless a model is passed per call
via `graph_model=`.

The [config docs](https://docs.cognee.ai) list `set_graph_model(model)` as "Set
graph extraction model", and its docstring says "Set the model used for graph
extraction." Neither matches the behavior.

## Steps to Reproduce

```python
import cognee
from pydantic import BaseModel

class MyGraphModel(BaseModel):
    # any custom graph schema
    ...

cognee.config.set_graph_model(MyGraphModel)

await cognee.add("some text")
await cognee.cognify()   # extraction runs under KnowledgeGraph, NOT MyGraphModel
```

## Expected Behavior

After `set_graph_model(MyGraphModel)`, `cognify()` (and other extraction entry
points) use `MyGraphModel` as the graph schema, the same as passing
`graph_model=MyGraphModel` per call.

## Actual Behavior

The configured model is ignored. Extraction uses the hardcoded `KnowledgeGraph`.

Root cause — the write and the read are disconnected:

- **Write:** `cognee/api/v1/config/config.py` — `set_graph_model()` does
  `graph_config.graph_model = graph_model`.
- **Reads of `graph_config.graph_model` across the whole tree:** only that write
  line, plus `GraphConfig.to_dict()` self-serialization. No extraction code reads
  it.
- **How extraction actually picks the model:** `cognee/api/v1/cognify/cognify.py`
  hardcodes `graph_model: BaseModel = KnowledgeGraph` (two signatures) and threads
  that *parameter* down to `extract_graph_from_data` → `extract_content_graph`.
  The parameter chain never consults `GraphConfig`.

So `set_graph_model` stores a value that only the config's own `to_dict()` echoes
back. The only working lever is the per-call `graph_model=` argument.

Verified on tag `v1.5.2` (`df6b549`) and current `main` (`a8f9760bb`): exactly one
occurrence of `graph_config.graph_model` in `config.py` (the write), and
`cognify.py` still defaults to `KnowledgeGraph`.

## Environment

- OS: macOS 15 (behavior is platform-independent)
- Python version: 3.13
- Cognee version: 1.5.2 (also reproduced on `main` @ `a8f9760bb`)
- LLM Provider: provider-independent (config-wiring bug, not provider-specific)
- Database: graph-backend-independent

## Additional Context

Suggested fix, one of:

1. Make extraction read the configured default when no per-call `graph_model=` is
   given — e.g. `cognify()` defaults `graph_model` from
   `get_graph_config().graph_model` instead of a hardcoded `KnowledgeGraph`, so
   `set_graph_model()` becomes the global default it is documented to be.
2. If a global default is intentionally unsupported, remove `set_graph_model` (and
   its docs row/docstring) to avoid advertising a no-op.

Option 1 matches the documented behavior and the pattern already used for other
`set_*_model` setters.

Nchaudry123:
## Summary

- Resolve the graph model from `config.set_graph_model()` when `cognify()` receives no explicit model.
- Preserve explicit `graph_model=` precedence.
- Preserve the default `KnowledgeGraph` behavior.
- Allow REST cognify requests with no custom schema to use the configured model.
- Correct graph-model annotations to describe Pydantic model classes.
- Add regression coverage for Python and REST entry points.

Closes #4632.

## Related work

This overlaps with #4633, but additionally fixes the REST cognify path, where an omitted schema was converted into an explicit `KnowledgeGraph`, bypassing the configured model.

This implementation also tests the real `config.set_graph_model()` behavior, explicit precedence, default compatibility, and REST propagation.

### Test evidence

**Configured graph-model regression tests - 4 passed**

Screenshot 2026-08-26 013805

**Broader cognify and graph-model suite - 20 passed**

Screenshot 2026-08-26 013840

**Pre-commit validation - all applicable hooks passed**

Screenshot 2026-08-26 013626

- `uvx pre-commit run --from-ref upstream/dev --to-ref HEAD`
- `uv run pytest cognee/tests/unit/api/v1/cognify/ cognee/tests/test_graph_model_utils.py -q`
- 20 focused tests passed
- Repository-wide Ruff lint and formatting checks passed


dexters1:
Thanks for the PR @Nchaudry123! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #4679, opened 2026-08-26
fix(llm): Stop retrying LiteLLM budget exhaustion (COG-6329)

GraDKh:
## Problem

A LiteLLM-proxy spend-cap rejection cannot clear by retrying, but cognee classified it as a transient rate limit.

Measured on the production gateway:

| | |
|---|---|
| HTTP requests per doomed logical LLM call | **~36** over ~270s |
| Share of all gateway traffic that was budget rejections | **93%** (2,904 of 2,969 in 15 min) |
| Tenants responsible | **6**, all with zero successful requests |

The amplification is three nested retry layers, only the outermost of which backs off:

| Layer | Attempts | Backoff |
|---|---|---|
| cognee tenacity (`retry_config`) | ~6 | 8→17→33→64→128s ✅ |
| instructor `MAX_RETRIES=2` | 2 | none (`wait_none`) |
| litellm/openai client | 3 | ~0.5s, ~0.8s |

Side effects: doomed calls held pipeline concurrency slots for ~4.5 minutes each, and the ~15-line traceback logged per rejection rotated the gateway's own logs down from 24h of history to ~4h.

## Why the classifier missed it

The rejection arrives as:

```
InstructorRetryException  →  RetryError  →  RateLimitError (status_code=429)
```

Every structured signal is destroyed on the way up, so all three existing branches of `is_budget_exhausted_error` are unreachable for this path:

- the outermost exception has no `status_code` and no `.response`;
- the client-side `RateLimitError` is a **different class** from `litellm.BudgetExceededError`;
- that `RateLimitError`'s `.response.json()` raises `JSONDecodeError` — the body is already consumed, so the proxy's `error.type == "budget_exceeded"` cannot be read.

Message text is the only signal that survives.

## Changes

**`exceptions.py`** — `is_budget_exhausted_error` walks the `__cause__` chain (deliberately not `__context__`, so an unrelated error raised while handling a budget error isn't misclassified) and adds message wording as a fourth, wrapper-proof case. Matching is **conjunctive** because `str(InstructorRetryException)` embeds the model's own partial completion — a single natural-language phrase would turn an ordinary validation-retry failure into a hard 402 for any document discussing an exceeded budget. All three LiteLLM sentence shapes are covered:

```
Budget has been exceeded! [Scope=...] Current cost: X, Max budget: Y   key/team/project/org/tag
ExceededBudget: [End ]User=<id> over budget. Spend=X, Budget=Y         internal-user, end-user
LiteLLM {Virtual Key|End User}: <id>, exceeded budget for model=<m>    per-model caps
```

Also adds `budget_exhaustion_detail()` (extracts just the provider sentence, bounded, so a hostile completion can't bloat the 402 body) and `raise_if_budget_exhausted()`.

**`retry_config.py`** — the check goes in `should_retry_llm_exception`, **not** `_TERMINAL_QUOTA_PATTERNS`. Putting it in the patterns tuple would also feed `is_quota_or_billing_error`, causing `raise_if_quota_error` to rewrite the 402 into a 422. A defensive `LLMPaymentRequiredError` short-circuit was added there too.

**`generic_llm_api/adapter.py`** — the budget check must run at the re-raise site: the `except` clause catching `InstructorRetryException` shadows the budget handler further down in the same `try`. It runs *before* the content-policy branch, so a rejection whose completion happens to mention a content policy isn't misrouted to the fallback path.

**`LLMGateway._fail_fast_on_quota`** — classification also happens at the single choke point every structured-output call passes through. Each instructor adapter has the same shadowing pattern, and the plain-text path (`response_model is str`) bypasses adapter budget handlers entirely, so `openai` / `azure_openai` / `gemini` / `bedrock` and `POST /api/v1/llm` were still surfacing a raw `InstructorRetryException` as HTTP 500. This gives every adapter and both paths the same 402.

**Matching is whole-sentence.** Detection and extraction share one regex, so fragments scattered across several KB of model output cannot trigger a match, and a positive classification always yields a detail. Identifiers in that detail (virtual-key hash, key alias, tenant IDs) are masked before it reaches the 402 body; spend and budget figures are kept, since they are the caller's own and are the actionable part.

## Verification

Against a local proxy returning the production 429 body:

| | before | after |
|---|---|---|
| HTTP requests | ~36 | **6** |
| Wall clock | ~270 s | **2.8 s** |
| Exception surfaced | `InstructorRetryException` | `LLMPaymentRequiredError` (402) |

Through the gateway choke point, the paths that previously returned HTTP 500:

| Path | after |
|---|---|
| `openai` adapter (unpatched, default provider) | 402, 6 requests, 2.76s |
| plain-text `response_model=str` | 402, 3 requests, 1.26s |

A control run with the new checks stubbed out reproduces 36 requests / 266s exactly, confirming this is the load-bearing change.

`cognee/tests/unit/infrastructure/llm/test_budget_exhaustion.py` adds 64 cases (0.11s, no network). It rebuilds the real exception chain with a populated `FailedAttempt` so the model's completion is embedded exactly as in production, and covers: the chain shape itself, all ten message variants, prose false-positives (bare and inside a completion), `__context__` not being followed, cycle and deep-chain handling, 402-not-downgraded-to-422, and bounded detail extraction.

Each assertion was mutation-tested — reverting any part of the fix is caught by at least one test.

Full suite: **255 passed, 7 skipped**. `ruff check` and `ruff format --check` clean.

## Scope / follow-ups

- The remaining 6 requests come from instructor (`MAX_RETRIES=2`) × the litellm client (3 tries). Closing that is a separate change.
- `openai`, `azure_openai`, `gemini` and `bedrock` still contain the same shadowing pattern in their own `except` clauses. It is no longer user-visible now that classification happens in `_fail_fast_on_quota`, but cleaning it up locally would be a reasonable follow-up.
- One residual false positive is accepted and pinned as a test: a document quoting a LiteLLM error verbatim is classified as budget exhaustion. The trade-off is deliberately asymmetric — a false negative resurrects the production storm, whereas this costs one already-failing call a wrong 402.

### DIFF
--- a/cognee/infrastructure/llm/LLMGateway.py
+++ b/cognee/infrastructure/llm/LLMGateway.py
@@ -5,6 +5,7 @@ from pydantic import BaseModel
 
 from cognee.infrastructure.llm import get_llm_config
 from cognee.infrastructure.llm.config import get_llm_context_config
+from cognee.infrastructure.llm.exceptions import raise_if_budget_exhausted
 from cognee.infrastructure.llm.retry_config import raise_if_quota_error
 from cognee.infrastructure.llm.structured_output_framework.litellm_instructor.llm.types import (
     TranscriptionReturnType,
@@ -80,14 +81,21 @@ async def _record_session_usage_after(
 
 
 async def _fail_fast_on_quota(coro: Coroutine) -> T:
-    """Convert provider quota/billing exhaustion into ``LLMQuotaExceededError``.
+    """Convert provider quota/billing exhaustion into an actionable error.
 
     Runs at the single choke point every structured-output call flows through,
     so it is provider- and framework-agnostic.
+
+    Budget exhaustion is checked first, and here rather than only in the
+    adapters: each instructor adapter catches ``InstructorRetryException`` in a
+    clause that precedes its own budget handler, and the plain-text path
+    (``response_model is str``) bypasses those handlers altogether. Classifying
+    at this choke point gives every adapter and both paths the same 402.
     """
     try:
         return await coro
     except Exception as error:
+        raise_if_budget_exhausted(error)
         raise_if_quota_error(error)
         raise
 
--- a/cognee/infrastructure/llm/exceptions.py
+++ b/cognee/infrastructure/llm/exceptions.py
@@ -1,3 +1,5 @@
+import re
+
 from cognee.exceptions.exceptions import CogneeValidationError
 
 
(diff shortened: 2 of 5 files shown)


Vasilije1990:
@GraDKh please open a Linear issue and solve tests

PR #4682, opened 2026-08-26
feat: Add transparent graph roots and chunk_attachment (SDK-163)

lxobr:
## Description

Lets a custom `graph_model` use a wrapper class that never reaches the graph, and lets a chunk link to every entity extracted from it instead of only the extraction root (SDK-163).

- A DataPoint class with `metadata["transparent"] = True` is skipped by the graph walk and its DataPoint children are promoted to roots, so one extraction can produce several top-level nodes, or none. The wrapper is never stored and never an edge endpoint.
- Non-DataPoint fields declared on a transparent class have nowhere to go once the wrapper is dropped. They are ignored, and each one warns once per class field rather than disappearing silently.
- New `cognify(chunk_attachment=...)`: `"all"` links each chunk once to every stored node reachable from the extracted root, so any entity is one hop from its source chunk. `"direct"` and omitting it keep the current single-root linkage, so existing runs are unaffected.
- `chunk_attachment` is rejected up front for an invalid value, a non-DataPoint `graph_model`, `temporal_cognify=True`, or a remote connection — `cognify()` forwards unknown keywords into the LLM call, where a bad value would be swallowed. It is allowed with `dry_run=True`, whose estimate is genuinely unchanged.
- The provenance task no longer attributes nodes to a root the walk did not store, which a transparent root now makes possible.
- The graph walk was reorganised to support the above: fields are classified once per node instead of twice, container resolution and field reading each moved into their own module, the recursion is synchronous (nothing in it touches I/O), and the unused `include_root` parameter is gone. `get_graph_from_model`'s public signature is otherwise unchanged.

## Acceptance Criteria

- A transparent wrapper class contributes no node and no edge to the stored graph.
- Its DataPoint children are stored as top-level nodes, including when a field points at one.
- `chunk_attachment="all"` puts every node reachable from a chunk's extracted root in that chunk's `contains`.
- Graphs with no transparent node and no `chunk_attachment` produce the same nodes and edges as before.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/api/v1/cognify/cognify.py
+++ b/cognee/api/v1/cognify/cognify.py
@@ -1,6 +1,6 @@
 import asyncio
 from pydantic import BaseModel
-from typing import Collection, Union, Optional
+from typing import Collection, Literal, Union, Optional
 from uuid import UUID
 
 from cognee.modules.cognify.config import get_cognify_config
@@ -9,6 +9,7 @@ from cognee.modules.cognify.routing import CognifyRoute, cognify_route_for
 from cognee.modules.ontology.ontology_env_config import get_ontology_env_config
 from cognee.shared.logging_utils import get_logger
 from cognee.shared.data_models import KnowledgeGraph
+from cognee.infrastructure.engine import DataPoint
 from cognee.infrastructure.llm import get_max_chunk_tokens
 
 from cognee.modules.pipelines import run_pipeline
@@ -71,6 +72,7 @@ async def cognify(
     embedding_config: Optional[EmbeddingConfig] = None,
     data_cache: bool = True,
     dry_run: bool = False,
+    chunk_attachment: Optional[Literal["direct", "all"]] = None,
     **kwargs,
 ):
     """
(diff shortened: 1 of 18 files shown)


lxobr:
**A note on the last commit.**

CI failed on `test_consolidate_entities_merges_real_graph`. This PR does not touch that test or the code it exercises.

That test creates two entities, "New York City" and "NYC", and expects cognee to notice they mean the same thing and merge them into one node. Cognee decides that by embedding both names and comparing them, so the test only passes when embeddings are real.

There is an env var, `MOCK_EMBEDDING`, that makes every embedding engine return fake vectors instead of calling the provider. Four test fixtures set it to `true` so their own tests run without an API key: `test_add_ingestion_regression.py`, `test_add_s3_call_counts.py`, `test_dataset_scoped_data.py` and `test_update_id_stability.py`. They set it with `os.environ` and never set it back.

Environment variables belong to the whole Python process, not to one test or one file. So once one of those fixtures runs, every test after it in that same process also gets fake embeddings. The consolidate test then compares two meaningless vectors, finds no match, and fails.

On dev this never happens. CI splits these tests into two shards, and each shard runs as its own process. On dev the fixtures that leak are in one shard and the consolidate test is in the other, so the leak cannot reach it. This PR adds tests, which changed how the tests divide between the two shards, and a leaking test ended up in the same shard as the consolidate test.

The commit makes those four fixtures set their env vars with `pytest.MonkeyPatch()` and call `undo()` when the fixture finishes, so the variables go back to what they were. It also clears the config caches at teardown, not just at setup, so later tests build their engines from the restored environment.

Checked locally: running the leaking test and the consolidate test together went from 1 failed to 2 passed, and each of the four files still passes on its own.

NMZivkovic (approved):

Issue #4684, opened 2026-08-26
LanceDB stores are never compacted: fragments accumulate per upsert and inflate storage, query cost and memory


Screddyice:
## Summary

LanceDB stores accumulate one table version and one data fragment per upsert and are never compacted. On a long-lived self-hosted deployment this grows without bound, inflating storage, query cost and process memory.

## Mechanism

`LanceDBAdapter.py` writes through `merge_insert("id")`:

```python
collection.merge_insert("id")   # LanceDBAdapter.py:522
```

Every `merge_insert` creates a new table version and a new data fragment. LanceDB does not compact on its own — that is what `table.optimize()` / `compact_files()` / `cleanup_old_versions()` are for.

Grepping the codebase, the only `optimize()` call is inside a one-off migration:

```
cognee/modules/migrations/versions/_vector_rekey.py:143:  await optimize()
```

Nothing in normal operation ever compacts, so fragments accumulate for the life of the deployment.

## Measured on a real deployment

Self-hosted `cognee/cognee:latest` (reports `1.5.3-local`), defaults throughout: `sqlite` relational, `lancedb` vector, `ladybug` graph, `fastembed` / `BAAI/bge-small-en-v1.5` at 384 dims. Roughly one week of steady agent writes.

| | before compaction | after |
|---|---|---|
| version files | 6,592 | 48 |
| data fragments | 6,557 | 27 |
| storage (`/cognee-storage/system`) | 937 MB | 411 MB |
| container anon memory, floor after restart | 1046 MiB | 718 MiB |

Worst single table:

```
EdgeType_relationship_name.lance   versions=1446  fragments=1451   269 MB
SessionQAVector_text.lance         versions= 878  fragments= 876    63 MB
TextSummary_text.lance             versions= 859  fragments= 857    60 MB
DocumentChunk_text.lance           versions= 859  fragments= 857    62 MB
Entity_name.lance                  versions= 828  fragments= 828    83 MB
```

The underlying corpus is **5.3 MB** of ingested data producing a graph of 2,461 nodes and 4,117 edges. Nearly all of the 937 MB was version and fragment overhead.

## Why it matters beyond disk

The container's memory climbed roughly 230 MiB/hour under normal use and hit its cgroup ceiling repeatedly, forcing reclaim. We first assumed a Python leak, then glibc arena retention, and `MALLOC_ARENA_MAX=2` did **not** fix it. Fragment count tracks the growth far better: every query has to open every fragment, and the count only ever rises.

[...]

PR #4686, opened 2026-08-26
perf(storage): skip re-embedding data points whose indexed field is unchanged

Screddyice:
Fixes the write amplification reported in #4684.

## The problem

`index_data_points` re-embeds and upserts **every** point it is handed, with no check against what the store already holds:

https://github.com/topoteretes/cognee/blob/main/cognee/tasks/storage/index_data_points.py

`add_data_points` calls it on every extracted node, so any pipeline that re-processes the same content pays a full re-embed and re-write that produces byte-identical rows.

On stores that version on write, that is not free. LanceDB upserts through `merge_insert`, so each redundant write creates a new table version **and a new data fragment** — and nothing in cognee compacts them (the only `optimize()` call in the tree is inside a one-off migration).

## Measured on a self-hosted deployment

```
120 seconds:  11 pipeline runs
              +21 LanceDB fragments
              + 0 rows
```

~630 fragments/hour of pure amplification. Left alone it reached:

| | |
|---|---|
| version files | 6,592 |
| data fragments | 6,557 across 20 tables |
| storage | 937 MB for a 2,461-node graph |
| worst single table | `EdgeType_relationship_name`, 1,451 fragments, 269 MB |

Offline compaction cut that to 48 versions / 27 fragments / 411 MB with **identical row counts** — which is what established the writes were redundant rather than useful.

The trigger in our case was `POST /api/v1/improve`: one call distilled a long agent session into **344 documents** and ran an ingest pipeline per document, re-ingesting all of them each time. The cost grows with session length, so it gets worse the longer an agent runs.

## The change

Compare each point's indexed field against the stored payload, and index only what differs. One `retrieve` by id per batch replaces an embedding call plus a write per point.

**It fails open, in the direction of doing the work.** Skipping happens *only* on positive evidence that the stored value is identical. Every other case indexes exactly as before:

- point the store does not know → index
- payload without the indexed field → index
- adapter that raises, or has no `retrieve` → index
- unexpected return shape → index

So a wrong guess can never leave a vector stale; the worst case is the current behaviour.

`retrieve` is on the shared `VectorDBInterface`, so this works across adapters rather than special-casing LanceDB.

## Config

`skip_unchanged_vector_writes` (default `True`) restores the previous unconditional behaviour when set to `False` — and disables the lookup entirely rather than performing it and ignoring the result.

Happy to flip the default to `False` if you would rather ship it opt-in first.

## Tests

Seven new, alongside the four existing ones in that file (**11 passed**):

- unchanged point is not re-indexed
- changed point is re-indexed
- point the store does not know is indexed
- payload missing the indexed field is indexed
- `retrieve` failure indexes everything
- mixed batch indexes only the changed member
- flag off performs no lookup at all

Wider suite: `4577 passed`. Twelve failures exist, and all twelve are **identical on clean `main`** — verified by running the same selections with the change stashed. Two directories (`eval_framework`, `databases/graph`) fail at collection on `main` for missing optional deps.

### DIFF
--- a/cognee/infrastructure/databases/vector/embeddings/config.py
+++ b/cognee/infrastructure/databases/vector/embeddings/config.py
@@ -83,6 +83,13 @@ class EmbeddingConfig(BaseSettings):
     # Total data points allowed in flight to the embedding engine during indexing.
     # Concurrent embedding requests = max(1, this // embedding_batch_size).
     embedding_max_concurrent_data_points: int = 150
+    #: Skip re-embedding a data point whose indexed field already matches what the
+    #: vector store holds. A redundant write costs an embedding call and produces
+    #: a byte-identical row; on stores that version on write (LanceDB upserts via
+    #: merge_insert) it also creates a table version and data fragment that
+    #: accumulate until something compacts them. Set to False to restore the
+    #: previous unconditional behaviour.
+    skip_unchanged_vector_writes: bool = True
     huggingface_tokenizer: Optional[str] = None
     # Some providers (e.g. NVIDIA NIM's nv-embed family) require an
     # "input_type" field in the embedding request body (typically "query" or
@@ -142,6 +149,7 @@ class EmbeddingConfig(BaseSettings):
             "embedding_input_type": self.embedding_input_type,
             "embedding_batch_size": self.embedding_batch_size,
             "embedding_max_concurrent_data_points": self.embedding_max_concurrent_data_points,
+            "skip_unchanged_vector_writes": self.skip_unchanged_vector_writes,
             "embedding_rate_limit_enabled": self.embedding_rate_limit_enabled,
             "embedding_rate_limit_requests": self.embedding_rate_limit_requests,
             "embedding_rate_limit_interval": self.embedding_rate_limit_interval,
--- a/cognee/tasks/storage/index_data_points.py
+++ b/cognee/tasks/storage/index_data_points.py
@@ -8,6 +8,56 @@ from cognee.shared.logging_utils import get_logger
 logger = get_logger("index_data_points")
 
 
+def _embeddable_value(data_point, field_name: str):
+    """The exact value that would be embedded for ``field_name``."""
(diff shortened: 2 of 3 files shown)


Screddyice:
Closing this. **My premise was wrong, and I want to correct it clearly rather than leave it sitting in your tracker.**

I claimed redundant `improve` calls re-index unchanged data points and amplify writes. I measured that on the live deployment and it does not hold.

**A/B on the same session, one improve with genuinely nothing changed between calls:**

| build | fragments added by a fully redundant improve |
|---|---|
| stock `main` | **+0** |
| this patch | **+0** |

Cognee already avoids the redundant work. `ingest_data` only resets `pipeline_status` when the content hash changes, and downstream stages honour it, so a second improve over identical content does not reach the indexing path at all.

**Where my original measurement went wrong:** I compared *total* fragment count across all 20 tables against the row count of **one** table (`EdgeType_relationship_name`). That table happened to be flat during my sample window while others were growing, which read as "fragments grow, rows don't". Measuring both properly across all tables under the same ambient load:

```
t=0    rows= 76043  fragments= 2877
t=180  rows= 76184  fragments= 2895
delta  rows=+141    fragments=+18      -> 7.8 rows per new fragment
```

That is ordinary batched writing, not amplification.

So this patch adds a `retrieve` per batch to prevent writes that were not happening. It is a cost with no demonstrated benefit, and I would rather withdraw it than have you spend review time on a fix for a non-problem.

**What does still hold, and is the useful half of #4684:** LanceDB stores are never compacted, so fragments accumulate monotonically with legitimate writes. Ours reached 6,592 version files and 6,557 fragments, 937 MB for a 2,461-node graph, with one table at 1,451 fragments. Offline `optimize()` took 12.5 seconds and cut it to 48 versions / 27 fragments / 411 MB with identical row counts. That part of the issue stands on its own and is worth fixing; the amplification framing in my follow-up comment there does not, and I have corrected it.

Apologies for the noise.

Issue #2538, opened 2026-04-03
Logging writes unbounded files to site-packages/logs/ without rotation


king-sheol:
## Description

Cognee writes log files directly into `site-packages/logs/` (inside the installed package directory) with no size limit, no rotation, and no cleanup mechanism.

## Impact

On our production system, a single log file grew to **22GB** over 3 days, and the total `logs/` directory reached **37GB** — filling the entire disk. This caused cascading failures in FalkorDB (AOF writes failed), SQLite (WAL writes failed), and the entire knowledge platform was down for 3+ hours.

## Environment

- Cognee v0.5.5
- `LOG_LEVEL=INFO` in config.env
- Multiple services calling cognee API (search, add, cognify) via HTTP daemon
- ~100+ API calls/hour from cron jobs and Telegram bot

## Log file details

```
$ ls -lhS .venvs/cognee/lib/python3.12/site-packages/logs/
-rw-rw-r-- 22G Apr  3 09:24 2026-03-31_13-13-39.log
-rw-rw-r-- 6.9G Mar 31 13:13 2026-03-30_12-07-11.log
-rw-rw-r-- 6.5G Mar 30 11:48 2026-03-29_16-13-41.log
-rw-rw-r-- 1.5G Mar 29 16:13 2026-03-29_12-45-30.log
```

## Expected behavior

1. Logs should NOT be written to `site-packages/` — this is a read-only package directory by convention
2. Log location should be configurable (e.g., `LOG_DIR` env var)
3. Log rotation should be built-in (e.g., `RotatingFileHandler` with max size)
4. At minimum, a size cap or daily rotation to prevent unbounded growth

## Workaround

We added a cron job to delete logs >100MB:
```bash
30 4 * * * find /path/to/.venvs/cognee/lib/python3.12/site-packages/logs -name "*.log" -size +100M -delete
```

king-sheol:
## Minimal reproduction

```python
# 1. Install cognee with default config
pip install cognee

# 2. Set LOG_LEVEL=INFO (default) and run any operation
import cognee
import asyncio

async def main():
    for i in range(100):
        await cognee.search("test query", top_k=3)

asyncio.run(main())

# 3. Check logs directory
import os
logs_dir = os.path.join(os.path.dirname(cognee.__file__), '..', 'logs')
print(f"Log dir: {os.path.abspath(logs_dir)}")
for f in os.listdir(logs_dir):
    size = os.path.getsize(os.path.join(logs_dir, f))
[...]

PR #4824, opened 2026-08-29
fix: PlainFileHandler never rotates, regression of #2538

### LINKED ISSUE #4823: Regression of #2538: PlainFileHandler still never rotates — emit() override never calls shouldRollover()/doRollover()
## Description

#2538 was closed by switching `PlainFileHandler`'s base class from `FileHandler` to
`RotatingFileHandler` and passing `maxBytes`/`backupCount` through. That fix doesn't
actually work: `PlainFileHandler.emit()` is a full override (needed for its structlog
dict-message formatting) that never calls `self.shouldRollover()` or `self.doRollover()`.
The string "rollover" does not appear anywhere in `logging_utils.py`. Rotation is
structurally unreachable regardless of `maxBytes`/`backupCount` being set correctly.

## Impact

On a long-running install with `LOG_FILE_NAME` pinned to one persistent path, the log
file grows without bound, same as the original report. Observed on my own install: 297MB
and climbing after ~2 days of normal use, zero `.1`/`.2` backup files ever created.

## Minimal reproduction

```python
import logging, os
from cognee.shared.logging_utils import PlainFileHandler

path = "/tmp/rotation_test.log"
handler = PlainFileHandler(path, maxBytes=2000, backupCount=2, encoding="utf-8")
logger = logging.getLogger("test")
logger.addHandler(handler)
logger.setLevel(logging.INFO)

for i in range(200):
    logger.info({"event": f"line {i}", "logger": "test"})

print(os.listdir("/tmp"))  # rotation_test.log.1 never appears
```

## Root cause

`PlainFileHandler.emit()` (`cognee/shared/logging_utils.py`) fully overrides
`RotatingFileHandler.emit()` to handle structlog dict-shaped records with custom
formatting. The base class's rotation trigger (`if self.shouldRollover(record):
self.doRollover()`) lives inside *its* `emit()`, which this override replaces entirely
without ever calling it. `maxBytes`/`backupCount` are accepted and stored (inherited
`__init__`), so the class looks capable of rotating, and the docstring says exactly that —
but nothing ever invokes the check.

## Suggested fix

Add a size check directly in `PlainFileHandler.emit()`, right after the existing
`if self.stream is None: self.stream = self._open()` guard:

```python
if self.maxBytes > 0:
    self.stream.seek(0, 2)
    if self.stream.tell() >= self.maxBytes:
        self.doRollover()
```

Deliberately not calling `self.shouldRollover(record)` directly — its stock
implementation estimates the pending message's length via `self.format(record)`, which
assumes a plain-string message. Since this handler's whole reason to exist is structlog
dict-shaped records (`record.msg` is often a `dict`), running the stock formatter just to
estimate a length is an avoidable way to raise on the very record the check is supposed
to protect.

Tested in isolation (`maxBytes=2000`, `backupCount=2`, 200 real structlog dict-message
calls): rotation fires correctly, stays bounded, zero exceptions, zero malformed lines.

## Environment

- Cognee v1.2.0

Happy to open a PR with this fix if useful.

BrightMine35:
## Summary
- `PlainFileHandler.emit()` fully overrides `RotatingFileHandler.emit()` (needed for its structlog dict-message formatting) and never calls `shouldRollover()`/`doRollover()`. `maxBytes`/`backupCount` are accepted and stored, and the class docstring claims automatic rotation, but nothing ever triggers it.
- This is a regression of #2538: that issue was closed by switching this class's base from `FileHandler` to `RotatingFileHandler`, but the pre-existing `emit()` override was never updated to actually call the rotation check, so the original unbounded-growth symptom recurs under a different shape (one huge file instead of many small ones).
- Confirmed locally: an install with `LOG_FILE_NAME` pinned to one path grew that file to 297MB with zero `.1`/`.2` backups ever appearing.
- Fix adds a direct stream-size check rather than calling `self.shouldRollover(record)` — its stock implementation estimates the pending message's length via `self.format(record)`, which assumes a plain-string message. Since this handler exists specifically for structlog dict-shaped records (`record.msg` is often a `dict`), running the stock formatter just to estimate a length risks raising on the very record the check is meant to protect.

Fixes #4823

## Test plan
- [x] `python -m py_compile cognee/shared/logging_utils.py`
- [x] Isolated test: 200 real structlog dict-message `logger.info({"event": ..., "logger": ...})` calls against a throwaway file with `maxBytes=2000, backupCount=2` — rotation fires correctly, `.1`/`.2` backups created, current file stays bounded, zero exceptions raised by the new check, zero malformed/truncated lines in any retained file.
- [ ] Not run against the full test suite in this environment — happy to adjust based on CI/maintainer feedback.

### DIFF
--- a/cognee/shared/logging_utils.py
+++ b/cognee/shared/logging_utils.py
@@ -160,6 +160,23 @@ class PlainFileHandler(logging.handlers.RotatingFileHandler):
             if self.stream is None:
                 self.stream = self._open()
 
+            # This method fully overrides RotatingFileHandler.emit() (needed
+            # for structlog-aware formatting below), which means the parent
+            # class's rotation trigger (shouldRollover()/doRollover(), called
+            # from ITS emit()) is never reached -- maxBytes/backupCount were
+            # accepted by __init__ but had no effect. Check size directly
+            # instead of calling self.shouldRollover(record): its stock
+            # implementation estimates the pending message's length via
+            # self.format(record), which assumes a plain-string message --
+            # this handler exists specifically for structlog dict-shaped
+            # records (record.msg is often a dict), so running the stock
+            # formatter just to estimate a length risks raising on the very
+            # record this check is meant to protect.
+            if self.maxBytes > 0:
+                self.stream.seek(0, 2)
+                if self.stream.tell() >= self.maxBytes:
+                    self.doRollover()
+
             # Extract the message from the structlog record
             if isinstance(record.msg, dict) and "event" in record.msg:
                 # Extract the basic message


Vasilije1990:
@BrightMine35 please resolve conflicts

Vasilije1990 (approved):

PR #4825, opened 2026-08-29
COG-6475 fix: classify budget exhaustion in embedding path

goran-radonic:
## Description

A LiteLLM spend cap cannot clear inside a retry window, but the embedding engines only
recognised terminal failures by exception class, so every doomed batch ran the full
`stop_after_delay(128)` ladder. `index_data_points` gates batches with a semaphore of
`max(1, embedding_max_concurrent_data_points // embedding_batch_size)`, which is 4 on
defaults, so each doomed batch also held a quarter of the embedding stage's concurrency for
that whole time. The caller then got "Verify EMBEDDING_ENDPOINT and provider settings",
which points at the wrong problem entirely.

COG-6329 (`e4ef19a2e`) fixed this on the LLM path. This is the embedding half:
`is_budget_exhausted_error` had zero call sites outside `cognee/infrastructure/llm/`.

The obvious patch does not work, and I checked before ruling it out. Adding
`litellm.exceptions.BudgetExceededError` to `retry_if_not_exception_type(...)` changes
nothing, for three independent reasons:

- `BudgetExceededError` subclasses plain `Exception`, so no base class in a tuple matches it;
- against a proxy, which is the only deployment where budgets exist, the client never
  receives that class. The proxy raises it server-side and the client maps the status onto an
  ordinary `RateLimitError`;
- the engines re-raise provider failures as `EmbeddingException(...) from error`, so by the
  time tenacity sees the failure the provider class is gone even when it would have matched.

So classification moves to a predicate over the `__cause__` chain, reusing the detector the
LLM path already relies on, and a budget rejection converts to the actionable 402 the API
already surfaces.

Three engines are covered. Ollama is included because `EMBEDDING_ENDPOINT` is never
validated for provider shape and a proxy's OpenAI-shaped body is accepted by its `"data"`
branch, so it works against a proxy until the cap. Fastembed is excluded because it embeds
in process and issues no requests at all.

Second commit, split so it can be reverted on its own: a rejected key on the
OpenAI-compatible engine ran the same ladder, and the obvious one-line fix would have been
dead code for the same wrapping reason as above. It now raises a cognee type that is both
terminal and inside the `CogneeApiError` family, so it fails on the first attempt without a
router's `except Exception` turning a configuration mistake into a 500.

## Acceptance Criteria

Measured end to end against a real budget-exhausted proxy, not inferred from source:
`ghcr.io/berriai/litellm:main-stable` with Postgres, a real virtual key with
`max_budget: 0.01` driven past its cap. One doomed batch:

| engine | | attempts | wall clock | HTTP requests | raised |
| --- | --- | --- | --- | --- | --- |
| LiteLLM | before | 7 | 138.4s | 21 | `EmbeddingException` 422 |
| LiteLLM | after | 1 | 1.6s | 3 | `LLMPaymentRequiredError` 402 |
| OpenAI-compatible | before | 7 | 135.9s | 21 | `EmbeddingException` 422 |
| OpenAI-compatible | after | 1 | 1.5s | 3 | `LLMPaymentRequiredError` 402 |

That run also settled two open questions from the ticket. The proxy answers 429, not the 400
the incident report described. And the 3 remaining requests per attempt are the provider
SDK's own `max_retries=2`, now tracked as COG-6478 rather than guessed at.

Which signal actually classifies differs by client, which is why this is a predicate and not
a type check: on the litellm path the response body is already consumed (`response.json()`
raises `JSONDecodeError`) so only the message regex works, while through the openai SDK the
body survives and `error.type == "budget_exceeded"` fires as well. Both are covered by tests.

Tests: 12 in `cognee/tests/unit/test_embedding_budget_fastfail.py` covering both budget
shapes on the litellm engine, the structured-body shape through the openai SDK, the 400
shape, the Ollama proxy shape, two transient negatives including prose that merely mentions a
budget, `CancelledError`, and the credentials contract (one attempt, 422, provider text
preserved). The message a real proxy sends is now pinned verbatim in the corpus in
`test_budget_exhaustion.py` as the one entry observed from a running proxy rather than read
from litellm's source.

Non-vacuity was checked by reverting the engine files and re-running: the budget test fails
after 2:11 with the wrong exception type.

Full `cognee/tests/unit`: 5179 passed, 69 skipped, 0 failed.

Not in this PR: budget rejections still count as provider-overload evidence, so the 900s
pacing window keeps rolling and throttles recovery after the budget clears (COG-6476, which
has a stronger claim on the incident's duration than anything fixed here); wrong exception
types on the remaining paths (COG-6477); the provider SDK's own retries (COG-6478). The
incident's symptom, event-loop saturation and probe timeouts, does not follow from this code
path and is not addressed here: retry backoff sleeps with `asyncio.sleep`, which yields the
loop. What this removes is the slot starvation, not the symptom.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/databases/exceptions/__init__.py
+++ b/cognee/infrastructure/databases/exceptions/__init__.py
@@ -10,6 +10,7 @@ from .exceptions import (
     UnsupportedProvenanceCapability,
     DatabaseNotCreatedError,
     EmbeddingContextWindowTooSmallError,
+    EmbeddingCredentialsError,
     EmbeddingException,
     MissingQueryParameterError,
     MutuallyExclusiveQueryParametersError,
--- a/cognee/infrastructure/databases/exceptions/exceptions.py
+++ b/cognee/infrastructure/databases/exceptions/exceptions.py
@@ -136,6 +136,28 @@ class EmbeddingContextWindowTooSmallError(EmbeddingException):
         super().__init__(message, name, status_code)
 
 
+class EmbeddingCredentialsError(EmbeddingException):
+    """
+    Raised when the embedding endpoint rejects the request's credentials.
+
+    Covers both directions: credentials the server does not accept (401) and
+    credentials it accepts but does not permit for this model or organization
+    (403). Neither can clear inside a retry window, so engines raise this
+    instead of the provider's own class: keeping the failure inside the
+    ``CogneeApiError`` family is what lets the API return an actionable 422
+    rather than a 500, while listing it as terminal is what stops the backoff
+    ladder. The provider's own message is carried through as *message*.
+    """
+
+    def __init__(
+        self,
+        message: str = "Embedding endpoint rejected the credentials.",
+        name: str = "EmbeddingCredentialsError",
+        status_code=status.HTTP_422_UNPROCESSABLE_CONTENT,
+    ):
+        super().__init__(message, name, status_code)
+
+
 class MissingQueryParameterError(CogneeValidationError):
     """
     Raised when neither 'query_text' nor 'query_vector' is provided,
--- a/cognee/infrastructure/databases/vector/embeddings/LiteLLMEmbeddingEngine.py
+++ b/cognee/infrastructure/databases/vector/embeddings/LiteLLMEmbeddingEngine.py
@@ -251,6 +251,14 @@ class LiteLLMEmbeddingEngine(EmbeddingEngine):
                 return handle_embedding_response(text, embedding_response, self.dimensions)
 
         except litellm.exceptions.BadRequestError as error:
+            # A spend cap can arrive as a 400 depending on how the proxy maps it, and this
+            # clause catches every BadRequestError before the conversion further down, so
+            # without this the same failure surfaces as a raw provider error on one
+            # transport status and as the 402 on another. Ahead of the length check because
+            # a budget rejection carries no length wording, so it would fall through to the
+            # bare re-raise below and never reach any conversion at all.
+            raise_if_budget_exhausted(error)
+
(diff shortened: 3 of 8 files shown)


linhongyu510 (commented):
The retry predicate correctly stops the ladder for all covered shapes, but the LiteLLM 400 budget path does not satisfy the PRs stated 402 error contract. `LiteLLMEmbeddingEngine.embed_text()` catches every `BadRequestError` before the generic branch that calls `raise_if_budget_exhausted()`. For a 400 carrying the recognised LiteLLM budget sentence, exact head `59f6f09c` therefore makes one attempt and then leaks `litellm.BadRequestError`; it never raises `LLMPaymentRequiredError`. The existing test explicitly expects that divergence, so the 12 green tests cannot detect the contract mismatch.

I added an assertion that this recognised budget shape must produce `LLMPaymentRequiredError` with `status_code == 402`; it fails on the PR head with the bare `BadRequestError`. Calling `raise_if_budget_exhausted(error)` at the start of the existing BadRequest handler, before the context-window check, makes that assertion pass. After updating the existing 400 test to the same 402 expectation, the full budget file is **13 passed** and Ruff passes. Ordinary non-budget bad requests still follow the existing branch, and context-window recovery is unchanged because the detector returns normally for those errors.

Please either convert this already-recognised 400 shape to 402 as the description promises, or explicitly narrow the public contract and rationale. I recommend the former so provider transport status does not change the API error family for the same budget exhaustion. Separately, the red rollup contains cancelled checks from an earlier run alongside a subsequent successful full matrix; those cancellations are not evidence of a code-test failure.

Vasilije1990 (approved):

PR #4078, opened 2026-07-15
fix(retrieval): enforce node-set scoping on hybrid neighborhoods and shared descriptions

lkzppm:
## Description

While testing dataset isolation with node sets on a shared Neo4j + PGVector deployment, I found two paths where a search scoped with `node_name=["A"]` puts content that belongs only to other node sets into the LLM context (reported in #4079):

1. **Hybrid entity lane neighborhood.** The entity vector search is filtered by node set correctly, but entities shared across datasets (deterministic UUIDs merge them, so `belongs_to_set` ends up like `["A", "B"]`) legitimately pass any single-set filter. `build_entities` then expands their one-hop neighborhood through `get_neighborhood`, which has no node set filter, and renders every neighboring edge as a context bullet. In my reproduction an A-scoped query about a shared vessel type returned "OP Atlantico has_vessel_type PSV" where "OP Atlantico" belongs only to dataset B, including edge_text sentences generated from B's chunks.

2. **Shared entity description, last writer wins.** `description` is a single global property on shared entity nodes and every dataset's re-extraction overwrites it, so a scoped search can render a description authored by another dataset's cognify run. GRAPH_COMPLETION's nodeset subgraph is otherwise tight (edges need both endpoints inside the set) but it projects `description`.

The fixes follow the same shape as the existing scoping in the retriever (for example the facts gating):

* `build_entities` now takes the node filter and drops neighborhood triples whose neighbor endpoint fails `payload_matches_node_filter`. Fail closed: a neighbor without a `belongs_to_set` list is dropped. The matched entity itself already passed the scoped vector search, and unscoped searches are unchanged.
* On scoped searches, both the hybrid entity lane and graph completion rendering suppress `description` unless the node provably belongs to exactly one node set. Single-owner entities keep their description, so unscoped answer quality is untouched and scoped answers only lose text that other datasets may have written. To make that decision at render time, scoped graph projections additionally project `belongs_to_set`.

One pre-existing test (`test_scoped_search_keeps_bullet_ranking_but_hides_facts`) used fixture neighbors without `belongs_to_set`; under the fail-closed rule those are now dropped, so the fixture nodes are tagged with the scoped set to keep testing what it was written for (bullet ranking).

Fixes #4079

## Acceptance Criteria

* Scoped hybrid searches no longer render edges to neighbors outside the requested node sets, and drop neighbors whose ownership cannot be proven. Covered by unit tests with mixed-set neighborhoods (OR and AND operators, self-loops, untagged neighbors).
[...]

PR #4827, opened 2026-08-29
fix(retrieval): preserve node-set tags on graph vectors

### LINKED ISSUE #4822: [Bug]: Triplet/EdgeType datapoints are created without belongs_to_set — node-set-filtered triplet search always returns nothing
### Bug Description

`Triplet` and `EdgeType` datapoints are created without `belongs_to_set`, so node-set-filtered searches over their collections (`Triplet_text`, `EdgeType_relationship_name`) can never match anything — silently, and at full-scan cost.

In `cognee/tasks/storage/add_data_points.py::_create_triplets_from_graph` the triplet datapoint is built as:

```python
triplets.append(
    Triplet(
        id=triplet_id,
        from_node_id=str(source_node_id),
        to_node_id=str(target_node_id),
        text=embeddable_text,
    )
)
```

Both endpoint nodes are available in `node_map` and carry `belongs_to_set`, but the tags are never propagated. Same pattern in `cognee/tasks/storage/index_graph_edges.py::create_edge_type_datapoints` — `EdgeType(relationship_name=text, number_of_edges=count)` with no tags.

The result on a real corpus (~800k triplets, cognee 1.4.2 + PGVector): **0 of 800,305 `Triplet_text` payloads carry any `belongs_to_set` value** (all empty arrays), while `Entity_name` payloads are correctly tagged (73,005/73,005). Every node-set-scoped triplet search (`node_name=[...]` → payload filter `(payload -> 'belongs_to_set') ?| tags`) therefore:

1. returns **nothing**, always, and
2. costs a near-full collection pass doing it (we measured 15–20 s per query on the unmatched filter before adding indexes; the correct result would have been found in ~100 ms).

Because the failure mode is an empty-but-valid result, it never throws — retrieval quietly degrades to no-graph-context for every filtered caller.

### Steps to Reproduce

1. `add` documents with `node_set=["A"]`, `cognify`.
2. Inspect the triplet collection: `SELECT payload->'belongs_to_set' FROM "Triplet_text" LIMIT 5;` → all `[]` (Entity payloads are tagged, so the pipeline's tagging itself works).
3. Run a search that reaches the triplet collection with `node_name=["A"]` → 0 results regardless of content.

### Expected behavior

Triplets (and EdgeTypes) inherit node-set membership from their endpoint nodes — e.g. the union of `source.belongs_to_set` and `target.belongs_to_set` — so scoped searches over pre-embedded triplets return the scoped subset instead of nothing.

We validated the union approach in production data by backfilling `Triplet_text.payload.belongs_to_set` via the deterministic triplet id (`uuid5(source_id + relationship_name + target_id)` from `generate_node_id`): scoped searches immediately returned correct results in ~75 ms.

### Suggested fix

In `_create_triplets_from_graph`, after resolving `source_node`/`target_node` from `node_map`:

```python
belongs_to_set = sorted(
    set(getattr(source_node, "belongs_to_set", None) or [])
    | set(getattr(target_node, "belongs_to_set", None) or [])
)
triplets.append(
    Triplet(
        id=triplet_id,
        from_node_id=str(source_node_id),
        to_node_id=str(target_node_id),
        text=embeddable_text,
        belongs_to_set=belongs_to_set or None,
    )
)
```

and the analogous propagation in `create_edge_type_datapoints` (union over the edges aggregated into each retrieval text).

### Environment

- cognee 1.4.2 (code unchanged in 1.5.3 — verified against the tag)
- PGVector adapter (pgvector 0.8.2, PostgreSQL 15/17), but the omission is backend-agnostic: the tags are dropped before any adapter sees the datapoint.

Related but distinct: #3914 / #4078 / #4079 cover node-set scoping at *query* time; this one is the *creation*-time gap that makes triplet-collection scoping structurally empty.

linhongyu510:
## Description

Fixes #4822.

Node-set tags were dropped before Triplet and EdgeType datapoints reached vector storage, so node-set-filtered graph retrieval silently returned no triplet context. This change propagates the deterministic union of endpoint node-set names through edge properties, Triplet datapoints, aggregated EdgeType datapoints, and both PostgreSQL and Neptune Analytics hybrid EdgeType payloads.

The change is limited to creation-time metadata propagation. It does not duplicate the query-time filtering work in #4078 or the lexical chunk filtering work in #4675.

## Tests

- `uv run pytest cognee/tests/unit/tasks/storage/test_add_data_points.py cognee/tests/unit/infrastructure/databases/test_index_graph_edges.py cognee/tests/unit/infrastructure/databases/test_postgres_hybrid_batching.py cognee/tests/unit/infrastructure/databases/test_neptune_analytics_hybrid.py -q` — 66 passed, 1 skipped
- `uv run ruff check <10 changed Python files>` — passed
- `uv run ruff format --check <10 changed Python files>` — passed
- `git diff --check origin/dev...HEAD` — passed

## AI assistance

TRAE was used to inspect the affected storage paths, check competing pull requests, and help prepare regression tests. I reviewed the complete diff and take responsibility for the contribution.

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/databases/hybrid/neptune_analytics/NeptuneAnalyticsAdapter.py
+++ b/cognee/infrastructure/databases/hybrid/neptune_analytics/NeptuneAnalyticsAdapter.py
@@ -12,7 +12,10 @@ from cognee.infrastructure.databases.exceptions import MutuallyExclusiveQueryPar
 from cognee.infrastructure.databases.graph.neptune_driver.adapter import NeptuneGraphDB
 from cognee.infrastructure.databases.vector.vector_db_interface import VectorDBInterface
 from cognee.infrastructure.engine import DataPoint
-from cognee.modules.graph.utils.prepare_edges_for_storage import get_edge_retrieval_text
+from cognee.modules.graph.utils.prepare_edges_for_storage import (
+    get_belongs_to_set_names,
+    get_edge_retrieval_text,
+)
 from cognee.modules.storage.utils import JSONEncoder
 from cognee.shared.logging_utils import get_logger
 from cognee.infrastructure.databases.vector.embeddings.EmbeddingEngine import EmbeddingEngine
@@ -586,11 +589,15 @@ class NeptuneAnalyticsAdapter(NeptuneGraphDB, VectorDBInterface):
 
         # Collect unique edge texts for embedding.
         edge_texts = []
+        belongs_to_set_by_text: dict[str, set[str]] = {}
         for edge in edges:
             props = edge[3] if len(edge) > 3 and edge[3] else {}
             edge_text = get_edge_retrieval_text(props.get("edge_text"), edge[2])
             if edge_text:
                 edge_texts.append(edge_text)
+                belongs_to_set_by_text.setdefault(edge_text, set()).update(
+                    get_belongs_to_set_names(props)
+                )
 
         edge_type_counts = Counter(edge_texts)
         if not edge_type_counts:
@@ -601,7 +608,7 @@ class NeptuneAnalyticsAdapter(NeptuneGraphDB, VectorDBInterface):
             IndexSchema(
                 id=str(EdgeType.id_for(text)),
                 text=text,
-                belongs_to_set=[],
+                belongs_to_set=sorted(belongs_to_set_by_text[text]),
             )
             for text in edge_type_counts
         ]
--- a/cognee/infrastructure/databases/hybrid/postgres/adapter.py
+++ b/cognee/infrastructure/databases/hybrid/postgres/adapter.py
@@ -31,7 +31,10 @@ if TYPE_CHECKING:
     from cognee.infrastructure.databases.vector.pgvector.PGVectorAdapter import PGVectorAdapter
 from cognee.modules.storage.utils import JSONEncoder
 from cognee.modules.graph.models.EdgeType import EdgeType
-from cognee.modules.graph.utils.prepare_edges_for_storage import get_edge_retrieval_text
+from cognee.modules.graph.utils.prepare_edges_for_storage import (
+    get_belongs_to_set_names,
+    get_edge_retrieval_text,
+)
 
 logger = get_logger()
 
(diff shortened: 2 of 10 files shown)


linhongyu510:
Rebased onto current `dev` (`45208c50e`) — the PR was showing `DIRTY`. New head `b9af546`, now `MERGEABLE`.

Two upstream changes intersected this branch, and one of them is a real scope reduction rather than a mechanical conflict:

**`c444e2d2f` removed the Postgres hybrid adapter entirely.** This PR had been propagating node-set tags through that adapter too. Rather than resurrect a deleted file through conflict resolution, I dropped both the adapter change and `test_postgres_hybrid_batching.py`. The Neptune adapter is untouched by that removal and keeps its tag propagation.

**Import blocks in `index_graph_edges.py` and the Neptune adapter were reordered upstream.** Merged `get_belongs_to_set_names` into the existing single import instead of leaving the duplicate `from ... import get_edge_retrieval_text` the automatic resolution produced.

The diff is now **8 files, `+130/-5`** (was 10 files, `+167/-7`).

Verification on the rebased head, re-run rather than carried over from the earlier pass:

```
pytest cognee/tests/unit/infrastructure/databases/test_index_graph_edges.py \
       cognee/tests/unit/infrastructure/databases/test_neptune_analytics_hybrid.py \
       cognee/tests/unit/tasks/storage/test_add_data_points.py
  52 passed, 1 skipped

ruff check cognee/        All checks passed
ruff format --check       19 files already formatted
```

No behavior change from the original review shape — same `get_belongs_to_set_names` helper, same propagation points, minus the adapter that no longer exists.

linhongyu510:
Same note as I left on #5024 about the red checks, since it's the same cause here:

The Integration failures are the fork-PR secrets gap, not this branch — the embedding engine is called with an empty model / `api_key=None`, so it errors out during setup before any test logic runs (`LLM Provider NOT provided. You passed model=` → `RuntimeError: Graph edge indexing error`). `dev` is green because it runs with those secrets present; the unit suites this PR actually touches pass locally (52 passed / 1 skipped, quoted above).

No action needed from me on the CI — it should clear on a maintainer re-run in a trusted context. Just flagging so it isn't mistaken for a code regression.

PR #4830, opened 2026-08-29
fix: get_or_create_dataset_database's existence check filters by owner_id, but dataset_id is the primary key

### LINKED ISSUE #4829: get_or_create_dataset_database's existence check filters by owner_id, but dataset_id is the table's primary key — false negatives crash /forget (and anything else routed through it) for non-owner users with full ACL grants
## Description

`_existing_dataset_database()` (`cognee/infrastructure/databases/utils/get_or_create_dataset_database.py`) checks for an existing row with:

```python
stmt = select(DatasetDatabase).where(
    DatasetDatabase.owner_id == user.id,
    DatasetDatabase.dataset_id == dataset_id,
)
```

But `DatasetDatabase.dataset_id` is that table's **primary key** (`cognee/modules/users/models/DatasetDatabase.py`: `dataset_id = Column(UUID, ..., primary_key=True)`) — one row per dataset, globally, by schema design. `owner_id` is a plain indexed column, not part of any composite key. Adding the `owner_id == user.id` condition can only ever produce a **false negative** (a `dataset_id` match is already unique on its own) — it can never change the result in a way that's correct, only in a way that's wrong.

## Impact

Any authenticated caller who has full ACL-granted permission on a dataset but is not its literal `owner_id` will get a crash instead of the correct existing row, the first time any code path calls `get_or_create_dataset_database()` for that dataset. Confirmed via `POST /api/v1/forget` with a single `dataId` + `datasetId` — this is not an authorization gap: the caller had all 4 of the same ACL permission rows on the dataset as the actual owner (full delegated access, correctly granted elsewhere in the system), yet this check still reported "no existing row," so the function fell through to `INSERT` and crashed on the dataset_id primary key.

## Minimal reproduction

1. Create a dataset as user A (e.g. via a dedicated agent identity with its own API key).
2. Grant user B a full ACL permission set on that dataset (own it not required — this is exactly what an ACL grant is for).
3. As user B, call anything that routes through `get_or_create_dataset_database()` for that dataset — e.g. `POST /api/v1/forget` with `dataId` + `datasetId` for an item already in that dataset.

## Actual result

```
sqlite3.IntegrityError: UNIQUE constraint failed: dataset_database.dataset_id
[SQL: INSERT INTO dataset_database (...) VALUES (...)]
```//api/v1/forget returns `{"error":"An error occurred during deletion."}`.

## Expected result

The existing `DatasetDatabase` row is found and returned, since `dataset_id` alone already uniquely identifies it and the caller has valid ACL permission on the dataset.

## Suggested fix

Drop the `owner_id` filter — check `dataset_id` alone, matching the schema's actual uniqueness semantics and respecting the ACL system the rest of cognee already uses for authorization instead of re-checking literal ownership on top of it:

```python
stmt = select(DatasetDatabase).where(
    DatasetDatabase.dataset_id == dataset_id,
)
```

Verified locally: the exact `/forget` call that failed before this change succeeds immediately after (`{"status":"success"}`, item confirmed removed), with no other behavior change — every other caller of `get_or_create_dataset_database()` that already had a matching owner_id continues to work identically, since `dataset_id` alone was always sufficient to identify the row.

## Environment

- Cognee `main` @ 690c0ec0 (also present in the released 1.2.0 line)

Happy to open a PR with this fix if useful (same as #4823/#4824).

BrightMine35:
## Summary
- `_existing_dataset_database()` filtered by `owner_id == user.id AND dataset_id == dataset_id`, but `DatasetDatabase.dataset_id` is that table's **primary key** — one row per dataset, globally, by schema design. The `owner_id` condition can only ever produce a false negative, never a correct change in result.
- Confirmed live via `POST /api/v1/forget`: a caller with full ACL-granted permission on a dataset (identical permission rows to the actual owner) but who wasn't the literal `owner_id` got `sqlite3.IntegrityError: UNIQUE constraint failed: dataset_database.dataset_id`, because the existence check reported "not found" and the function fell through to `INSERT` against a row that already existed since the dataset was first created.
- Fix: filter on `dataset_id` alone, matching the schema's real uniqueness semantics and respecting the ACL system the rest of cognee already uses for authorization.

Fixes #4829

## Test plan
- [x] `python -m py_compile` on the changed file
- [x] Live repro: the exact `/api/v1/forget` call that failed with the constraint error before this change succeeded immediately after (`{"status":"success"}`, item confirmed removed from the relational table), with no other observable behavior change
- [ ] Not run against the full test suite in this environment — happy to adjust based on CI/maintainer feedback

### DIFF
--- a/cognee/context_global_variables.py
+++ b/cognee/context_global_variables.py
@@ -195,8 +195,23 @@ class DatabaseContextManager:
         data_root_directory = os.path.join(
             base_config.data_root_directory, str(user.tenant_id or user.id)
         )
+        # dataset_database.owner_id, not the calling user.id: get_or_create_dataset_database()
+        # correctly resolves the dataset's real DatasetDatabase row for an ACL-granted caller
+        # who isn't the literal owner, but Ladybug/Kuzu's on-disk graph file was created under
+        # the OWNER's directory (.cognee_system/databases/{owner_id}/{graph_database_name}),
+        # not the caller's. Building this path from the caller's own id redirects a non-owner
+        # to their own, likely-nonexistent directory instead of the dataset's real graph file.
+        # Reported by @linhongyu510 in review of PR #4830: reproduced on head 9edd8b12 with a
+        # context test where caller_id != dataset_database.owner_id -- the resolved path
+        # contained caller_id and the owner-path assertion failed. Confirmed independently with
+        # a live repro against real data: /forget on a dataset owned by a different identity
+        # reported success (the relational Data/dataset_data rows were correctly deleted) while
+        # never touching the real graph at all, since it resolved to a path that didn't exist --
+        # 295 of 580 real graph nodes for that dataset were left orphaned as a direct result.
+        # LanceDB does not share this symptom -- its vector_database_url is already absolute,
+        # not reconstructed from a directory + user id.
         databases_directory_path = os.path.join(
-            base_config.system_root_directory, "databases", str(user.id)
+            base_config.system_root_directory, "databases", str(dataset_database.owner_id)
         )
 
         # Set vector and graph database configuration based on dataset database information
--- a/cognee/infrastructure/databases/utils/get_or_create_dataset_database.py
+++ b/cognee/infrastructure/databases/utils/get_or_create_dataset_database.py
@@ -55,9 +55,23 @@ async def _existing_dataset_database(
     """
     db_engine = get_relational_engine()
 
+    # dataset_id is this table's PRIMARY KEY (DatasetDatabase.py:
+    # `dataset_id = Column(UUID, ..., primary_key=True)`) -- one row per
+    # dataset, globally, by schema design. Filtering by owner_id here can
+    # never help (a dataset_id match is already unique on its own) and can
+    # only produce false negatives for a caller who has full ACL-granted
+    # permission on the dataset but isn't its literal owner_id -- confirmed
+    # live: a caller had all 4 of the same ACL permission rows as the actual
+    # owner on a dataset it didn't literally own, yet this filter still
+    # reported "no existing row", causing get_or_create_dataset_database()
+    # to attempt a fresh INSERT and crash with
+    # `UNIQUE constraint failed: dataset_database.dataset_id`
+    # (hit via /api/v1/forget on a single data item). Filtering on
+    # dataset_id alone matches the schema's actual uniqueness semantics and
+    # respects the ACL system the rest of cognee already uses for
+    # authorization, instead of re-checking literal ownership on top of it.
     async with db_engine.get_async_session() as session:
         stmt = select(DatasetDatabase).where(
-            DatasetDatabase.owner_id == user.id,
             DatasetDatabase.dataset_id == dataset_id,
         )
         existing: DatasetDatabase = await session.scalar(stmt)
--- a/cognee/infrastructure/databases/utils/get_or_create_dataset_database.py
+++ b/cognee/infrastructure/databases/utils/get_or_create_dataset_database.py
@@ -55,21 +55,22 @@ async def _existing_dataset_database(
     """
     db_engine = get_relational_engine()
 
-    # dataset_id is this table's PRIMARY KEY (DatasetDatabase.py:
-    # `dataset_id = Column(UUID, ..., primary_key=True)`) -- one row per
-    # dataset, globally, by schema design. Filtering by owner_id here can
-    # never help (a dataset_id match is already unique on its own) and can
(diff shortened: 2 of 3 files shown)


linhongyu510 (commented):
The primary-key lookup fixes the immediate duplicate INSERT, but the cross-owner path described by this PR still resolves the wrong physical graph database on the default local backend. After `_existing_dataset_database()` returns the actual owner row, `DatabaseContextManager.apply_database_context_variables()` builds `graph_file_path` from the current caller: `.../databases/{user.id}/{dataset_database.graph_database_name}`. Ladybug/Kuzu create and delete their files under `dataset_database.owner_id`, so an ACL-granted non-owner is redirected to its own directory instead of the owners existing graph.

I reproduced this on exact head `9edd8b12` with a context test where `caller_id != dataset_database.owner_id`: the resolved path contained `caller_id` and the assertion for the owner path failed. Changing only that directory component to `str(dataset_database.owner_id)` made the same test pass. LanceDB does not expose the same symptom because its stored `vector_database_url` is already absolute.

Please include this owner-path propagation (or persist/use an absolute graph path) and add a regression that enters the database context as an ACL-granted non-owner on Ladybug/Kuzu. Otherwise `/forget` may stop raising `UNIQUE constraint failed` yet still operate against a missing or newly created graph file instead of the shared dataset. I would also keep the ACL check explicit at the caller boundary, since UUID inputs pass through `get_unique_dataset_id()` unchanged.

BrightMine35:
@linhongyu510 Thank you for catching this — you were right, and it wasn't theoretical.

**Fix pushed** (`db754521`): `apply_database_context_variables()` now builds `databases_directory_path` from `dataset_database.owner_id` instead of the calling `user.id`, exactly as you diagnosed. Confirmed with your reproduction shape (context entered with `caller_id != owner_id`) and separately against this repo's real relational data — same result both ways.

**Real-world impact**: this bug had already caused actual data loss on a live install I run. A `/forget` call as a non-owner caller against an ACL-granted dataset reported success — the relational `Data`/`DatasetData` rows really were deleted — but the graph delete silently no-opped because it resolved against a path that didn't exist under the caller's own directory instead of the owner's. 295 of 580 real graph nodes for that dataset were left orphaned with no error surfaced anywhere. I'm disclosing this because it's the clearest evidence I have that this needed exactly the fix you asked for, not just the regression test. Remediated on my end (backup, isolate the orphaned rows, rebuild from the one real remaining source doc) — not a cognee-side action item, just noting the fix is validated against a real incident, not only synthetic repro.

**On the caller-boundary ACL point** (`f838893`): confirmed `get_unique_dataset_id()` does pass a UUID through unchanged with no check of its own, as you said. Traced the current caller: `forget.py` already calls `get_authorized_dataset(user, dataset_ref, "delete")` (or `get_authorized_dataset_by_name`) before it ever reaches `get_or_create_dataset_database()`, and that resolves through `get_specific_user_permission_datasets()`, which does a real ACL lookup keyed by permission type and raises `PermissionDeniedError` if the caller lacks it — not just an ownership check. So `/forget` itself isn't exposed. But you're right that the shared utility has no such guarantee built in for whichever caller comes next, so I added an explicit docstring warning on `get_or_create_dataset_database()` stating it performs no authorization check and that callers must do their own, pointing at `forget.py` as the pattern to follow. Didn't want to bolt an ACL check onto that function itself without knowing every current caller's expectations — seemed like the honest scope was "document the contract" rather than silently changing behavior underneath callers I haven't all audited.

**Regression test added** (`91f6e433`): `test_graph_file_path_uses_dataset_owner_not_caller` in `test_context_global_variables.py`, entering the context as an ACL-granted non-owner on the Ladybug/Kuzu path as you asked. Verified it fails on the pre-fix code (reverted just that one line locally, confirmed the exact failure) and passes with the fix. Also had to add `owner_id` to an existing test's fake `DatasetDatabase` that was missing it entirely — harmless before this PR since nothing read it, but `apply_database_context_variables()` now reads it unconditionally.

Let me know if you'd rather see the ACL-boundary note expressed differently (e.g. an assertion instead of a docstring), happy to adjust.

dexters1:
Function is intended to check if dataset exists for the owner, not a general dataset check for all users

BrightMine35:
Thanks for taking this the rest of the way — #4850's mechanism-level fix (deriving the owner independently rather than trusting caller-supplied storage ids, plus the optional explicit `permission_type` gate on `set_database_global_context_variables`) is a more thorough answer to both issues than what I had here, especially on @linhongyu510's caller-boundary point. Good catch on `_get_dataset_owner_id` having no authorization of its own in the bot review too.

Closing this out on my end — appreciate the review time from both of you.

PR #4833, opened 2026-08-30
Fix/lancedb periodic compaction

### LINKED ISSUE #4684: LanceDB stores are never compacted: fragments accumulate per upsert and inflate storage, query cost and memory

SahilArate:
## Type of Change

- [ x] Bug fix (non-breaking change that fixes an issue)
- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

- [ x] **I have tested my changes thoroughly before submitting this PR** (See `CONTRIBUTING.md`)
- [ x ] **This PR contains minimal changes necessary to address the issue/feature**
- [ x ] My code follows the project's coding standards and style guidelines
- [ x] I have added tests that prove my fix is effective or that my feature works
- [ x] All new and existing tests pass
- [ x ] I have searched existing PRs to ensure this change hasn't been submitted already
- [ x ] I have linked any relevant issues in the description
- [  x] My commits have clear and descriptive messages

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/databases/vector/config.py
+++ b/cognee/infrastructure/databases/vector/config.py
@@ -36,6 +36,14 @@ class VectorConfig(BaseSettings):
     vector_db_host: str = ""
     vector_db_subprocess_enabled: bool = True
     vector_pool_args: Union[str, None] = None
+    # Writes-per-table threshold at which accumulated LanceDB fragments/
+    # versions are folded back down via `optimize()`. Every `merge_insert`
+    # mints a new table version *and* a new data fragment, and LanceDB never
+    # compacts on its own -- left unbounded this inflates storage, query
+    # cost and process memory without limit on a long-lived deployment
+    # (see https://github.com/topoteretes/cognee/issues/4684). Ignored by
+    # backends other than LanceDB. `0` disables background compaction.
+    vector_db_compaction_write_interval: int = 25
 
     model_config = SettingsConfigDict(env_file=".env", extra="allow")
 
@@ -113,6 +121,7 @@ class VectorConfig(BaseSettings):
             "vector_db_password": self.vector_db_password,
             "vector_db_host": self.vector_db_host,
             "vector_db_subprocess_enabled": self.vector_db_subprocess_enabled,
+            "vector_db_compaction_write_interval": self.vector_db_compaction_write_interval,
         }
 
 
--- a/cognee/infrastructure/databases/vector/lancedb/LanceDBAdapter.py
+++ b/cognee/infrastructure/databases/vector/lancedb/LanceDBAdapter.py
@@ -3,6 +3,7 @@ import copy
 import inspect
 import threading
 import types
+from datetime import timedelta
 from collections import OrderedDict
 from os import path
 from uuid import UUID
@@ -13,6 +14,7 @@ from lancedb.pydantic import LanceModel, Vector
 from typing import List, Optional, Union, get_args, get_origin, get_type_hints
 
 from cognee.infrastructure.databases.exceptions import MissingQueryParameterError
+from cognee.infrastructure.databases.vector.config import get_vectordb_config
 from cognee.infrastructure.engine import DataPoint
 from cognee.infrastructure.engine.utils import parse_id
 from cognee.infrastructure.files.storage import get_file_storage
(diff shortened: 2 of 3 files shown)


Vasilije1990:
@SahilArate please read contributing MD, open PR to dev

SahilArate:
Thanks for informing @Vasilije1990  that fixed now. Retargeted to dev, and I found an old unrelated commit had gotten mixed in from my local branch by mistake; that's removed too, so this PR now only contains the LanceDB compaction fix. Ready for review whenever you get a chance. if there is any wrong please let me know!
Thank u!

dexters1:
Thanks for the PR @SahilArate! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

ArmagedonFlamer:
Thanks @SahilArate for taking on #4684 and for the careful write-up and tests. As written the patch has three issues we could reproduce: the new setting is exported into the vector-engine factory, which rejects it; the default subprocess table proxy's optimize() takes no arguments, so the cleanup call fails silently and compaction never runs there; and forcing retention to zero deletes versions a concurrent reader may still be on. Rather than round-trip all of that here, we carried your commit into #5150 with your authorship and sign-off intact and fixed those points on top, so we're closing this one in its favour. The interval-based approach is yours and is what #5150 builds on. 
We'd be glad to see your next contribution.

PR #4835, opened 2026-08-30
fix(dashboard): render operation-level activity rows without crashing

### LINKED ISSUE #4834: Local dashboard crashes when activity rows have null pipeline_name
### Description

The local Cognee dashboard crashes in AgentActivityTerminal when the activity endpoint returns an operation-level PipelineRun row.

The backend intentionally stores non-pipeline operations such as recall, remember, improve, search, and forget with pipeline_name set to null and operation_name populated. The frontend still types pipeline_name as a required string and calls toLowerCase() unconditionally.

This is reproducible on v1.5.3, and the affected code is also present on current main.

### Reproduction

1. Start the local backend and frontend.
2. Perform a recall or another operation recorded through record_operation.
3. Open the local dashboard.
4. The dashboard renders the error boundary.

### Actual result

Runtime TypeError:

    Cannot read properties of null (reading 'toLowerCase')

Location:

    cognee-frontend/src/ui/elements/AgentActivityTerminal.tsx:313

Affected expression:

    r.pipeline_name.toLowerCase()

The GET /api/v1/activity/pipeline-runs response includes rows where pipeline_name is null. The endpoint currently does not return operation_name, even though that is the identifying field for these rows.

### Expected result

The dashboard should render operation-level activity without crashing.

### Root cause

- cognee/modules/operations/record_operation.py deliberately creates PipelineRun rows with pipeline_name=None and operation_name populated.
- cognee/api/v1/activity/routers/get_activity_router.py includes those rows and serializes pipeline_name as null, but omits operation_name.
- AgentActivityTerminal assumes pipeline_name is always a string.

### Suggested fix

Align the activity response and frontend type:

1. Include operation_name in the pipeline-runs response.
2. Type pipeline_name as string | null.
3. Normalize the display/filter name from pipeline_name or operation_name before calling toLowerCase().
4. Add a focused regression test with an operation-level row where pipeline_name is null.

A null guard that only drops these rows would prevent the crash, but would also hide recall/search operation activity from the terminal.

xu-jia-ming:
## Summary

Operation-level `PipelineRun` rows (recall, remember, search, ...) are deliberately stored with `pipeline_name = null` and `operation_name` populated (`record_operation._write_operation_row`). The API already returns `operation_name` on `dev`, but `AgentActivityTerminal` still types `pipeline_name` as a required string and calls `.toLowerCase()` on it unconditionally, so the first operation-level row renders the error boundary (`Cannot read properties of null (reading 'toLowerCase')`). This PR fixes the frontend half.

## Changes

- `cognee-frontend/src/ui/elements/runPipelineName.ts` (new, pure module): `runPipelineName(r) = r.pipeline_name ?? r.operation_name ?? ""` plus `pipelineLabel` moved next to it.
- `AgentActivityTerminal.tsx`: `pipeline_name` typed `string | null`, `operation_name` added; the event filter and the label now derive the display name via `runPipelineName`. Operation rows are labeled like their pipeline counterparts (`cognee.recall`, `cognee.search`) and survive the search/recall filter instead of being dropped.
- `cognee-frontend/jest.config.json` (new, minimal): the repo ships `__tests__` suites and jest dev-dependencies but no jest config, so the new test was not runnable; this wires `ts-jest` + the `@/` alias. Nothing else runs jest today, so it changes no existing pipeline.
- `cognee-frontend/src/ui/elements/__tests__/runPipelineName.test.ts`: regression tests (null fallback, both-null, no-crash on `.toLowerCase()`, label parity).

## How was this tested?

- `npx jest --config jest.config.json` → 5/5 passed (scoped to the new suite).
- `npx tsc --noEmit` → 27 pre-existing project errors both with and without this change; zero errors reference the touched files.
- Backend change is a serialized-field addition (`operation_name` exists on the model, `PipelineRun.py:48`); `ast.parse` clean. No backend unit suite covers this router (none exists).

Fixes #4834

## Additional context

A null-guard that only dropped these rows would also hide recall/search operation activity from the terminal; falling back to `operation_name` keeps it visible, matching the issue's suggested direction. Commit is DCO-signed.

### DIFF
--- a/cognee-frontend/src/ui/elements/AgentActivityTerminal.tsx
+++ b/cognee-frontend/src/ui/elements/AgentActivityTerminal.tsx
@@ -15,7 +15,9 @@ import { SEARCH_SESSION_PREFIX } from "@/modules/sessions/getSessions";
 export interface PipelineRun {
   id: string;
   pipeline_name: string | null;
-  status: string;
+  // The activity API serializes status as null when run.status is unset
+  // (get_activity_router.py), so every .includes(...) read must coalesce.
+  status: string | null;
   dataset_id: string | null;
   dataset_name: string | null;
   owner_id: string | null;
@@ -422,7 +424,7 @@ export function AgentActivityTerminal({
       return ev.kind === "agentQuery";
     }
     if (termFilter === "errors") {
-      if (ev.kind === "run") return ev.r.status.includes("ERRORED");
+      if (ev.kind === "run") return (ev.r.status ?? "").includes("ERRORED");
       if (ev.kind === "session") return ev.s.effective_status === "failed" || ev.s.error_count > 0;
       return false;
     }
@@ -836,8 +838,8 @@ export function AgentActivityTerminal({
               // ── Operational pipeline run: a compact, non-expandable marker ──
               if (ev.kind === "run") {
                 const r = ev.r;
-                const isError = r.status.includes("ERRORED");
-                const isRunning = r.status.includes("STARTED") || r.status.includes("INITIATED");
+                const isError = (r.status ?? "").includes("ERRORED");
+                const isRunning = (r.status ?? "").includes("STARTED") || (r.status ?? "").includes("INITIATED");
                 const outcome: Outcome = isRunning ? "running" : isError ? "error" : "done";
                 // Defensive fallback only — the push above already guarantees
                 // pipeline_name is set for every "run" event reaching here.


Vasilije1990:
@xu-jia-ming can you manage the conflicts

xu-jia-ming:
Done — the branch was rebuilt as a single commit on top of the current `dev` head (f324b0d) and GitHub now reports it as mergeable with no conflicts. (The earlier state you saw predates a force-push that fixed the branch base.) Diff is the 4 frontend files: the nullable typing, the `runPipelineName` fallback helper with its tests, and the minimal jest config that makes them runnable.

acherifi:
Ran into exactly this on a self-hosted 1.5.3 deployment (first `recall` through the MCP took the dashboard down with `Cannot read properties of null (reading 'toLowerCase')`), so +1 on the fix.

One gap worth closing in the same PR: operation-level rows also have `status = null`. `_write_operation_row` sets `status=None`, and `get_activity_router.py` on `dev` serializes it as `"status": run.status.value if run.status else None` (line 179). With this PR those rows now pass the search/recall filter, but the render path still reads `status` unguarded on `dev`:

- `AgentActivityTerminal.tsx:425` `if (ev.kind === "run") return ev.r.status.includes("ERRORED");`
- `AgentActivityTerminal.tsx:839` `const isError = r.status.includes("ERRORED");`
- `AgentActivityTerminal.tsx:840` `const isRunning = r.status.includes("STARTED") || r.status.includes("INITIATED");`

So the first operation row that reaches the list will throw the same TypeError one field over (`reading 'includes'`). Suggest typing `status: string | null` alongside `pipeline_name` and coalescing those three reads (`(r.status ?? "")`), or mapping `outcome` onto a status-like string for operation rows since they carry `outcome` instead of `status`.

xu-jia-ming:
Thanks for the report and the +1, @acherifi — great catch on `status`.

I checked `get_activity_router.py` on `dev` and confirmed operation-level rows serialize `"status": null` in two places (the run serializer at line 81 and `root.get("status")` at line 119), so the same TypeError fires one field over exactly as you described.

Pushed in 3f510ed70, taking your first suggestion since the operation rows in the activity payload don't expose an `outcome` field the frontend could map:

- `PipelineRun.status` is now typed `string | null`
- added a `runStatus()` helper next to `runPipelineName()` that coalesces to `""`, and routed all three unguarded reads through it (`AgentActivityTerminal.tsx` error filter + the isError/isRunning outcome derivation)
- operation rows now render with the "done" outcome badge instead of crashing; added regression tests asserting `runStatus(row).includes("ERRORED")` doesn't throw and returns `false` for a null-status row

Jest: 8/8 pass in the touched suite (the 4 pre-existing failing suites fail identically at the merge base — missing resolver modules, unrelated to this change).

xu-jia-ming:
Conflict resolved — and while rebasing I noticed 064abd7fa landed an upstream fix for the `pipeline_name` half (the `kind` discriminator), so I dropped my superseded helper approach entirely.

The PR is now a single minimal commit on top of current `dev` carrying only the `status` half from @acherifi's report: `PipelineRun.status` is typed `string | null` and the three `.includes(...)` reads (errors filter + isError/isRunning derivation) coalesce with `?? ""`, matching the existing `pipeline_name ?? operation_name` defensive style. Note this still matters post-064abd7fa: the API serializes `status: null` whenever `run.status` is unset — including pipeline rows, which do reach the render path.

Verified on the rebuilt branch: jest 473/473 pass; `tsc --noEmit` reports the same 22 pre-existing errors as the merge base, none in the touched file.

xu-jia-ming:
CI triage after the rebase:

- **DCO check**: was failing because the PR body lacked the required statement — added it, check now passes.
- **Code Quality (Ruff) / Community Tests Status**: pre-existing on `dev` and unrelated to this PR — the failing files (`.github/scripts/motherduck_nightly_etl.py`, `cognee-mcp/src/__init__.py`, ...) are untouched by this change, and the same failures reproduce on other open PRs against current `dev` (e.g. the 06:13 community run on #4904's branch reports the same BLE001/I001 findings). Push-triggered `Test Suites` runs on `dev` don't execute the Ruff job, which is why `dev` itself looks green.
- The bulk of the `test_suites.yml` matrix fails within ~4s with zero executed steps — consistent with the fork-PR secret/runner restrictions rather than code.

This PR touches a single `.tsx` file; happy to re-run anything if useful.

PR #4836, opened 2026-08-30
fix(neo4j): count graph metrics in Cypher and stop retrying OOM

### LINKED ISSUE #4832: [Bug]: GET /datasets/graph-summary materializes entire Neo4j graph and triggers OOM errors
### Bug Description

`GET /api/v1/datasets/graph-summary` exhausts Neo4j transaction memory pool on large graphs by materializing the complete graph instead of performing lightweight count queries.

## Root Cause

The endpoint calls:
```
get_datasets_graph_summary()
  → graph_engine.get_graph_metrics(include_optional=False)
    → Neo4jAdapter.get_model_independent_graph_data()
```

Which executes memory-intensive Cypher queries:
```cypher
MATCH (n:`__Node__`) 
RETURN collect(n) AS nodes

MATCH (n:`__Node__`)-[r]->(m:`__Node__`) 
RETURN collect([n, r, m]) AS elements
```

The second query materializes the entire graph—all nodes and edges with all their properties—into a single Neo4j aggregation result. On large graphs, this allocates multiple GiB of transaction memory within seconds, eventually failing with:

```
Neo.TransientError.General.MemoryPoolOutOfMemoryError
dbms.memory.transaction.total.max threshold reached
```

## Retry Amplification Issue

Because `Neo4jAdapter.query()` is decorated with `@deadlock_retry()`, and the decorator retries **all** errors containing `Neo.TransientError`, this deterministic OOM error is retried **10 times consecutively**, each time materializing the entire graph again.

Result: a single graph-summary request can consume 30+ GiB of transaction memory across retries and eventually crash Neo4j.

## Caching Ineffective

Since the initial computation deterministically fails, the `GraphMetrics` cache entry is never written. Every subsequent request to `/datasets/graph-summary` triggers the same retry storm.

### Steps to Reproduce

1. Deploy Cognee with Neo4j 5.26.29 Community and a moderately-large graph (10k+ nodes and edges)
2. Call `GET /api/v1/datasets/graph-summary`
3. Observe Neo4j transaction memory growth in logs
4. Transaction fails with `MemoryPoolOutOfMemoryError` after 3-5 seconds
5. `@deadlock_retry()` decorator retries the same query 9 more times
6. Each retry attempt materializes the complete graph again
7. Subsequent requests trigger the same retry storm (no cache, first attempt failed)

### Expected Behavior

`/datasets/graph-summary` should obtain node and edge counts using bounded-memory count queries that complete in < 1 second and use O(1) memory regardless of graph size.

For Neo4j, this should be:
```cypher
MATCH (n:`__Node__`)
RETURN count(n) AS num_nodes

MATCH (n:`__Node__`)-[r]->(m:`__Node__`)
RETURN count(r) AS num_edges
```

This pattern already works correctly in Neptune's adapter, proving it is a viable solution.

Additionally, `MemoryPoolOutOfMemoryError` should not be retried—it is a structural problem with the query, not a transient race condition like actual deadlocks.

### Actual Behavior

**Observable transaction memory growth** (from Neo4j logs on moderately-large graph):
```
0.12 GiB after 0.06s
1.49 GiB after 0.72s
3.68 GiB after 2.07s
5.71 GiB after 3.37s
[Transaction fails with MemoryPoolOutOfMemoryError]
[deadlock_retry() restarts the same query]
```

**Request behavior:**
- First request: 10 retry attempts × full graph materialization = eventual Neo4j crash or timeout
- Subsequent requests: same retry storm (GraphMetrics cache never written because first attempt failed)
- Other concurrent queries: may timeout or fail due to memory contention

**Impact:** API endpoint becomes a denial-of-service vector for large graphs; even polling for status updates can crash Neo4j.

### Environment

- OS: Linux (Docker Compose deployment)
- Python version: 3.x (not specified in original report)
- Cognee version: 1.5.0-local (also reproduced by source inspection in released 1.5.3)
- Database: Neo4j 5.26.29 Community
- Storage: PostgreSQL + PGVector
- Graph size: Moderately-large (10k+ nodes and edges)

### Logs/Error Messages

```shell
Neo.TransientError.General.MemoryPoolOutOfMemoryError
dbms.memory.transaction.total.max threshold reached


Transaction is retried 10 times via `@deadlock_retry()` decorator, materializing the entire graph on each attempt:

**Relevant code locations:**
- Problematic method: `cognee/infrastructure/databases/graph/neo4j_driver/adapter.py#L1772-L1791` (`get_model_independent_graph_data`)
- Graph metrics endpoint: `cognee/api/v1/datasets/routers/get_datasets_router.py#L509-L646` (`get_datasets_graph_summary`)
- Deadlock retry decorator: `cognee/infrastructure/databases/graph/neo4j_driver/adapter.py` (decorates `query()` method)
```

## Comparative: Neptune Implementation (Correct)

Neptune's adapter already implements this correctly using COUNT queries:
```python
query_string = f"""
    MATCH (n :{self._GRAPH_NODE_LABEL})
    WITH count(n) AS nodeCount
    MATCH (a :{self._GRAPH_NODE_LABEL})-[r]->(b :{self._GRAPH_NODE_LABEL})
    RETURN nodeCount AS numVertices, count(r) AS numEdges
"""
```
See: `cognee/infrastructure/databases/graph/neptune_driver/adapter.py#L1475-L1495`

This proves the pattern works and should be adopted for Neo4j.

## Related Issues

Related to #4233 (full-graph projection through remember → improve → memify), but this report is distinct:
- Triggered by: `GET /api/v1/datasets/graph-summary` → `get_graph_metrics()` → `get_model_independent_graph_data()`
- Distinct problem: OOM retry amplification on a lightweight endpoint, not a heavy computation pipeline

## Proposed Fixes

**Priority 1 (Required):**
1. Add lightweight `get_graph_counts()` method to `GraphDBInterface` that returns only `(num_nodes, num_edges)` using COUNT queries
2. Implement `get_graph_counts()` in `Neo4jAdapter` with native COUNT Cypher (mirroring Neptune's pattern)
3. Update `/datasets/graph-summary` endpoint to call `get_graph_counts()` instead of `get_graph_metrics()`
4. Exclude `MemoryPoolOutOfMemoryError` from `deadlock_retry()` decorator—immediately re-raise (not a retryable error)

**Priority 2 (Recommended):**
5. Update `get_model_independent_graph_data()` to avoid `collect()` when only counts are needed
6. Add regression test: mock Neo4j adapter to verify `/graph-summary` endpoint never calls full-graph projection methods

**Priority 3 (Follow-up):**
7. Audit all callers of `get_model_independent_graph_data()` and `get_graph_metrics()` for similar OOM vulnerabilities

### Pre-submission Checklist

xu-jia-ming:
## Summary

`GET /v1/datasets/graph-summary` exhausted the Neo4j transaction memory pool on large graphs, and then amplified the damage — two causes, one problem (#4832):

1. **Materializing to count.** `Neo4jAdapter.get_graph_metrics()` called `get_model_independent_graph_data()` — `collect(n)` and `collect([n, r, m])` — purely to compute `num_nodes = len(...)` / `num_edges = len(...)`. Those aggregates load every node and edge with all properties into a single transaction result (multiple GiB on 10k+ node graphs). Node/edge counts are now plain `count()` aggregations, mirroring what the ladybug adapter's `get_graph_metrics` already does. `get_model_independent_graph_data()` itself is untouched (it has no other callers besides the old metrics path, but it stays a public adapter method).

2. **Retry amplification.** `MemoryPoolOutOfMemoryError` arrives as `Neo.TransientError.General.MemoryPoolOutOfMemoryError`, so `deadlock_retry` matched the `"Neo.TransientError"` retry rule and re-ran the identical query up to 10 more times — each retry re-allocating the same transaction memory (the issue measured 30+ GiB across one request). The decorator now re-raises it immediately: it is deterministic, unlike deadlocks and genuine transient errors.

## Changes

- `cognee/infrastructure/databases/graph/neo4j_driver/adapter.py` — `get_graph_metrics` counts via Cypher `count(n)` / `count(r)` scoped to the same `BASE_LABEL` as before.
- `cognee/infrastructure/databases/graph/neo4j_driver/deadlock_retry.py` — `MemoryPoolOutOfMemoryError` excluded from retries.
- `cognee/tests/unit/infrastructure/databases/graph/test_neo4j_graph_metrics_counts.py` (new) — asserts metrics are computed without `get_model_independent_graph_data`, counts come from aggregation queries, `mean_degree` math, and the empty-graph path.
- `cognee/tests/unit/infrastructure/databases/graph/neo4j_deadlock_test.py` — new test asserting the OOM error is raised after exactly one call (no retry); existing deadlock/DatabaseUnavailable behavior untouched.

## How was this tested?

```
$ pytest cognee/tests/unit/infrastructure/databases/graph/test_neo4j_graph_metrics_counts.py \
         cognee/tests/unit/infrastructure/databases/graph/neo4j_deadlock_test.py \
         cognee/tests/unit/infrastructure/databases/graph/test_neo4j_has_node.py -q
11 passed
```

Without the source fixes, 3 of these fail (both metrics tests and the OOM no-retry test). A live Neo4j repro of the GiB-scale allocation would need the reporter's deployment; the unit tests pin the code-level behavior (no graph materialization call, count aggregations issued, no retry on OOM). Commit is DCO-signed.

Fixes #4832

### DIFF
--- a/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
+++ b/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
@@ -2285,13 +2285,21 @@ class Neo4jAdapter(GraphDBInterface):
             input flag.
         """
 
-        nodes, edges = await self.get_model_independent_graph_data()
+        # Count in Cypher instead of materializing the graph via
+        # get_model_independent_graph_data()'s collect(n) / collect([n, r, m]):
+        # those aggregates load every node and edge (with all properties) into a
+        # single transaction result, which exhausts the Neo4j transaction memory
+        # pool on large graphs just to take len() of it.
+        node_count = await self.query(f"MATCH (n:`{BASE_LABEL}`) RETURN count(n) AS count")
+        edge_count = await self.query(
+            f"MATCH (n:`{BASE_LABEL}`)-[r]->(m:`{BASE_LABEL}`) RETURN count(r) AS count"
+        )
+        num_nodes = node_count[0]["count"] if node_count else 0
+        num_edges = edge_count[0]["count"] if edge_count else 0
+
         graph_name = "myGraph"
         await self.drop_graph(graph_name)
         await self.project_entire_graph(graph_name)
-
-        num_nodes = len(nodes[0]["nodes"])
-        num_edges = len(edges[0]["elements"])
 
         mandatory_metrics = {
             "num_nodes": num_nodes,
--- a/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
+++ b/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
@@ -2298,18 +2298,28 @@ class Neo4jAdapter(GraphDBInterface):
         num_edges = edge_count[0]["count"] if edge_count else 0
 
         graph_name = "myGraph"
-        await self.drop_graph(graph_name)
-        await self.project_entire_graph(graph_name)
+        # The GDS projection is needed only by the connected-component and optional
+        # metrics. Projecting an LLM-extracted graph can fail outright — thousands
+        # of distinct relationship types, some with non-identifier characters,
+        # break the interpolated projection query — so the count-only path must
+        # not pay for (or fail on) it (#4832); component fields are None there.
+        if include_optional:
+            await self.drop_graph(graph_name)
+            await self.project_entire_graph(graph_name)
 
         mandatory_metrics = {
             "num_nodes": num_nodes,
             "num_edges": num_edges,
             "mean_degree": (2 * num_edges) / num_nodes if num_nodes != 0 else None,
             "edge_density": await get_edge_density(self),
-            "num_connected_components": await get_num_connected_components(self, graph_name),
+            "num_connected_components": await get_num_connected_components(self, graph_name)
+            if include_optional
+            else None,
             "sizes_of_connected_components": await get_size_of_connected_components(
                 self, graph_name
-            ),
+            )
+            if include_optional
+            else None,
         }
 
         if include_optional:
--- a/cognee/infrastructure/databases/graph/neo4j_driver/deadlock_retry.py
+++ b/cognee/infrastructure/databases/graph/neo4j_driver/deadlock_retry.py
@@ -52,6 +52,11 @@ def deadlock_retry(max_retries=10):
                         raise  # Re-raise the original error
 
                     error_str = str(error)
+                    if "MemoryPoolOutOfMemoryError" in error_str:
+                        # Deterministic: retrying the identical query re-allocates
+                        # the same transaction memory and fails the same way, so
+                        # each retry only amplifies the memory pressure.
+                        raise
                     if "DeadlockDetected" in error_str or "Neo.TransientError" in error_str:
                         await wait()
                     else:
--- a/cognee/tests/unit/infrastructure/databases/graph/neo4j_deadlock_test.py
+++ b/cognee/tests/unit/infrastructure/databases/graph/neo4j_deadlock_test.py
@@ -95,3 +95,23 @@ if __name__ == "__main__":
         await test_database_unavailable_retries_match_neo4j_error()
 
     asyncio.run(main())
+
+
+@pytest.mark.asyncio
+async def test_memory_pool_out_of_memory_is_not_retried():
+    """MemoryPoolOutOfMemoryError arrives as Neo.TransientError.General.*, but it
+    is deterministic: retrying the identical query re-allocates the same
+    transaction memory, so each retry only amplifies the memory pressure."""
+    mock_function = AsyncMock(
+        side_effect=Neo4jError(
+            "Neo.TransientError.General.MemoryPoolOutOfMemoryError: "
+            "dbms.memory.transaction.total.max threshold reached"
+        )
+    )
+
+    wrapped_function = deadlock_retry(max_retries=5)(mock_function)
+
+    with pytest.raises(Neo4jError, match="MemoryPoolOutOfMemoryError"):
+        await wrapped_function(self=None)
+
+    assert mock_function.await_count == 1, "the OOM error must not be retried"
--- a/cognee/tests/unit/infrastructure/databases/graph/test_neo4j_graph_metrics_counts.py
+++ b/cognee/tests/unit/infrastructure/databases/graph/test_neo4j_graph_metrics_counts.py
@@ -1,0 +1,64 @@
+from unittest.mock import AsyncMock
+
+import pytest
+
+from cognee.infrastructure.databases.graph.neo4j_driver.adapter import Neo4jAdapter
+
+
+def _metrics_adapter(num_nodes: int, num_edges: int) -> Neo4jAdapter:
+    """Adapter wired so every metric helper finds the key it expects."""
+    adapter = object.__new__(Neo4jAdapter)
+    adapter.get_model_independent_graph_data = AsyncMock()
+    adapter.drop_graph = AsyncMock()
+    adapter.project_entire_graph = AsyncMock()
+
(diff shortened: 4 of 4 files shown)


linhongyu510 (changes requested):
Blocking scope issue at `3cb0c0a9`: this PR currently contains the full, unrelated dashboard fix from #4835 as its first commit (`9f73b8fb`). The PR description lists four Neo4j files, but the actual diff against `main` is 9 files and also changes the frontend activity terminal, Jest config, and activity API. Please rebase/cherry-pick the Neo4j commit directly onto current `main` (or retarget as an explicit stack) so #4836 can be reviewed and merged independently of #4835.

The intended Neo4j change itself looks sound: the count queries preserve the old `BASE_LABEL` scope, avoid materializing node/edge properties, and the OOM branch exits before retry. Fresh local validation on Python 3.12: the three targeted modules passed (`11 passed`); Ruff check and format-check passed on the four Neo4j-related files; `git diff --check origin/main...HEAD` is clean. No live large-graph Neo4j reproduction was performed. Once the unrelated commit is removed, I see no blocker in the stated fix.

xu-jia-ming:
The unrelated commit is already gone — your review landed against the pre-rebuild head (`3cb0c0a`). The branch was rebuilt as a single cherry-picked commit on top of the current `dev` head (`f324b0d`) and the PR retargeted to `dev`: it now shows 1 commit and exactly the four Neo4j files (adapter, deadlock_retry, and the two test files), independently mergeable from #4835.

I re-ran the validation on the rebuilt branch: `test_neo4j_graph_metrics_counts.py` + `neo4j_deadlock_test.py` → 9 passed; ruff check/format clean on the four files.

nagelm:
Ran this fix's approach against a real-world graph and found one remaining failure mode on the summary path, worth noting for #4832:

**The GDS projection still runs for `include_optional=False`.** `get_graph_metrics` keeps `drop_graph` + `project_entire_graph` unconditionally (needed for `num_connected_components` / `sizes_of_connected_components`). On LLM-extracted graphs the dynamic relationship-type space explodes — ours is ~24k nodes / ~96k edges with **5,548 distinct relationship types**, 7 of which have non-identifier characters (e.g. `adds_~400mb_per_job`, `measured_docker.img_growth`) that break the interpolated projection query. So even with the counts fixed, `GET /datasets/graph-summary` (which calls `get_graph_metrics(include_optional=False)` and stores only `num_nodes`/`num_edges`) still can't compute or cache its result on such graphs — it just fails faster now.

Since the summary path needs only the two counts (+ `mean_degree`/`edge_density`, which are also plain Cypher), gating the projection and the component metrics on `include_optional` would let graph-summary work without projecting at all. We're running exactly that as a local addition on top of this PR's diff and it restores the endpoint (metrics columns are nullable, so `None` for the component fields round-trips fine). Happy to file it as a follow-up PR if that split is preferred.

xu-jia-ming:
@nagelm Great catch — incorporated in the new head rather than a follow-up, since it's the same endpoint and issue (#4832).

The projection and the connected-component metrics are now gated on `include_optional`: the count-only path that `GET /datasets/graph-summary` uses runs no `drop_graph`/`project_entire_graph` at all, and returns `None` for the two component fields (the endpoint stores only the counts; `mean_degree`/`edge_density` are plain Cypher and stay populated). The full path (`include_optional=True`) still projects exactly as before.

Regression tests added:
- `test_graph_metrics_count_only_path_skips_the_gds_projection` — asserts neither projection call is awaited and the component fields are `None` for `include_optional=False`
- `test_graph_metrics_full_path_still_projects` — asserts both projection calls still happen for `include_optional=True`

Full suite on the branch: `test_neo4j_graph_metrics_counts.py` + `neo4j_deadlock_test.py` → 11 passed; ruff check/format clean. Your local patch should now be redundant — if your graph can still reproduce a failure on this head, happy to dig further.

PR #4838, opened 2026-08-30
fix(neo4j): make has_edge return a bool as the interface declares

### LINKED ISSUE #4837: Neo4jAdapter.has_edge returns the raw result list, so it is always truthy — memify cross-connect writes zero edges
## Bug Description

`Neo4jAdapter.has_edge()` (`cognee/infrastructure/databases/graph/neo4j_driver/adapter.py:1047-1048`) returns `self.query(...)`'s result directly:

```python
edge_exists = await self.query(query, params)
return edge_exists
```

The query is `RETURN COUNT(relationship) > 0 AS edge_exists` — an aggregation that **always returns exactly one row**, `[{"edge_exists": False}]` when the edge does not exist. So the method returns a non-empty list in both cases: truthy in both cases, and never a `bool` as the `GraphDBInterface.has_edge -> bool` contract declares (every other adapter — turso, postgres_demo, hybrid postgres — returns a real bool).

## Impact

`cognee/tasks/memify/cross_connect_entities.py:183` (`_drop_existing_edges`):

```python
if not await graph_engine.has_edge(source_id, target_id, relationship_name):
    guarded.append(edge)
```

`not <non-empty list>` is always `False`, so `guarded` is always empty and the pipeline calls `graph_engine.add_edges([])` — **the memify cross-connect stage writes zero inferred edges on Neo4j**, silently (no error, no log). Any caller using `await has_edge(...) is False` is also broken.

## Steps to Reproduce

No server needed — mock `query` as the driver would answer:

```python
from unittest.mock import AsyncMock
adapter = object.__new__(Neo4jAdapter)
adapter.query = AsyncMock(return_value=[{"edge_exists": False}])
res = await adapter.has_edge("a", "b", "rel")
# res == [{'edge_exists': False}]  — truthy, not False
```

## Expected behavior

`has_edge` returns a `bool`: `bool(results[0]["edge_exists"]) if results else False`.

Happy to send a PR with regression tests.

xu-jia-ming:
## Summary

`Neo4jAdapter.has_edge()` returned `self.query(...)`'s raw result list. The edge-existence query is an aggregation that always returns exactly one row — `[{"edge_exists": False}]` when the edge is absent — so the method was **truthy in both cases** and never returned the `bool` that `GraphDBInterface.has_edge -> bool` declares (the turso, postgres_demo, and hybrid postgres adapters all return real bools).

Downstream, `cross_connect_entities._drop_existing_edges` guards with `if not await graph_engine.has_edge(...)`: `not <non-empty list>` is always `False`, so every candidate edge was dropped and **the memify cross-connect stage wrote zero inferred edges on Neo4j, silently**.

Fixes #4837

## Changes

- `cognee/infrastructure/databases/graph/neo4j_driver/adapter.py` — `has_edge` now returns `bool(results[0]["edge_exists"]) if results else False`.
- `cognee/tests/unit/infrastructure/databases/graph/test_neo4j_has_edge.py` (new) — bool-True when present, bool-False + `not result` when absent (the caller-facing check), and False on an empty result. All three fail on `main`.

## How was this tested?

```
$ pytest cognee/tests/unit/infrastructure/databases/graph/test_neo4j_has_edge.py -q
3 passed
```

Without the fix, all 3 tests fail (`[{'edge_exists': False}]` is not `False`). The mock mirrors what the driver returns for the aggregation (`COUNT(...) > 0 AS edge_exists` always yields one row). Commit is DCO-signed.

### DIFF
--- a/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
+++ b/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
@@ -1044,8 +1044,12 @@ class Neo4jAdapter(GraphDBInterface):
             "to_node_id": str(to_node),
         }
 
-        edge_exists = await self.query(query, params)
-        return edge_exists
+        results = await self.query(query, params)
+        # The query is an aggregation and always returns exactly one row
+        # ({"edge_exists": False} when absent); returning the raw result list made
+        # every call truthy, so callers like cross_connect_entities'
+        # `if not has_edge(...)` never saw a missing edge.
+        return bool(results[0]["edge_exists"]) if results else False
 
     async def has_edges(self, edges):
         """
--- a/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
+++ b/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
@@ -1034,7 +1034,7 @@ class Neo4jAdapter(GraphDBInterface):
             - bool: True if the edge exists, otherwise False.
         """
         query = f"""
-            MATCH (from_node: `{BASE_LABEL}`)-[:`{edge_label}`]->(to_node: `{BASE_LABEL}`)
+            MATCH (from_node: `{BASE_LABEL}`)-[relationship: `{edge_label}`]->(to_node: `{BASE_LABEL}`)
             WHERE from_node.id = $from_node_id AND to_node.id = $to_node_id
             RETURN COUNT(relationship) > 0 AS edge_exists
         """
--- a/cognee/tests/unit/infrastructure/databases/graph/test_neo4j_has_edge.py
+++ b/cognee/tests/unit/infrastructure/databases/graph/test_neo4j_has_edge.py
@@ -1,0 +1,38 @@
+from unittest.mock import AsyncMock
+
+import pytest
+
+from cognee.infrastructure.databases.graph.neo4j_driver.adapter import Neo4jAdapter
+
+
+@pytest.mark.asyncio
+async def test_has_edge_returns_bool_true_when_present():
+    adapter = object.__new__(Neo4jAdapter)
+    adapter.query = AsyncMock(return_value=[{"edge_exists": True}])
+
+    assert await adapter.has_edge("node-a", "node-b", "relates_to") is True
+
+
+@pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


linhongyu510 (changes requested):
Blocking issue: the production Cypher still references an unbound variable. The pattern is `-[:`{edge_label}`]->`, but the return expression is `COUNT(relationship) > 0`; no `relationship` variable is introduced. I reproduced the exact generated query against Neo4j 5 and it raises `Neo.ClientError.Statement.SyntaxError: Variable relationship not defined`. Naming the relationship in the pattern (for example `-[relationship:`{edge_label}`]->`) makes the absent-edge query return `[{"edge_exists": false}]` as intended. The added AsyncMock tests all pass, but they bypass Cypher parsing and therefore cannot catch this blocker. Please bind the relationship (or use a valid count expression) and add a Neo4j-backed regression that executes `has_edge` for both present and absent edges. Local audit at `40f841f3`: focused unit tests 10 passed; Ruff check and format check passed; `git diff --check` passed.

xu-jia-ming:
Good catch — fixed in the new head.

- The pattern now binds the variable: `-[relationship: \`{edge_label}\`]->`, so `COUNT(relationship)` is valid Cypher. I verified against a live Neo4j 5 container (anonymous auth, `neo4j:5` image): `has_edge` on a seeded `__Node__` pair returns `True` for the present edge, `False` for an absent target, and `False` for a present pair under a different label — all as plain bools.

```
has_edge(present)     -> True
has_edge(absent)      -> False
has_edge(wrong label) -> False
```

- On the regression-test ask: the repo's unit suite has no live-Neo4j fixture (the community-dataset tests mock the container manager), so I added `test_has_edge_query_binds_the_relationship_variable`, which pins the exact defect — it fails if the pattern stops binding the relationship variable, which is precisely what made the query invalid Cypher. It cannot catch arbitrary future syntax errors the way a server-backed test would; if you'd like, I can add a testcontainers-based variant gated behind an env var in a follow-up.

xu-jia-ming:
Thanks @linhongyu510 — your review caught a real Cypher error my mock-based tests couldn't see, which made this fix actually correct rather than just cosmetically so. Appreciate the thorough triage.

PR #4851, opened 2026-08-31
fix(eval): rank BEAM tau-b system-only events as reference ties

Madhatter2099:
## Description

While reading through the BEAM evaluation metrics, at first I just wanted to understand how questions like `event_ordering` are scored. The docstring at the top of the file states that it mirrors `compute_metrics.py` from the upstream BEAM repository, so I pulled down the upstream code to compare.

At first I suspected `_kendall_tau_b` itself — it is written in pure Python without SciPy, so I thought the tied-pair counting might be off. I ran 5,456 random cases (4,875 of them containing ties) against `scipy.stats.kendalltau(variant="b")` and diffed the results: zero deviation. That function is correct; the problem is not there.

The issue became clear when I looked at the calling code. Upstream ranks both sequences through the same `to_rank()`, so any member not present in a given sequence falls back to `tie_rank`. The system side in cognee follows this:

```python
sys_ranks = [sys_rank_map.get(u, tie_rank) for u in union]
```

But the reference side was written as:

```python
ref_ranks = list(range(len(union)))
```

`union` contains system-only events — steps the answer invented that do not exist in the reference at all. `list(range(...))` gives them ascending, genuine reference ranks, which is equivalent to asserting that the reference already contained these events, in exactly that order. As a result the reference side never produces ties.

Using upstream's `to_rank()` plus SciPy as ground truth, I ran 60,000 random alignments: 29,071 of them (48.5%) disagreed with the upstream score, and every single one was inflated — zero were deflated. In the worst case a score of 0.0564 was reported as 0.2000, an inflation of 255%. Being strictly one-directional is what tells me this is a systematic bias rather than floating-point noise, and the direction makes cognee look better on BEAM than it actually is.

The trigger is not exotic: it fires whenever the answer contains two or more events absent from the reference. With only one such event, `tie_rank` has no second member to tie with, so the score is unchanged and the bug stays invisible — I suspect that is why it went unnoticed, and it is why I added a test pinning that boundary specifically.

I also found that the same function used two different criteria for "is this a match": the `union` construction used `alignment.get(i, -1) < 0`, while `sys_rank_map` used `0 <= ref_idx < n_reference`. When the LLM returns an out-of-range index (say `{"1": 7}` when the reference holds only two events), the F1 side counts it as unmatched while the ranking side silently drops it, so that system event never takes part in the comparison at all. I unified the predicate into a single `_matched_reference_index()`. This shares its root cause with the main bug — system-only events not being recognised correctly in the rank construction. Note that this one can push a score either way: with malformed indices included, 4,003 of the 60,000 cases came out too low rather than too high, the worst reporting 0.5333 as 0.0000.

I also extracted the rank construction from `a_measure()` into a pure function, `_build_rank_vectors()`. The reason is practical rather than cosmetic: it sat between two LLM calls, so it was untestable without mocking, and the module had no test coverage at all. With it extracted I added 14 unit tests. After the patch, both scenarios above agree with the upstream implementation exactly — 0 mismatches.

One caveat worth flagging: this changes `event_ordering` scores, so historical numbers are no longer directly comparable to new ones. The direction is always downward, regressing toward the upstream definition; no answer scores higher than it did before.

## Acceptance Criteria

* `_build_rank_vectors()` now ranks both sequences against the shared union the way upstream `to_rank()` does, so system-only events tie on the reference side instead of receiving ascending ranks.
* After the patch, the metric agrees with `scipy.stats.kendalltau(variant="b")` over upstream's rank construction on **80,000 randomised alignments** (including malformed LLM indices) with **0 mismatches**, worst absolute difference 0.0.
* **14 new unit tests** in `cognee/tests/unit/eval_framework/test_kendall_tau.py`; this module previously had no test coverage. `test_system_only_events_tie_in_the_reference_ranks` is the regression guard for the main bug, and `test_single_system_only_event_needs_no_tie` pins the boundary that kept it hidden.
* Existing suite unaffected: `pytest cognee/tests/unit/eval_framework/` → **83 passed**.
* `ruff check` and `ruff format --check` pass using the version pinned in `.pre-commit-config.yaml` (v0.15.11).

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

No public API changes. It does change computed `event_ordering` scores, always downward toward the upstream definition — see the caveat at the end of the description.

## Screenshots

result3
result2

Commands used, both run against commit `d1d2778`:

```
pytest cognee/tests/unit/eval_framework/test_kendall_tau.py -v     # 14 passed
```

and the full local check:

```
ruff check / ruff format --check (v0.15.11)      All checks passed! / 2 files already formatted
pytest cognee/tests/unit/eval_framework/test_kendall_tau.py        14 passed
pytest cognee/tests/unit/eval_framework/                           83 passed
pytest cognee/tests/integration/ --collect-only                    418 tests collected
differential vs upstream reference (scipy)                         0 mismatches, worst 2.22e-16
```

A note on the integration suite: it collects cleanly (418 tests, no collection errors), but executing it needs a live `LLM_API_KEY`, which I do not have set locally. This change touches only pure rank-vector arithmetic and no LLM call path, so it is covered by the unit tests above. Happy to run anything else you would like to see.

## Pre-submission Checklist

- [x] All new and existing tests pass (unit; integration collects cleanly but needs a live `LLM_API_KEY` to execute — see Screenshots)
- [ ] I have linked any relevant issues in the description (no existing issue covers this)

## DCO Affirmation


Vasilije1990:
I had a discussion with our research lead, and we understood that this is only impacting a small subset of questions in another subset of questions and would not change the results. I'm closing this.

PR #4854, opened 2026-08-31
fix: Verify the schema matches the models (SDK-526)

NMZivkovic:
## Problem

Both migration paths only **assert** the schema reached the target — neither verifies it.

```python
if await _relational_schema_exists():
    await run_relational_migrations(...)     # upgrade, assumes each revision did its job
else:
    await get_relational_engine().create_database()   # create_all from the ORM models
    await run_relational_stamp("head", ...)           # records the chain, runs NOTHING
```

The fresh-database shortcut is sound *provided the models and the Alembic head describe the same schema*. Nothing checks that.

In cognee cloud they diverged for nine days. The Alembic tree had gained `a7f3c9e1b5d2` (14 columns on `pipeline_runs`) while the image still bundled a cognee whose `PipelineRun` model had 8. **148 tenant databases** were provisioned with an 8-column table stamped at head — a schema Alembic will never migrate again, because it considers that revision applied.

Every signal read healthy the whole time: initContainer `exit 0`, `alembic current` = head, logs saying `Relational migrations applied (target head)`. It surfaced eleven days later as `column pipeline_runs.user_id does not exist` in a customer's `cognify` call.

## Change

`verify_relational_schema()` compares the live schema against the same `Base.metadata` that `create_database` builds from, and reports the tables and columns the models declare that the database lacks. It runs on **both** paths, before the graph/vector chain reads those tables.

`MIGRATION_SCHEMA_CHECK` selects behaviour: `strict` raises, `warn` (default) logs, `off` skips.

## Two deliberate defaults, both open to challenge

**Default is `warn`, not `strict`.** A fleet that already carries drift would fail closed on every affected tenant the moment this lands — turning a latent problem into an outage. `warn` surfaces the drift immediately (it would light up all 148 today) and the flip to `strict` is a one-line change once a fleet is clean. If you'd rather ship fail-closed from the start, say so and I'll flip it.

**Inspection failures are swallowed**, logged as a warning and treated as "no drift". A safety net must not become a new way for startup to fail — that lesson is fresh: adding embedding credentials to the migrations initContainer converted a crash into an unbounded hang.

## Testing

13 new tests reflecting a **real SQLite database** rather than mocking the inspector, so a change that stops detecting drift fails here.

Mutation-checked — each of these turns the suite red:
| Mutation | Result |
|---|---|
| drift detection always returns `[]` | 2 failed |
| remove the call from `apply_all_migrations` | 3 failed |
| `strict` no longer raises | 1 failed |

Existing suites: 64 migration tests pass; `cognee/tests/unit` shows no new failures (the 15 in a full-suite run are pre-existing event-loop pollution — `modules/retrieval` passes 518/518 in isolation on both this branch and `dev`).

### DIFF
--- a/cognee/modules/migrations/startup.py
+++ b/cognee/modules/migrations/startup.py
@@ -242,6 +242,113 @@ async def _relational_schema_exists() -> bool:
     return "users" in tables or "alembic_version" in tables
 
 
+_SCHEMA_CHECK_MODES = ("strict", "warn", "off")
+
+
+def _schema_check_mode() -> str:
+    """Read ``MIGRATION_SCHEMA_CHECK`` dynamically (same contract as
+    ``ENABLE_AUTO_MIGRATIONS``): ``strict`` raises on drift, ``warn`` logs it,
+    ``off`` skips the comparison entirely."""
+    mode = os.getenv("MIGRATION_SCHEMA_CHECK", "warn").strip().lower()
+    if mode not in _SCHEMA_CHECK_MODES:
+        logger.warning(
+            "Unknown MIGRATION_SCHEMA_CHECK=%r; using 'warn'. Valid values: %s.",
+            mode,
+            ", ".join(_SCHEMA_CHECK_MODES),
+        )
+        return "warn"
+    return mode
+
+
+async def _collect_relational_schema_drift() -> list[str]:
+    """Tables and columns ``Base.metadata`` declares that the live database lacks.
+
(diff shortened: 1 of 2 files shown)


NMZivkovic:
Tracked in [SDK-526](https://linear.app/cognee/issue/SDK-526/verify-the-relational-schema-matches-the-models-after-migration) (SDK, Cycle 67).

PR #4856, opened 2026-08-31
fix(embeddings): recover from Vertex per-model batch-size rejections in embed_text

lallenlowe:
## Description

Follow-up to #4569. While testing that fix against a real Vertex deployment, we found Vertex enforces two separate batch limits, and each rejects with its own error message:

- the per-prediction instance cap: `2048 instance(s) is allowed per prediction. Actual: 4777`, already recovered by #4569
- the per-model batchSize cap: `Unable to submit request because it included too many instances. ... a batchSize value of 1234 but the supported range is from 1 (inclusive) to 251 (exclusive).` — **not** recovered

A batch under the first cap but over the second (e.g. 1,234 names against `gemini-embedding-001`'s 250 limit) fell through the recovery path and was re-raised, so `embed_text` still crashed on large graphs even with #4569 applied.

This widens `_EMBED_LENGTH_ERROR_RE` to also match the second phrasing, so the existing recursive split lands the request under whichever cap is binding. No behavior change for any other BadRequest.

## Acceptance Criteria

- A batch-size rejection phrased the per-model way (the verbatim QA error text) is recovered by the existing split-and-retry path instead of re-raised.
- The per-prediction phrasing from #4569 and the OpenAI phrasing keep working (existing tests cover both).
- Genuinely bad requests still fail fast (narrow alternation added; nothing else widened).
- New unit test fails without the fix and passes with it.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots

**Unit tests** (`test_embedding_context_window_fallbacks.py`, including the new `test_litellm_embedding_splits_batch_on_vertex_model_batch_limit_error`):

unit tests passing

**Integration test** (1,234 entity names, the exact batch size that failed on our QA deployment, embedded through Vertex `gemini-embedding-001` via a LiteLLM-compatible gateway, exit 0):

integration test passing

## Pre-submission Checklist

- [x] My code follows the project's coding standards and style
- [x] I have added tests that prove my fix is effective or that it works

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/databases/vector/embeddings/LiteLLMEmbeddingEngine.py
+++ b/cognee/infrastructure/databases/vector/embeddings/LiteLLMEmbeddingEngine.py
@@ -41,8 +41,15 @@ logger = get_logger("LiteLLMEmbeddingEngine")
 # BadRequestError (e.g. OpenAI 400 "maximum input length is 8192 tokens"). Match
 # those by message so the split/pool recovery below can handle them too. Kept
 # narrow to length/token-limit phrasings so genuinely-bad requests still fail fast.
+# Vertex reports oversized batches two ways depending on which limit is hit:
+# the per-prediction instance cap ("2048 instance(s) is allowed per prediction")
+# and the per-model batch cap ("a batchSize value of 1234 but the supported
+# range is from 1 (inclusive) to 251 (exclusive)" -- "too many instances").
 _EMBED_LENGTH_ERROR_RE = re.compile(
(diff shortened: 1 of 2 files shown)


lallenlowe:
Checked the failing CI jobs, they're infra-side (runners dying with no steps executed) and a missing `LLM_API_KEY` CI secret affecting other PRs today too, not this diff; the suites that did run (pre-commit/lockfile, MCP, integration-infra) all pass.

PR #4859, opened 2026-09-01
feat(graph): Add typed edges for custom models (SDK-162)

lxobr:
## Description
Custom graph models can declare sibling relationships as typed `Edge` fields. The LLM returns identity strings; cognee resolves them to ordinary stored edges.

[SDK-162](https://linear.app/cognee/issue/SDK-162/add-explicit-relationship-extraction-for-custom-graph-models)

- `Edge` is generic over source type, target type, and relationship-name mode (field name, enum, or free-form string).
- A nested `FromIdentity` field lets the LLM name a related node by its identity value instead of nesting a full object.
- Extraction compiles `list[Edge[...]]` fields into flat source/target rows, attaches resolved `Edge` objects on the built tree, and skips a row whose endpoint is missing.
- The model walk reads edges from `DataPoint` field lists and builds the stored node itself (relationship fields stripped). Nested DataPoints and `(Edge, DataPoint)` tuples still work.
- The custom-model guide covers nested groups, `FromIdentity` roles, and fixed / enumerated / free-form edges. The default `KnowledgeGraph` extraction path is unchanged.
- **Note on edge counts:** stored edge identity is now (source, target, relationship_name) instead of per declaring field, so re-ingesting existing data can report fewer edges than before — deduplication, not data loss.
- **SDK-only for now:** the JSON-schema round trip (`POST /cognify`/`/remember` with a model schema) cannot carry `Edge` generics or `FromIdentity` markers, so typed edges work through the Python SDK only; HTTP support is a follow-up.

## Acceptance Criteria
- [ ] A custom model can declare a sibling relationship as `list[Edge[Person, Person]]`
- [ ] Fixed names come from the field; enumerated and free-form names come from the LLM
- [ ] An unresolved endpoint is skipped and logged; the chunk still stores
- [ ] Nested DataPoints and hand-authored `(Edge, target)` tuples still produce the same edges

## Type of Change
- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation


cognee/shared/graph_model_utils.py
```diff
@@ -13,8 +15,279 @@
...
+
+def _instance_at(root: DataPoint, path: tuple[Any, ...]) -> Any:
+    current: Any = root
+    for step in path:
+        current = current[step] if isinstance(step, int) else getattr(current, step)
+    return current
+
+
+def _lookup_endpoint(index: dict, endpoint_type: type[DataPoint], identity_value: str):
+    return index.get((endpoint_type, endpoint_type.id_for(identity_value)))
```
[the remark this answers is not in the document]
lxobr: Not critical, created a follow up.

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -199,53 +143,35 @@ def _walk_data_point(
...
-        if edge_key not in state.added_edges:
-            edges.append(
-                (
-                    data_point.id,
-                    target.id,
-                    relationship_name,
-                    _create_edge_properties(
-                        data_point.id, target.id, relationship_name, edge_metadata
-                    ),
+    field_edges = data_point.get_edges_from_fields()
```
NMZivkovic: **[Critical] The walk crashes on `copy_model()`-built nodes**

This calls DataPoint-only methods (`get_edges_from_fields` / `get_fields_without_edges`), but `copy_model()` (`cognee/modules/storage/utils/__init__.py:49`) does `create_model(..., __base__=ConfiguredBase)` where `ConfiguredBase(BaseModel)` — not `DataPoint`. The pre-PR `iter_fields()` was BaseModel-agnostic.

This fails an existing test on the branch:

```
pytest cognee/tests/unit/tasks/storage/test_chunk_ownership_exactness.py
# branch:      1 failed, 54 passed
#              AttributeError: 'DocumentChunk' object has no attribute 'get_edges_from_fields'
# merge-base 6c02245:  5 passed
```

That test documents the COGX-import path (a chunk rebuilt from an export). Same crash reachable from `unwrap_transparent_nodes.py:94/98` and `_graph_node_from` at `:69`.

cognee/infrastructure/engine/models/Edge.py
```diff
@@ -24,8 +31,66 @@ class Edge(BaseModel):
...
+        field_name: str,
+        target: DataPoint | None = None,
+    ) -> Self:
+        """Return an Edge with source, target and relationship_type filled from context.
+
+        Source falls back to ``owner``, name to ``self.relationship_type or field_name``.
+        When both ``self.target`` and the ``target`` argument are set, the argument
+        wins: it is the tuple form's target at the point of use.
+        """
+        resolved_source = self.source if self.source is not None else owner
```
NMZivkovic, line 84: **[Critical] An omitted `source` silently becomes the field's owner**

`resolved_source = self.source if self.source is not None else owner`, and `source: Optional[Source]` accepts `None`, so pydantic cannot catch the omission.

```python
SocialGraph(people=[alice, bob], friends_with=[Edge(target=bob)])
# walks to: SocialGraph --friends_with--> Bob
# expected: Alice --friends_with--> Bob
```

The declared `Source` generic says a container→Person edge is impossible, but that is exactly what gets written. For a custom `graph_model` the container root is an unnamed `PeopleGraph` node, so it becomes the subject of *every* such relationship. No error, no warning.

`source=` is the one field the LLM path always fills, so no test covers the hand-written case. Contrast three lines down: a missing **target** raises a loud `ValueError`. Suggest making the same omission loud, or narrowing the annotation.

cognee/modules/graph/utils/unwrap_transparent_nodes.py
```diff
@@ -79,38 +91,15 @@ def unwrap_transparent(
...
-        for target in iter_targets(edge_targets):
-            for node in unwrap_transparent(target, _active):
-                if str(node.id) not in seen:
-                    seen.add(str(node.id))
-                    resolved.append(node)
-
+        if edge.source is not data_point:
+            _warn_foreign_source_edge(data_point, field_name)
+            continue
+        resolved.extend(unwrap_transparent(edge.target, _active))
```
NMZivkovic: **[Critical] `unwrap_transparent` lost its `seen` dedup set**

Pre-PR this loop kept `seen = set()` and appended only ids it had not returned. `resolved.extend(unwrap_transparent(edge.target, _active))` has no dedup, and `_active` only guards true cycles (same object on the current path), not diamonds.

Reproduced with `Box(a=child, b=child)` chains that all resolve to **one** leaf:

| depth | results | time |
|---|---|---|
| 4 | 16 | — |
| 8 | 256 | 14.7 ms |
| 12 | 4,096 | 307 ms |
| 16 | 65,536 | 10,977 ms |

Pre-PR every case returns 1. Each duplicate then costs a `claim_edge` f-string plus a full `_walk_data_point` frame.

cognee/shared/graph_model_utils.py
```diff
@@ -13,8 +15,279 @@
...
+        target=target_node,
+        relationship_type=name,
+    ).normalize(source_node, field_name, target=target_node)
+
+
+async def _attach_edge_rows(root: DataPoint, rows: dict) -> None:
+    stored = await collect_stored_data_points(root)
+    index = {(type(node), node.id): node for node in stored}
+    for (path, field_name), row_dicts in rows.items():
+        owner = _instance_at(root, path)
```
NMZivkovic: **[Critical] Edge owners are located by a pre-validation path applied to the post-validation tree**

The positional path is computed on the raw LLM dump but walked against the validated model, so any `field_validator` that reorders or filters a list mis-attaches edges.

Reproduced with a `@field_validator('cliques')` that merely **sorts by label**:

```
clique 'alpha' (members C,D)  received friends_with=[(A,B)]
clique 'zeta'  (members A,B)  received friends_with=[(C,D)]
```

Silent cross-attachment — no error, wrong graph written. With a validator that *drops* an item, `_instance_at` (`:235`, `current[step]`) raises an uncaught `IndexError`; `content_graph_to_data_point` does not catch it and neither does `extract_content_graph`, so the whole cognify run dies.

Suggest keying rows by owner identity rather than positional path.

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -199,53 +143,35 @@ def _walk_data_point(
...
+    nodes.append(_graph_node_from(data_point, field_edges))
+
+    for _field_name, edge in field_edges:
+        for target in unwrap_transparent(edge.target):
```
NMZivkovic, line 187: **[Major] `edge.source` is never unwrapped, so a transparent source writes a dangling edge**

`edge.target` is unwrapped here but `edge.source` is not. `Edge.source` is new in this PR, so a transparent container can now legally *be* a source.

```python
Holder(links=[Edge(source=<transparent Team>, target=bob)])
# emits (team.id, bob.id, 'links')
# but _walk_data_point(team) unwraps the container and stores only its children
# -> team.id is in no stored node
```

`ensure_default_edge_properties` (`prepare_edges_for_storage.py:100`) does not filter dangling edges, so this reaches the adapter: on Neo4j/Kuzu the MERGE either drops the edge or mints a phantom label-less node, and `_build_fallback_edge_text` generates text for an unknown id.

`test_transparent_foreign_source_edge_is_skipped_whole` covers a container *owning* an edge, not appearing *as* a source.

examples/guides/custom_graph_model.py
```diff
@@ -1,39 +1,49 @@
...
-class Activity(DataPoint):
+class Role(DataPoint):
     name: str
     metadata: dict = {"index_fields": ["name"], "identity_fields": ["name"]}
 
 
 class Person(DataPoint):
     name: str
-    likes: List[Activity] | None = None
+    is_a: Annotated[Role, FromIdentity()] | None = None
```
NMZivkovic, line 22: **[Major] The shipped guide's `Annotated[...] | None` is silently ignored**

`Annotated[Role, FromIdentity()] | None` keeps the annotation **inside** the Optional, so pydantic never surfaces the marker. Verified on this exact model:

```python
Person.model_fields['is_a'].metadata   # []
from_identity_fields(Person)           # {}
```

`is_a` then renders as `{"anyOf":[{"$ref":"#/$defs/Role"},{"type":"null"}]}` where `$defs/Role` carries all 18 DataPoint infra fields plus `$defs` for `DataPoint` and `MetaData` — a 6825-char schema containing `"format": "uuid"`, exactly what `DATAPOINT_INFRA_FIELDS.isdisjoint(...)` asserts against elsewhere.

`Annotated[Role | None, FromIdentity()]` works. Two gaps behind it: `_replace_datapoint_types` (`graph_model_utils.py:309`) has no `Annotated` branch, and the `_FromIdentity` check reads only `field_info.metadata`. Worth fixing the guide **and** raising when `_FromIdentity` appears in the annotation tree but not in `field_info.metadata` — otherwise the next person writes the same thing.

cognee/shared/graph_model_utils.py
```diff
@@ -13,8 +15,279 @@
...
+        rows.update(nested_rows)
+
+    for name in edges:
+        if name not in converted:
+            converted[name] = []
+
+    return converted, rows
+
+
+def _descend_llm_dump(value: Any, annotation: Any, path: tuple[Any, ...]) -> tuple[Any, dict]:
```
NMZivkovic: **[Major] `_descend_llm_dump` skips unions, dicts and Annotated inners**

Returning the raw value for multi-arm unions, dict-valued fields and `Annotated` inners means edge-row payloads nested under those are never rewritten to `[]`, and reach `model_validate` as raw source/target dicts.

```python
class OrgUnion(DataPoint):
    items: list[Union[Team, Note]]      # Team has friends_with: list[Edge[Person, Person]]
# rows == {}
# OrgUnion.model_validate(dump) raises:
#   items.0.Team.friends_with.0.source: Input should be a valid dictionary or instance of Person
```

`content_graph_to_data_point` does not catch it, so `extract_content_graph` raises and the chunk's whole graph extraction fails. Same crash for `dict[str, Team]` and `list[Annotated[Team, Embeddable()]]`.

cognee/shared/graph_model_utils.py
```diff
@@ -13,8 +15,279 @@
...
+def _from_identity_llm_annotation(annotation: Any) -> Any:
+    origin = get_origin(annotation)
+    if origin in (list,):
+        return list[str]
+    if origin in (Union, types.UnionType) and type(None) in get_args(annotation):
+        return str | None
+    return str
+
+
+def _list_edge_inner(annotation: Any) -> type[Edge] | None:
```
NMZivkovic: **[Major] `_list_edge_inner` matches only a bare `list[...]`, so Optional edge fields are invisible**

Changing **only** the default:

```python
friends_with: list[Edge[P, P]] = []          # edge_field_types == ['friends_with']
friends_with: list[Edge[P, P]] | None = None # edge_field_types == []
```

The schema then becomes `$defs/Edge_P_P_Union_str__NoneType__` whose `source`/`target` `$ref` the **real** `P` DataPoint — the LLM is asked to fill `id`, `created_at`, `version`, `metadata`, `belongs_to_set`, `weight`, and `$defs` carries two definitions of the same class. Same for `best_friend: Edge[Person, Person] | None`. No warning; the dump-rewriting and row-resolution passes skip the field too.

Note `DataPoint._edges_in_field` *does* accept a bare `Edge` value, so the walk and the extraction layer disagree on which shapes are supported.

cognee/shared/graph_model_utils.py
```diff
@@ -13,8 +15,279 @@
...
+
+
+def _instance_at(root: DataPoint, path: tuple[Any, ...]) -> Any:
+    current: Any = root
+    for step in path:
+        current = current[step] if isinstance(step, int) else getattr(current, step)
+    return current
+
+
+def _lookup_endpoint(index: dict, endpoint_type: type[DataPoint], identity_value: str):
```
NMZivkovic: **[Major] `_lookup_endpoint` can never resolve a base- or subclass-typed endpoint**

The key is the exact runtime class **and** `endpoint_type.id_for(...)`, but `DataPoint.id_for` (`DataPoint.py:183`) namespaces the uuid5 by `cls.__name__`:

```python
uuid5(NAMESPACE_OID, f"{cls.__name__}:{joined}")
# Person.id_for('Alice') != Employee.id_for('Alice')
```

so an `isinstance` check on the dict key would fix nothing. Reproduced:

```python
class G(DataPoint):
    people: list[Employee]
    friends_with: list[Edge[Person, Person]]
# root.friends_with == []   (only logger.warning('Skipping unresolved edge on friends_with'))
```

Declaring an edge against a base type is the natural polymorphic pattern, and it yields zero edges on every chunk with no exception and no count. `test_subclass_endpoint_is_skipped` currently enshrines this as intended. Suggest indexing stored nodes by normalized identity value and selecting with `isinstance`.

cognee/shared/graph_model_utils.py
```diff
@@ -13,8 +15,279 @@
...
+    """Return the Edge[...] class if this is list[Edge[...]], else None."""
+    if get_origin(annotation) not in (list,):
+        return None
+    inner = get_args(annotation)[0]
+    if isinstance(inner, type) and issubclass(inner, Edge):
+        return inner
+    return None
+
+
+def _edge_type_args(inner: type[Edge]) -> tuple[type[DataPoint], type[DataPoint], Any]:
```
NMZivkovic: **[Major] `_edge_type_args` breaks three legal `Edge` annotations**

It reads `__pydantic_generic_metadata__['args']` off the class only, without checking the args are types or walking the MRO. All three verified:

1. **Bare `list[Edge]`** — `args == ()` → `InvalidReferenceTypeError` (HTTP 400). Pre-PR the same model built fine (1771-char schema), so this is a regression for existing custom models. The raise also fires from `cognee/modules/cognify/estimator.py:217` and `graph_model_to_graph_schema()`.
2. **`list[Edge["Person","Person"]]`** — the only way to write a self-referential typed edge in a class body — keeps raw strings in that metadata even after `model_rebuild()`, so `_single_identity_field` does `'Person'._get_identity_fields()` → uncaught `AttributeError`.
3. **`class WeightedFriendship(Edge[Person, Person]): since: int | None`** then `friends: list[WeightedFriendship]` → 400, even though `get_graph_from_model` handles it and stores `since` as an edge property. Subclassing `Edge` to attach typed metadata is the obvious extension path, so the walk supports it while the schema builder bans it.

The error message names neither the field nor the owning model.

cognee/shared/graph_model_utils.py
```diff
@@ -13,8 +15,279 @@
...
+    index = {(type(node), node.id): node for node in stored}
+    for (path, field_name), row_dicts in rows.items():
+        owner = _instance_at(root, path)
+        source_type, target_type, row_model = _edge_field_spec(type(owner), field_name)
+        built = []
+        for row in row_dicts:
+            edge = _resolve_edge_row(row, row_model, index, source_type, target_type, field_name)
+            if edge is not None:
+                built.append(edge)
+        setattr(owner, field_name, built)
```
NMZivkovic: **[Major] `setattr` here builds an object cycle and skips validation**

Two problems on this line.

**1. Object cycle.** The Edges being attached hold endpoints from the same tree, so a model declaring its edge field on one of its own endpoint types self-references:

```python
class Person(DataPoint):
    name: str
    knows: list[Edge[Person, Person]] = []
# LLM row {'source':'Alice','target':'Bob'} on Alice:
alice.knows[0].source is alice          # True
alice.model_dump()                       # ValueError('Circular reference detected (id repeated)')
```

The graph walk survives (edge fields are stripped from the stored copy), but any `to_json` / `to_dict` / `model_dump` — logging, telemetry, snapshot/export, a cached LLM payload — raises instead of returning. Even without a cycle, `root.model_dump()` now re-serializes both endpoints in full inside every edge, inflating output to O(edges × node size).

**2. Validation bypass.** `setattr` skips validation, so the `Literal` constraint the generated row model exists to enforce is never applied to the stored value.

cognee/infrastructure/engine/models/DataPoint.py
```diff
@@ -374,3 +377,47 @@ def from_dict(cls, data: dict[str, Any]) -> "DataPoint":
...
+            if isinstance(item, tuple) and len(item) == 2 and isinstance(item[0], Edge):
+                edge_metadata, targets = item
+                if isinstance(targets, DataPoint):
+                    edges.append(edge_metadata.normalize(self, field_name, target=targets))
+                    continue
+                if isinstance(targets, list) and targets and isinstance(targets[0], DataPoint):
+                    for inner in targets:
+                        edges.append(edge_metadata.normalize(self, field_name, target=inner))
+                continue
+            if isinstance(item, Edge) and item.target is not None:
```
NMZivkovic: **[Major] A targetless `Edge` is handled three inconsistent ways**

`Edge[A,B].target` is Optional and defaults to `None`, so all three constructions are legal:

- **single field** → stored as a node **property** holding a serialized DataPoint subtree
- **list field** → silently dropped
- **tuple field** → loud `ValueError` in `Edge.normalize`

Verified: `G(friends_with=[Edge(source=alice)])` yields 1 node, 0 edges, and a `friends_with` node property holding alice's full JSON dump (id, metadata, all infra fields) — alice is never stored as a node and nothing warns. `[Edge(source=a,target=b), Edge(source=a)]` produces exactly one edge, the second discarded with zero diagnostics.

The deleted `extract_field_relationships.iter_fields` docstring stated the invariant explicitly ("storing it would store DataPoints") and enforced it by classifying on the field *before* target resolution. On Neo4j, `SET n += node.properties` rejects non-primitive array elements, failing the whole `add_nodes` batch.

cognee/modules/graph/utils/unwrap_transparent_nodes.py
```diff
@@ -58,6 +53,23 @@ def _warn_dropped_field(data_point: DataPoint, field_name: str, value: Any) -> N
     )
 
 
+def _warn_foreign_source_edge(data_point: DataPoint, field_name: str) -> None:
```
NMZivkovic, line 59: **[Major] This warning and `_warn_dropped_field` mutually suppress each other**

`_warn_foreign_source_edge` reuses the `_WARNED_DROPPED_FIELDS` set **and** the identical `(class qualname, field name)` key as `_warn_dropped_field`, so whichever fires first silences the other.

Reproduced: a transparent `Crowd` whose `friends_with` first holds a non-edge truthy value claims the key via `_warn_dropped_field`; a later `Crowd(friends_with=[Edge(source=alice, target=bob)])` then returns `nodes == []`, `edges == []` and emits **zero** warnings (expected 2, got 1). A whole relationship vanishes with no diagnostic.

This function also drops the `if field_name in DataPoint.model_fields: return` guard the other one has. Suggest its own set, or a key including the warning kind — or just use `cognee.shared.logging_utils.warn_once`, which this hand-rolls.

cognee/infrastructure/engine/models/Edge.py
```diff
@@ -24,8 +31,66 @@ class Edge(BaseModel):
...
+                "target": resolved_target,
+                "relationship_type": resolved_name,
+            }
+        )
+
+    def to_properties(self) -> dict[str, Any]:
+        """Edge metadata for storage, excluding source, target and relationship_type."""
+        data = self.model_dump(
+            exclude_none=True,
+            exclude={"source", "target", "relationship_type"},
```
NMZivkovic: **[Major] `to_properties()` drops `relationship_type` — an unversioned on-disk change**

Pre-PR `_create_edge_properties` did `edge_metadata.model_dump(exclude_none=True)`. A/B on the same input:

```
pre-PR : [edge_text, relationship_name, relationship_type, source_node_id,
          target_node_id, updated_at, weight, weight_conf, weights]
post-PR: relationship_type is gone
```

This is the mainline cognify path — `expand_with_nodes_and_edges.py:124` builds `Edge(relationship_type='contains', ...)` for every DocumentChunk link, and the ontology/temporal/rdf paths do the same. Downstream:

- `cognee/modules/visualization/preprocessor.py:1361` reads it from the stored blob and now emits `None`
- `tasks/memify/global_context_index/load.py:115`'s fallback arm becomes dead code
- a single graph ends up carrying both edge shapes, so user Cypher filtering on `relationship_type` silently matches only pre-PR edges

If the removal is intended it needs a migration note; if not, it should go back into the property bag.

NMZivkovic (commented):
## Automated review — cognee `/code-review max`

Reviewed at max effort across five angles (correctness, removed-behaviour audit, cross-file tracing, efficiency, conventions). **Every finding below was reproduced by running code against this branch**, not inferred from reading the diff. 26 inline comments follow, tagged Critical / Major / Minor.

### The one that blocks merge

An existing test fails on this branch:

```
pytest cognee/tests/unit/tasks/storage/test_chunk_ownership_exactness.py
# this branch:         1 failed, 54 passed
# merge-base 6c02245:  5 passed
```

### Theme

Most of the top findings are **silent wrong-graph** bugs rather than crashes: an omitted `Edge(source=)` is rewritten to the field's owner; edge rows are attached by a positional path computed pre-validation; and an edge declared against a base class can never resolve, dropping every row behind a single `logger.warning`. All three write an incorrect graph with no error surfaced.

### One finding that has no line in the diff

**Typed edges are effectively SDK-only.** `graph_model_to_graph_schema` round-trips drop both edge and `FromIdentity` fields, so `/cognify` and `/remember` called with a `graph_model_schema` silently lose typed edges. Two currently-green tests lock this in as intent rather than flagging it as a known hole — worth confirming whether that is deliberate for this PR.

### Verified clean

`ruff check`, `ruff format --check` and pre-commit all pass; no import cycle introduced; the deleted `get_data_points_from_model` and `extract_field_relationships` had no remaining importers; the new CI job is correctly wired; and `visited_properties` was already per-root pre-PR, so dropping the sharing is a no-op.

---

cognee/infrastructure/engine/models/DataPoint.py
```diff
@@ -374,3 +377,47 @@ def from_dict(cls, data: dict[str, Any]) -> "DataPoint":
...
+        Targets stay raw: a transparent container is still a container here.
+        """
+        return [
+            (field_name, edge)
+            for field_name, field_value in self
+            if field_name != "metadata"
+            for edge in self._edges_in_field(field_name, field_value)
+        ]
+
+    def get_fields_without_edges(self) -> list[tuple[str, Any]]:
```
NMZivkovic: **[Major] `get_fields_without_edges()` rebuilds every Edge a second time (1.7–2.6× slower walk)**

`_walk_data_point` computes `get_edges_from_fields()` at `get_graph_from_model.py:149`, then `_graph_node_from` (`:69`) calls this, which walks every field again and re-builds and re-`normalize()`s every Edge only to discard it. `unwrap_transparent` (`:94`, `:98`) pays it a third time.

Measured independently by three passes:

| workload | pre-PR | post-PR |
|---|---|---|
| 60×25 chunks (4704 nodes / 9000 edges) | 0.0776 s | 0.1339 s |
| 30×400 warm | 0.098 s | 0.256 s |
| 30×400 cold | — | 4.2× |

`model_construct` is called 24,000 times for 12,000 edges. Two secondary costs in the same loop: recursion is entered for both `edge.source` and `edge.target` even when already claimed (2,080 vs 529 frames — pre-PR checked `added_nodes` *before* recursing), and `unwrap_transparent(edge.target)` allocates a throwaway 1-element list per edge where pre-PR did per-*field* work and returned the same list object when nothing was transparent.

Worth noting this is the file whose header comment documents a prior +50 MB/cycle incident in this exact walk.

cognee/shared/graph_model_utils.py
```diff
@@ -13,8 +15,279 @@
...
+            continue
+        fields[name] = _edge_type_args(inner)
+    return fields
+
+
+def _row_class_name(field_name: str) -> str:
+    return "".join(part.title() for part in field_name.split("_")) + "Edge"
+
+
+@lru_cache(maxsize=128)
```
NMZivkovic: **[Minor] `@lru_cache` on `_edge_row_model_for` pins DataPoint classes forever**

The cache keys on DataPoint **classes**, so it retains them: 128/200 dynamically-created classes still alive after `del` + `gc.collect()`, with `hits=1, misses=201`.

`graph_schema_to_graph_model` mints fresh classes per API request, so on that path this is ~100% miss plus pure retention — the exact pattern `_SIMPLE_MODEL_CACHE` was written to replace.

cognee/shared/graph_model_utils.py
```diff
@@ -13,8 +15,279 @@
...
+    """Map field name to the DataPoint type a FromIdentity field refers to."""
+    fields: dict[str, type[DataPoint]] = {}
+    for name, field_info in model.model_fields.items():
+        if any(isinstance(meta, _FromIdentity) for meta in field_info.metadata):
+            fields[name] = _reference_target_type(field_info.annotation)
+    return fields
+
+
+def _single_identity_field(target_type: type[DataPoint], owning_field: str) -> str:
+    identity = target_type._get_identity_fields()
```
NMZivkovic: **[Minor] Raw `AttributeError` for a non-DataPoint `FromIdentity` target**

`Annotated[str, FromIdentity()]` — a plausible reading of the marker's name — kills the pipeline with an uncaught `AttributeError` instead of the `InvalidReferenceTypeError` this module raises elsewhere.

cognee/shared/graph_model_utils.py
```diff
@@ -13,8 +15,279 @@
...
 from cognee.api.v1.search import SearchType
-from cognee.infrastructure.engine import DataPoint
+from cognee.infrastructure.engine import DataPoint, Edge
+from cognee.infrastructure.engine.models.FieldAnnotations import _FromIdentity
+from cognee.modules.engine.utils import generate_edge_name
+from cognee.modules.graph.utils.get_graph_from_model import collect_stored_data_points
 from cognee.shared.logging_utils import ERROR, setup_logging
+from cognee.tasks.graph.exceptions import InvalidReferenceTypeError
+
+logger = logging.getLogger(__name__)
```
NMZivkovic: **[Minor] Raw `logging.getLogger` against the repo convention**

AGENTS.md: "Use shared logging utilities in `cognee.shared.logging_utils`."

Worth fixing here specifically because this logger's one call site is the **only** signal that an extracted edge was silently dropped (the unresolved-endpoint case above).

cognee/shared/graph_model_utils.py
```diff
@@ -13,8 +15,279 @@
...
+        if name not in DataPoint.model_fields and name not in identity and info.is_required()
+    ]
+    if offending:
+        raise InvalidReferenceTypeError(
+            f"{target_type.__name__} on {owning_field} has required fields that cannot "
+            f"be filled from an identity string: {offending}"
+        )
+
+
+def _from_identity_llm_annotation(annotation: Any) -> Any:
```
NMZivkovic: **[Minor] `_from_identity_llm_annotation` returns a scalar for a list annotation**

`Annotated[list[Role] | None, FromIdentity()]` yields `str | None` instead of `list[str] | None`. Round-trip then raises `ValidationError: Input should be a valid list`.

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -52,6 +51,32 @@ def _simple_model_for(data_point_type, excluded_fields):
...
+    for field_name, _edge in field_edges:
+        if field_name == "belongs_to_set":
+            node_properties[field_name] = _belongs_to_set_names(data_point.belongs_to_set or [])
+            continue
+        stripped_field_names.add(field_name)
+    return _simple_model_for(type(data_point), stripped_field_names)(**node_properties)
+
+
 def _create_edge_properties(
     source_id: str, target_id: str, relationship_name: str, edge_metadata: Optional[Edge]
```
NMZivkovic: **[Minor] Edge dedup key moved from field name to relationship name**

This fixes one collapse (locked in by a new test) but introduces the mirror case: two different fields carrying the same explicit `relationship_type` to the same target now collapse into one, with iteration order deciding whose `edge_text` / `weight` survives.

It also silently changes edge counts when existing corpora are re-ingested.

cognee/infrastructure/engine/models/DataPoint.py
```diff
@@ -374,3 +377,47 @@ def from_dict(cls, data: dict[str, Any]) -> "DataPoint":
...
+    def get_edges_from_fields(self) -> list[tuple[str, "Edge"]]:
+        """Return ``(field_name, edge)`` for every field that expands into graph edges.
+
+        Targets stay raw: a transparent container is still a container here.
+        """
+        return [
+            (field_name, edge)
+            for field_name, field_value in self
+            if field_name != "metadata"
+            for edge in self._edges_in_field(field_name, field_value)
```
NMZivkovic: **[Minor] Empty vs non-empty `list[Edge]` produce two different node shapes**

An **empty** `list[Edge[...]]` is stored as an Edge-typed node property, while a non-empty one is stripped — so the same class writes two node shapes and doubles its `_SIMPLE_MODEL_CACHE` footprint.

`_attach_edge_rows` produces exactly this state whenever all rows fail to resolve, which the base/subclass endpoint issue above makes routine rather than rare.

cognee/modules/graph/utils/get_graph_from_model.py
```diff
@@ -52,6 +51,32 @@ def _simple_model_for(data_point_type, excluded_fields):
     return model
 
 
+def _belongs_to_set_names(belongs_to_set: list[Any]) -> list[str]:
```
NMZivkovic, line 56: **[Minor] `_belongs_to_set_names` is recomputed per edge, not per field**

200 NodeSets → 200 calls scanning 40,000 items, 199 of the results overwritten. Pre-PR: 1 call.

cognee/shared/graph_model_utils.py
```diff
@@ -13,8 +15,279 @@
...
+    fields: dict[str, tuple[type[DataPoint], type[DataPoint], Any]] = {}
+    for name, field_info in model.model_fields.items():
+        inner = _list_edge_inner(field_info.annotation)
+        if inner is None:
+            continue
+        fields[name] = _edge_type_args(inner)
+    return fields
+
+
+def _row_class_name(field_name: str) -> str:
```
NMZivkovic: **[Minor] `_row_class_name` uses `str.title()`, so distinct fields collide**

`HTTP_link` and `http_link` both become `HttpLinkEdge`. Colliding classes share `__module__`, so pydantic emits `cognee__shared__graph_model_utils__LinkedToEdge__1` as the type name the **LLM** sees.

cognee/infrastructure/engine/models/Edge.py
```diff
@@ -24,8 +31,66 @@ class Edge(BaseModel):
...
+
     weight: float | None = None
     weights: dict[str, float] | None = None
-    relationship_type: str | None = None
     properties: dict[str, Any] | None = None
     edge_text: str | None = None
+
+    @model_validator(mode="before")
+    @classmethod
+    def _accept_edge_instance(cls, value):
```
NMZivkovic, line 60: **[Minor] `_accept_edge_instance` drops `__pydantic_extra__`**

Rebuilding from `dict(value.__dict__)` loses extras, so an `Edge` subclass with `extra='allow'` loses its metadata entirely on rebuild.

cognee/infrastructure/engine/models/Edge.py
```diff
@@ -1,9 +1,16 @@
...
 
+from cognee.infrastructure.engine.models.DataPoint import DataPoint
 
-class Edge(BaseModel):
+Source = TypeVar("Source", bound=DataPoint, default=DataPoint)
+Target = TypeVar("Target", bound=DataPoint, default=DataPoint)
+RelationshipType = TypeVar("RelationshipType", default=Optional[str])
+
+
+class Edge(BaseModel, Generic[Source, Target, RelationshipType]):
```
NMZivkovic, line 13: **[Minor] `Edge` is not on the public surface, and its docstring predates the PR**

**Public surface.** `Edge` and `FromIdentity` are absent from `cognee.low_level` and `cognee/__init__.py` — the PR's own example and e2e test import them from `cognee.infrastructure.engine` alongside `from cognee.low_level import DataPoint`. Those are not even the same hierarchy: `cognee.low_level.DataPoint` is `ExtendableDataPoint` while `Edge`'s TypeVars bind the base `DataPoint`, so the documented usage mixes two class trees.

**Docstring.** It still documents only the tuple form, and says nothing about the three generics, `source`/`target`, `normalize()`'s fallback rules, or `to_properties()`'s exclusion set. The omitted-source and targetless-Edge behaviours are invisible at runtime, so this docstring was the only place those rules could have been stated.

cognee/tests/test_custom_model_relationships.py
```diff
@@ -0,0 +1,73 @@
...
+        "Alice and Bob like playing board games together."
+    )
+    custom_prompt = (
+        "Extract all people mentioned in the text. "
+        "For each person, extract ALL activities they like, including shared activities. "
+        "Also extract friendships between people."
+    )
+
+    await cognee.add(text)
+    await cognee.cognify(graph_model=PeopleGraph, custom_prompt=custom_prompt)
```
NMZivkovic, line 58: **[Minor] The new e2e test stops at cognify and never searches**

It asserts against the graph engine directly, so the new CI job stays green if edges are written but **unretrievable** — precisely the failure mode the dangling-source and unresolved-endpoint findings produce.

cognee/modules/graph/utils/get_graph_from_model.py
lxobr: Fixed in abc3e1459 — classification is now free functions over any pydantic model
(`modules/graph/utils/field_edges.py`). The chunk-ownership test is green at the tip.

cognee/infrastructure/engine/models/Edge.py
lxobr: The fallback is intentional (an edge declared on a node is from that node), but you're
right it's wrong on a root container. Fixed in 3e17b1c92: falling back on a
parametrized edge checks the owner against the declared `Source` and raises on
mismatch. Docs updated.

cognee/modules/graph/utils/unwrap_transparent_nodes.py
lxobr: Confirmed — reproduced your exact numbers. Fixed in 3e17b1c92: dedup restored, and
the traversal prunes too (pre-PR also explored exponentially, it just deduped the
output — 0.56s at depth 16; now 0.3ms).

cognee/shared/graph_model_utils.py
lxobr: The crash half is fixed in abc3e1459: a row that cannot be attached now drops with a
warning instead of dying with an uncaught IndexError. The silent reorder case is
real; the right fix is keying rows by the owning node's identity instead of list
position — the same identity-index machinery as the subclass-endpoint finding on
`_lookup_endpoint`. We'll file one follow-up ticket covering both: it changes how
rows are addressed end-to-end, too large to bolt onto this PR.

cognee/modules/graph/utils/get_graph_from_model.py
lxobr: Agreed — fixed in 3e17b1c92: the edge is skipped with a warning. Unwrapping a source
is ambiguous (which child would be the subject?).

examples/guides/custom_graph_model.py
lxobr: Fixed in abc3e1459 — markers are found anywhere in the annotation tree now, so the
guide's spelling works as written. Both spellings tested.

cognee/shared/graph_model_utils.py
lxobr: Fixed in abc3e1459 — the inverse pass descends all of them now, tests per shape.

cognee/shared/graph_model_utils.py
lxobr: Agreed — fixed in 3e17b1c92: Optional is peeled, and anything else that would leak a
raw `Edge` into the LLM schema raises at build time instead.

cognee/shared/graph_model_utils.py
lxobr: Agreed, and you're right that isinstance on the key can't fix it — `id_for` is
class-namespaced, so a subclass instance's id differs from the one the declared base
type derives. A real fix needs an index keyed by identity value plus a rule for
ambiguous matches; that's a design change, so we'll file a follow-up ticket (together
with the row-attachment reorder case) rather than rush it into this PR.

cognee/shared/graph_model_utils.py
lxobr: Point by point:

1. String endpoints and the error message: fixed in abc3e1459 — strings resolve
   against the owning model and its module, and the error now names the field and
   the model.
2. Bare `list[Edge]`: keeping the raise. Pre-PR it only "worked" by leaking the raw
   Edge/DataPoint schema into the LLM prompt — the same leak you flagged for
   `list[Edge[P, P]] | None` — so accepting it would reintroduce that.
3. Edge subclasses: not supporting them — subclassing Edge isn't a supported
   surface; `properties`/`weights` are the supported way to attach edge metadata.

cognee/shared/graph_model_utils.py
lxobr: Two separate points:

1. Validation is not skipped: every row is validated against the generated row model
   (Literal relationship names included) before any Edge is built, so `setattr` only
   ever attaches already-validated values.
2. The object cycle is real — a self-referential edge makes `model_dump()` on the
   returned tree raise. Nothing in the pipeline dumps that tree today, so it cannot
   bite yet; we'll file a follow-up ticket to break the cycle before anything does.

cognee/infrastructure/engine/models/DataPoint.py
lxobr: Agreed on the bad case — fixed in 3e17b1c92: a leftover Edge that carries a DataPoint
is dropped with a warning, instead of embedded as a property. Metadata-only Edge
stays a property.

cognee/modules/graph/utils/unwrap_transparent_nodes.py
lxobr: Agreed — fixed in 3e17b1c92: the warning kind is part of the key, and both private
registries were replaced by the shared `warn_once` from `logging_utils` (recurrences
demote to DEBUG rather than going silent).

cognee/infrastructure/engine/models/Edge.py
lxobr: Agreed — the key is written again in 3e17b1c92, so stored edges are unchanged on
disk. Migrating the three readers to the first-class `relationship_name` touches
stored-data consumers beyond this PR, so we'll file that as a follow-up ticket.

cognee/infrastructure/engine/models/DataPoint.py
lxobr: Reproduced (1.8×/2.0× locally). Fixed in 3e17b1c92: field classification runs once
per node (one pass answers both "which fields are edges" and "which are values"),
behind a never-edges type fast path; claim-check before recursing; `belongs_to_set`
hoisted; the per-node debug log removed (it alone cost ~19%). Now 0.90×/1.20× vs the
pre-PR walk — the typical workload beats it; the residual on edge-heavy loads is the
Edge object per relationship. Full numbers in the commit message.

cognee/shared/graph_model_utils.py
lxobr: Fair. It's bounded at 128 entries, so it's a cap on small classes rather than a
leak — leaving it in this PR. The follow-up work on endpoint resolution touches this
module anyway; we'll revisit the caching there.

cognee/shared/graph_model_utils.py
lxobr: Fixed in abc3e1459 — unsupported FromIdentity shapes raise
`InvalidReferenceTypeError` naming the field and the supported spellings.

cognee/shared/graph_model_utils.py
lxobr: Agreed — switched to `get_logger` (3e17b1c92).

cognee/shared/graph_model_utils.py
lxobr: Fixed in abc3e1459 — resolves to `list[str] | None`. Tested.

cognee/modules/graph/utils/get_graph_from_model.py
lxobr: Intended: stored edge identity is (source, target, relationship_name), so the same
triple declared through two fields is one edge. You're right about the visible
effect, though — re-ingesting existing data can report fewer edges than the old key
produced. We'll add a note to the PR description so the count change isn't mistaken
for data loss.

cognee/infrastructure/engine/models/DataPoint.py
lxobr: True, and leaving it: the empty case stores a scalar `[]`, so the two shapes differ
in that one property and nothing downstream branches on it. The main way to reach
the empty shape is every row failing to resolve, which the endpoint-resolution
follow-up ticket we're filing (the `_lookup_endpoint` subclass finding) will make
rarer — we'll note this shape split on that ticket.

cognee/modules/graph/utils/get_graph_from_model.py
lxobr: Agreed — fixed in 3e17b1c92: computed once per node.

cognee/shared/graph_model_utils.py
lxobr: Agreed — fixed in 3e17b1c92: capitalizes without lowercasing (`HTTP_link` → `HTTPLinkEdge`).

cognee/infrastructure/engine/models/Edge.py
lxobr: Only reachable through an Edge subclass with `extra="allow"`, and Edge subclassing
isn't a supported surface — `properties`/`weights` are the supported metadata path
(same call as on the `_edge_type_args` thread). Leaving it.

cognee/infrastructure/engine/models/Edge.py
lxobr: Agreed on both — fixed in 3e17b1c92: `Edge`/`FromIdentity` are exported from
`cognee.low_level` and the docstring rewritten (generics, endpoints, fallback rules).

cognee/tests/test_custom_model_relationships.py
lxobr: Agreed — fixed in 3e17b1c92: the e2e searches after cognify.

lxobr:
Thanks — this caught real bugs. Note the review ran against `42d9cc15d`; the two
commits after it (abc3e1459 and 3e17b1c92) fix the blocker (the copy_model walk
crash — the chunk-ownership test is green at the tip), the inverse-pass container
gaps, the FromIdentity spelling bugs, the string-endpoint resolution, and the rest
of the agreed findings; details on each thread.

"Typed edges are SDK-only" is intended for this PR — the JSON-schema round trip
can't carry the Edge generics or FromIdentity markers. We'll document the limitation
and file a follow-up ticket for HTTP support.

NMZivkovic (approved):

Issue #402, opened 2024-12-31
Ollama endpoint seems not to work


Vasilije1990:
Hello Vasilije,

I just returned from my holiday, and decided to give cognee a go. I am interested to test the local model integration instead of with any third-party. As such, as per docs (https://docs.cognee.ai/core_concepts/local_models), I added the option to use the Ollama service hosted on the same dev machine:
cognee.config.llm_provider = 'ollama'

Using the same code on the same site, I get the following exception when I tried to run:

-------
Running cognify to create knowledge graph...

InvalidValueError: LLM API key is not set. (Status code: 422)Coroutine task errored: `extract_graph_from_data`
('LLM API key is not set.', 'InvalidValueError')
Traceback (most recent call last):
 File "/Users/myacc/codes/python/cognee/lib/python3.11/site-packages/cognee/modules/pipelines/operations/run_tasks.py", line 116, in run_tasks_base
 task_result = await running_task.run(*args)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 File "/Users/myacc/codes/python/cognee/lib/python3.11/site-packages/cognee/tasks/graph/extract_graph_from_data.py", line 19, in extract_graph_from_data
 chunk_graphs = await asyncio.gather(
 ^^^^^^^^^^^^^^^^^^^^^
 File "/Users/myacc/codes/python/cognee/lib/python3.11/site-packages/cognee/modules/data/extraction/knowledge_graph/extract_content_graph.py", line 7, in extract_content_graph
 llm_client = get_llm_client()
 ^^^^^^^^^^^^^^^^
 File "/Users/myacc/codes/python/cognee/lib/python3.11/site-packages/cognee/infrastructure/llm/get_llm_client.py", line 37, in get_llm_client
 raise InvalidValueError(message="LLM API key is not set.")
cognee.exceptions.exceptions.InvalidValueError: ('LLM API key is not set.', 'InvalidValueError')Async generator task errored: `extract_chunks_from_documents`
('LLM API key is not set.', 'InvalidValueError')
-------

Even though I am using a local LLM, it is still asking for LLM api key. When I simply added some random key, the client ends up reaching out to OpenAI, and fails.

I searched your Discord community and there was a post from some months ago with a similar issue as mine. But I don't see any specific suggestions on how to resolve the issue to use just Ollama without any OpenAI reliance.

If you have any suggestion, I will be keen to explore.

Thanks,
Kennedy

Vasilije1990:
[...]

PR #4869, opened 2026-09-01
SDK-551 feat(migrations): chain-born databases — frozen base schema in the initial migration, no more stamp

dexters1:
**Linear:** [SDK-551](https://linear.app/cognee/issue/SDK-551) · cloud adoption: cloud-backend #402 ([CLO-698](https://linear.app/cognee/issue/CLO-698), [CLO-699](https://linear.app/cognee/issue/CLO-699)).

## Problem

The fresh-database bootstrap runs `Base.metadata.create_all()` and then **stamps** Alembic head — asserting, without ever checking, that the models and the migration chain describe the same schema. When that assertion silently broke (a chain carrying the `pipeline_runs` operation-record extension paired with models that predate it), fresh databases were born disagreeing with the revision they were stamped at, `alembic upgrade head` could never repair them, and the damage surfaced days later as `UndefinedColumn` at runtime. A second consequence of `create_all`: what a fresh database contains depends on **which models the creating process happened to import** — two processes can build two different schemas and stamp both as head.

This PR removes the stamp and makes every new database **chain-born**: the chain becomes the single source of schema truth, and `alembic_version` only ever records revisions that actually executed.

## What changed

**1. The initial migration (`8057ae7329c2`) now carries the frozen base schema.** Historically empty (the chain was a delta log; base tables only ever came from `create_all`), it now contains DDL for all 32 base tables + 64 indexes, rendered by Alembic's own autogenerate engine from the **certified** model surface at chain head `d1e2f3a4b5c6` (a real chain-migrated database was verified to match the models exactly, zero drift, before freezing). Two properties are load-bearing:

- **Frozen means frozen.** The body is a snapshot, never regenerated from live models — a live-model body would be a moving target that silently absorbs any model change whose migration is missing. Future schema changes ship as NEW migrations at the head, exactly as before.
- **Everything is existence-guarded.** Each table (with its indexes) is created only when absent (`if "users" not in _existing:`), so the body is idempotent, safe over partially-existing legacy databases, and inert on every database where the revision is already recorded — Alembic considers it applied and never runs it there. Existing deployments are untouched by construction.

**2. `create_database()` builds the schema by running the chain.** It keeps its storage prep (SQLite directory, pgvector extension), then runs `alembic upgrade head` instead of `create_all`. All creation call sites — the bootstrap fresh branch, `create_db_and_tables()`, the API-lifespan fallback — converge on this one path. `script_location` is threaded through (falls back to `COGNEE_ALEMBIC_PATH`, then the packaged chain), so vendored-chain deployments build from *their* chain.

**3. The stamp is gone.** `run_relational_stamp` is removed; nothing writes a revision that didn't execute. A pleasant side effect: the fresh/existing probe (`_relational_schema_exists`) stops being safety-critical — a wrong "empty" verdict now just replays the same guarded chain an existing database would, instead of authorizing a stamp.

## Timing (the cost of honest bookkeeping)

Median of 5 fresh-database builds each, measured on this branch (Apple Silicon; Postgres = pgvector/pg17 in Docker, fresh database per run):

| Dialect | Chain-born (this PR) | create_all + stamp (before) | Delta |
|---|---|---|---|
| SQLite | **103 ms** | 49 ms | +54 ms (2.1×) |
| Postgres | **429 ms** | 270 ms | +159 ms (1.6×) |

The cost is per **database creation**, not per startup — an existing database takes the unchanged upgrade path.

## Verification

- **Empty → chain, SQLite and Postgres**: builds the complete schema — every model table and column, zero name-level diffs against `Base.metadata` (34 tables incl. `alembic_version` and `notebooks`), bookkeeping at head.
- **Idempotency**: a second `create_database()` and a full re-`upgrade` are no-ops.
- **Legacy heal**: a database built by raw `create_all` with NO alembic bookkeeping (the pre-Alembic shape) replays the chain from base — every guarded revision no-ops against what exists, and honest bookkeeping lands.
- New integration tests pin all three properties (`cognee/tests/integration/infrastructure/relational/test_chain_born_database.py`, real SQLite); the bootstrap unit tests are updated to the new contract (fresh → `create_database`, nothing stamped, `run_relational_stamp` gone).
- One real dialect bug was caught and fixed during Postgres verification: the render initially froze `server_default=false()` as SQLite's compiled literal `sa.text('0')`, which Postgres rejects for a boolean — the frozen body now uses the portable `sa.false()`.

## Behavior changes to review consciously

- **Fresh databases now contain the chain's full truth** regardless of which models the creating process imported: the certified 32-table surface plus `notebooks` (the one chain-only legacy table — model removed, drop migration never written). Previously a process that hadn't imported e.g. the integrations models built a smaller schema and stamped it as head anyway.
- **Database creation is ~2× slower** (table above) — visible in test suites that prune/recreate, negligible for real deployments.
- `downgrade()` of the initial revision stays a no-op, preserving the existing contract that downgrade-to-base never drops the base schema.

## Relation to other work

- **#4862** (CI lockstep guard) — complementary: the guard proves at PR time that chain and models agree; this PR makes fresh databases *come from* the chain, so the agreement is enforced by construction at birth. The guard remains the tripwire for the other direction (a model change with no migration).
- **#4849** (probe fail-open fix) — this PR defuses the same hazard from the other side: even a wrong "fresh" verdict can no longer mis-stamp anything.
- **#4854** (runtime drift detector) — unchanged and still valuable for existing databases.
- The Aug-2026 cloud mis-stamp incident (CLO-688) is the motivating case: under this design, the poisoned image pairing would have **built** `pipeline_runs.user_id` from its chain instead of stamping a claim about it.

## Postgres hardening (second commit, from the first CI run)

The Postgres e2e suites surfaced two real regressions the SQLite tests could not:

1. **`DuplicateObjectError: type "pipelinerunstatus" already exists`.** Postgres enum types outlive `DROP TABLE`, so a recreated database (prune + setup) already has them, and Alembic's `create_table` emits an unconditional `CREATE TYPE`. The frozen initial migration now creates the types up front with `checkfirst` and marks the columns `create_type=False` — the chain's own pattern (`211ab850ef3d`).
2. **`KeyError: 'config'` from Alembic's `EnvironmentContext.__exit__`.** Alembic installs module-global proxies for the duration of a command; two commands in concurrent worker threads corrupt each other. Before this PR that could not happen (`create_database` used `create_all`, no Alembic); now `create_db_and_tables()` and `run_migrations()` can both run the chain concurrently. A process-wide lock serializes every Alembic command, and a regression test reproduces the failure without it.

## Hand verification of the resulting database state (Postgres, pgvector/pg17)

- **Prune + recreate cycle** (`create_database` → `delete_database` → `create_database`) succeeds; `alembic_version` = `d1e2f3a4b5c6`; both enum types present exactly once.
- **Deep comparison** of a chain-born database against a `create_all`-born one — column types and nullability, primary keys, foreign keys, unique constraints, indexes, across all 32 tables — is **identical except one index**, where the chain is *more* correct than the model (below). One diff was found and fixed on the way: the frozen DDL now names `users.parent_user_id`'s foreign key `fk_users_parent_user_id_users`, exactly as migration `7c5d4e2f8a91` names it, so that migration's existence guard recognizes it instead of adding a second, auto-named copy.

### Finding for a separate fix: `PipelineRun.__table_args__` is assigned twice

`cognee/modules/pipelines/models/PipelineRun.py` declares `__table_args__` at line 29 (`ix_pipeline_runs_dataset_pipeline_created_at`) and again at line 48 (`ix_pipeline_runs_created_at_id`); the second silently overrides the first, so **`create_all` has never created the composite index** — only migration `d1e2f3a4b5c6` adds it to upgraded databases. Chain-born databases get it. The model fix is a one-line merge of the two tuples; it is deliberately not folded into this PR.

## Third commit: the unit-test matrix

The full unit matrix (all Python versions, all OSes) failed on two things the targeted suites had not exercised:

1. **`create_database()` on a directly constructed adapter built the wrong database.** `env.py` resolved the engine from the global `get_relational_engine()`, so a test fixture's `SQLAlchemyAdapter(f"sqlite:///{tmp}").create_database()` ran the chain against the global default DB and left its own empty (`no such table: data`). The adapter docstring had called this out as a caveat — that was a design flaw, not a caveat: an instance method must act on its instance. Fixed with Alembic's standard handoff: `run_relational_migrations(..., engine=...)` puts the adapter in `config.attributes["relational_engine"]`, `env.py` prefers it over the global engine, and `create_database` passes `self`. A regression test pins the contract (own database built, global untouched).
2. **`graph_metrics.created_at` frozen as `server_default=sa.text("now()")`** — Alembic's renderer emits the model's `func.now()` that way, and SQLite rejects it at INSERT time (`unknown function: now()`). Now `sa.func.now()` (portable: `CURRENT_TIMESTAMP` on SQLite, `now()` on Postgres). Together with the `sa.false()` case, the rule for the frozen body is: server defaults are SQLAlchemy constructs, never compiled literals — both slips came from the renderer, and both were caught by tests on the *other* dialect.

## Fourth commit: the reconcile revision (new head `1c22e6cb5aec`)

Review testing with databases created by the **released** 1.5.3 and 1.5.0 wheels found the one gap chain-born creation leaves for *existing* databases: a database bootstrapped by the old `create_all` + `stamp head` holds exactly the tables the creating process had imported — `import cognee` registers 30 of the 32 model tables (`integration_credentials` and `sync_operations` are only registered by the API routers) — and the head stamp means their creation migrations never run. Nothing heals it afterwards: the initial revision counts as applied, and `create_database()` now runs the chain (a no-op at head) where it used to be an incremental `create_all`.

`1c22e6cb5aec_reconcile_to_frozen_schema` is the new **last** revision, the same design as the cloud's `70b277baf6da`: it carries the same frozen snapshot as the initial revision (derived mechanically from that revision's DDL — never from live models — plus the one chain-provided index the model surface lacks, `ix_pipeline_runs_dataset_pipeline_created_at`), and applies it **additively and idempotently**: missing tables are created whole with their indexes, missing columns and indexes are added to existing tables, nothing is dropped or altered. It plans first and aborts **before any change** if a primary key or a NOT NULL column without a server default is missing on a populated table (that case needs a dedicated backfill migration). SQLite specifics: `ADD COLUMN` rejects function defaults such as `now()`, so column adds go through Alembic batch mode, rebuilding the table only when such a default is involved, with named FK constraints as batch mode requires.

**Verified** (both SQLite and Postgres, real wheels): a 1.5.3-created database (31 tables, stamped `c7e2a9b4d1f3`) and a 1.5.0-created one (31 tables, no `pipeline_runs.user_id`) upgrade to **33 tables at head** and match a chain-born database on every table, column, type, nullability, default, PK, FK and index — the only differences are the dead chain-only `notebooks` table (no model) and the *name* of the `users.parent_user_id` FK. A chain-born database passes through as a no-op. Fresh-build cost on Postgres: ~420 ms (the reconcile adds one inspection pass, ~10–20 ms).

Five real-SQLite integration tests (`test_reconcile_to_frozen_schema.py`) pin the contract: chain-born no-op — also for a second database in the same process; partial-surface legacy convergence; missing columns return with their server defaults including the `now()` rebuild path; a nullable FK column returns with its constraint; the NOT NULL abort changes nothing.

**Cloud note:** when cloud-backend re-locks its pin to a cognee build containing this, its mirror check will list `1c22e6cb5aec`; the cloud chain's own `70b277baf6da` is a superset (45-table union), so the right move there is to whitelist it in `scripts/cognee_migration_whitelist.txt` rather than mirror it.

## Review round (`650a310d6`): one frozen module, dialect from the bind

Two review comments, both accurate, both fixed:

1. **One source for the frozen surface.** The initial revision and the reconcile carried the same 800-line schema in two encodings (imperative `op.create_table` vs declarative `sa.Table`). `cognee/alembic/frozen_schema.py` is now the single source both build from: the initial revision creates whatever frozen table is missing (`create_missing_tables`), the reconcile converges an existing database to it (`reconcile`: tables, columns, indexes; plan-first, NOT NULL abort; SQLite batch mode). Both revisions are now a few lines; the test helper imports the module instead of loading a revision by path. `frozen_metadata(dialect)` builds a fresh `MetaData` per call, so one process can migrate databases of different dialects. **Verified**: chain-born databases before and after the refactor are identical on SQLite and Postgres — every column, type, nullability, default, PK, FK, unique constraint and index. Because a single module makes "regenerate from the models" easier — the one failure the lockstep guard cannot see, since chain and models would then agree — `tests/unit/test_frozen_schema_seal.py` pins a dialect-independent **fingerprint** of the surface (formatting-insensitive, schema-sensitive, mutation-tested). Changing it is a deliberate re-certification, in the same commit.
2. **`1d0bb7fede17` decided its dialect from the globally configured engine**, not from `op.get_bind()`. Pre-existing, but the `engine=` handoff made it reachable: a SQLite adapter created while the config points at Postgres emitted `ALTER TYPE` against SQLite and `create_database()` failed. Fixed to the bind's dialect (the only revision in the chain with that pattern; its unused model imports are gone too). New integration test: a SQLite adapter built in a child process whose configured engine is Postgres — fails on the old revision with exactly that error, passes now.

32 migration tests pass (`test_chain_born_database`, `test_reconcile_to_frozen_schema`, the seal, `test_run_migrations`, the lockstep guard).


cognee/tests/integration/infrastructure/relational/test_chain_born_database.py
dexters1, line 1: This should be deleted before merge

cognee/tests/integration/infrastructure/relational/test_reconcile_to_frozen_schema.py
dexters1, line 1: Should be deleted before merge

siillee (dismissed):

PR #4872, opened 2026-09-02
fix(ci): Point the nightly Ollama endpoint at the native route (SDK-533)

NMZivkovic:
The nightly has failed every night since 2026-08-22. `Ollama Tests / run_ollama_test` is red in all 12 of those runs, and on 08-30 (run 33289344265) and 08-31 (run 33352541264) it was the only failing job. This restores it.

## What is broken

main's `test_ollama.yml` sets `LLM_ENDPOINT: "http://localhost:11434/v1/"`. That was correct while litellm took the OpenAI-compatible route for the configured model `phi4`. The v1.5.1 release merge (`c659c51cc`, 2026-08-21) brought `_qualify_model()` to main (`litellm_native/get_native_client.py:43`), which rewrites a bare `phi4` to `ollama/phi4`. That moves litellm onto its native Ollama route, which appends `/api/generate` to `api_base`. Every LLM call in the job now resolves to `http://localhost:11434/v1/api/generate`, which 404s.

The nightly immediately before that merge (run 32442942321, sha `fd5045f6b`) was green on this job. The nightly at the merge itself (run 32548431304, sha `c659c51cc`) was the first red one. Every nightly since has been red on it.

From job 100109026799 (run 33585616743, event=schedule, branch=main, sha=`e8dd3d93d`):

```
MaskedHTTPStatusError: Client error '404 Not Found' for url 'http://localhost:11434/v1/api/generate'
OllamaError: 404 page not found
litellm.APIConnectionError: OllamaException - 404 page not found
```

Everything before the final step succeeded, including the in-workflow smoke checks: the `/v1/chat/completions` curl returned a real phi4 completion and `/api/embed` returned a real vector. Both models pulled and ran fine on the runner. The only thing wrong is the base URL litellm is handed.

## Why base=main and not dev

dev already has the fix. #4861 (merge `b94908f54`) changed this exact line on dev on 2026-09-01. That merge is not an ancestor of main. GitHub fires `schedule:` events only on the default branch, and `nightly_tests.yml` pins no ref and performs no checkout in its test jobs — all 16 entries are local reusable calls (`uses: ./.github/workflows/*.yml`), which GitHub resolves from the calling workflow's ref. So a scheduled nightly loads its whole job graph, including this workflow, from main. The fix on dev has never once been executed by the nightly.

So this has to be a base=main PR. There is nothing left to land on dev. After this change main's copy of `test_ollama.yml` is byte-identical to dev's, so the backport adds no new drift and the next release merge has nothing to reconcile here.

## The change

One line in `.github/workflows/test_ollama.yml`:

```diff
-          LLM_ENDPOINT: "http://localhost:11434/v1/"
+          LLM_ENDPOINT: "http://localhost:11434"
```

Nothing else. The `/v1/models` readiness poll (line 40) and the `/v1/chat/completions` smoke curl (line 57) stay as they are: they address Ollama's OpenAI-compatible surface directly, and both pass today. `EMBEDDING_ENDPOINT` at line 105 is already correct and matches dev.

The alternatives are worse. Unsetting `LLM_ENDPOINT` works (litellm defaults to `http://localhost:11434`) but creates fresh main/dev drift. Switching `LLM_PROVIDER` to `custom` to stay on the OpenAI-compat route is a larger behavioural change that stops testing the path dev actually uses.

## What this does NOT fix

The nightly will still be red tomorrow morning. On run 33585616743 two jobs failed, and this fixes one of them. `Performance — War and Peace (cloud)` failed with:

```
Cognify FAILED: Remote cognify failed (402): {"detail":"Insufficient credits to run cognify.
Only $10.00 of credits remain. Add credits and try again.","reason":"insufficient_credits",
"operation":"cognify","remaining_usd":10.0}
```

`Determine status` in `nightly_tests.yml` ANDs 16 job results including `needs.perf-wap-cloud.result`, so the run stays red until that arm passes. No change in this repo creates credits; that one is a billing action, tracked in SDK-533.

Deliberately not adding a skip or `continue-on-error` for the 402: a credits-conditional green would hide a genuine remote-cognify regression, and the arm most likely to be refused is the most expensive and most valuable one.

Also outside this PR and tracked in the same ticket: a non-idempotent tenant-creation retry in `bench_cognee.py` that accounts for four of the five recent cloud-arm failures, a missing `all_passed` filter in the MotherDuck regression view that lets a failed run's placeholder timings become a baseline, and Postgres pool exhaustion in the arm.

## Verification

Draft until this is done. `test_ollama.yml` is `on: workflow_call:` only, so it cannot be dispatched on its own, and dispatching the whole nightly on this branch means ~1.5h and the cloud credits that are already exhausted. To verify: temporarily add `workflow_dispatch:` to this file on the PR branch, dispatch it alone, confirm the log reaches `Coroutine task completed: add_data_points` → `Pipeline run completed` → a `graph_completion` result, then drop the trigger before merge.

Worth doing rather than trusting the dev run. The reference green run (33512258137, job 99870579329, Ollama green) is on a dev-based branch that is 164 ahead / 13 behind main, and main's library has moved ~260 files since main's own last green Ollama run. The reasoning here is URL arithmetic that does not depend on the branch, and main's real-LLM perf arms pass nightly through the same adapter, so the residual risk is small and Ollama-specific. It is not zero.

### DIFF
--- a/.github/workflows/nightly_tests.yml
+++ b/.github/workflows/nightly_tests.yml
@@ -81,32 +81,32 @@ jobs:
       mock_memories_key: nightly_ci_artifacts/performance_test_artifacts/mock_war_and_peace.json
     secrets: inherit
 
-  # Henkel technical datasheets: 164 real PDFs (~1.35M chars). Complements the
+  # Technical datasheets: 164 real PDFs (~1.35M chars). Complements the
   # existing corpora on document *count* -- War and Peace is one long document,
   # the 50-doc set is short synthetic ones. This is many medium-sized real-world
   # documents, which is the shape most user datasets actually have.
   #
   # No Rust arms yet: those read fixtures from topoteretes/cognee-rs
   # (scripts/perf/fixtures/<label>/), so they land with that repo's PR.
-  perf-henkel-llm:
-    name: Performance — Henkel (real LLM)
+  perf-datasheets-llm:
+    name: Performance — Datasheets (real LLM)
     uses: ./.github/workflows/performance_report.yml
     with:
       mode: llm
-      label: henkel
+      label: datasheets
       runs: '3'
-      memories_key: nightly_ci_artifacts/performance_test_artifacts/henkel.json
+      memories_key: nightly_ci_artifacts/performance_test_artifacts/datasheets.json
     secrets: inherit
 
-  perf-henkel-mock:
-    name: Performance — Henkel (mock LLM)
+  perf-datasheets-mock:
+    name: Performance — Datasheets (mock LLM)
     uses: ./.github/workflows/performance_report.yml
     with:
       mode: mock_llm
-      label: henkel
+      label: datasheets
       runs: '3'
-      memories_key: nightly_ci_artifacts/performance_test_artifacts/henkel.json
-      mock_memories_key: nightly_ci_artifacts/performance_test_artifacts/mock_henkel.json
+      memories_key: nightly_ci_artifacts/performance_test_artifacts/datasheets.json
+      mock_memories_key: nightly_ci_artifacts/performance_test_artifacts/mock_datasheets.json
     secrets: inherit
 
   # 27x-multiplied capture of the same book (~100k graph nodes/edges per run):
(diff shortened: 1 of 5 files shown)


NMZivkovic:
Closing this. It targeted `main`, which was the wrong call.

The reasoning was that the nightly is a scheduled workflow, so it runs on the default branch, and the endpoint fix already on `dev` (#4861) therefore never reaches it. That much is true. But hand-carrying dev→main fixes through individual PRs is not the right instrument, and this PR demonstrated why: it went red on `MCP Tests` for a completely unrelated reason, namely that `main`'s `test_mcp.yml` references `inputs.python-version` without declaring it, so `setup-python` falls back to 3.10 and litellm 1.98.0's `from typing import NotRequired` fails. That input declaration was added on `dev` in #4664 and, like the endpoint fix, has not reached `main`. Getting this PR green would have meant backporting that too, and then whatever surfaced next.

Nothing here is lost:

- The endpoint change is already on `dev` at the same value, so the commit was a no-op there.
- The corpus label rename is open against `dev` in #4876.

The underlying issue — `main` being 173 commits behind `dev` while the nightly schedule runs on `main` — is tracked in SDK-533 and wants a release-level decision rather than a CI patch.

PR #4876, opened 2026-09-02
fix(ci): Make the nightly self-consistent and stop the flaky/failing jobs (SDK-533)

NMZivkovic:
The nightly has failed every night since 2026-08-22. This PR makes it self-consistent and stops each of the jobs that were failing or flaking, so that after this merges to `dev` and the next release carries it to `main`, the nightly runs the code it should and goes green — subject to two items no PR can deliver, listed at the end.

Five commits, one per fix. Each is independently revertable.

## What was broken

The nightly is a `schedule:` workflow. GitHub fires those only against the default branch's copy of the file, and every one of its 16 `uses: ./...` calls resolves at that same commit. So the cron on `main` has only ever tested released code, and a CI fix merged to `dev` — the Ollama endpoint (#4861), the MCP python-version input (#4664) — stays invisible to it until a release merge. That is why it stayed red for twelve nights after being fixed.

On top of that, three jobs were failing or flaking for their own reasons: the Windows unit and library tests wedge and get killed at 60 minutes on about a third of runs; the datasheets postgres perf arm exhausts its connection pool; and the cloud perf arms fail on a non-idempotent tenant-creation retry, then report search timings for a graph that was never built.

## 1. Nightly scheduler

New `nightly_scheduler.yml` keeps the cron on `main` and dispatches `nightly_tests.yml` at the ref that should be tested: `main` daily at 18:00 UTC — 19:00 CET winter / 20:00 CEST summer, inside the 17:00–22:00 CET peak-US window on both sides of the DST switch, so the cloud perf arms measure the platform under real load — and `dev` weekly on Sunday, so the nightly-only jobs (the 3.11–3.13 unit matrix, the search-DB version sweep, the telemetry pipeline) still exercise in-development code. Trade-off, stated: a dev-only break in those nightly-only jobs is found within a week rather than a day; the classes moved there have zero recorded incidents since 2025-06. It works because `workflow_dispatch` is one of the two events `GITHUB_TOKEN` may raise that still creates a run. `schedule:` is removed from `nightly_tests.yml` in the same commit so the release merge is atomic and nothing fires twice.

The dispatch endpoint returns 204 with no body, so a lost dispatch looks like success and a nightly that never starts posts nothing. The shim records the newest existing run id before dispatching and polls for one newer than it; a confirm step without that baseline re-finds last night's run and passes unconditionally. A dispatch failure alerts Slack.

Two nightlies now exist, so everything that reaches a human or the warehouse says which one it is: a `cadence` input, the branch in the Slack header and preview text, and a provenance stamp (branch, sha, run id, attempt, event) written into every perf report. `v_perf_runs` has always selected `git_sha` and `run_id`; nothing ever wrote them.

In MotherDuck, `suite` stays byte-identical so no dashboard chart breaks; a branch-qualified `series` is added and the regression view partitions on it, otherwise a weekly main run becomes `rn=1` against seven dev baselines. The regression view also drops runs where `all_passed` is false: a failed run's placeholder timing (a 402's 2.05s cognify) was being reported as a 99.6% improvement against a 500s baseline. Since two refs would otherwise make `CREATE OR REPLACE VIEW` last-writer-wins, `PERF_REFRESH_VIEWS` lets the weekly main run load rows without redefining views.

Verified with a local DuckDB: failed runs excluded from latest and baseline, dev and main independent series, the Slack builder renders with and without the new fields, the jq stamp leaves `.stats` untouched. Removing the `all_passed` filter reproduces the phantom -99.6%.

## 2. Windows: bounded, retried, instrumented, and a false pass closed

This does not claim to fix the hang. The unit step has produced four outcomes from a byte-identical environment (normal 15-17 min; 60-min silent wedge; 14s silent exit-0; 14s silent exit-127), and the library step stops at the same import-time log line in both observed hangs. The one shared property is that the step emitted nothing. The PATH guard added for this in 46bd2fe45 did not fix it — PATH is not consulted for extension DLL resolution on Python 3.8+, and every hang since has "Dropped 1 PATH entry" in its log. Its comments are corrected; the guard is kept.

What this does: a Windows unit job has reported SUCCESS in 14 seconds having run nothing (job 97050693924). Exit status does not prove the suite ran, so each attempt now asserts on the junit report (floor 4000 vs 5066 collected) and fails regardless of rc when it is missing or short. The library step uses a completion sentinel the same way. A retry fires only on that "emitted nothing" signature — a step that produced a junit report ran the suite, so its failure is real and is never retried; no genuine bug and no flaky test can hide behind this. Two attempts fit the 60-minute ceiling; a pass-on-retry is a workflow warning. A process-table snapshot every five minutes, from inside the still-running step (an `if: always()` step cannot start until the step ends, which on a wedge is never), image names only. What it shows on the next wedge selects the follow-up.

Every bash step was extracted from the rendered YAML and checked with `bash -n`, and the retry gate was driven through all five outcomes with a fake `uv`. The first cut had a heredoc terminator that YAML's indent-stripping left at column 2, which would have swallowed the pytest invocation and made every Windows job exactly the silent no-op this exists to catch.

## 3. postgres_demo graph adapter: pool exhaustion

`QueuePool limit of size 5 overflow 10 reached` on the datasheets real-LLM postgres arm (jobs 99042050518, 98411831471). Not a leak — all 32 acquisitions are `async with`. Three causes, each with a test that fails when its fix is removed:

- Every concurrent writer checked a connection out and *then* blocked on `pg_advisory_xact_lock`, so 164 writers pinned the whole pool while one made progress. A process-local write gate puts the queue in front of the pool. Peak sessions 50 → 1.
- `initialize()` ran a full `create_all(checkfirst=True)` reflection on every `get_graph_engine()` and every `get_graph_metadata()` — hundreds per batch, each on a pooled connection. That is the read-path failure. Memoized per adapter; 50 → 1.
- The adapter honoured only `poolclass == "nullpool"` and discarded every sizing key, so it always ran the stock 5/10/30s pool. Now resolved properly, defaulting to 2/20/280s — PGVector's per-dataset sizing, not SqlAlchemyAdapter's 5/35, which would put the CI container at 102 connections against `max_connections=100`.

Both locks are per running loop: the adapter is process-cached and outlives the loop it was built on.

## 4. Benchmark harness: tenant adoption, and no timings for work not done

Tenant creation retried a lost response by re-POSTing the same name. But `POST /tenants` commits the row before provisioning starts, so the retry gets `503 already exists` — six occurrences, on two transports (aiohttp's 300s timeout, and an intermediary reset at 272-294s, so a longer client timeout alone would still fail half the time). Both shapes now look the tenant up by name and adopt it; `GET /tenants/me/with-status` runs the identical query to the duplicate check, so when the 503 fires the lookup must find it.

Two leaks closed with it: a 404 on the id we hold was treated as "gone", but the controller's rollback re-creates the same name under a new id which may go healthy, and abandoning it leaks a billed tenant; and `_create_cloud_tenant` raised without its id, so teardown was unreachable on exactly the branch that leaks.

Phase 3 reported SUCCESS on an empty graph: job 100109026857 shows three 402s on cognify each followed by `Search GRAPH_COMPLETION : 2.30s [success]`, and those numbers were the p50/p90/p99 posted to Slack. The phase is skipped when ingest did not succeed and the timing omitted rather than zeroed. Changes no pass/fail outcome; the run was already failed. The report tolerates the resulting mixed runs (it raised `KeyError` or silently dropped the metric depending on run order, and rendered a missing metric as `0.000s`).

Tests drive the real `_create_cloud_tenant` against an in-process aiohttp controller, so a lost response is a genuine transport close. Against pristine dev they reproduce the nightly's verbatim 503 and the leaks.

## 5. Corpus label

The third perf corpus was labelled after its source rather than its content. Renamed to `datasheets` throughout; the benchmark is unchanged.

## Not fixed here

- **Cloud credits.** The War and Peace cloud arm is refused with `402 insufficient_credits, remaining_usd: 10.0`. No repo change creates credits. It may not be depletion either: `remaining_usd` has read exactly 10.0 across sequential runs for seventeen days while other arms spend. Top up, enable auto-recharge, watch two nights.
- **S3 objects for the corpus rename.** The three arms now read `datasheets.json` / `mock_datasheets.json`; the objects must be copied to the new names before this reaches a nightly, or those arms fail on a missing key.
- The Windows hang's mechanism. Instrumented, not diagnosed.
- Per-row provenance round-trips in the postgres adapter (the throughput lever), the synchronous traceback renderer that turns two timeouts into 49, the stale-tenant reaper (designed, flag-gated, not enabled here), and back-merging main→dev after hotfixes.

## Things a reviewer should push on

- `GITHUB_TOKEN` dispatch: the docs say `workflow_dispatch` always creates a run, but the org could disallow elevating to `actions: write`. Fallback is one line: swap to `REPO_DISPATCH_PAT_TOKEN`.
- Pool sizing 2/20 vs 5/35 — either is defensible only with the write gate; neither alone.
- The 4000 junit floor.
- The shim means nothing nightly-exercises `main`'s code except the weekly run. Released artifacts stay covered by `docker_validation_nightly.yml` and `release_test.yml`.

## Release-day note

Until the release merge, `main` still carries the old `schedule:` and keeps running the stale nightly. The merge removes it and adds the shim atomically. The first shim-dispatched nightly will be red on the cloud arm until the credits are sorted; say so when it posts.


.github/workflows/e2e_tests.yml
```diff
@@ -93,36 +93,6 @@ jobs:
           EMBEDDING_API_KEY: ${{ secrets.EMBEDDING_API_KEY }}
         run: uv run python ./cognee/tests/test_telemetry.py
 
-  run-telemetry-pipeline-test:
```
dexters1, line 116: If I am not mistaken this is the only test we have to confirm telemetry is working properly in our CI

The rest disable telemetry completely 

Can we move this one into release_test.yml

There we store tests we need to run before the release, but don't want to run on every PR


For the rest of the tests I am not sure how much we need them, but they can also be moved to the release_test.yml until we are sure we do not need them

The agent_registry and custom graph delete test worry me a bit as I do not know if what they test is covered in another place

.github/workflows/nightly_tests.yml
```diff
@@ -40,6 +50,68 @@ jobs:
     uses: ./.github/workflows/test_llamacpp.yml
     secrets: inherit
 
+  # The PR gate runs the unit suite on the interpreter edges (3.10, 3.14) and
+  # on windows/macos 3.13 -- the four environments with version- or OS-specific
+  # fixes in the history. The three middle Linux versions run here instead.
+  # Same reusable workflow, same extras as the PR gate's ubuntu legs.
+  unit-matrix-mid:
```
dexters1: Might make sense to move these test as well into the release_test.yml so we do not spam the nightly_ci reports too much if that is alright with you

dexters1 (approved):

PR #4881, opened 2026-09-02
fix: Pass max_completion_tokens to litellm (SDK-538)

NMZivkovic:
## Problem

[SDK-538](https://linear.app/cognee/issue/SDK-538/pass-max-completion-tokens-to-litellm-in-the-native-adapter). `NativeLiteLLMAdapter` accepts and stores `max_completion_tokens` — `get_native_client` computes it as `min(model limit, llm_max_completion_tokens)` (default 16384) — but **never sends it**: all three `litellm.acompletion` call sites pass only `{**self.llm_args, **kwargs}`.

On reasoning models (`gpt-5-mini` — the OSS default and cloud's OpenAI bridge) nothing caps output, and reasoning tokens bill as output. The CLO-679 COGS analysis (adversarially verified against the repo's own `estimator.py` calibration, a ~20× reasoning-output multiplier measured on a real default-config run) put default-config cognify at **~$21–24 per 1M ingested tokens** vs a **~$1.72 floor** with output bounded. This unsent cap is most of that gap, and it gates $1/1M (CLO-679) being margin-positive on the OpenAI path.

## Change

One guarded line at the adapter's single kwargs-merge origin (`acreate_structured_output`), covering every call path (plain-str, schema-native, prompted-JSON fallback):

```python
if not ({"max_completion_tokens", "max_tokens"} & merged_kwargs.keys()):
    merged_kwargs["max_completion_tokens"] = self.max_completion_tokens
```

The stand-down matters: cloud tenant charts cap via `LLM_ARGS max_tokens` (24000 / 48000), and sending both aliases is an API error on some providers. Explicit `llm_args` / call kwargs always win.

## Verification

- 4 new tests assert the kwargs **actually received** by a patched `litellm.acompletion`: stored cap on the schema-native and plain-str paths, no injection when `max_tokens` is present, explicit override wins — 32/32 green
- Mutation-checked: replacing the injection with `pass` turns exactly the two direct-assertion tests red
- Behaviour change to expect after deploy: structured outputs can now be truncated at 16384 completion tokens where they previously ran unbounded — at the default chunk size the expected output is far below the cap, but flagging it for review

### DIFF
--- a/cognee/infrastructure/llm/structured_output_framework/litellm_native/native_adapter.py
+++ b/cognee/infrastructure/llm/structured_output_framework/litellm_native/native_adapter.py
@@ -371,6 +371,13 @@ class NativeLiteLLMAdapter:
         fallback model before raising ``ContentPolicyFilterError``.
         """
         merged_kwargs = {**self.llm_args, **kwargs}
+        # The stored cap is min(model limit, llm_max_completion_tokens) — see
+        # get_native_client. Without it in the request, a reasoning model bills
+        # unbounded reasoning output. Inject it as the base layer unless the
+        # caller already caps via either alias: sending max_tokens AND
+        # max_completion_tokens together is an API error on some providers.
+        if not ({"max_completion_tokens", "max_tokens"} & merged_kwargs.keys()):
+            merged_kwargs["max_completion_tokens"] = self.max_completion_tokens
 
         # A plain string needs no schema — skip structured output entirely.
         if response_model is str:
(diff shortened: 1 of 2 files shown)


cognee/infrastructure/llm/structured_output_framework/litellm_native/native_adapter.py
```diff
@@ -478,6 +478,16 @@ async def acreate_structured_output(
         fallback model before raising ``ContentPolicyFilterError``.
         """
         merged_kwargs = {**self.llm_args, **kwargs}
+        token_limit_keys = {"max_completion_tokens", "max_tokens"}
```
dexters1, line 481: Should we clean up or add support for this env var as well then?

Image

For some reason we added it as an explicit config parameter instead of relying on these parameters to be defined through LLM_ARGS

PR #4882, opened 2026-09-02
fix: Read EMBEDDING_API_BASE as endpoint alias (SDK-539)

### LINKED ISSUE #4871: Custom EMBEDDING_API_BASE not respected by LiteLLM routing
## Description
When configuring Cognee with a custom OpenAI-compatible embedding endpoint (SiliconFlow, local vLLM, etc.), the EMBEDDING_API_BASE is ignored by LiteLLM. Instead, LiteLLM routes to OpenAI default endpoint.

## Steps to Reproduce
export EMBEDDING_PROVIDER=openai
export EMBEDDING_MODEL=openai/BAAI/bge-m3
export EMBEDDING_API_KEY=sk-xxx
export EMBEDDING_API_BASE=https://api.siliconflow.cn/v1

## Expected
LiteLLM should use EMBEDDING_API_BASE for embedding requests.

## Actual
Requests go to https://api.openai.com/v1/embeddings (404 error).

## Workaround
Use EMBEDDING_PROVIDER=fastembed or EMBEDDING_PROVIDER=ollama for local embeddings.

## Environment
Cognee 1.5.3, Python 3.12/3.14, WSL2 Ubuntu

NMZivkovic:
Fixes #4871 · [SDK-539](https://linear.app/cognee/issue/SDK-539/accept-embedding-api-base-as-an-alias-for-embedding-endpoint)

## Problem

Configuring a custom OpenAI-compatible embedding endpoint via `EMBEDDING_API_BASE` (SiliconFlow, local vLLM — the name the litellm/OpenAI ecosystem uses) was **silently ignored**: requests routed to `api.openai.com` and 404'd.

Verified root cause: the engine is innocent — `LiteLLMEmbeddingEngine` forwards `api_base=self.endpoint` unconditionally, so custom bases already work under `EMBEDDING_PROVIDER=openai`. The whole failure is that `EMBEDDING_API_BASE` was not a name cognee reads anywhere, and `EmbeddingConfig`'s `extra="allow"` swallowed it without a warning — endpoint stayed `None`, litellm defaulted to OpenAI. A config mistake surfaced as a confusing 404 on the wrong host.

## Change

- `embedding_endpoint` accepts both names via `AliasChoices("EMBEDDING_ENDPOINT", "EMBEDDING_API_BASE")` — first alias wins when both are set
- `populate_by_name=True` so programmatic `EmbeddingConfig(embedding_endpoint=...)` construction keeps working (used by preflight/doctor tests)
- `.env.template` documents the alias

## Verification

- 4 new tests drive **real env parsing** (monkeypatched environment, `_env_file=None`), including the issue's exact repro: `EMBEDDING_API_BASE=https://api.siliconflow.cn/v1` alone → endpoint populated
- 72/72 green across the embedding-config, preflight, context-globals and LLM-config suites (the field-name construction consumers)
- Mutation-checked: removing `"EMBEDDING_API_BASE"` from the alias turns the repro test red

## Follow-up candidates (out of scope, noted on SDK-539)

Same alias for `LLM_ENDPOINT`/`LLM_API_BASE`, and a startup warning for unknown `EMBEDDING_*`/`LLM_*` vars so `extra="allow"` stops eating typos silently.

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -102,7 +102,7 @@ STRUCTURED_OUTPUT_FRAMEWORK="litellm_native"
 #  Tune these when using non-default embedding providers.
 ################################################################################
 
-#EMBEDDING_ENDPOINT=""
+#EMBEDDING_ENDPOINT=""   # EMBEDDING_API_BASE is accepted as an alias
 #EMBEDDING_API_VERSION=""
 #EMBEDDING_MAX_COMPLETION_TOKENS=8191
 #EMBEDDING_BATCH_SIZE=36
--- a/cognee/infrastructure/databases/vector/embeddings/config.py
+++ b/cognee/infrastructure/databases/vector/embeddings/config.py
@@ -1,5 +1,6 @@
 from typing import Optional
 from functools import lru_cache
+from pydantic import AliasChoices, Field
 from pydantic_settings import BaseSettings, SettingsConfigDict
 
 from cognee.shared.logging_utils import get_logger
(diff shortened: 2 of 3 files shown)


.env.template
```diff
@@ -102,7 +102,7 @@ STRUCTURED_OUTPUT_FRAMEWORK="litellm_native"
 #  Tune these when using non-default embedding providers.
 ################################################################################
 
-#EMBEDDING_ENDPOINT=""
+#EMBEDDING_ENDPOINT=""   # EMBEDDING_API_BASE is accepted as an alias
```
dexters1, line 105:
```suggestion
#EMBEDDING_ENDPOINT="" 
```

PR #4889, opened 2026-09-03
refactor(recall): Simplify query router (SDK-545)

Vasilije1990:
## Summary

Rewrites the rule-based query router behind `recall()` as an ordered first-match rule table, removes dead weight, and makes auto-routing consistent across SDK, CLI, and REST.

Linear: [SDK-545](https://linear.app/cognee/issue/SDK-545/simplify-recall-query-router-and-tighten-auto-routing)

## Why

`cognee/api/v1/recall/query_router.py` was a 232-line weighted regex scorer with a negation window, a module-level override counter that nothing reads, and confidence metadata no caller consumes. It also ran on every explicit-`query_type` call just to feed that counter.

Probing 18 realistic queries showed it sending ordinary questions to strategies that are slower and often worse than the HYBRID default on a default-built graph:

| Query | Old route | Problem |
|---|---|---|
| "When was the company founded?" | TEMPORAL | Default graphs have no event nodes (`temporal_cognify=False`), so this pays an extra LLM call for interval extraction and falls back to graph-only triplets |
| "Why is the deploy failing?" | GRAPH_COMPLETION_COT | Up to 4 extra LLM round trips |
| "Refactor plan for the billing service" | CODING_RULES | Reads only the `coding_agent_rules` nodeset; empty on ordinary graphs |
| "What is the relationship between price and demand?" | CONTEXT_EXTENSION | HYBRID already includes entity neighbourhoods |

## What changed

**Router** (232 → 109 lines). Six named rules, first match wins, `RouteDecision(search_type, rule)`:

| Rule | Routes to |
|---|---|
| Leading Cypher clause (`MATCH`, `OPTIONAL MATCH`, `RETURN`, `CREATE`, `MERGE`, `UNWIND`) | CYPHER |
| Quoted phrase, or exact / verbatim / literal | CHUNKS_LEXICAL |
| summarize / overview / tl;dr / main points | GRAPH_SUMMARY_COMPLETION |
| Year range, preposition + year, decade, ISO date, "timeline" | TEMPORAL |
| "coding rules" / "code review guidelines" phrases | CODING_RULES |
| everything else | HYBRID_COMPLETION |

Design rule: auto-routing may only pick a strategy that is at least as good as HYBRID on a default-built graph and does not add LLM calls without an unambiguous signal. COT and CONTEXT_EXTENSION are no longer auto-picked; they stay reachable via explicit `query_type`.

**`recall()`**: router skipped when `query_type` is explicit; docstring fallback corrected to HYBRID_COMPLETION; router logs at DEBUG with the rule name only, never the query text.

**CLI**: `cognee-cli recall` without `--query-type` now lets the SDK route instead of pinning HYBRID_COMPLETION, and prints the type that actually ran.

**REST**: `searchType` now defaults to `null`, so an omitted field auto-routes exactly like the SDK and CLI. REST was the last surface pinning `HYBRID_COMPLETION`. As in the SDK, an omitted type with `sessionId` and no datasets lets a session hit short-circuit the graph search; passing a value disables that. The CLI `--api-url` path, the frontend recall client, and the agent prompts follow the server default and print or pass the resolved type.

**Cypher rule**: anchored to a leading clause keyword only. The unanchored `--(` / `)--` alternatives could fire on natural language mid-sentence (review nit). Negative tests added.

**Docs**: new `docs/recall-vs-search.md` (CLAUDE.md already linked it but it did not exist) with the rule table, the exclusions and why, and a per-surface bypass matrix. CLAUDE.md router wording updated.

**Tests**: moved from the stale `cognee/tests/unit/api/v2/` path to `cognee/tests/unit/api/v1/recall/test_query_router.py` as a 43-query golden table plus negative invariants (no TEMPORAL without a date token, no CODING_RULES without the phrase, never COT / CONTEXT_EXTENSION). One CLI test added.

## Behaviour change

Callers who relied on "why ..." auto-routing to chain-of-thought now get HYBRID answers. Passing `query_type=SearchType.GRAPH_COMPLETION_COT` restores the old behaviour.

REST clients that omitted `searchType` previously got `HYBRID_COMPLETION` and now get the router. Most queries still resolve to `HYBRID_COMPLETION`; the exceptions are the six rules above. Pass `"searchType": "HYBRID_COMPLETION"` to keep the old behaviour. A body with `sessionId`, no datasets, and no `searchType` now searches the session cache first, matching `cognee.recall(query, session_id=...)`.

## Test plan

- [x] `uv run pytest cognee/tests/unit/api/v1/recall` — 160 passed (112 router)
- [x] `uv run pytest cognee/tests/cli_tests/cli_unit_tests cognee/tests/unit/api/recall/test_recall_search_type_default.py cognee/tests/unit/api/v1/recall/test_query_router.py cognee/tests/unit/api/test_query_required_on_recall_and_search.py` — 284 passed
- [x] `uv run pytest cognee/tests/cli_tests/cli_unit_tests/test_cli_commands.py cognee/tests/unit/api/v1/session/test_session_memory_flows.py` — pass
- Note: three tests in `test_recall_api.py` / `test_recall_tools.py` fail when the recall, session, CLI, and public-memory suites run in one pytest invocation. Reproduced identically on the base commit (`36d6c8f82`) in a scratch worktree; pre-existing ordering interaction, not from this change.

## Out of scope

Unifying this router with the LLM-based `FEELING_LUCKY` selector (`select_search_type`), which has a different default (RAG_COMPLETION) and a stale prompt. Follow-up.


NMZivkovic (commented):
Nice cleanup. The first-match rule table is much easier to reason about than the weighted scorer, skipping the router when `query_type` is explicit (and no longer logging the raw query at INFO) is a real improvement, and the bounded year pattern and the anchored, case-sensitive Cypher rule remove old false positives like "top 2000 customers" or "Match these two contracts". `cloud_client.recall` omitting `search_type` also means `serve()` users finally get the same routing as local `recall()`, and the golden table with negative invariants makes future routing changes reviewable.

What I'd sort out before merging:

1. **The conflict with dev in `recall.py` can't be resolved by picking a side** (inline).
2. **Defaulting REST `searchType` to null turns on the session short-circuit for the bundled agent prompts, and almost any shared word counts as a session hit** (inline on the router and `prompts.ts`).
3. **REST, the CLI and the frontend now auto-route raw Cypher, write clauses included**, so a request with only read permission can run `MATCH (n) DETACH DELETE n` (inline on the Cypher rule).

Smaller routing notes are inline too: a bare "exact"/"literal", year-like numbers going to `TEMPORAL`, the broadened coding-rules phrases, and the CLI mixing session rows into `-d` recalls, plus a few nits. Outside the diff, the `cognee-cli` skill still says the CLI defaults `--query-type` to `GRAPH_COMPLETION` (and lists 7 of its 10 search types); this PR updated the cognee-docker skill but not that one.

On CI, the failing Python 3.10 shard isn't from this PR. `test_forget_endpoint_not_found.py` patched `cognee.api.v1.forget.routers...` by dotted path, which 3.10 resolves through the re-exported `forget` function. Dev fixed that test in #4630 a few minutes after this run started, so merging dev (needed for the conflict anyway) clears it.

cognee/api/v1/recall/recall.py
```diff
@@ -628,19 +628,12 @@ async def _run_graph() -> list[RecallResponse]:
...
 
-                local_query_type = query_type
-                if local_query_type is not None:
-                    if auto_route:
-                        from cognee.api.v1.recall.query_router import record_override, route_query
-
-                        result = route_query(query_text)
-                        routed_type = result.search_type
-                        record_override(routed_type, local_query_type)
+                if query_type is not None:
```
NMZivkovic, line 631: This block conflicts with dev (#4929, the GLiNER backend), and neither side of the conflict is safe on its own. Dev added `elif not llm_available(llm_config): local_query_type = SearchType.CHUNKS` between the explicit and auto-route branches, and it still imports `record_override` inside the explicit branch, which this PR deletes.

- Taking this side drops the no-LLM fallback, so a keyless setup auto-routes to completion types that can't answer, and the `llm_available` import (which merges cleanly) is left unused. Nothing on dev tests that fallback, so CI won't notice.
- Taking dev's side keeps the function-local `from ... import record_override`, so every `recall(query_type=...)` with the default `auto_route=True` raises `ImportError` at call time, while importing the module still works.

The merged version should be: explicit type, then the no-LLM `CHUNKS` fallback, then `route_query`, then `HYBRID_COMPLETION`, plus a test for the no-LLM branch. `docs/recall-vs-search.md` should also mention that without an LLM an omitted type and `auto_route=False` both end up as `CHUNKS`.

cognee/api/v1/recall/routers/get_recall_router.py
```diff
@@ -22,13 +22,17 @@
...
 class RecallPayloadDTO(InDTO):
-    # Default is HYBRID_COMPLETION. Pass ``search_type: null`` explicitly
-    # to opt into auto-routing (the new ``cognee.recall`` default).
+    # Omitted or null means auto-route, matching ``cognee.recall`` and the CLI.
+    # As in the SDK, a null search_type with session_id and no datasets also
+    # lets a session hit short-circuit the graph search; pass a value to pin a
+    # strategy and disable that. See docs/recall-vs-search.md.
     search_type: SearchType | None = Field(
-        default=SearchType.HYBRID_COMPLETION,
+        default=None,
```
NMZivkovic, line 30: With `search_type` now defaulting to null, a body with only `query` and `session_id` takes the "session hit short-circuits graph" path in `recall()`. That's exactly the body the bundled agent prompts tell agents to send (comment on `prompts.ts`). And a session hit is very loose: the session search counts an entry as a hit if it shares any token of two or more characters with the query, with no stopword filtering, so "What is the deadline for project Apollo?" matches an unrelated earlier QA through "is" and "the". The graph search is then skipped, and the caller gets old QA rows instead of an answer from the graph.

The same default also changes bodies that do set datasets: `session_id` + datasets + no type now returns session entries ahead of the graph results (`source: "session"`, no `text` or `search_type`), where it used to be graph-only, and that isn't mentioned in the DTO comment, the description or the docs. A client reading `results[0].text` breaks.

Before this PR the REST default was `HYBRID_COMPLETION`, which kept both cases on the graph. I'd keep the two effects separate: auto-route when the type is omitted, but only pull in session entries when the caller asks for it (an explicit `scope`), or at least require a meaningful overlap (stopwords filtered, a minimum share of the query) before a session entry counts.

cognee-frontend/src/data/prompts.ts
```diff
@@ -63,7 +63,7 @@ Content-Type: application/json
 Body: {"query": "<user question>", "session_id": "<session-id>"}
 \`\`\`
 
-For targeted retrieval, add "search_type" to the recall body — one of: HYBRID_COMPLETION (default), GRAPH_COMPLETION, CHUNKS, GRAPH_SUMMARY_COMPLETION.
+The query is auto-routed to a search strategy by default. For targeted retrieval, add "search_type" to the recall body — one of: HYBRID_COMPLETION, GRAPH_COMPLETION, CHUNKS, GRAPH_SUMMARY_COMPLETION.
```
NMZivkovic, line 66: The documented recall body here (`query` + `session_id`, no datasets or `search_type`) is the one that now hits the session short-circuit. Since these prompts also have agents store facts as session entries, the graph effectively stops being searched after the first stored entry: "What is the shipping policy?" returns a stored "What is the refund policy?" entry through "what", "is", "the" and "policy". Adding `"datasets": ["<dataset>"]` to the documented body keeps auto-routing and runs session and graph together, or pin `"search_type": "HYBRID_COMPLETION"`. The same text is repeated in the other prompts in this file.

cognee/api/v1/recall/query_router.py
```diff
@@ -1,233 +1,112 @@
...
+
+# (rule name, pattern, search type). First match wins.
+_RULES: tuple[tuple[str, re.Pattern, SearchType], ...] = (
+    # Anchored to a leading Cypher clause keyword, case-sensitive. Relationship
+    # syntax such as ``)--(`` is not matched on its own: real Cypher always opens
+    # with a clause, and an unanchored alternative would fire mid-sentence.
     (
-        re.compile(r"(^MATCH\s|^RETURN\s|^CREATE\s|^MERGE\s|--\(|\)--)"),
+        "cypher_syntax",
+        re.compile(r"^(?:OPTIONAL\s+MATCH|MATCH|RETURN|CREATE|MERGE|UNWIND)\s"),
```
NMZivkovic, line 56: On the SDK, all-caps `MATCH`/`CREATE`/`MERGE` already routed to Cypher, but with REST, the CLI and the frontend now auto-routing by default, this rule opens raw Cypher to any request without a `searchType`. `MATCH (n) DETACH DELETE n` and `CREATE (x:Note {t: 1}) RETURN x` both route here, `CypherSearchRetriever` runs the query as written through `graph_engine.query`, and the search only checks read permission. So a read-only user, or an agent passing along injected text, can wipe or modify the graph with no Cypher opt-in in the request, unless `ALLOW_CYPHER_QUERY` is turned off (it's on by default). I'd not auto-route to `CYPHER` at all and keep it behind an explicit `searchType`, or at least only route statements with no `CREATE`/`MERGE`/`DELETE`/`SET`/`REMOVE`.

cognee/cli/commands/recall_command.py
```diff
@@ -118,17 +142,17 @@ async def run_recall():
...
                             "query_text": args.query_text,
-                            "query_type": query_type,
                             "datasets": args.datasets,
                             "top_k": args.top_k,
                             "system_prompt_path": (
                                 args.system_prompt or "answer_simple_question.txt"
                             ),
                             **session_kwargs,
                         }
+                        if args.query_type is not None:
```
NMZivkovic, line 154: Without `-t` this now omits `query_type`, and with `-s` plus `-d` that changes which sources run: `recall()` treats `session_id` + datasets + no type as "session AND graph, both contribute", where the old pinned `HYBRID_COMPLETION` kept it graph-only. `cognee-cli recall "What is the deadline?" -s s1 -d proj` now prints old session QA rows first as raw reprs, labelled `auto`, and the graph answer as "Result 2", because the resolved type is read from `results[0]` and session entries don't carry one. The `--api-url` path in `api_dispatch.py` changed the same way, and the `--session-id` help still says that with `-d` or `-t` it only adds session history to the search context. Passing `scope=["graph"]` when `-d` is given without `-t`, and reading the type from the first result that has one, would keep the old output.

cognee/api/v1/recall/query_router.py
```diff
@@ -1,233 +1,112 @@
...
-# Scores accumulate per search type; highest total wins.
-
-_RULES: list[tuple[re.Pattern, SearchType, float]] = [
-    # --- Cypher: raw query syntax (high confidence) ---
+    search_type: SearchType
+    rule: str
+
+
+# Years 1500-2099. Excludes ticket numbers, ports, and counts like "1000 users".
+_YEAR = r"(?:1[5-9]|20)\d{2}"
```
NMZivkovic, line 35: Any number from 1500 to 2099 counts as a year here, so counts and prices route to `TEMPORAL`: "Apartments between 1500 and 2000 sq ft", "Which laptops cost 1500-2000 dollars?" and "Show orders by 2000 customers" all do, despite the comment above about excluding counts. And for real years on a default graph (`temporal_cognify=False`), `TemporalRetriever` still makes its LLM time-extraction call and falls back to triplets without document passages, which is the reason the doc gives for not routing bare "when"/"after". I'd only route to `TEMPORAL` when the graph actually has temporal events, or keep just the unambiguous forms (between YEAR and YEAR, decades, ISO dates) and reject a following unit word.

cognee/api/v1/recall/query_router.py
```diff
@@ -1,233 +1,112 @@
...
-    (
+        "quoted_phrase",
         re.compile(r'^"[^"]+"$'),
         SearchType.CHUNKS_LEXICAL,
-        8.0,
     ),
     (
-        re.compile(r"\b(exact|verbatim|literal|word.for.word)\b", re.IGNORECASE),
+        "exact_match_intent",
+        re.compile(r"\b(?:exact|verbatim|literal|word.for.word)\b", re.IGNORECASE),
```
NMZivkovic, line 66: A bare "exact" or "literal" anywhere in the question wins before every later rule, and `CHUNKS_LEXICAL` returns BM25-ranked chunk payloads with no LLM answer. "What was the exact revenue in 2024?", "What is the exact deadline for the filing?" and "don't summarize, just tell me the exact number" all route here. The old SDK router did the same for simple cases, but REST and CLI callers who used to get a HYBRID answer now get raw chunks too, and the number may not be in the top ones. I'd require explicit lexical intent (`exact phrase/wording/quote/match`, `verbatim`, `word for word`) and add these questions to the negative invariants.

cognee/api/v1/recall/query_router.py
```diff
@@ -1,233 +1,112 @@
...
-        5.0,
+        "explicit_time_range",
+        re.compile(_TEMPORAL_PATTERNS, re.IGNORECASE),
+        SearchType.TEMPORAL,
     ),
     (
+        "coding_rules_intent",
         re.compile(
-            r"\b(what (connects|links|ties)|path between|degree of separation)\b",
+            r"\b(?:coding (?:rules?|standards?|conventions?|guidelines?)"
```
NMZivkovic, line 85: `CODING_RULES` reads only the `coding_agent_rules` node set, which no core pipeline writes (only the memify coding-rules example does), so on an ordinary graph it returns nothing, as `docs/recall-vs-search.md` itself notes. Broadening the phrases to coding standards/conventions/guidelines and the code review checklist means a team that ingested `CONTRIBUTING.md` and asks "Where are our coding standards?" now gets no results, where `HYBRID_COMPLETION` would answer from the document, and REST and the CLI now auto-route by default. I'd route here only when that node set exists (or fall back to HYBRID on an empty result), or keep the narrower pre-PR phrases.

cognee/api/v1/recall/skill_gate.py
```diff
@@ -24,6 +23,18 @@
 # side-channel next to the main answer, not the answer itself.
 DEFAULT_SKILL_GATE_TOP_K = 3
 
+# Suppress skill matches preceded by a negation within this many characters.
+_NEGATION = re.compile(r"\b(not|n't|no|never|without|lack)\b", re.IGNORECASE)
```
NMZivkovic, line 27: Nit, carried over from the router: `n't` can never match here, because in "don't" there's no word boundary between the "o" and the "n", so contractions are never treated as negation. `(?:\b(?:not|no|never|without|lack)\b|n't\b)` would fix it, or just drop `n't`.

docs/recall-vs-search.md
```diff
@@ -0,0 +1,88 @@
...
+
+`recall()` is the memory API's read path. It wraps `search()` and adds three
+things:
+
+1. **Query routing.** When `query_type` is omitted, a rule-based classifier
+   picks the search strategy. No LLM call is involved, so routing is free.
+2. **Session memory as a source.** `scope` selects `graph`, `session`,
+   `trace`, `session_context`, or a list of them. With a bare `session_id`, a
+   session hit short-circuits the graph search.
+3. **Normalized results.** Every entry carries `_source` (`"graph"`,
```
NMZivkovic, line 11: Nit: the SDK objects and the REST JSON use `source`, not `_source` (`RecallResponse` is discriminated on `source`), so a client written from this doc gets nothing back. The new `TestRecallDispatch` fixture uses `"_source": "graph"` too, a shape the server never sends, which hides that the CLI's existing `results[0].get("_source") == "session"` checks never match.

cognee/tests/unit/api/v1/recall/test_query_router.py
```diff
@@ -0,0 +1,161 @@
...
+]
+
+
+@pytest.mark.parametrize(("query", "expected"), GOLDEN, ids=[q for q, _ in GOLDEN])
+def test_golden_routes(query, expected):
+    assert route_query(query).search_type == expected
+
+
+@pytest.mark.parametrize(("query", "_"), GOLDEN, ids=[q for q, _ in GOLDEN])
+def test_router_only_picks_routable_types(query, _):
```
NMZivkovic, line 83: Nit: this is parametrized over `GOLDEN`, whose expected types are all routable already, so it can only fail when `test_golden_routes` fails for the same row. A new rule that sends a query outside the golden table to `GRAPH_COMPLETION_COT` would pass both. Asserting on the rule table itself (every rule's search type plus the default is in `ROUTABLE_TYPES`) would actually guard the invariant.

cognee/cli/commands/recall_command.py
```diff
@@ -43,7 +62,10 @@ def configure_parser(self, parser: argparse.ArgumentParser) -> None:
             "-t",
             choices=SEARCH_TYPE_CHOICES,
             default=None,
-            help="Search mode (default: HYBRID_COMPLETION)",
+            help=(
+                "Search mode. Omit to auto-route the query "
```
NMZivkovic, line 66: Nit: auto-routing can print `GRAPH_SUMMARY_COMPLETION`, `TEMPORAL` or `CODING_RULES`, but none of them is in `SEARCH_TYPE_CHOICES`, so a user who tries to pin what they just saw with `-t` gets "invalid choice". They're also missing from `COMPLETION_SEARCH_TYPES`, so their answers print as "Result i:" instead of "Response:".

lxobr (dismissed):

PR #4891, opened 2026-09-03
Expose similarity score from ChunksRetriever.get_completion_from_context

GrowDev1:
## Summary

`ChunksRetriever` already computes a similarity score for each result on `ScoredResult`, but `get_completion_from_context` never returns it in the chunk payload — so a caller doing hybrid/fusion ranking across retrievers has no way to see it.

This spreads `found_chunk.score` into the returned chunk dict, matching the existing idiom already used in `temporal_retriever.py`, and matching what `normalize_search_payload.py`'s `_score_from()` already expects to find when present.

`completion_retriever.py` is intentionally left untouched — it returns LLM-generated text, not chunk dicts, so injecting a score there would corrupt the prompt rather than add metadata.

## Test plan

- [x] Added `test_get_completion_from_context_includes_score`
- [x] Added `test_get_completion_from_context_empty_returns_empty_list`
- [x] Existing `chunks_retriever_test.py` suite passes (12/12)
- [x] `ruff check` / `ruff format --check` clean
- [x] Full retrieval test directory run — no new failures introduced (pre-existing unrelated failures confirmed via `git stash`/re-run: missing LLM key and no live Postgres in this environment)

### DIFF
--- a/cognee/modules/retrieval/chunks_retriever.py
+++ b/cognee/modules/retrieval/chunks_retriever.py
@@ -66,7 +66,10 @@ class ChunksRetriever(BaseRetriever):
         """
         # TODO: Do we want to generate a completion using LLM here?
         if retrieved_objects:
-            chunk_payloads = [found_chunk.payload for found_chunk in retrieved_objects]
+            chunk_payloads = [
+                {**found_chunk.payload, "score": found_chunk.score}
+                for found_chunk in retrieved_objects
+            ]
             return chunk_payloads
         else:
             return []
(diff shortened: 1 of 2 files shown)


GrowDev1:
Following up on CI status here — the failing checks all appear to share one root cause, not 170 separate issues: GitHub withholds repository secrets (LLM/embedding provider credentials) from workflow runs triggered by a pull request from a fork, which is how this PR is submitted. Each failing job's annotation confirms this directly, e.g.:

> Error when evaluating 'secrets'. Secret LLM_PROVIDER is required, but not provided while calling...

The one workflow that doesn't require secrets (`Community Tests (No Secrets)`) already passes cleanly (2391 passed, 38 skipped).

Would a maintainer be able to approve/re-run the secret-gated workflows for this PR? Happy to make any code changes needed in the meantime — just flagging that the current failures don't look like something fixable from the PR side, since they're gated on repo-level secret access rather than anything in the diff.

Vasilije1990:
Fixed this up for you a bit and merged

PR #4897, opened 2026-09-03
chore(ci): Move nightly-only tests to release tests (SDK-533)

dexters1:
Review follow-up for #4876, addressing the two review comments there plus the follow-up asks to move the search-DB sweep and the Ollama/llama-cpp suites as well.

## What moves

Into `release_test.yml` (runs on PRs to `main` and on `workflow_dispatch`):

- **`unit-matrix-mid`** (3.11/3.12/3.13 on ubuntu) — out of `nightly_tests.yml`. Same reusable workflow and extras as before.
- **`search-db-versions`** (3.10/3.12/3.13 sweep of the four search backend combos) — out of `nightly_tests.yml`.
- **`ollama-tests`** and **`llamacpp-tests`** — out of `nightly_tests.yml`. With these gone the nightly is the perf arms only, which is what its Slack report and the MotherDuck load are built around.
- **`run-telemetry-pipeline-test`** — out of `nightly_tests.yml`. This is the only CI job that runs the pipeline with telemetry on; everything else disables it.
- **`test-deletion-on-default-graph`**, **`test-deletion-on-custom-graph`**, **`run_agent_registry_tests`** — restored from the `e2e_tests.yml` jobs the parent PR deleted outright, until it is confirmed their coverage exists elsewhere.

The moved jobs carry the same `if:` guard as the rest of the file (fork PRs skipped). They install from the checked-out source via `cognee_setup`, so the `cognee_version` input does not apply to them; noted in the block comment.

## Nightly cleanup

- `nightly_tests.yml`: the five jobs removed, plus their `needs:` entries, the status-gate lines, and the `STATUS_OLLAMA` / `STATUS_LLAMACPP` / `STATUS_UNIT_MID` env. What remains is the perf arms and the notify job.
- `build_nightly_slack_blocks.py`: the header's suite-status line is gone; the header is now emoji, summary, ran-at, branch, cadence, and commit.
- Comments in `nightly_scheduler.yml` and `test_suites.yml` that pointed at the nightly for these suites now point at the release tests.

## Verification

- YAML parses; pre-commit (yaml check, ruff, ruff format) passes on all changed files.
- Slack builder smoke-tested with the reduced header.
- `release_test.yml` dispatched manually on this branch: run 33766754503 went green on attempt 2 (47 passed, 2 skipped by design); every moved job passed on the first attempt. Details in the PR comment.

### DIFF
--- a/.github/scripts/build_nightly_slack_blocks.py
+++ b/.github/scripts/build_nightly_slack_blocks.py
@@ -13,8 +13,8 @@ Input (env):
                `metrics` output; empty/unparseable is treated as `{}`, which is
                what a failed job produces. A `tenant_create` key adds the
                cloud-only tenant row.
-  STATUS_*     header fields (emoji, summary, ran_at, ollama, llamacpp,
-               branch, cadence, sha — the last three fall back if unset).
+  STATUS_*     header fields (emoji, summary, ran_at, branch, cadence,
+               sha — the last three fall back if unset).
   RUN_URL      link target for the footer.
 
 Output: `blocks=<compact JSON>` appended to $GITHUB_OUTPUT (stdout if unset).
@@ -98,8 +98,6 @@ def main():
         f"{env.get('STATUS_SUMMARY', '')}\n"
         f"*Ran at:* `{env.get('STATUS_RAN_AT', '')}`  •  "
         f"*Branch:* `{branch}` (`{cadence}`)  •  *Commit:* `{sha}`\n"
-        f"*Ollama:* `{env.get('STATUS_OLLAMA', '')}`  •  "
-        f"*Llama-cpp:* `{env.get('STATUS_LLAMACPP', '')}`\n"
     )
 
     blocks = [header]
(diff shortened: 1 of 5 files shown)


dexters1:
Manual `release_test.yml` dispatch on `fed38f728`: https://github.com/topoteretes/cognee/actions/runs/33766754503

Green on attempt 2 — 47 jobs passed, 2 skipped by design (Load Tests, Dev Canary Release). Every moved job passed on the first attempt: Ollama, llama-cpp, all 12 search-DB sweep legs, the 3.11–3.13 unit matrix (unit shards plus build/library/soft-delete/custom-graph-delete legs), telemetry pipeline, both delete tests, and agent registry.

The single attempt-1 failure was the 3.12 build test's `pip install *.whl` getting an empty PyPI index response for aiohttp (`from versions: none`); same step passed on 3.11 and 3.13 in the same run, and the re-run passed.

NMZivkovic (approved):

PR #4911, opened 2026-09-04
fix(serve): Proxy update() to remote instance and carry pinned data ids

Vasilije1990:
## Summary

After `cognee.serve(url=..., api_key=...)`, `update()` ran against the **local** store: it resolved the local default user, ran a local `datasets.delete_data` against the remote dataset UUID, and failed with `PipelineRunFailedError: Dataset '<uuid>' not found` while the remote document stayed untouched. `datasets.list_data()` had the same gap. This PR closes both, plus the two related `serve()` gaps from the report.

### 1. `update()` and `list_data()` proxy to the remote
- `CloudClient.update()` sends `PATCH /api/v1/update` exactly as the route expects: `data_id` / dataset / `chunk_level_diff` as query params, the new content as the multipart `data` file, `node_set` as repeated form fields. It is an in-place replace — never a local delete + remote add, which would mint a new id.
- `CloudClient.list_data()` wraps `GET /api/v1/datasets/{id}/data`; `datasets.list_data()` routes through it and returns attribute-accessible rows.
- The remote branch runs **before** any local work in `update()`; parameters the route has no slot for (`graph_model`, `custom_prompt`, store configs, …) log a warning instead of being dropped silently.

### 2. `DataItem.data_id` reaches the server
- New `data_ids` form field on `POST /api/v1/add` and `POST /api/v1/remember`, same positional JSON-array convention as `labels` / `external_metadata` (`null` = let the server mint). Ingestion already resolves pins inside the target dataset and refuses foreign ones, so this is wiring only.
- `CloudClient.add()` / `remember()` previously dropped `DataItem` wrappers entirely. They now unwrap the payload and send `data_ids` / `labels` / `external_metadata`, plus `node_set` and `datasetId` (also dropped before). Local `/abs/path` and `file://` strings upload the file bytes, since the server can't see the caller's filesystem.
- **Old-server safety:** FastAPI ignores unknown form fields, so a server without `data_ids` would silently mint its own id. The client compares the ids the response reports against the pins and raises with an actionable message on a mismatch; a response with no ids (background run) logs a warning.

### 3. `dataset_name` on `update()`
- SDK `update()` and `PATCH /api/v1/update` accept `dataset_name` as an alternative to `dataset_id` (exactly one required; 422 on the route otherwise), matching every other entry point. Locally a name resolves only among datasets the caller may **write** to and never creates one. `dataset_id` stays optional-with-default so existing positional callers are unchanged.

## Test plan
- [x] `cognee/tests/unit/api/v1/serve/test_update_remote_routing.py` — wire format, upload helper, remote mode does zero local work, `dataset_name` forwarding and local resolution
- [x] `cognee/tests/unit/api/v1/serve/test_cloud_client_pinned_ids.py` — `data_ids`/labels/metadata/node_set/datasetId on add+remember, path uploads, pin-mismatch error and no-ids warning
- [x] `cognee/tests/unit/api/v1/test_add_remember_data_ids.py` — route parsing/pairing, 400s on malformed/mismatched ids and on pins with `session_id`/`content_type`, update route `dataset_name` + exactly-one check
- [x] All unit tests touching the changed modules: 603 passed, 7 skipped; `pre-commit run` clean
- [ ] Not exercised against a live remote instance from this environment

### DIFF
--- a/.github/workflows/e2e_tests.yml
+++ b/.github/workflows/e2e_tests.yml
@@ -88,6 +88,28 @@ jobs:
       - name: Run Migration/Model Lockstep Guard
         run: uv run pytest cognee/tests/e2e/migrations/test_migration_model_lockstep.py -v
 
+  run-serve-remote-routing-e2e:
+    name: serve() remote routing e2e
+    runs-on: ubuntu-22.04
+    container: ${{ inputs.ci-image != '' && fromJSON(format('{{"image":"{0}","credentials":{{"username":"{1}","password":"{2}"}}}}', inputs.ci-image, github.actor, github.token)) || null }}
+    steps:
+      - name: Check out
+        uses: actions/checkout@d23441a48e516b6c34aea4fa41551a30e30af803  # v6.1.0
+
+      - name: Cognee Setup
+        uses: ./.github/actions/cognee_setup
+        with:
+          python-version: '3.11.x'
+
+      # Real API server in a subprocess + the SDK after cognee.serve(url=...):
+      # remember → datasets.list_data → update (chunk-level and full paths, by
+      # id and by name) must all land on the server, never the SDK's local
+      # store. Mocked LLM and embeddings: no secrets.
+      - name: Run serve() remote routing suite
+        env:
+          ENV: 'dev'
+        run: uv run pytest cognee/tests/e2e/serve/ -v
+
   run-telemetry-test:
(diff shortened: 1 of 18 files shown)


dexters1:
Closing PR, issue will be resolved here:

https://github.com/topoteretes/cognee/pull/4916

Update was missing from serve

PR #4920, opened 2026-09-04
feat: Make improve() an explicit orchestrator (SDK-416)

Vasilije1990:
## Description

`improve()` had grown into nine steps written at different times, chained by `if` statements. Stages ran when they could not do anything (feedback weights with `DEFAULT_FEEDBACK_INFLUENCE=0`, triplet enrichment with `triplet_embedding=False`), nobody could see which stages ran (the return value was stage 8's memify output, a held lock returned `{}`), three stages redid all their LLM work on every call, four stages received the dataset by name although the top of the function resolves it to an id precisely to avoid the shared-dataset retargeting hazard, and `run_in_background` was forwarded into individual pipelines so the lock could be released while writes were in flight.

This PR makes `improve()` an explicit orchestrator over an ordered registry of the same nine stages. **No stage changes how it computes anything**: the weight formula, learning rate and truth-factor bounds are frozen. What changes is the glue: a stage declines work it cannot do, skips work it already did, and reports which of the two happened.

Design: the goal state reviewed with Lazar (Linear SDK-416, SDK-157; supersedes the unmerged #4408 / SDK-448). The plan and audit are published as a Claude artifact and kept out of the repo, as with #4408.

### What changes

**New package `cognee/modules/improve/`** (every type is improve-local; nothing is added beside an existing pipeline-layer concept)

- `ImproveStage`: `name`, `kind` (`session` | `graph`), `fatal`, `gate(inputs) -> skip reason | None`, `run(inputs) -> StageResult`. A stage is a gate, a call into the existing helper, and a result mapping. No retries or ordering inside stages.
- `StageResult` / `ImproveResult`: statuses reuse `PipelineRunInfo`'s vocabulary (`completed`, `already_completed`, `errored`) plus `skipped` with a required reason. A pipeline-backed stage carries its `PipelineRunInfo` in `result.run` and takes its status from it. `ImproveResult.memify_run` nests the legacy memify return for one release.
- `ImproveRunInputs`: one frozen bundle (user, resolved dataset id and row, session ids, config, capabilities) handed to every stage. No `run_in_background` field; the runner owns background mode. `OperationContext` stays the context.
- `GraphCapabilities`: per-adapter probe for `feedback_weights` / `truth_state` (and `supports_incremental_chunk_updates` when an adapter declares it), resolved once per run. Stages 1 and 7 now gate with `backend_unsupported` instead of failing into a swallowed warning.
- `ImproveConfig` (env prefix `IMPROVE_`): only the loop's own knobs — `IMPROVE_AUTO_ENABLED`, `IMPROVE_DEBOUNCE_ENTRIES`, `IMPROVE_DEBOUNCE_SECONDS`, `IMPROVE_STAGES_DISABLED`, `IMPROVE_FEEDBACK_ALPHA`. Shared knobs stay where they live.
- `DEFAULT_STAGES` registry with `after=` dependencies validated at import and pinned by a test (4 → 5 → 7 → 8). `operations_catalog` generates its improve rows from it. `memify_task_registry` is untouched (it serves the public `memify(tasks=[...])` API).
- `graph_changes.py`: stage 8 reports `already_completed` when `pipeline_runs` shows no completed write pipeline for the dataset since the last improve. It never counts the graph.

**`cognee/api/v1/improve/improve.py`** is now the orchestrator: same public signature; dataset resolved once and only the id/row passed on; one lock claim keyed to the run (every session id given, or `dataset:<id>` when none) held across the whole chain including background mode; a lost claim returns every stage `skipped: lock_held`; `persist_session_qa` is the only `fatal=True` stage (a failure stops the chain and re-raises with the partial result on `exc.improve_result`); remote passthrough forwards `session_ids`, `build_*` and `run_in_background` (it silently dropped them); PostHog telemetry hashes session ids.

**Surfaces** return the same `ImproveResult` with the same options: router (`response_model`, `build_truth_subspace`, `feedback_alpha`, `run_in_background`), CLI (`--build-global-context-index`, `--build-truth-subspace`, one line per stage), MCP `improve` (`node_name`, `build_*`) and MCP `remember` (`self_improvement`, revives #4246). The advertised "sync graph back into session" stage that never existed is removed from the router, MCP and `remember()` docstrings.

**`remember()`** attaches `RememberResult.improve` / `.improve_error`; an improve failure after a successful cognify no longer flips the remember to `errored`. The session-path auto-improve is debounced (`cognee/api/v1/remember/auto_improve_debounce.py`, state kept in the same internal session-context row pattern the persist watermark uses). New `cognee/infrastructure/background_tasks.py` registry with `cognee.wait_for_background_tasks()`; the API server drains it on shutdown.

**Stage fixes, all bookkeeping (no formula changes)**

- Feedback weights: per-QA applied element ids and an attempt cap in the existing `memify_metadata`; deleted ids are pruned and the row is marked processed, so a missing node no longer re-applies the same feedback on every run. Rows with no `used_graph_element_ids` are marked processed. The implicit `referenced_qa_rating` is used with half alpha when no explicit score exists; `feedback_text` reaches extraction and the distillation prompt.
- One state-row watermark helper (generalized from `session_persist_watermark.py`) now serves Q&A persistence, agent-context extraction, **agent-trace persistence** (was `last_n_steps=None`, re-cognifying the whole trace history each run) and **distillation** (was a full curator/writer LLM fan-out each run). Lesson titles no longer carry the run date so re-accepted lessons dedup at `add()`.
- Distillation gate uses clamped net helpfulness instead of `harmful_count == 0`.
- Truth build: no `[0.0]*k` coords persisted for a failed embedding batch; chunk truth states are written at epoch N+1 before the centroids so a mid-build failure leaves epoch N live; capability is probed before any embedding call.
- The concurrent and sequential turn paths now write the same QA row (`context`).
- The per-turn rating prompt section is appended whenever `AUTO_FEEDBACK` is on (not only with personalization); per-tool-call LLM trace summaries are gated on `AUTO_FEEDBACK` and default off.

**Removed**: frequency weights end to end (pipeline, task, interface stubs, registry entry, catalog row, `session.add_frequency_weights`; no adapter implemented them and nothing read them), `get_session_memify_tasks` (no callers), the `try/except Exception → False` catch-alls in `add_feedback` / `delete_feedback` (SDK-341; `False` now means only not-found or caching-disabled), the stale `pr_body.md`. `persist_agent_trace_feedbacks` now enters `set_database_global_context_variables` with `async with` like its siblings. Duplicated literals (node-set names, alpha default, confidence gate, `is_auto_feedback_enabled`) collapse onto `cognee/modules/improve/constants.py` and the shared predicate.

### Verification

- `pre-commit run --all-files`: clean.
- `pytest cognee/tests/unit` (excluding `eval_framework` and `neo4j_deadlock_test.py`, which need extras not installed locally): **5884 passed, 38 skipped, 0 failed**.
- `pytest cognee/tests/cli_tests`: 237 passed.
- `cognee-mcp/tests` (main venv with `fastmcp<4`, excluding the benchmark): 85 passed.
- Session/memify integration suites (`test_session_persistence_memify_integration.py`, `test_feedback_weights_memify_integration.py`, session manager fs/sql/redis, concurrent and sequential turn flows, session SDK, user_preferences full loop): **113 passed** plus 6 passed in the feedback-weights file. One integration test (`test_feedback_weights_mixed_success_keeps_false`) encoded the old compounding behaviour and was rewritten to the new contract (`..._prunes_and_applies_once`). The two in-memory Redis fakes gained `llen`, which the real client has and the watermark path now calls.

### Deliberately not in this PR

- Learned weights surviving a full re-cognify on nodes (B5), atomic counter increments across cache adapters (B7), the ablation harness and any default flip of `DEFAULT_FEEDBACK_INFLUENCE` (Phase D), lesson lifecycle, truth-subspace keep-or-trim.
- Frequency weights are deleted rather than implemented (open PR #3466 does the latter; decision recorded in the plan).
- Linear/GitHub hygiene from the plan (mark SDK-448 as not merged, close #3466, delete `feature/distill_mem`).

### Notes for reviewers

- `improve()` now returns `ImproveResult`, not a dict. Callers that indexed the memify dict find it under `result.memify_run`. `stages_run` is gone; `result.stage_summary()` is the tracing value.
- Relates to #4874: its `supports_incremental_chunk_updates` flag is read by `GraphCapabilities` when present. Its incremental delete path does not invalidate session turns (separate finding, raised on that PR).


Vasilije1990:
CI triage for the first run:

- **Unit tests 3.10.x (3 shards)** — caused by this PR, fixed in c6f08f3e1. `test_auto_feedback_gate.py` patched `cognee.infrastructure.session.get_session_manager.get_session_manager` by dotted string; the `session` package re-exports the function under the same name as its submodule, and Python 3.10's `mock.patch` resolves the path by getattr and lands on the function. The test now patches the module object from `importlib.import_module`.
- **Test permissions / Turso Multi-User** (both run `test_permissions.py::test_improve_permission_matrix`) — caused by this PR, fixed in c6f08f3e1. The test asserted memify is awaited once; the triplet-enrichment stage is now gated on `triplet_embedding` (off by default) and reports `skipped`. The test opens the gate with a patched cognify config so its "the stage receives the authorized dataset" assertion still holds.
- **Code Quality** — not this PR. `uv run ruff check .` resolves ruff 0.16.0 from the refreshed `uv.lock` (11f7a1b67, today) and reports ~11k pre-existing findings repo-wide (FURB/TRY/EXE/DTZ…). Every branch that ran today fails the same way (e.g. `fix-update-serve-routing`: 10905 errors). Pre-commit's pinned ruff passes. Needs a separate fix on dev (pin ruff in `uv.lock` or align the job with pre-commit).

cognee/modules/improve/capabilities.py
```diff
@@ -0,0 +1,108 @@
...
+        )
+
+
+def _overrides_interface(engine: Any, method_name: str) -> bool:
+    """True when ``engine``'s class implements ``method_name`` beyond the interface stub."""
+    from cognee.infrastructure.databases.graph.graph_db_interface import GraphDBInterface
+
+    interface_method = getattr(GraphDBInterface, method_name, None)
+    if isinstance(engine, GraphDBInterface):
+        adapter_method = getattr(type(engine), method_name, None)
```
NMZivkovic, line 57: This probe returns False for every real graph engine. `get_graph_engine()` always hands back a `_GraphEngineHandle`, and the cached value inside it is a second proxy (`_LeasedValueProxy`). Both forward `__class__`, so the `isinstance(engine, GraphDBInterface)` check passes, but `type(engine)` is still the handle class, which doesn't define `set_node_truth_state` or the feedback-weight methods.

On Ladybug that means `feedback_weights` and `build_truth_subspace` are skipped as `backend_unsupported` (a direct `build_truth_subspace()` call returns `skipped` too, and logs "graph adapter _GraphEngineHandle has no truth-state support"), and Neo4j loses `feedback_weights` the same way. Both ran before this PR. `getattr(engine.__class__, method_name, None)` resolves through both proxies to the real adapter class. `test_capabilities.py` patches `get_graph_engine` to return a plain adapter, which is why it passes. A test that goes through the real handle would catch this.

Smaller, while you're here: `resolve_graph_capabilities` runs on every improve before the lock claim, so every run, lock losers included, leases the graph engine and (with backend access control on, the default) takes a dataset-queue slot and does the dataset and owner lookups, even though only stages 1 and 7 read the result. Resolving it lazily inside those two gates would avoid that.

cognee/infrastructure/locks/session_lock.py
```diff
@@ -135,3 +135,47 @@ async def release_improve_lock(session_id: str) -> None:
...
+def improve_lock_keys(session_ids: Iterable[str] | None, dataset_id: Any) -> tuple[str, ...]:
+    """The claim keys for one improve run: its session ids plus its dataset id.
+
+    Every run claims the dataset key, session-fed or not, so a session-keyed
+    bridge run and a dataset-keyed run over the same dataset exclude each
+    other — improves for one dataset serialize; an overlapping claim loses and
+    returns ``lock_held``.
+    """
+    sessions = tuple(session_id for session_id in (session_ids or ()) if session_id)
+    return (*sessions, f"dataset:{dataset_id}")
```
NMZivkovic, line 181: Every run now claims `dataset:<id>` on top of its session ids, in one process-wide set. Two things follow:

- Session state is scoped per `(user_id, session_id)`, but the keys are raw session ids, so two users who both have a session called `chat` still block each other, even on their own datasets.
- Two `remember(session_id=...)` bridges for different sessions on the same dataset now exclude each other, and so do a bridge and the sessionless improve at the end of `remember()`. With the default debounce every session remember fires a bridge, and it holds the dataset key through Q&A cognify, agent-context extraction and distillation, so any other session writing to that dataset in that window loses.

The loser gets every stage `skipped: lock_held`: nothing is persisted until that session's next winning bridge, nothing retries, and remember still logs "bridged to permanent graph" (with `IMPROVE_DEBOUNCE_*` set, the loser has also used up its debounce window). Before this PR only single-session runs locked. Adding the user id to session keys fixes the first case. For the dataset key, waiting for the claim (or retrying the loser) instead of skipping would keep runs serialized without dropping bridges.

cognee/modules/improve/graph_changes.py
```diff
@@ -0,0 +1,96 @@
...
+        last_improve_conditions = [
+            PipelineRun.dataset_id == dataset_id,
+            PipelineRun.operation_name == IMPROVE_OPERATION_NAME,
+            # "succeeded" excludes failed runs (their work may not have
+            # happened) and "noop" rows (a lost lock claim, an all-skipped
+            # run — nothing ran, so nothing to watermark). Accepted gap: a
+            # session run whose enrichment stage was skipped by config still
+            # records "succeeded", so after enabling triplet_embedding the
+            # backlog stays gated until the dataset's next write pipeline.
+            PipelineRun.outcome == "succeeded",
```
NMZivkovic, line 63: The honest outcomes help a lot here: errored runs are `failed`, lock-held and all-skipped runs are `noop`, and a background run closes its row at the end. A few ways this can still mark enrichment done when it didn't happen:

- A run where other stages ran but stage 8 didn't still records `succeeded`, because only an all-skipped run becomes `noop`. The `remember(session_id=...)` bridge on the default config is one (stage 2 runs, stage 8 is skipped as `triplet_embedding_disabled`), and so is a session or `build_global_context_index` run with `triplet_enrichment` in `IMPROVE_STAGES_DISABLED`. After `TRIPLET_EMBEDDING` is turned on, a plain `improve()` reports `already_completed` until the next write pipeline. The comment above calls this an accepted gap, but nothing tells the user why nothing was embedded.
- The row is written after stage 9, and stages 8 and 9 take the dataset lock separately. With `build_global_context_index=True`, a `cognify` queued behind stage 8 gets the lock before stage 9, and its write counts as already enriched.
- `WRITE_PIPELINE_NAMES` misses `custom_pipeline` (the default for `cognee.run_custom_pipeline`) and `presort_graph_pipeline`. `add_data_points` doesn't embed triplets by default, so edges written that way never get embedded.
- It also misses in the other direction: stages 2 and 3 log a completed `memify_pipeline` row even when their watermarks find nothing new, so an improve with `session_ids` never reports `already_completed`, and with `TRIPLET_EMBEDDING=true` a repeat session improve re-embeds every triplet in the dataset.

Recording whether stage 8 actually completed (and using that run's start time), and counting pipeline runs that really wrote to the graph rather than a list of names, would cover these.

NMZivkovic (commented):
Thanks, this is a big step for `improve()`. Explicit stages that report what they did, the dataset id passed down instead of the name, and one place that owns the lock and background mode all make this much easier to reason about. I reviewed it at `8a6193ff`, so Lazar's hardening pass is included: background improves registering for the shutdown drain, the CLI waiting for `--background`, `failed`/`noop` operation rows, errored truth builds reporting `errored`, and the stage-name validation all look right.

The ones I'd fix before merging, roughly by impact:

1. **The capability probe turns off feedback weights and truth-subspace builds on real backends** (capabilities.py).
2. **Every run now claims the dataset lock, so concurrent session bridges on the same dataset skip each other and drop work** (session_lock.py).
3. **Stage 8's "no writes since the last improve" check can still skip enrichment that never ran** (graph_changes.py).
4. **Distillation marks entries as done when its LLM calls failed** (distill.py).
5. **`CACHING`/`AUTO_FEEDBACK` are frozen at import time**, which breaks the `cognee-cli demo` pin and the tests that turn auto-feedback off (session_manager.py).
6. **`import cognee` fails on pydantic-settings older than 2.7** (config.py).

Also inline: trace persistence (dataset scope and capped windows), trace summaries that are now empty, feedback-weight bookkeeping across datasets, the `serve()` passthrough, the shutdown drain vs. container kill timeouts, the feedback gate, truth-build epoch consistency, and feedback text in logs.

Smaller things:
- The 1-5 rating question is now appended on every analyzed turn whenever `AUTO_FEEDBACK` is on (about 18% more system prompt), while CLAUDE.md still lists it among what `PERSONALIZATION_ENABLED=false` turns off, and `base_config.py` and `.env.template` still say that flag keeps a default deployment byte-identical.
- Session ids are hashed in the `cognee.remember`/`cognee.improve` telemetry events but still sent raw by `cognee.session.add_qa`, `cognee.recall` and `POST /v1/search`, so "only a fingerprint leaves the process" doesn't hold yet. Adding `session_id` to `_sanitize_nested_properties` would cover every event and replace both `_hash_session_id` copies.
- A typo in `IMPROVE_STAGES_DISABLED` now fails every `improve()` at call time, so auto-improve and session bridging stop with only an `improve_error` and a log line, and `POST /improve` hides the reason behind a generic 409. Validating it once at startup (or in `ImproveConfig`) would surface it.
- One invalid `IMPROVE_*` value (say `IMPROVE_FEEDBACK_ALPHA=0` or a negative debounce) makes every `remember()` raise a `ValidationError` before the data is ingested or the session entry written: `auto_improve_enabled()` builds `ImproveConfig` outside any `try`, and `lru_cache` doesn't cache the exception.
- With `--api-url --background`, the CLI prints "Improvement started in background!" for any result that has stages, so a run skipped with `lock_held` is reported as started. The local path already checks `status == "running"`.
- `ExtractAgentContextStage` can't report errors: `extract_pending_agent_context` catches everything and returns `[]`, so its `errored` branch never fires, and its test only passes because it mocks the extractor to raise.
- Trace persistence now loads the whole trace for its snapshot on every non-idle run, so periodic persistence with `persist_session_trace_after=N` reads every stored step at each trigger instead of N.
- The new MCP `improve` only has `@log_usage`, not `@registry.tool`, so no MCP client can reach it, although the description lists it as a surface.
- The router documents 409 when `persist_session_qa` fails, but the usual failure (a task error that memify reports as `PipelineRunErrored`) is raised as `CogneeSystemError` and reaches clients as a 500. Only a non-Cognee exception gets the 409, with a generic body, and neither response carries the partial `ImproveResult`.
- CLAUDE.md says `already_completed` means "nothing new since the stage's watermark", but only `triplet_enrichment` returns it. `persist_session_qa`, `persist_agent_traces`, `extract_agent_context` and `distill_sessions` report `completed` when there is nothing new.
- `memify_metadata` now holds lists and ints, which the pre-PR model rejects, so during a rolling deploy (or after a rollback) old pods fail to read any session containing a rated QA row the new feedback-weight task has processed. Today that's sessions touched by `apply_feedback_weights_pipeline`; once stage 1 works it's every improved session with ratings.
- The distillation gate now uses net helpfulness while `update_user_preferences` still requires `harmful_count == 0` through `is_context_entry_usable`, and the new comments in `session_context_models.py` and `session_distillation/models.py` (plus that function's docstring) still say distilled entries must never have been rated harmful and go through the same check.
- The distillation watermark only helps when nothing is new. With a single new gated entry, `propose_lessons` still gets every gated entry and the full QA history, so the curator/writer fan-out the description says is gone runs again and re-proposes lessons that were already published (a reworded one won't be caught by `add()` dedup). Curating only the undistilled entries would fix that. On idle runs it also loads and dumps the whole QA history before the `no_new_entries` exit and re-reads every context row for the watermark. Moving the QA load past both exits would keep those cheap, which matters because it runs for every session after each `remember(session_id=...)` with the default debounce.
- The schema catalog gives `distill_sessions` `pipeline_name = "cognify_pipeline"`, and the schema view marks a link as observed when a type's most common source pipeline matches, so any cognify-built graph shows "distill sessions" producing its entity types even when no distillation ever ran.
- Cleanup: `is_auto_feedback_enabled` is duplicated in `feedback_detection.py` next to the `SessionManager` predicate that calls itself the one gate, the CLI and MCP each format `ImproveResult`, the enrichment run is kept three times and serialized twice, and `try_acquire_improve_lock`/`release_improve_lock` have no callers left.

Housekeeping: the dev merge cleared the earlier conflict. On CI, the Windows 3.13 unit shard hangs until its 15-minute timeout on both attempts without producing any test results (shard 3/3 at `8a6193ff`, shard 1/3 at the previous head), while the same shards pass on the other open PRs, so something on this branch likely hangs on Windows. The new background-task and drain code would be my first suspect, and a per-test timeout on that job would name the test.

cognee/modules/session_distillation/distill.py
```diff
@@ -382,25 +400,73 @@ async def publish_distilled_lessons(
...
+        logger.info(
+            "Session %s: all %d gated entries already distilled into dataset %s",
+            scope.session_id,
+            len(context_entries),
+            scope.dataset_id,
+        )
+        return scope.result("no_new_entries")
+
     proposed = await propose_lessons(qa_rows, context_entries)
     if not proposed:
```
NMZivkovic, line 461: `propose_lessons` returns `[]` both when the curator finds nothing durable and when every curator call failed (the batch error is caught and logged as "failed open"), and this branch advances the watermark in both cases. So one LLM outage (quota, auth, or a 429 window longer than the retries) marks every gated entry as distilled. Later runs return `no_new_entries` without calling the LLM, so for a finished session those entries are never distilled. The same happens at `if not accepted:` when the writer calls fail, and after a partial publish, where the lost batches get marked too.

The watermark docstring says a failed run leaves it untouched. `propose_lessons` and the writer need to tell "failed" apart from "nothing to keep" (raise, or return a flag) before this advances.

cognee/infrastructure/session/session_manager.py
```diff
@@ -308,12 +314,16 @@ def is_session_available_for_completion(self, user_id: str | None) -> bool:
...
     def is_auto_feedback_enabled(self) -> bool:
-        """Return True if caching and automatic turn-feedback analysis are both enabled."""
-        cache_config = CacheConfig()
+        """Return True if caching and automatic turn-feedback analysis are both enabled.
+
+        The one predicate for the session layer's auto-feedback gate: retrievers,
+        turn handling and the improve stages all ask here rather than re-reading
+        the cache config themselves.
+        """
+        cache_config = get_cache_config()
```
NMZivkovic, line 326: Switching these gates from `CacheConfig()` to the `@lru_cache`d `get_cache_config()` freezes `CACHING` and `AUTO_FEEDBACK` at import time, because the Ladybug adapter calls `get_cache_config()` at module level, so `import cognee` already fills the cache. Anything that sets the env var after importing is now ignored. At this head, after `import cognee`, setting `AUTO_FEEDBACK=false` leaves `is_auto_feedback_enabled()` returning True:

- `cognee-cli demo` imports cognee and then sets `AUTO_FEEDBACK=false` so keyless machines don't fire a turn-analysis LLM call per query. That pin no longer works.
- `test_search_db.py`, `test_library.py` and `test_cogx_roundtrip.py` set `AUTO_FEEDBACK=False` after import to keep turn analysis out of retrieval assertions, so they now run with it on.

Retrievers still read `CacheConfig()` fresh, so the two disagree within one search, and the new `feedback_detection.is_auto_feedback_enabled` has the same frozen read. I'd go back to a fresh read here.

cognee/modules/improve/config.py
```diff
@@ -0,0 +1,79 @@
...
+    IMPROVE_DEBOUNCE_SECONDS=0         # ... or after T seconds
+    IMPROVE_STAGES_DISABLED=a,b        # csv of stage names to skip
+    IMPROVE_FEEDBACK_ALPHA=0.1         # learning rate for feedback weights, (0, 1]
+"""
+
+from functools import lru_cache
+from typing import Annotated, Any
+
+from pydantic import field_validator
+from pydantic_settings import BaseSettings, NoDecode, SettingsConfigDict
```
NMZivkovic, line 23: `NoDecode` first shipped in pydantic-settings 2.7.0, but `pyproject.toml` still allows `>=2.2.1`. This import runs during `import cognee` (through `cognee.api.v1.improve`), so any environment that resolves 2.2.1 to 2.6.1 now fails with `ImportError: cannot import name 'NoDecode'`. The lock file pins 2.15, so CI won't catch it. Either raise the floor to 2.7.0, or keep the field a plain `str` and split it in a property.

cognee/modules/agent_memory/decorator.py
```diff
@@ -41,7 +41,7 @@ def agent_memory(
     session_id: str | None = None,
     user: User | None = None,
     dataset_name: str | None = None,
-    session_trace_summary: bool = True,
+    session_trace_summary: bool = False,
```
NMZivkovic, line 44: With this default flipped (and the summary also gated on `AUTO_FEEDBACK` in `add_agent_trace_step`), each successful decorator trace step stores the fallback line `<qualname> succeeded.` (failed ones get `<qualname> failed. Reason: <error>.`). Stage 3 still reads `session_feedback` (`raw_trace_content=False`, no gate), and so does periodic persistence, so `improve()` now runs `add` plus `cognify` LLM extraction over lines like `search_docs succeeded.`, and the `agent_trace_feedbacks` node set loses the content the summaries used to carry.

The notes here and in `add_agent_trace_step` that nothing is lost hold for stage 4, which reads the return value, but not for stage 3. Either have stage 3 read the raw return values when there's no summary, or skip windows that only contain fallback lines.

cognee/infrastructure/session/session_persist_watermark.py
```diff
@@ -46,51 +205,90 @@ class SessionPersistWindow:
...
+    await SESSION_PERSIST_WATERMARK.write_count(
+        session_manager, user_id, session_id, persisted_qa_count
+    )
+
+
+# -- Stage 3: persisted agent trace steps ------------------------------------
+
+# Per (user, session), NOT per dataset — same single-target-dataset model as
+# the Q&A watermark above; see "Scope" in the module docstring.
+TRACE_PERSIST_WATERMARK = StateRowWatermark(
```
NMZivkovic, line 226: The module docstring now calls the missing dataset in this key a known limitation, on the premise that session content bridges into the one dataset its `remember(session_id=...)` calls target. That premise doesn't hold for traces: `@agent_memory` writes them, and periodic persistence sends them to `dataset_name or "main_dataset"`, unrelated to `remember()`. So once periodic persistence has moved this row, `improve(dataset="project", session_ids=[s])` persists nothing (or only newer steps) into `project`, while before this PR stage 3 read every step each time. If per-dataset rows are too much for this PR, it's worth a follow-up issue and a note in the `improve()` docs, since users will see an empty stage 3 with no hint why.

cognee/tasks/memify/extract_agent_trace_feedbacks.py
```diff
@@ -79,37 +119,100 @@ async def extract_agent_trace_feedbacks(
...
+                        len(normalized_trace_values),
+                        content_label,
+                        window_size,
+                        total_trace_count,
+                    )
+                    yield TracePersistWindow(
+                        user_id=user_id,
+                        session_id=session_id,
+                        text=f"Session ID: {session_id}\n\n" + "\n".join(normalized_trace_values),
+                        persisted_trace_count=total_trace_count,
```
NMZivkovic, line 214: The one-snapshot read closes the count/fetch race, but a capped window still advances the watermark to the total. When periodic persistence falls behind (one failed cognify, or a second decorated function without persistence sharing the session and moving the count past a multiple of N), the next trigger reads only the newest `last_n_steps`, and the older pending steps are sealed: `improve()` then sees nothing pending and can't backfill them. The docstring says this is exactly what `last_n` did before, but before this PR nothing sealed those steps and `improve()` re-read them. Taking the oldest N pending steps and advancing to `effective + N` would keep the cap without losing steps. `test_last_n_steps_caps_the_pending_window` pins the current behavior.

cognee/tasks/memify/apply_feedback_weights.py
```diff
@@ -155,64 +202,126 @@ async def _process_feedback_item(
...
+    if pruned:
+        logger.warning(
+            "Feedback QA %s (session %s): %d graph element(s) no longer exist and were dropped: %s",
+            qa_id,
+            session_id,
+            len(pruned),
+            pruned,
+        )
+
+    qa_success = not failed
```
NMZivkovic, line 275: The new prune-and-mark path treats an id missing from `get_weights()` as deleted and marks the row done. The comment where ids are pruned calls the cross-dataset case a deliberate trade, but it has real costs:

- Session QA rows don't carry a dataset, and with access control on this reads only the target dataset's graph. A row rated during a search over dataset B gets all its ids pruned by `improve(dataset="A", session_ids=[s])` and is marked applied, so B's own improve skips it and B's weights never move. Before this PR the row stayed pending and B's run applied it.
- On Ladybug, `get_nodes` swallows query errors and returns `[]`, so a transient read failure (a closed adapter, lock contention) prunes every node id (edge reads raise instead) and marks node-only rows (RAG rows) done for good. A mixed row whose edge writes succeed is also marked done with its nodes dropped.
- The re-rating check compares only the score, so an explicit rating equal to an earlier implicit one counts as already applied, and the half-alpha implicit update stands.
- If a graph call raises after the node weights were written, nothing is saved, so the same nodes move again on every run, the attempt cap never trips, and the rest of the batch is aborted. That path predates this PR, but the constants docstring says a partial row now moves each element exactly once.

`improve()` doesn't reach this today: stage 1 is skipped as `feedback_influence_zero` on the default `DEFAULT_FEEDBACK_INFLUENCE=0`, and as `backend_unsupported` by the probe when it's set. Direct `apply_feedback_weights_pipeline` callers (the feedback demos) hit it now, and `improve()` will once both gates are fixed. Recording pruned ids separately instead of counting them as applied, and pruning only ids that are confirmed deleted, would cover the first two.

cognee/api/v1/improve/improve.py
```diff
@@ -45,512 +86,393 @@ async def improve(
...
+    node_name: list[str] | None,
+    session_ids: list[str],
+    build_global_context_index: bool,
+    build_truth_subspace: bool,
+    run_in_background: bool,
+    overrides: dict,
+) -> ImproveResult:
+    """Forward every option to the configured server, which runs the same
+    stages and hands back its ``ImproveResult``."""
+    payload = await remote_client.improve(
```
NMZivkovic, line 435: In `serve()` mode these options still don't reach the server, despite the docstring above. `get_remote_client()` only ever returns a `CloudClient`, and `CloudClient.improve(dataset, **kwargs)` (untouched here) copies only `run_in_background` and `node_name` into the POST body. So `cognee.improve("docs", session_ids=["s1"], build_truth_subspace=True)` sends `{"dataset_name": "docs"}`, the server skips every session stage with `no_session_ids`, and the caller gets a normal-looking result. Only the mocked client in `test_improve_orchestration.py` sees the forwarded options.

Related: a remote `run_in_background=True` comes back with `finished=false` and no local task, so `result.status` stays `running` and `await result.wait()` returns immediately. `remember()`'s remote branch has the same gap for the new `self_improvement` field (and `session_ids`), so `self_improvement=False` is ignored in `serve()` mode.

cognee/api/client.py
```diff
@@ -108,6 +112,19 @@ async def lifespan(app: FastAPI):
...
     yield
 
+    # Let in-flight background work (background remember runs, the session
+    # improve bridge) finish before the engines below are torn down under it.
+    # Bounded so a stuck task cannot hold the process hostage; nothing is
+    # cancelled on timeout, it is only reported.
+    from cognee.infrastructure.background_tasks import wait_for_background_tasks
+
+    logger.info("Shutting down: draining background tasks")
+    if not await wait_for_background_tasks(timeout=BACKGROUND_DRAIN_TIMEOUT_SECONDS):
```
NMZivkovic, line 122: This can wait up to 30 seconds before the engines are closed, but the shipped image runs gunicorn without `--graceful-timeout` (default 30s), docker compose uses Docker's 10s stop timeout, and the Helm chart uses Kubernetes' 30s default. If a background remember, session bridge or `run_in_background` improve (now registered here too) is still running at shutdown, the worker is SIGKILLed inside this wait, so the engine close and Ladybug WAL checkpoint below never run, which is exactly what that block is there to prevent. Before this PR the close ran right after requests drained.

I'd keep the default drain well under the smallest kill window, or raise gunicorn's `--graceful-timeout` and the compose and chart stop timeouts to match.

cognee/modules/improve/stages.py
```diff
@@ -0,0 +1,408 @@
...
+class FeedbackWeightsStage(BaseStage):
+    """Stage 1: move ``feedback_weight`` on the graph elements scored answers used."""
+
+    name = "feedback_weights"
+    needs_sessions = True
+
+    def gate(self, inputs: ImproveRunInputs) -> str | None:
+        from cognee.base_config import get_base_config
+
+        if get_base_config().default_feedback_influence <= 0:
```
NMZivkovic, line 43: This gate reads the global `DEFAULT_FEEDBACK_INFLUENCE`, but `recall()` and `search()` take `feedback_influence` per call, the way the feedback demo recalls. `base_config` even warns against a positive global, since it defers every hybrid search to graph completion. So a user who rates answers and recalls with `feedback_influence=0.8` never gets weights applied through `improve()`, where before this PR the stage ran whenever `session_ids` were given. I'd gate on session ids only and leave the influence decision to read time.

cognee/modules/truth_subspace/build.py
```diff
@@ -268,41 +388,54 @@ def build_for_epoch(truth_epoch: int):
...
-            }
+            return live_epoch_result(
+                nodes_skipped=nodes_skipped, error=f"persisting alignments failed: {error}"
+            )
 
         nodes_scored = sum(1 for ok in write_result.values() if ok)
 
+        # Step 6: ... and only then move the centroids to N+1. Until this write
+        # lands, epoch N is what the reranker sees.
+        commit_error = await commit_centroids()
```
NMZivkovic, line 420: If `commit_centroids()` fails here, every chunk has already been rewritten in place with epoch N+1 coordinates while the live centroids stay at N, and nothing restores the old coordinates. The stage now reports it as errored, which is good, but the reranker only applies `truth_factor` when a chunk's epoch matches the live epoch, so truth reranking is off for the rescored corpus until a later build succeeds. The description's "a mid-build failure leaves epoch N live" is only true for the centroids.

Related, a few lines up: `nodes_scored` is counted but never checked. If `set_node_truth_state` returns all `False` (Ladybug does when `get_nodes` swallows an error, or when the chunks were deleted meanwhile), the N+1 centroids are still committed, no chunk carries the live epoch, and the build returns `completed` with `nodes_scored=0`. Skipping the commit when nothing was written, like the `if not scored` guard does for embeddings, would cover it.

cognee/tasks/memify/apply_feedback_weights.py
```diff
@@ -155,64 +202,126 @@ async def _process_feedback_item(
...
+            MEMIFY_METADATA_FEEDBACK_WEIGHTS_APPLIED_EDGE_IDS_KEY: sorted(applied_edges),
+            MEMIFY_METADATA_FEEDBACK_WEIGHTS_APPLIED_SCORE_KEY: feedback_score,
+            MEMIFY_METADATA_FEEDBACK_WEIGHTS_ATTEMPTS_KEY: attempts,
+        },
     )
 
     logger.info(
-        "Processed feedback QA %s from session %s (nodes=%d, edges=%d, applied=%s)",
+        "Processed feedback QA %s from session %s (source=%s, alpha=%s, nodes=%d, edges=%d, "
+        "moved=%d, pruned=%d, failed=%d, attempt=%d, applied=%s, feedback_text=%r)",
```
NMZivkovic, line 302: This INFO line now logs `feedback_text=%r`, which is the user's own words: explicit feedback text, or on the implicit path the verbatim follow-up message the rating came from. It goes unredacted to stderr and the default log file. The old line only had ids and counts. I'd drop the text here, or log its length.

cognee/modules/improve/capabilities.py
lxobr: Confirmed and reproduced — fixed with your `engine.__class__` suggestion (plus the adapter name string) and pinned by a test through a real handle; the probe also moved after the lock claim so losers never pay for it.

cognee/infrastructure/locks/session_lock.py
lxobr: Confirmed — session keys are now user-scoped, and a lock-held bridge logs the skip honestly and refunds its debounce window so the next remember() retries (your "retry the loser" option). Waiting on the dataset key changes the background-mode claim contract, so I'd take that in its own pass.

cognee/modules/improve/graph_changes.py
lxobr: Confirmed on all four — the watermark is now stage-8-scoped as you suggested (stamped only when stage 8 completed, compared at its start time with the stamped run's own pipelines excluded), the persist stages no longer log empty memify rows, and both missing pipeline names are on the write list.

cognee/modules/session_distillation/distill.py
lxobr: Confirmed, partial publish included — failed calls now surface distinctly, the run publishes what survived (date-free documents dedup on retry) and raises instead of advancing the watermark.

cognee/infrastructure/session/session_manager.py
lxobr: Confirmed and reproduced — all three gates read a fresh `CacheConfig()` again, pinned by a post-import env-toggle test.

cognee/modules/improve/config.py
lxobr: Confirmed (reproduced on 2.6.1) — floor raised to 2.7.0.

cognee/modules/agent_memory/decorator.py
lxobr: Confirmed (worse with AUTO_FEEDBACK off — stage 4 skips too, so nothing read the values) — went with your first option: stage 3 substitutes the return value for fallback-only steps, and the "nothing is lost" notes are corrected.

cognee/infrastructure/session/session_persist_watermark.py
lxobr: Confirmed — caveat added to the module docstring and improve()'s session_ids docs. The per-dataset trace watermark row goes to a follow-up issue as you suggested (the distill watermark's id-suffix pattern fits cleanly).

cognee/tasks/memify/extract_agent_trace_feedbacks.py
lxobr: Confirmed — the window now takes the oldest N pending and advances only past them, so a backlog drains across triggers; the pinned test now pins the draining behavior.

cognee/tasks/memify/apply_feedback_weights.py
lxobr: Confirmed on all four — pruned ids now keep the row pending (bounded by the attempt cap) so the dataset that has them consumes it, a raise persists the bookkeeping and spares the rest of the batch, and re-rating compares (score, source). The Ladybug `get_nodes` error-swallow is shared adapter surface, so that root cause goes in its own change.

cognee/api/v1/improve/improve.py
lxobr: Confirmed — the server DTO already accepted all four fields, so `CloudClient.improve` now forwards them (with a client-level payload test), and remember() forwards `self_improvement` and raises on `session_ids` like `extractor`. Remote background observability needs a status endpoint, so the docstring says fire-and-forget for now and the endpoint goes to a follow-up.

cognee/api/client.py
lxobr: Confirmed — the drain default is 8s and compose sets `stop_grace_period: 15s`, so drain plus engine close fit the smallest shipped window.

cognee/modules/improve/stages.py
lxobr: Confirmed — the influence check is gone; stage 1 gates on session ids and backend capability only (landed with the probe fix, since the two gates masked each other).

cognee/modules/truth_subspace/build.py
lxobr: Confirmed on both — the commit is now skipped when nothing was scored (mirroring the embeddings guard) and the docstrings no longer claim rescored chunks stay valid. Restoring old coordinates on commit failure would need a read-before-write snapshot; given the guard plus the errored status I'd leave that out unless you feel strongly.

cognee/tasks/memify/apply_feedback_weights.py
lxobr: Confirmed — it logs the length now (scanned the diff; this was the only such line).

lxobr:
Thanks for the review. Every point was correct, each fix is its own small commit, and the inline replies carry the details. The latest dev is merged in.

The smaller items are all fixed too, with one framing correction: MCP `improve` is deliberately not a tool (the hardening test pins the tool set), so the fix there is the CLAUDE.md wording. Follow-ups tracked separately: the trace-snapshot range read, the per-dataset trace watermark, a status endpoint for remote background improves, the Ladybug `get_nodes` error swallowing, the waiting version of the dataset lock, `memify_metadata` rolling-deploy compat (release notes), and preferences on net helpfulness.

The branch also carries a round of boundary cleanup on top: stage 8 now owns its watermark stamp, the improve lock survives cancellation, remember() reads `ImproveResult.lock_held` instead of decoding stages itself, the remote client derives its option surface from `MEMIFY_PASSTHROUGH_KEYS`, and `run_info` merges append-style.

Windows CI: the job already runs a 300 second per-test timeout and a py-spy watchdog, so the next run's snapshots should name the hang.

lxobr (approved):

PR #4940, opened 2026-09-05
SDK-572 fix(mcp): remove unreachable list_data helper

Vasilije1990:
## Summary

Closes SDK-572.

`cognee-mcp/src/server.py` defined `list_data(dataset_id=None)` decorated with `@log_usage(function_name="MCP list_data", log_type="mcp_tool")` but never registered it with the tool registry, and nothing in the server called it. Since `fix(cognee-mcp): expose only memory tools` the operational helpers are intentionally internal, but this one was not used internally either, so it was dead code with two user-visible side effects:

- its own output told users to call `list_data(dataset_id="...")`, which the server answers with `Unknown tool: 'list_data'`
- the `get_document` docstring described a "list_data lookup" step that does not exist

Found by the MCP client journey (`cognee/tests/journeys/test_mcp_journey.py`, #4939), whose direct call to the tool returned `Unknown tool`.

### Change

- Delete `list_data` (128 lines) and its decorator.
- Drop the stale mention from the `get_document` docstring.
- Remove `list_data` from the README's list of internal helpers.

No tool surface change: `test_mcp_exposes_only_memory_tools` and the tool-search tests are unaffected.

### Worth a product call (not done here)

`save_interaction`, `search`, `get_document`, `get_chunk_neighbors`, `delete_dataset`, `prune` and `improve` are in the same state: usage-logged, unregistered, and with no internal call sites. `cognify` and `delete` are used by `remember` and `forget`. Deleting the rest, or re-exposing them as hidden searchable tools like `cognify_status`, is tracked in SDK-572's notes.

## Test plan

- [x] `pytest cognee-mcp/tests` → 80 passed
- [x] `ruff check` / `ruff format --check` clean
- [x] No remaining `list_data` references in `cognee-mcp` (`list_datasets` on the client is a different method and untouched)

### DIFF
--- a/cognee-mcp/README.md
+++ b/cognee-mcp/README.md
@@ -476,7 +476,7 @@ docker run \
 **API Mode behavior:**
 The MCP server intentionally exposes only the memory API: `remember`, `recall`, and `forget`.
 In API mode these tools call the Cognee API server endpoints directly. Operational helpers such as
-`cognify`, `search`, `list_data`, `delete`, `prune`, `improve`, and document retrieval helpers are
+`cognify`, `search`, `delete`, `prune`, `improve`, and document retrieval helpers are
 kept internal and are not exposed as MCP tools.
 
 ## 💻 Basic Usage
--- a/cognee-mcp/src/server.py
+++ b/cognee-mcp/src/server.py
@@ -750,7 +750,7 @@ async def get_document(
     """
     Retrieve a complete source document and its chunks from the knowledge graph.
 
-    Use this after a CHUNKS search or list_data lookup when you need the full source
+    Use this after a CHUNKS search when you need the full source
     context around a result. If a chunk ID is provided instead of a document ID, the
     tool resolves the chunk's parent document and returns that document.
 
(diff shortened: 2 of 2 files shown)


ayush00git:
Hii, I was looking at the helpers this PR calls to remove/hide and noticed that the `save_interaction` is already
broken end-to-end.

**`save_interaction` always fails at the rule-association step**

`save_user_agent_interaction` calls 
```py
                    await add_rule_associations(
                        data=data,
                        rules_nodeset_name="coding_agent_rules",
                        context={
                            "user": user,
                            "dataset": dataset,
                            "data": added_data,
                        },

```
but the name is bound to the real cognee function, import (around server.py line 67)
```py
try:
    from cognee.tasks.codingagents.coding_rule_associations import (
        add_rule_associations,
        get_existing_rules,
    )
except ModuleNotFoundError:
    from .codingagents.coding_rule_associations import (
        add_rule_associations,
        get_existing_rules,
    )

```
prefers cognee over the vendored fallback. That function's signature is `(data, rules_nodeset_name, user_prompt_location, system_prompt_location, ctx=None)`, so every call raisesa TypeError. The error goes silent into `_task_errors`, and the tool is reporting success while the rule association silently never happened.

If the helper is being dropped, then its fine.but if it is kept or re-exposed later, the call site needs to pass `user_prompt_location`/`system_prompt_location` instead of `context=`.

ayush00git:
Here's a regression test, hope it would help verifying this qiuck.

```python
"""save_interaction passes `context=` to cognee's add_rule_associations, which has no such
parameter -> TypeError on every call, swallowed into _task_errors."""

import asyncio
import inspect
from types import SimpleNamespace

import pytest

import src.server as server
from cognee.tasks.codingagents import coding_rule_associations as real_module


class _DirectClient:
    use_api = False

    async def add(self, *args, **kwargs):
        return None

    async def cognify(self, *args, **kwargs):
        return None


@pytest.mark.asyncio
async def test_save_interaction_calls_add_rule_associations_with_a_valid_signature(monkeypatch):
    # The MCP server imports the real cognee implementation, not the vendored copy.
    assert server.add_rule_associations.__module__ == real_module.__name__

    user = SimpleNamespace(id="user-1", tenant_id=None)
    dataset = SimpleNamespace(id="dataset-1", name="main_dataset")
    data_item = SimpleNamespace(id="data-1")

    async def fake_get_default_user():
        return user

    async def fake_get_datasets_by_name(name, user_id=None):
        return [dataset]

    async def fake_get_last_added_data(dataset_id):
        return data_item

    real_signature = inspect.signature(real_module.add_rule_associations)
    bound_calls = []

    async def signature_checking_stub(*args, **kwargs):
        # Bind against the real signature (raises TypeError on mismatch, like the
        # real call would) without running the LLM step.
        bound_calls.append(real_signature.bind(*args, **kwargs))
        return None

    monkeypatch.setattr(server, "cognee_client", _DirectClient())
    monkeypatch.setattr(server, "get_default_user", fake_get_default_user)
    monkeypatch.setattr(server, "get_datasets_by_name", fake_get_datasets_by_name)
    monkeypatch.setattr(server, "get_last_added_data", fake_get_last_added_data)
    monkeypatch.setattr(server, "add_rule_associations", signature_checking_stub)
    monkeypatch.setattr(server, "_task_errors", {})

    await server.save_interaction(data="User asked X. Agent answered Y.")
    await asyncio.gather(*list(server._background_tasks))

    recorded = list(server._task_errors.get("main_dataset", []))
    assert recorded == [], f"save_interaction background task failed: {recorded}"
    assert len(bound_calls) == 1, "add_rule_associations was never called successfully"
    assert bound_calls[0].arguments["rules_nodeset_name"] == "coding_agent_rules"


```

PR #4941, opened 2026-09-05
fix: Sanitize identifiers in telemetry aggregates (COG-6288)

Vasilije1990:
## Description

The first scheduled run of the telemetry-insights Action (run 33947189669, Sep 5, 05:26 UTC) **failed closed by design**: the extraction's privacy guard found an identifier-shaped value in `provider_stack_daily.csv` (line 1573) and refused to publish. Root cause: user-configurable `llm_model` strings can embed UUIDs (Azure/vLLM deployment names), long hex ids (fine-tune ids), or emails. The guard worked — nothing was uploaded — but the job can never go green until the source columns are sanitized.

## Changes

1. **SQL-side sanitizer** (`_san`): replaces UUID / long-hex / email substrings with typed tokens (`<uuid>` / `<hex>` / `<email>`) in every user-configurable string column — llm/graph/vector/relational provider, `llm_model`, `endpoint`, and `version`. Aggregate keys stay analytically useful (`azure/dep-<uuid>-gpt4` still groups) without carrying identifiers.
2. **Stronger guard backstop**: adds a bare long-hex cell pattern and makes the uuid/key-hash patterns case-insensitive, so the job still fails closed on anything the sanitizer misses.

## Testing

Local DuckDB end-to-end: hostile model strings (uuid deployment name, hex fine-tune id, email, `ak_` key hash) all sanitize to guard-passing tokens; clean names (`openai/gpt-5-mini`) pass untouched; raw identifiers still trip the guard (backstop verified). Ruff clean.

## Note for reviewers

The cron executes **main's** copy of the workflow — this fix takes effect on the schedule only after the next dev→main sync. Until then the daily run keeps failing closed (safe, just red). The missing `telemetry-insights` label (the other latent failure — the analyze step files issues with it) has been created in the repo directly.

Fixes COG-6288 follow-up

### DIFF
--- a/.github/scripts/telemetry_aggregate_extract.py
+++ b/.github/scripts/telemetry_aggregate_extract.py
@@ -65,8 +65,30 @@ _IDENT = (
 )
 # Surface the event came from: 'sdk' (default), 'cloud', 'cli', ... Safe enum.
 _ORIGIN = "coalesce(json_extract_string(properties, '$.telemetry_origin'), 'unknown')"
+
+
+def _san(expr: str) -> str:
+    """Scrub identifier-shaped substrings from a user-configurable string column.
+
+    Provider/model/endpoint/version values come from customer config and can
+    embed UUIDs (deployment names), long hex ids, or even email addresses —
+    the first scheduled run tripped the privacy guard on exactly this (an
+    identifier inside an llm_model string). Replace them with typed tokens so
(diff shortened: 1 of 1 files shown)


ArmagedonFlamer:
Closing this in favour of #5034, which fixed the same failure in the same file with whole-value redacted bucketing plus regression tests, and is already on dev.

PR #4966, opened 2026-09-08
SDK-577 fix: Recover interrupted pipeline runs

goran-radonic:
## Description

Startup recovery only looked at `cognify_pipeline`, so an add killed mid-flight
(SIGKILL, OOM, pod eviction) kept its `DATASET_PROCESSING_STARTED` row forever.
Nothing was left alive to write a terminal status, and the only thing that
clears it is the next add to that same dataset, so a dataset nobody adds to
again never recovers and `GET /v1/datasets/status?pipeline=add_pipeline` reports
it as processing indefinitely. SDK-577.

Two things I got wrong on the first pass, both worth writing down.

I started by doing what the ticket asked: reset the stale status to INITIATED.
Then I checked what that status actually does downstream and it does not hold
up. `check_pipeline_run_qualification` only runs when `use_pipeline_cache=True`,
and add, cognify and memify all pass False, so a stale STARTED row never blocked
a re-run in the first place. And INITIATED is an in-progress status to every
consumer: `pollDatasetStatus.ts` keeps it in `IN_PROGRESS_STATUSES` and resolves
only on COMPLETED or ERRORED, `DatasetDetailPage` treats it exactly like
STARTED. So a reset swaps one non-terminal status for another and the poller
keeps waiting on a run that can never finish. Recovery now closes the abandoned
run as ERRORED carrying a new `AbandonedPipelineRunError`, reusing the dead
run's own ids so its history reads as one run, which is also the row the run
would have written had it raised instead of being killed.

I also picked candidates with the existing latest-run-per-dataset query. That
answers a status question, where only the newest row matters. Recovery needs
every run a crash left open, and a background batch writes all its STARTED rows
up front, so one crash can abandon several runs of the same dataset and
pipeline. The new `get_unclosed_pipeline_runs` selects STARTED rows that have no
terminal row for their own run id, which also means a run recovery already
closed is never picked up again.

Writing a terminal status where callers never used to see one has two readers
that needed to learn the difference between a killed run and a failed one:

- `graph_warmup` counted any non-add ERRORED row as a failed build, so recall
  would answer "Memory build failed" over an intact graph, and the abandoned row
  would become the reported cause instead of an older real one. Abandoned rows
  are excluded from that verdict now, NULL `error_class` included, since `!=`
  alone would have dropped legacy rows.
- the graph-count cache keys on the latest cognify run's id and the closing row
  reuses the dead run's id, so counts cached before the rollback would be served
  from then on. Only a COMPLETED run's counts are served now, which also keeps a
  mid-run dataset reporting live numbers.

One part is separable. Entering a dataset's database context takes a queue
permit before it can fail, and a failed entry never released it, so a sweep over
datasets whose owner user was deleted could exhaust the queue or block startup
outright. That is pre-existing, but recovery is the caller that hits it in a
loop. Happy to split `context_global_variables.py` and its test into their own
PR if you would rather review it separately.

The module moved to `cognee/modules/pipelines/recovery.py`, since it is not
cognify-specific any more, and the entry point is now
`recover_stale_pipeline_runs_on_startup`. The tenant pod's lifespan wiring for
CLO-365 needs the new name at the pin bump.

## Acceptance Criteria

- A dataset whose add was killed mid-flight reports a terminal status after a
  restart instead of STARTED forever, and can be added to and cognified again.
- Every run one crash abandoned is recovered, not only the newest per dataset
  and pipeline, and a closed run is not reopened on the next boot.
- Cognify still has its partial graph unwound before its run is closed, and a
  rollback that fails leaves the run open on purpose for the next startup.
- A dataset with an intact graph is not reported as build_failed because an
  unrelated run was abandoned, and a genuine failure is still the cause that
  gets reported.
- Recovery cannot take the API down: a failing lookup costs that one run, and
  the lifespan call is guarded.

Proof: full unit suite green locally, 5790 passed / 70 skipped. New tests are 15
in `test_recovery.py` (each on its own tmp sqlite engine, so the global sweep
never reaches the developer's own database), 5 in
`test_unclosed_pipeline_runs.py`, 3 in the graph_warmup probe, 1 in the
graph-count tests, and 1 for the queue permit release, which I checked fails
without the fix. The integration test that runs the rollback against a real
graph now also asserts the closing row and that a second boot does not repeat
it. I did not run the integration or e2e legs locally, they need API keys, so
those are on CI.

`COGNEE_STALE_RUN_RECOVERY_MIN_AGE_SECONDS` is documented in `.env.template`,
no longer takes the server down when it is misparsed, and is floored at 60s so a
typo cannot turn recovery into a sweep that closes live runs.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):


cognee/api/client.py
```diff
@@ -100,9 +100,16 @@ async def lifespan(app: FastAPI):
     from cognee.modules.users.methods import get_default_user
 
     await get_default_user()
-    from cognee.modules.cognify.recovery import recover_stale_cognify_runs_on_startup
+    from cognee.modules.pipelines.recovery import recover_stale_pipeline_runs_on_startup
 
-    await recover_stale_cognify_runs_on_startup()
+    try:
+        await recover_stale_pipeline_runs_on_startup()
+    except Exception:
```
dexters1: We have to avoid using these statements:

except Exception

It blocks us from seeing and finding real issues with the codebase, only specific exceptions can be caught

In this case it is only acceptable to have them catch the specific dataset recovery attempt, but then it has to be logged and properly notify us what and why happened at the recovery attempt of a specific dataset, this it too broad and can mask too much different failures unrelated to a specific dataset recovery attempt.

cognee/modules/pipelines/recovery.py
```diff
@@ -0,0 +1,263 @@
...
+                    user=run_user,
+                    started_at=getattr(pipeline_run, "started_at", None),
+                )
+            logger.info(
+                "Startup recovery closed abandoned %s run %s as ERRORED (dataset=%s).",
+                pipeline_name,
+                pipeline_run.pipeline_run_id,
+                pipeline_run.dataset_id,
+            )
+        except Exception as error:
```
dexters1: This is too broad of an exception handling, unexpected exceptions need to raise an Exception and each exception for each dataset needs to be caught and reported

cognee/modules/pipelines/recovery.py
```diff
@@ -0,0 +1,263 @@
...
+                continue
+
+            rollback_handler = rollback_handlers.get(pipeline_name)
+
+            if rollback_handler is not None:
+                # The dataset's own graph/vector databases are entered only to
+                # unwind partial data: a pipeline with nothing to unwind would
+                # otherwise provision them just to write a relational row.
+                async with set_database_global_context_variables(dataset.id, dataset.owner_id):
+                    await rollback_handler(
```
dexters1: This feature could significantly reduce container start time. Going through datasets one by one and recovering them will take a long time and it blocks server start not sure how important that is for the Cloud team.

Additionally recovery should be started in parallel for multiple datasets to speed up the process

Dataset pipeline related operations are queued in Cognee, so for example calling add/cognify/add/cognify in parallel will orchestrate the calls to run sequentially, I do not know if the rollback mechanism adheres to this queue @siillee can check and add it if it does not, but with that change we don't need to handle recovery before server start and we can start the server and start the recovery in the background which will block other operations on the dataset until recovery is finished

dexters1 (changes requested):
Broad exception handling should almost never be used and when used it should only catch minimal areas, only exceptions for each dataset recovery operations attempt can be catched but they need to be reported for each dataset so we have a status on each recovery attempt for each dataset. The rest of the recovery system should not be included in the exception catching. If a specific dataset recovery fails the rest of the datasets should still try and be recovered.

We need to use a fail-fast approach with exceptions because it:
  - Makes bugs easier to find
  - Keeps errors close to their source
  - Prevents invalid state from spreading
  - Simplifies code by avoiding deeply nested error handling

Claude code for some reason keeps forcing broad exception handling in all the code it writes, just so it can pass tests but it creates a nightmare for future product maintenance 

Additionally, feature now blocks server start it needs to be moved to happen after server starts and @siillee should help here to make sure no other dataset operations for the dataset can run until the recovery process is finished in the background for the dataset.

dexters1:
@siillee can you review this one? Since you created the rollback mechanism

.env.template
```diff
@@ -290,6 +290,19 @@ VECTOR_DATASET_DATABASE_HANDLER="lancedb"
...
+#  On API startup, a run left non-terminal by a killed process (SIGKILL, OOM,
+#  pod eviction) is rolled back where its pipeline has a rollback policy and
+#  then closed as errored, so a dataset stops being reported as processing a
+#  run that can never finish. A run is only treated as abandoned once it has
+#  been non-terminal for longer than this, which is what keeps the sweep from
+#  closing a run still executing in another replica. Raise it if your jobs
+#  legitimately run longer than an hour. Values below 60 are raised to 60.
+################################################################################
+
+#COGNEE_STALE_RUN_RECOVERY_MIN_AGE_SECONDS=3600
```
dexters1: Lets remove this from .env.template it should not be configurable by users and is too much info for them

cognee/context_global_variables.py
```diff
@@ -219,6 +219,26 @@ async def apply_database_context_variables(
...
+            await self._apply_dataset_databases(dataset, user_id, permission_type)
+        except BaseException:
+            # Everything below the slot acquisition can raise: a deleted
+            # dataset owner (get_user), a provisioning or connection failure.
+            # When it does, __aenter__ never returns, so __aexit__ never runs
+            # and the permit would be held for the life of the task. In a
+            # lifespan task that is the life of the process, and enough of
+            # them wedge every later ensure_slot (SDK-577 found this through
+            # startup recovery looping over datasets).
+            await dataset_queue().release_slot_for(dataset)
```
dexters1, line 232: Nice addition, extracted this out to a different PR here:

https://github.com/topoteretes/cognee/pull/4973/changes

NMZivkovic (commented):
## Review

Read the whole diff and verified the load-bearing claims against the branch rather than the diff. This is careful work — the candidate selection rewrite (every unclosed run, not the latest per dataset), reusing the dead run\s ids so its history reads as one run, the `error_class` NULL guard in `graph_warmup`, and the `data_info` passthrough are all correct and well-argued in the code comments. The PR description explaining the two things you got wrong on the first pass is genuinely useful.

**One real defect**, inline: the per-dataset recovery count is dropped on the floor, so the summary log always reports `0 ... closed`. Nothing else I found rises above minor.

### Cross-PR: this collides with #4993

This PR **deletes** two files that #4993 **modifies**:

| file | #4966 | #4993 |
|---|---|---|
| `cognee/modules/cognify/recovery.py` | deleted | edited (`reset_pipeline_run_status` call site) |
| `cognee/tests/unit/modules/cognify/test_recovery.py` | deleted | edited |

Whichever lands second conflicts. Worth deciding the order deliberately, because there is a semantic interaction too: #4993\s headline motivation for keying the reset marker off the run rather than the acting user is *"startup recovery hits that every time, since it acts as the dataset owner"* — and this PR removes recovery\s `reset_pipeline_run_status` call entirely, replacing the reset with a terminal ERRORED row. Once this lands, that motivation survives only for the shared-dataset case in `reset_dataset_pipeline_run_status`. Still a correct fix, but the justification changes.

Also note the branch is currently **CONFLICTING** against `dev`.

### Interaction with #4983

#4983 reports a stale `STARTED` row as `"ABANDONED"` at read time; this PR writes a real `ERRORED` row for the same run. Since `/activity/pipeline-runs` does not dedupe by `pipeline_run_id`, a recovered run will surface as **both** an `ABANDONED` row and an `ERRORED` row. Combined with the separate issue I raised on #4983, the two features want a shared decision about what the feed shows per run.

---

cognee/modules/pipelines/recovery.py
```diff
@@ -0,0 +1,429 @@
...
+    logger.info(
+        "Recovery found %d abandoned pipeline run(s) across %d dataset(s).",
+        candidate_count,
+        len(runs_by_dataset),
+    )
+
+    rollback_handlers = _rollback_handlers()
+    slots = asyncio.Semaphore(_max_concurrent_dataset_recoveries())
+
+    async def _recover_dataset_within_bound(dataset_id: UUID) -> None:
```
NMZivkovic: **[Major] `_recover_dataset_within_bound` drops the count, so the summary always reports 0**

It is annotated `-> None` and never returns `_recover_dataset`\s result, so every element of `outcomes` is `None` and the `sum(... isinstance(outcome, int))` on line 419 has nothing to add up. The closing log line reports **0 closed** on every successful sweep.

```python
async def _recover_dataset() -> int: return 3      # closes 3 runs
async def _within_bound() -> None:                  # the shape here
    await _recover_dataset()
outcomes = await asyncio.gather(*(_within_bound() for _ in range(4)), return_exceptions=True)
# outcomes -> [None, None, None, None]
# sum(o for o in outcomes if isinstance(o, int)) -> 0   (12 runs were actually closed)
```

That log line is the only operator-visible summary of the whole feature, and it will read `Recovery finished: 0 of 12 abandoned run(s) closed` right after twelve `Recovery closed abandoned ... as ERRORED` lines — which reads like a failure and will send someone debugging a working sweep.

The tests all assert database state (`_rows(... status=ERRORED)`), which is the right thing to assert and is why none of them catch this. One-line fix: `return await _recover_dataset(...)`, and drop the `-> None`.

cognee/modules/pipelines/recovery.py
```diff
@@ -0,0 +1,429 @@
...
+    that dataset's lock rather than racing its recovery. The lock is
+    process-local, so that exclusion holds within this process only.
+
+    Nothing here is wrapped in error handling except the two steps of one
+    dataset's attempt, which are reported per dataset: the read below is the
+    sweep itself, and if it fails there is no recovery to speak of, so it
+    propagates to the caller that started this.
+    """
+    recovery_candidates = await get_unclosed_pipeline_runs()
+    datasets_by_id, users_by_id = await _load_datasets_and_users(recovery_candidates)
```
NMZivkovic: **[Minor] Datasets and users are loaded before the staleness filter**

`_load_datasets_and_users(recovery_candidates)` runs against *all* unclosed runs, then the loop below discards the ones inside `STALE_RUN_MIN_AGE_SECONDS`. On a busy instance the runs that are genuinely in flight are exactly the ones most likely to be unclosed at boot, so this reads rows for the set it is about to throw away.

Filtering by threshold first and loading against the survivors would keep the batching win the docstring measured (2.7s → 1.2s) while shrinking the input.

cognee/modules/data/methods/get_datasets_graph_counts.py
```diff
@@ -177,7 +177,20 @@ async def get_datasets_graph_counts(
...
 
-        cached = cached_metrics.get(latest_run.pipeline_run_id)
+        # Only a COMPLETED run's counts are trustworthy. The latest run may be
+        # one the startup recovery closed as ERRORED, which reuses the dead
+        # run's pipeline_run_id, so its counts were cached before that run's
+        # rollback deleted the nodes they describe (SDK-577). Recomputing on a
+        # non-COMPLETED run also keeps a mid-run dataset reporting real counts
+        # instead of zero. Same guard the other latest-run readers already
+        # apply (live_updates, edge-evidence lookup).
+        from cognee.modules.pipelines.models import PipelineRunStatus
```
NMZivkovic: **[Minor] Import sits inside the per-dataset loop**

`from cognee.modules.pipelines.models import PipelineRunStatus` is inside the `for dataset in ...` body, so it re-executes per dataset. It is only a `sys.modules` lookup, but the guard itself is a good change and reads better with the import at module scope (or at least above the loop) — the other readers this mirrors do it that way.

cognee/modules/pipelines/methods/get_unclosed_pipeline_runs.py
```diff
@@ -0,0 +1,70 @@
...
+    scope it.
+    """
+    if dataset_ids is not None and not dataset_ids:
+        return []
+
+    # Aliased so the NOT IN subquery keeps its own FROM: sharing the outer
+    # table would let SQLAlchemy correlate it and silently change the test to
+    # "this row is not itself terminal".
+    closed = aliased(PipelineRun)
+    closed_run_ids = select(closed.pipeline_run_id).filter(closed.status.in_(_TERMINAL_STATUSES))
```
NMZivkovic: **[Minor] `NOT IN` subquery is unguarded against NULL, and unindexed on status**

Two small robustness points on the anti-join:

1. If `closed_run_ids` ever yields a NULL, SQL `NOT IN` returns the empty set and recovery silently stops finding anything. Today the `status.in_(_TERMINAL_STATUSES)` filter makes that unreachable (operation rows carry NULL status and are excluded), so this is insurance rather than a live bug — `.filter(closed.pipeline_run_id.isnot(None))` costs nothing and removes the footgun class entirely. `NOT EXISTS` would too.
2. `pipeline_run_id` is indexed, but `status` is not, so on an append-only table this scans every terminal row at every boot. Worth watching once the table is large; a partial index on `status` would fix it if it shows up.

cognee/api/client.py
```diff
@@ -98,15 +121,30 @@ async def lifespan(app: FastAPI):
...
-    from cognee.modules.cognify.recovery import recover_stale_cognify_runs_on_startup
+    from cognee.modules.pipelines.recovery import recover_abandoned_pipeline_runs
 
-    await recover_stale_cognify_runs_on_startup()
+    # Recovery of runs a previous process abandoned runs in the background, so
+    # a boot that finds work to do does not hold the port closed while it does
+    # it. Each dataset is recovered under that dataset's lock, so operations
+    # arriving for a dataset queue behind its recovery instead of racing it
+    # (in this process: the lock is asyncio, see infrastructure/locks).
+    recovery_task = asyncio.create_task(recover_abandoned_pipeline_runs())
```
NMZivkovic, line 132: **[Question] Recovery now overlaps live traffic — worth confirming the intent for non-cognify pipelines**

Moving the sweep off the boot path is the right call for startup latency, and the per-dataset `asyncio` lock is the right guard. Two consequences worth stating explicitly since they are new:

- A client polling a dataset can now observe `STARTED → ERRORED` *while the server is serving*, where previously recovery had finished before the port opened. For `cognify_pipeline` the rollback runs first so the graph is consistent, but the status transition is now visible mid-session.
- For `incremental_update_pipeline` and `migration_import_pipeline` — which `_rollback_handlers()` correctly documents as having no handler — the run is now closed as ERRORED while its partial graph writes stay in place. The docstring is admirably honest about this, but closing a run as terminal while leaving data behind is a state a reader of `error_class` cannot distinguish from a cleanly-unwound cognify.

Is leaving those two uncovered acceptable for this PR, or should they be excluded from the sweep until they have handlers?

siillee (dismissed):

cognee/api/client.py
goran-radonic: Both consequences are real, and the second one had a fix in it, so I took that
part.

The two states shared one message, which is what made them indistinguishable.
`AbandonedPipelineRunError` now takes `rolled_back`, so the closing row says
either "the data it wrote was rolled back, so the dataset is back to its
previous state" or "<pipeline> has no rollback policy, so whatever it wrote
before it died is still in the dataset". `error_class` stays the same for both,
since the class means the run was killed and that is true either way, and
nothing machine-readable consumes it today anyway (graph_warmup excludes this
class wholesale, the frontend never reads the field). A test pins the split. If
you would rather have a second class so it is machine-readable, that is one
line, say the word.

On the question itself: I would not exclude them. Excluding them restores the
exact bug this PR closes, since their runs would sit STARTED forever with the
same data left behind, only now invisible. Closed-with-a-note is strictly more
honest than never-closed.

Handlers for them are their own ticket rather than a dictionary entry.
`incremental_update_pipeline` writes the new chunks and then deletes the ones
they replace, so unwinding a run killed between those two steps deletes the new
chunks while the old ones are already gone, which leaves a hole in the document
instead of repairing it. That needs its own logic and its own tests.

cognee/modules/pipelines/recovery.py
```diff
@@ -0,0 +1,429 @@
...
+        *(_recover_dataset_within_bound(dataset_id) for dataset_id in runs_by_dataset),
+        return_exceptions=True,
+    )
+
+    closed = sum(outcome for outcome in outcomes if isinstance(outcome, int))
+    logger.info(
+        "Recovery finished: %d of %d abandoned run(s) closed across %d dataset(s).",
+        closed,
+        candidate_count,
+        len(runs_by_dataset),
```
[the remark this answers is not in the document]
goran-radonic: Not doing this one, and the reason is in the thread above: the block was in
`.env.template` and was removed at review request, because this is an operator
escape hatch rather than a user knob.

The catch about the PR description is right though, and that is the actual
defect here: the body still said the variable is documented in `.env.template`
after the block came out. Fixed the description instead of the file.

NMZivkovic (commented):
Re-reviewed at `c4f89469`. All five points from my last review are addressed, and each fix has a test that would catch it regressing:

- **Summary count** — `_recover_dataset_within_bound` returns the count now; `test_the_closing_summary_counts_what_it_closed` and `test_the_summary_does_not_count_a_dataset_that_failed` pin both directions.
- **Read before filter** — the staleness filter runs first, and a boot with nothing abandoned costs one query (`test_a_boot_with_nothing_abandoned_reads_no_datasets`).
- **Import in the loop** — hoisted above it. Writing down why it can't go to module scope (the cycle `test_methods_package_has_no_import_cycle` pins) is the right call.
- **NULL in the anti-join** — guarded, with a test that actually inserts the NULL-id terminal row.
- **Non-cognify pipelines** — agreed on keeping them in the sweep. Excluding them just makes the same leftover data invisible, and closed with an honest message beats never closed. The `rolled_back` split in the message is enough for now; I wouldn't add a second error class until something needs to tell the two apart programmatically.

Nothing new from me. Two things still stand between this and merge, neither of them code on this branch:

1. dexters1's changes-requested review is still open — his threads are resolved, so it needs his re-review.
2. The collision with #4993 is unchanged: this PR deletes `cognee/modules/cognify/recovery.py` and `cognee/tests/unit/modules/cognify/test_recovery.py`, and #4993 edits both. #4993 has no blocking reviews, so landing it first and rebasing this one onto it still looks like the cheaper order.

NMZivkovic (dismissed):

cognee-mcp/src/server.py
```diff
@@ -888,6 +888,50 @@ def _agent_scoped_default_dataset() -> str:
...
+    try:
+        from cognee.modules.operations import ORIGIN_MCP
+        from cognee.modules.pipelines.recovery import recover_abandoned_pipeline_runs
+    except ImportError:
+        logger.warning(
+            "Installed cognee has no startup recovery (SDK-577) — a pipeline "
+            "run this server started and never finished stays reported as "
+            "processing until the same dataset is added to again. Remove "
+            "this guard once cognee-mcp requires a cognee release that ships "
+            "cognee.modules.pipelines.recovery."
```
dexters1, line 928: Lets wait for the cognee release to be done before we merge this and then we can remove the exception catching as it wont be needed anymore. Maybe seperate these changes into a new PR for now just for MCP changes

cognee/api/client.py
```diff
@@ -122,11 +142,40 @@ async def lifespan(app: FastAPI):
...
+    # Recovery first, and cancelled rather than drained: the sweep is resumable
+    # by design (it only closes runs that still have no terminal row), so
+    # losing the work in flight costs nothing, while waiting for it would hold
+    # shutdown for a sweep nobody needs to finish. It is also cancelled before
+    # the drain below because it holds a per-dataset lock while it works, and a
+    # draining task on that dataset would otherwise sit behind it for the whole
+    # timeout.
+    if not recovery_task.done():
+        recovery_task.cancel()
+        with suppress(asyncio.CancelledError):
```
dexters1, line 176: Will this information be logged so users are aware recovery tasks were killed?

cognee/api/client.py
```diff
@@ -122,11 +142,40 @@ async def lifespan(app: FastAPI):
...
+    # (in this process: the lock is asyncio, see infrastructure/locks).
+    #
+    # Created last, right before yield: everything above can still fail the
+    # boot (a bad secret, a bad IMPROVE_* value), and a task created before
+    # that point would dangle on a failed boot — the generator frame holding
+    # its only strong reference tears down without ever reaching the shutdown
+    # code that cancels it, leaving the task to the event loop's own weak
+    # reference (#4312).
+    recovery_task = asyncio.create_task(recover_abandoned_pipeline_runs())
+    recovery_task.add_done_callback(_report_recovery_outcome)
```
dexters1, line 160: Good change, just to make sure

Are all the dataset that recovery needs to be ran for locked behind the dataset lock?

So users can't do operations in a dataset that needs to have its recovery started (Specifically asking about the situation in which we need to rollback more datasets than there is room in the dataset queue)

ArmagedonFlamer:
Thanks @goran-radonic for the careful work here. Dev now covers this through #5090 and #5104: every pipeline is swept, every run without a terminal row is closed as ERRORED with AbandonedPipelineRunError on the dead run's own ids, and completed documents survive the rollback

PR #4968, opened 2026-09-08
fix(neo4j): return true edge direction from get_edges

### LINKED ISSUE #4967: Neo4jAdapter.get_edges returns incoming edges with source/target inverted (traversal-side order, not true edge direction)
## Bug

`Neo4jAdapter.get_edges` (`cognee/infrastructure/databases/graph/neo4j_driver/adapter.py`) matches relationships undirected:

```cypher
MATCH (n: `{BASE_LABEL}`{id: $node_id})-[r]-(m)
RETURN n, r, m
```

and returns tuples in the order `(n.id, m.id, {"relationship_name": ...})`.

Because `n` is always bound to the *queried* node, an edge that is actually stored as `(other) -[:REL]-> (queried node)` comes back with the endpoints **inverted** — the queried node is reported as the source. Example verified against a live Neo4j 5 container:

```
stored:   (node-b) -[:reports_to]-> (node-a)
get_edges("node-a") → [('node-a', 'node-b', {'relationship_name': 'reports_to'})]
expected:            [('node-b', 'node-a', {'relationship_name': 'reports_to'})]
```

## Impact

The main consumer is the memory consolidation pipeline (`cognee/memify_pipelines/consolidate_entity_descriptions.py`): `get_entity_neighborhood` → `format_edges` renders these tuples as "neighbor —relationship→ node" prompt material for entity-description consolidation. For every **incoming** relationship the direction is silently reversed, so the LLM consolidates entity descriptions over swapped endpoints. The Postgres hybrid adapter (`cognee/infrastructure/databases/hybrid/postgres/adapter.py:139`) forwards to the same method.

Note the outgoing/incoming asymmetry also makes `get_edges` inconsistent with `has_edges`/`add_edge`, which all operate in true edge direction.

## Expected behavior

`get_edges` should return `(source_id, target_id, ...)` in true edge direction regardless of traversal side. Fix: project `startNode(r).id AS source_id` in the query and use it as the tuple's first element.

## Environment

- cognee dev @ 6c022454e
- neo4j Python driver 5.x against Neo4j 5 (Docker `neo4j:5`)

xu-jia-ming:
## TLDR

Problem this solves:

- `get_edges` returned incoming edges with source/target swapped (traversal-side order, not true edge direction)
- The entity-description consolidation pipeline therefore rendered every incoming relationship over swapped endpoints

How it solves it:

- Project `startNode(r).id AS source_id` and build the tuple from the true source, keeping direction consistent with `has_edges`/`add_edge`

Fixes #4967

## Pre-Submission checklist

- [x] I have added meaningful tests
- [x] The handful of test files covering my change pass locally, e.g. `uv run pytest tests/test_litellm/<your_test_file>.py -v` → here: `pytest cognee/tests/unit/infrastructure/databases/graph/test_neo4j_get_edges.py`
- [x] My PR passes all required CI/CD checks (to be observed on this PR)
- [x] My PR's scope is as isolated as possible; it only solves 1 specific problem
- [ ] I have received a Greptile Confidence Score of at least 4/5 before requesting a maintainer review

## Proof of fix

Verified against a live Neo4j 5 container (Docker `neo4j:5`), seed: `node-a`, `node-b`, edges `node-a -[:works_on]-> node-b` and `node-b -[:reports_to]-> node-a`:

### Before (6c022454e, dev)

```
get_edges("node-a") =
  [('node-a', 'node-b', {'relationship_name': 'works_on'}),
   ('node-a', 'node-b', {'relationship_name': 'reports_to'})]   # incoming edge inverted
```

### After (7526c3eff, this PR)

```
get_edges("node-a") =
  [('node-a', 'node-b', {'relationship_name': 'works_on'}),
   ('node-b', 'node-a', {'relationship_name': 'reports_to'})]   # true direction
get_edges("node-b") =
  [('node-a', 'node-b', {'relationship_name': 'works_on'}),
   ('node-b', 'node-a', {'relationship_name': 'reports_to'})]   # consistent from both sides
```

## Type

🐛 Bug Fix

## Final Attestation

- [x] The tests check the right things, including the edge cases, and regressions in the respective real-world customer use-cases are not possible after this PR

New tests (`cognee/tests/unit/infrastructure/databases/graph/test_neo4j_get_edges.py`): the direction regression asserts an incoming edge is reported `(true_source, queried_node)`; the query-shape test asserts `startNode(r).id AS source_id` is projected. Both fail at the previous head and pass with the fix. The touched suite plus the neighboring neo4j suites run 10/10 locally.

### DIFF
--- a/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
+++ b/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
@@ -1310,7 +1310,9 @@ class Neo4jAdapter(GraphDBInterface):
         return [
             (
                 result["source_id"],
-                result["m"]["id"] if result["n"]["id"] == result["source_id"] else result["n"]["id"],
+                result["m"]["id"]
+                if result["n"]["id"] == result["source_id"]
+                else result["n"]["id"],
                 {"relationship_name": result["r"][1]},
             )
             for result in results
--- a/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
+++ b/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
@@ -1292,17 +1292,27 @@ class Neo4jAdapter(GraphDBInterface):
         Returns:
         --------
 
-            A list of edges connecting to the specified node, represented as tuples of details.
+            A list of edges connecting to the specified node as
+            (source_id, target_id, {"relationship_name": name}) tuples in true
+            edge direction, regardless of which endpoint the traversal started
+            from.
         """
+        # `r` serializes through .data() as (start_props, type, end_props), so
+        # the type is read positionally; start(n) keeps the true source id even
+        # when the undirected MATCH traverses the edge backwards.
         query = f"""
         MATCH (n: `{BASE_LABEL}`{{id: $node_id}})-[r]-(m)
-        RETURN n, r, m
+        RETURN n, r, m, startNode(r).id AS source_id
         """
 
         results = await self.query(query, dict(node_id=node_id))
 
         return [
-            (result["n"]["id"], result["m"]["id"], {"relationship_name": result["r"][1]})
+            (
+                result["source_id"],
+                result["m"]["id"] if result["n"]["id"] == result["source_id"] else result["n"]["id"],
+                {"relationship_name": result["r"][1]},
+            )
             for result in results
         ]
 
--- a/cognee/tests/unit/infrastructure/databases/graph/test_neo4j_get_edges.py
+++ b/cognee/tests/unit/infrastructure/databases/graph/test_neo4j_get_edges.py
@@ -1,0 +1,83 @@
+import pytest
+
+pytest.importorskip("neo4j")
+
+from cognee.infrastructure.databases.graph.neo4j_driver.adapter import BASE_LABEL, Neo4jAdapter
+
+
+def _make_adapter() -> Neo4jAdapter:
+    return Neo4jAdapter(
+        "bolt://unused",
+        graph_database_allow_anonymous=True,
+        driver=object(),
+    )
+
+
+@pytest.mark.asyncio
+async def test_get_edges_reports_true_edge_direction_for_incoming_edges():
+    """get_edges walks the relationship undirected, so an edge stored as
+    (other) -[:reports_to]-> (queried node) surfaces with n = the queried node.
+    The contract is (source_id, target_id, ...) in TRUE edge direction (the
+    consolidation pipeline renders 'neighbor —relationship→ node'), so the
+    source must come from startNode(r), not from which endpoint the traversal
+    happened to bind first."""
+    adapter = _make_adapter()
+    adapter.query = _FakeQuery(
+        [
+            {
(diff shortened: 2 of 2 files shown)


xu-jia-ming:
CI triage:

- `Code Quality / ruff format` was flagging our long conditional in `adapter.py` — fixed in 5a3d6a9e9 (the exact reformat ruff asked for), so the format half now passes.
- The remaining `Code Quality / Ruff check` failure is pre-existing on `dev` and unrelated to this PR: all findings are in files this PR never touches (`.github/scripts/motherduck_nightly_etl.py` BLE001, `cognee-mcp/src/*` I001, `tools/*.py`, `cognee/eval_framework/beam/eval/beam_eval_adapter.py`, ...). Our two files (`adapter.py`, `test_neo4j_get_edges.py`) have zero findings. Same failures reproduce on other open PRs (#4835 triage), while `dev` push runs skip the Ruff job, which is why `dev` itself looks green.
- The mass ~5s failures across the test matrix are the fork-PR secret restrictions (jobs exit before executing any steps).

Touched files are limited to `cognee/infrastructure/databases/graph/neo4j_driver/adapter.py` + its new test file; the touched suite runs 10/10 locally.

xu-jia-ming:
Follow-up on the updated run: the matrix jobs now execute further (instead of the 0-step startup exits) and fail mid-run with `LLMAPIKeyNotSetError: LLM API key is not set (422)` — the E2E suites drive a full cognify pipeline, which needs the repo's LLM secrets that fork PRs don't receive. Same signature as the already-merged #4838's branch and #4835's run (their Neo4j/E2E jobs failed identically before merge), so it's not introduced by this change. The jobs that can run without secrets (basic checks, Docker builds, lockfile/pre-commit format checks relevant to the touched files) pass; `Code Quality / Ruff check` findings remain the pre-existing `dev` ones in untouched files.

dexters1:
Thanks for the PR @xu-jia-ming! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

ArmagedonFlamer:
Thanks @xu-jia-ming for the careful diagnosis in #4967 and for verifying the fix against a live Neo4j container. Since this PR was opened, #4613 rewrote entity-description consolidation to read incident edges through get_connections() and take direction from each edge's own source_node_id, so the swapped-endpoint bug no longer reaches that pipeline. 
On current dev get_edges has no callers left in library code, and this patch would also leave Neo4j returning id tuples where Ladybug returns node dicts. We are closing this rather than merging a change to an unused method, but the investigation itself was solid and appreciated. 
Work like this is exactly what keeps cognee getting better, and we'd be really glad to see more from you.

xu-jia-ming:
Thanks for the thorough explanation — makes sense. I confirmed on current `dev` (`fa116e32f`) that `get_edges` has no library-code callers left after the #4613 consolidation rewrite, so there's nothing user-visible left to fix and closing is the right call.

Good to know the direction information now flows from each edge's own `source_node_id` through `get_connections()` — that's a cleaner design than reconstructing it in the adapter.

Appreciate the kind words; I have a few more Neo4j/graph-layer ideas from auditing this adapter, so you'll likely see more PRs from me.

PR #4970, opened 2026-09-08
ci: Mock the example jobs and move real-model jobs to nightly (SDK-533)

NMZivkovic:
The examples suite made real OpenAI calls on every PR push — ~13 jobs, ~45 pushes a weekday — to run zero-assert "the documented example still runs" scripts. This PR keeps every one of those checks on the gate while removing the paid model calls, and moves the jobs whose value **is** the real model call to the nightly.

## Mocked on the gate (10 jobs)

All example jobs except docling-slim (already LLM-free) now run through the new `cognee/tests/utils/run_mocked.py`: `install_mocks({})` turns the existing cassette-replay infrastructure into a generic mock — every structured-output call returns its response model's default instance, embeddings zero-vector via the built-in `MOCK_EMBEDDING` switch, and chunking/graph/vector writes stay fully real. The jobs carry inert `mock-not-used` keys; no LLM secret is loaded at all.

That required one infra fix: the fallback was `response_model()`, which raises for models with required fields (the memify example hits this). `_default_instance` now synthesizes a valid instance for arbitrary models — scalars, lists, dicts, Optionals, Literals, enums, UUIDs, nested models — pinned by unit tests including an end-to-end `install_mocks({})` gateway call, mutation-checked (reverting the fallback fails the pin test).

Verified locally keyless (any real call would 401): custom-pipeline, memify, feedback (with `CACHE_BACKEND=fs`), temporal, ontology, agentic, permissions — all pass end to end. Metrics/S3/docling variants rely on the same mechanism and are validated by this PR's own run.

## Moved to the nightly (6 jobs)

`test-multimedia-example` and `test-eval-example` cannot be mocked — real transcription/evaluation is the test. They move to `nightly_tests.yml` (daily against main, weekly against dev), wired into its status gate, together with the Bedrock auth-variant trio and OpenRouter from `test_llms` — those differ from the remaining gate canaries (Gemini, Anthropic, Fastembed) only in credential plumbing. Blocks are verbatim minus the reusable-workflow `container:` line (no `ci-image` input exists in the nightly).

**Placement note for the reviewer:** the 2026-09-03 restructure made the nightly perf-only and moved correctness suites to `release_test.yml`. These six land in the nightly per Nikola's direction; if the perf-only split should hold, relocating them to `release_test.yml` is a straightforward block move — flagging rather than deciding unilaterally.

## Net effect

Real-token CI spend per PR push drops ~70% (the `OPENAI_API_KEY`-consuming job count goes from ~25 to ~8: OS-matrix multimedia steps, notebooks, BAML, and the three provider canaries' embeddings). Nothing a merge proves is lost: the mocked jobs still catch every API-shape regression they ever caught, and the real-model coverage runs daily instead of ~45×/day.


dexters1:
@NMZivkovic should we move this to the release_test github actions? Or do you think those actions do not run often enough? (in which case we can also run them weekly and give a report in a new slack channel)

I'd ideally like to keep the nightly CI report small in slack it is already hard to keep track of the nightly report as is

NMZivkovic:
Agreed — moved them to `release_test.yml` in 8c4cfdeb9, so nothing is added to the nightly Slack message. That also puts them where the suites you moved on the 3rd already live (unit-matrix-mid, search-db-versions, ollama, llamacpp, telemetry, the delete tests), rather than inventing a second home for "runs less often than every push".

The tradeoff, so it's on the record: release_test runs on dispatch and release PRs rather than daily, so these six lose daily cadence. If that turns out to matter for the Bedrock/OpenRouter provider contracts specifically — those are the ones that catch an SDK-vs-provider wire-format break — your weekly-plus-new-channel idea is the better home and I'm happy to wire it.

Two other things landing on this PR from Nikola's review, so you're not surprised by the diff:

- The perf arms get slimmed: the five `50_small_documents` arms are deleted (same shape as datasheets, which is real data), `war_and_peace` moves to weekly, and datasheets becomes the only daily arm. That shrinks the daily Slack table from 14 arms to 3, which should help the readability problem more than the six jobs did.
- While verifying that, I found the corpus rename's S3 step was never done: `nightly_ci_artifacts/performance_test_artifacts/` still holds old file names, and the workflows on dev now ask for `datasheets.json`. Once the shim reaches main, every datasheets arm fails on a missing key — and after this change datasheets *is* the daily nightly. Flagged to Nikola with the copy commands.

.github/workflows/nightly_tests.yml
```diff
@@ -196,6 +176,12 @@ jobs:
     secrets: inherit
   perf-rust-wap:
     name: Performance — Rust SDK — War and Peace (mock LLM)
+    # Weekly only. The daily cadence runs the datasheets corpus alone; War and
+    # Peace -- and perf-wap-large-mock, a 27x-inflated replay of the same book
+    # and the slowest job in the nightly -- is a much larger budget for the
+    # same signal, so it earns a slot once a week. `manual` and any unset
+    # cadence run EVERYTHING: a hand dispatch exists to validate the whole graph.
+    if: ${{ inputs.cadence != 'daily' }}
```
dexters1: Lets keep the mock LLM runs as part of the nightly runs

NMZivkovic: Agreed, done in 002f3967a — `perf-wap-mock` and `perf-rust-wap` are back on the daily cadence.

You're right for a reason my first cut missed: the gate that matters is model spend, not which corpus a job reads. A mock arm bills nothing and its timings carry no API latency, so it's the cleanest night-over-night signal we produce — exactly the series that wants a datapoint every night rather than 52 a year. They're cheap too, from the last full nightly: `perf-wap-mock` 4-5 min, `perf-rust-wap` 8 min. The weekly gate now reads as "arms that cost real model or cloud budget", and I rewrote the rationale comments to say that instead of the old corpus-based story.

**One deliberate exception, so it's your call rather than my oversight:** `perf-wap-large-mock` stays weekly. It is a mock arm, so it's technically in scope for what you asked. But it ran **70 min** on the last nightly — nine times the next-slowest mock arm and longer than every real-LLM arm — and its corpus is the 27x-inflated replay, not real data. Daily felt like buying runtime rather than signal, especially since it's the one arm Nikola's "daily = real data only" call was aimed at. Say the word and I'll un-gate it too; it's a one-line change.

Daily is now 5 arms (3 datasheets + the 2 mock War and Peace) against 14 before the restructure, so the Slack report stays small — which was your original ask.

The status step and the Slack renderer needed no changes: both already treat a cadence-skipped arm generically, so nothing there is pinned to which arms happen to be gated.

dexters1 (approved):

PR #4971, opened 2026-09-08
feat: Add item-level processing status endpoint

siillee:
## Description

Adds `GET /api/v1/datasets/{dataset_id}/processing-status`, which returns
item-level incremental completion counts for a dataset:

```json
{ "total": 12, "completed": 9, "pending": 3 }
```

GET /datasets/status only reports the dataset-level pipeline-run state. Operators asking
"which items are done vs pending?" previously had to query Data.pipeline_status directly.

Changes

- New route in get_datasets_router.py. Optional ?pipeline= query param, default
  cognify_pipeline. 404 when the dataset is missing or not readable by the caller, 409 on
  unexpected errors, matching the existing dataset routes.
- is_data_item_completed() in DataItemStatus.py: one predicate for the per-item stamp.
  Accepts the legacy bare string DATA_ITEM_PROCESSING_COMPLETED and the dict form
  {"status": "DATA_ITEM_PROCESSING_COMPLETED", ...} planned for the hash rollout.
- get_dataset_processing_status() in cognee/modules/data/methods/: counts over
  get_dataset_data(). NULL status columns and stamps for other datasets or pipelines count
  as pending.
- Existing readers switched to the predicate: the skip check in run_tasks_data_item.py
  and is_data_processed() in publish_updated_data.py. Behaviour for string stamps is
  unchanged; the skip check also gains the NULL guard the other readers already had.

Backward compatibility

- Additive only: new route, new exports, no changes to the Data table, migrations, or writers.
- The predicate is a strict superset of the old equality check, so every existing stamp still
  counts as completed.
- The route cannot shadow /status or /status/progress (literal second segment).

Out of scope

- Per-item errored status (not persisted today).
- CLI / MCP wrappers for the new endpoint.

Tests

- test_data_item_status.py: both representations, plus negatives (None, other strings,
  dict without a completed status).
- test_get_dataset_processing_status.py: empty dataset, fully completed, mixed, legacy string
  and dict side by side, NULL status, other-dataset/other-pipeline stamps, pipeline override.
- test_dataset_processing_status_router.py: auth enforced, 200 body, 404, 409, 422,
  ?pipeline= forwarded.
- Import-cycle guard extended to the new helper. Existing pipeline, incremental-update, and
  datasets router suites pass unchanged (203 tests).

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation


dexters1:
Do we also return which ones are in the completed state which are in the pending state as additional information? With their data_id?

siillee:
> Do we also return which ones are in the completed state which are in the pending state as additional information? With their data_id?

@dexters1 yes, a later commit adds this as an opt-in via `?include_items=true`, but we can make it default. It's something I meant to discuss with the team anyways

dexters1:
> > Do we also return which ones are in the completed state which are in the pending state as additional information? With their data_id?
> 
> @dexters1 yes, a later commit adds this as an opt-in via `?include_items=true`, but we can make it default. It's something I meant to discuss with the team anyways

If it's a dictionary return value always it can always be on, don't see a reason right now to make it opt-in / opt-out unless it changes the return value type

NMZivkovic (approved):

PR #4977, opened 2026-09-08
fix(pgvector): scope prune to vector-owned tables

### LINKED ISSUE #4956: [Bug]: prune_system(vector=True) drops the entire relational schema on PGVector — metadata=False offers no protection
## Summary

On a PostgreSQL + PGVector deployment, `cognee.prune.prune_system(vector=True, metadata=False)` deletes **every table in the `public` and `public_staging` schemas**, including `data`, `datasets`, `users` and `alembic_version`. The documented safety flag `metadata=False` does not prevent this, because the destruction does not go through the metadata path at all — it goes through the *vector* path.

This is data loss with the documented safety flag set, on a configuration the docs present as supported.

## What the documentation promises

[`python-api/prune`](https://docs.cognee.ai/python-api/prune) documents the parameter as:

> **`metadata`** (bool, default `False`) — Delete the relational metadata database (datasets, data records, pipeline state). Defaults to `False` to prevent accidental loss of metadata.

and describes `vector` as:

> **`vector`** (bool, default `True`) — Delete all data from the vector store (e.g., LanceDB, Qdrant).

Both examples name vector stores that live **outside** the relational database. Nothing on the page warns that with PGVector the two are the same database — or, as shown below, the same *engine*.

## Root cause

`PGVectorAdapter` inherits from `SQLAlchemyAdapter` and, when the relational store is PostgreSQL, **borrows the relational engine outright** ([`PGVectorAdapter.py#L158-L161`](https://github.com/topoteretes/cognee/blob/e93a4f0c76e6af27142c25189f14c200e1749d51/cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py#L158-L161)):

```python
elif relational_db.engine.dialect.name == "postgresql":
    # If postgreSQL is used and not backend access control we must use the same engine and sessionmaker
    self.engine = relational_db.engine
    self.sessionmaker = relational_db.sessionmaker
```

Its `prune()` then delegates to the inherited `delete_database()` ([`PGVectorAdapter.py#L865-L868`](https://github.com/topoteretes/cognee/blob/e93a4f0c76e6af27142c25189f14c200e1749d51/cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py#L865-L868)):

```python
async def prune(self):
    """Drop all vector collection tables and reset cached reflection metadata."""
    self._metadata.clear()
    await self.delete_database()
```

And `SqlAlchemyAdapter.delete_database()` reflects and drops **everything in the schema**, with no filter on vector collections ([`SqlAlchemyAdapter.py#L762-L771`](https://github.com/topoteretes/cognee/blob/e93a4f0c76e6af27142c25189f14c200e1749d51/cognee/infrastructure/databases/relational/sqlalchemy/SqlAlchemyAdapter.py#L762-L771)):

```python
schema_list = ["public", "public_staging"]
for schema_name in schema_list:
    await connection.run_sync(metadata.reflect, schema=schema_name)
    for table in metadata.sorted_tables:
        drop_table_query = text(
            f'DROP TABLE IF EXISTS {schema_name}."{table.name}" CASCADE'
        )
        await connection.execute(drop_table_query)
    metadata.clear()
```

The docstring says "Drop all vector collection tables". The implementation drops all tables, full stop.

`prune_system` then never reaches its own `metadata` branch for this damage — the relational schema is already gone by the time that flag is consulted.

## Affected configuration

The vulnerable branch is taken when **all** of the following hold:

- relational provider is PostgreSQL,
- vector provider is `pgvector`,
- no per-dataset schema isolation (`self.schema` unset),
- backend access control disabled, or enabled with the same database name.

This is the standard single-database Postgres deployment.

## Reproduction

```python
import asyncio, cognee

async def main():
    await cognee.add("hello world", dataset_name="demo")
    await cognee.cognify(datasets=["demo"])

    # metadata=False is documented as preserving the relational database
    await cognee.prune.prune_system(
        graph=True, vector=True, metadata=False, cache=False
    )

asyncio.run(main())
```

This is the exact call we ran. The destruction comes from the vector path, so
`vector=True` alone should be sufficient to trigger it — but we did not test that
variant in isolation, and I would rather not claim it untested.

Then, against the same database:

```sql
SELECT count(*) FROM pg_tables WHERE schemaname = 'public';
```

**Expected:** the relational tables survive; only `DocumentChunk_text`, `Entity_name`, `TextSummary_text` and the other `{Type}_{field}` collections are dropped.

**Actual:** `0`. `data`, `datasets`, `users`, `pipeline_runs` and `alembic_version` are gone.

The API server then answers `500` on every endpoint until it is restarted, and the first error surfaces as:

```
record_operation: failed to persist prune_system record
  asyncpg.exceptions.UndefinedTableError: relation "pipeline_runs" does not exist
```

which is the operation-logging step failing to write into a table its own call had just dropped.

## Impact

Raw ingested files under `DATA_ROOT_DIRECTORY` survive (`prune_data()` is a separate call), so content is recoverable — but the registry is not. Lost: every `Data` row, every dataset, and the mapping from stored file to document name and dataset. Recovery requires re-creating the dataset and re-ingesting every source, and the dataset UUID changes, breaking anything that referenced it by id.

We hit this on a production knowledge base of 140 documents. Restarting the service rebuilds the schema (`run_migrations` correctly treats a database with neither `users` nor `alembic_version` as fresh), and `POST /api/v1/datasets` re-creates the dataset, but everything else had to be re-ingested.

## Suggested fix

Either of these would have prevented it:

1. Make `PGVectorAdapter.prune()` drop only its own collection tables, matching its docstring — the adapter already knows which collections it manages, and `SQLAlchemyAdapter.delete_database()` is the wrong primitive to reuse when the engine is borrowed.
2. Refuse to run at all when `self._owns_engine` is `False`: a borrowed engine means the vector store shares the relational database, and a whole-schema drop is never the intended scope there.

A documentation note would help but is not sufficient on its own — the current behaviour makes `metadata=False` actively misleading rather than merely underspecified.

## Environment

- cognee `1.5.4` (latest release; the code above is identical on `main` at `e93a4f0`)
- Docker image `cognee/cognee:1.5.4`
- Relational: PostgreSQL 17 (`pgvector/pgvector:pg17`), single database shared with the vector store
- Vector: `pgvector`
- Graph: Neo4j 5.26
- Embeddings: Ollama / `nomic-embed-text`

## Related

`#3347` (closed) also concerns `prune_system` on PostgreSQL, but for stale asyncpg prepared statements after a prune/setup cycle — a different failure mode, and not data loss.

pandasuwu:
Fixes #4956.

## Description

`PGVectorAdapter.prune()` currently calls the inherited `SqlAlchemyAdapter.delete_database()`, which drops every reflected table in the `public` and `public_staging` schemas.

This becomes a problem when PostgreSQL is being used as both the vector and relational provider. In that setup, `PGVectorAdapter` borrows the relational provider's engine rather than owning one itself (`_owns_engine = False`). As a result, calling `prune()` on the vector adapter can also drop relational tables such as `users`, `datasets`, `acls`, and `alembic_version`.

This means that `prune_system(vector=True, metadata=False)` can still delete metadata, despite `metadata=False` being intended to preserve it. The issue is also reachable through the public SDK, since `cognee.prune.prune_system()` defaults to `metadata=False`.

The main issue here is that the vector adapter does not currently distinguish between an engine it owns and one that is shared with the relational provider.

## Changes

`prune()` now branches on `_owns_engine`:

- **Owned engine** - unchanged. That database is dedicated to vectors, so dropping the schema is the intended behaviour.
- **Borrowed engine** - drops only tables whose column shape matches what `create_collection` builds: a pgvector-typed `vector` column alongside `id` and `payload`, scoped to `current_schema()`. It logs what it dropped and from which schema.

`delete_database()` is deliberately left alone, since other adapters rely on its current behaviour.

Also documented the `metadata=False` contract on `prune_system` and on its SDK wrapper, since that's the invariant that was violated.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Acceptance Criteria

`prune_system(vector=True, metadata=False)` must leave the relational schema intact on PGVector while still dropping the vector collections. Evidence below.

## Testing

Verified against `pgvector/pgvector:pg16` with Postgres serving as both the relational and the vector backend, driven through the public entry point the issue reports — `cognee.prune.prune_system(vector=True, metadata=False)`, with the schema built by the real migrations. The issue's reproducer uses `add()` + `cognify()` to get collections into the database; I create them directly instead, so the run needs no LLM and the only thing being measured is the prune.

Before the fix:

```
BEFORE (37 tables in public)
AFTER  (0 tables in public)

relational tables destroyed : ['acls', 'alembic_version', 'data', 'datasets', 'users']
STATUS: BUG PRESENT via prune_system(metadata=False)
```

After the fix, same database, same call:

```
BEFORE (37 tables in public)
AFTER  (35 tables in public)

Dropped: DocumentChunk_text, Entity_name
relational tables destroyed : <none>
vector collections left     : <none>
STATUS: OK - prune_system(metadata=False) honoured its contract
```

The same before/after holds one level down, calling `PGVectorAdapter.prune()` directly:

```
BEFORE prune (37 tables in public):
  DocumentChunk_text, Entity_name, acls, alembic_version, data, dataset_configurations,
  dataset_database, datasets, edges, global_database_version, graph_metrics, ... users

AFTER prune (0 tables in public):
  <none>
```

After the fix, same database, only the two collection tables go:

```
AFTER prune (35 tables in public):
  acls, alembic_version, data, dataset_configurations, dataset_database, datasets, edges,
  global_database_version, graph_metrics, ... users

Dropped by prune(): DocumentChunk_text, Entity_name
```

New file: `cognee/tests/unit/infrastructure/databases/vector/test_pgvector_prune_scope.py`. Four tests against a real Postgres, skipping when it's unreachable, following the pattern in `test_update_payload.py`. They let `__init__` pick the branch on its own rather than forcing `_owns_engine` after construction, since the branch selection is part of what's under test.

- borrowed engine preserves `users` / `datasets`
- borrowed engine preserves `alembic_version` and its recorded revision
- owned engine still drops everything (pins the existing behaviour)
- a table that merely *looks* vector-ish is left alone: same column names but no pgvector type, or a pgvector column with no `payload`

With the fix: 4 passed. Reverting only the adapter change: 3 failed, 1 passed, the survivor being the owned-engine test.

```
pytest cognee/tests/unit/infrastructure/databases/   →  639 passed, 14 skipped
ruff format --check / ruff check (pinned 0.15.11)    →  clean
```

I haven't run `cognee/tests/test_library.py` locally. It runs on SQLite + LanceDB + Kuzu, where `PGVectorAdapter` is never imported, so it can't reach anything this PR touches — the PGVector runs above are the ones that cover the change. Leaving it to CI.

## Review Notes

* `_owns_engine` is used to determine whether the vector adapter is sharing its database with the relational layer. This works regardless of how `prune()` is reached, including through the hybrid Postgres adapter.

* The table check uses column shape rather than the PascalCase naming convention because this is a deletion path. Accidentally leaving a vector table behind is safer than accidentally dropping a relational table like `users`.

* I added a log line recording which tables the borrowed-engine path drops, and from which schema, so it's easy to audit if the detection ever behaves unexpectedly.

* Another option would be to refuse pruning when the engine is borrowed and require `metadata=True`. I didn't use that approach because `prune_system` calls vector pruning before dropping the relational database, so raising there would also break `prune_system(metadata=True)`.

## Screenshots

**1. New regression tests (4 passed)**

**2. Same tests against the pre-fix adapter (3 failed, 1 passed) — proves they catch the bug**

**3. `cognee/tests/unit/infrastructure/databases/` — 639 passed, 14 skipped**

**4. Before/after through `prune_system(metadata=False)` — 37 → 0 tables vs 37 → 35**

## DCO Affirmation

### DIFF
--- a/cognee/api/v1/prune/prune.py
+++ b/cognee/api/v1/prune/prune.py
@@ -9,6 +9,11 @@ class prune:
 
     @staticmethod
     async def prune_system(graph=True, vector=True, metadata=False, cache=True):
+        """Drop the backing stores. Note ``metadata`` defaults to False here.
+
+        The default therefore clears the graph, the vectors and the cache while
+        keeping the relational schema (users, datasets, ACLs, migration state).
+        """
         await _prune_system(graph, vector, metadata, cache)
 
 
--- a/cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py
+++ b/cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py
@@ -37,6 +37,12 @@ QUERY_BATCH_SIZE = 1000
 # connections close on release instead of idling.
 _ACCESS_CONTROL_DEFAULT_POOL_ARGS = {"pool_size": 2, "max_overflow": 20}
 
+# Stamped as a table comment on every collection this adapter creates, and the only
+# thing prune() trusts when the engine is shared with the relational layer. Column
+# shape can't stand in for it: an app table holding an id, a JSON payload and an
+# embedding looks exactly like a collection (#4956). Compared exactly, not by prefix.
+_COLLECTION_OWNERSHIP_MARKER = "cognee:pgvector-collection:v1"
+
 
 class IndexSchema(DataPoint):
     """
@@ -93,6 +99,9 @@ class PGVectorAdapter(SQLAlchemyAdapter, VectorDBInterface):
         self.VECTOR_DB_LOCK = asyncio.Lock()
         self._write_locks: dict[str, asyncio.Lock] = {}
         self._metadata = MetaData()
+        # Collections already marked by this instance, so the backfill costs one
+        # statement per collection rather than one per create_collection() call.
+        self._marked_collections: set[str] = set()
         # True when this adapter created its own engine and must dispose it on close().
         # False when the engine is borrowed from the relational adapter.
         self._owns_engine: bool = False
@@ -327,11 +336,18 @@ class PGVectorAdapter(SQLAlchemyAdapter, VectorDBInterface):
                             await connection.run_sync(
                                 Base.metadata.create_all, tables=[PGVectorDataPoint.__table__]
                             )
+                            # Same transaction as the DDL, so the table and its
+                            # ownership marker can never exist without each other.
+                            await self._mark_collection_owned(connection, collection_name)
+                            self._marked_collections.add(collection_name)
                     # Reflect AFTER the DDL transaction commits so
                     # _metadata is never populated for a table that
                     # might be rolled back.
                     async with self.engine.begin() as connection:
                         await connection.run_sync(self._metadata.reflect, only=[collection_name])
+                    return
+
+        await self._backfill_collection_marker(collection_name)
 
     @retry(
         retry=retry_if_exception_type((DeadlockDetectedError, DBAPIError)),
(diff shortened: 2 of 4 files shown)


linhongyu510:
Thanks for taking this on. I independently checked the borrowed-engine classifier against a disposable `pgvector/pgvector:pg17` database and found one remaining data-loss boundary: column shape is not proof of adapter ownership.

I created the real collection shape and an unrelated application table with the same required columns plus one extra column:

```sql
CREATE TABLE "Entity_name" (
  id uuid PRIMARY KEY, payload json, vector vector(4)
);
CREATE TABLE user_embeddings (
  id uuid PRIMARY KEY, payload json, vector vector(4),
  created_at timestamptz DEFAULT now()
);
```

Running `_OWNED_VECTOR_TABLES_SQL` returned both:

```text
 Entity_name
 user_embeddings
```

So `prune()` would still delete an unrelated user table whenever it legitimately stores an `id`, JSON payload, and pgvector embedding. The new near-miss test covers missing/wrong columns, but not this exact-shape collision. This is especially important because the path is intended to protect a shared relational schema from destructive ownership mistakes.

Could the borrowed-engine path use an authoritative collection registry/metadata instead of structural inference, and fail closed when ownership cannot be proven? I would add a regression where an unrelated exact-shape table (preferably with extra columns) survives while a registered Cognee collection is removed.

Reproduction environment: disposable PostgreSQL 17 + pgvector container; the classifier query completed successfully and returned both rows.

linhongyu510:
I verified the revised marker-based implementation independently on the latest head (`ba8241a2b`) using a disposable `pgvector/pgvector:pg17` database. The ownership behavior now matches the requested fail-closed boundary, including exact-shape application tables, forged markers, unmarked legacy collections, and marker backfill.

Runtime results:

```text
pytest test_pgvector_prune_scope.py: 8 passed
pytest cognee/tests/unit/infrastructure/databases/vector/: 58 passed, 14 skipped
git diff --check: clean
```

I did find the current head fails the repository Ruff gate with one `BLE001` in `_backfill_collection_marker` and seven `SIM117` findings in the new test file. I prepared a minimal signed-off cleanup commit that narrows the catch to `exc.SQLAlchemyError` and combines the nested async context managers, without changing the tested behavior:

https://github.com/linhongyu510/cognee/commit/5ff7b273b

```bash
git cherry-pick 5ff7b273b
```

After that commit:

```text
ruff check: passed
ruff format --check: passed
PGVector regression: 8 passed
vector unit directory: 58 passed, 14 skipped
```

dexters1:
Thanks for the PR @pandasuwu! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #4983, opened 2026-09-08
SDK-591 fix: Report abandoned pipeline runs at read time

goran-radonic:
## Description

Pipeline runs stuck at `STARTED` after a worker crashed showed up in the Activity feed as "RUNNING" forever, with elapsed time growing unbounded (observed 29h, 98h, 100h+ in cognee-saas-frontend). `get_pipeline_runs` had no `ABANDONED` state and no staleness check at all, unlike the `/v1/sessions` endpoint which already handles this via `get_effective_status_sql()`.

This adds a read-time staleness check to `/activity/pipeline-runs`: a `pipeline_runs` row stuck at `DATASET_PROCESSING_STARTED` for longer than `PIPELINE_RUN_ABANDON_AFTER_SECONDS` (default 1800s / 30 min) is now reported as `"ABANDONED"` in the API response. Nothing is written back to the row and `PipelineRunStatus` (a native Postgres enum) is left untouched, so this needs no Alembic migration — same approach as the sessions side.

Scope: only the read-time detection (parts 1+2 of the linked ticket). The worker-side termination path (writing a terminal status on crash/timeout) is a separate, larger piece of work and is intentionally left out of this PR.

Ran this through an adversarial review pass on myself before opening the PR (4 reviewers across lifecycle/config/error-paths/tests angles, then a verifier). Caught and fixed:
- the threshold env var accepted negative/zero/absurdly large values — negative made almost every in-flight run instantly report ABANDONED, and a huge value crashed the endpoint with an uncaught `OverflowError`. Now clamped with a sane default fallback.
- the boundary test didn't actually pin `<` vs `<=` at the threshold — added a precise boundary test.
- the docstring implied a row transitions in place when a straggler worker finishes; in reality each writer always inserts a new row, so the ABANDONED row stays ABANDONED forever and a separate terminal row appears later. Reworded to match reality and cross-reference the existing dedup guidance.
- added a docstring caveat that this is a heuristic (no heartbeat column exists), so a legitimately slow pipeline can read as ABANDONED too — worker-side heartbeat is future work.

Linear: SDK-591

## Acceptance Criteria

- A `pipeline_runs` row stuck at `DATASET_PROCESSING_STARTED` past the configured threshold is reported as `"ABANDONED"` by `GET /activity/pipeline-runs`, without any write to the row or the `PipelineRunStatus` enum.
- Fresh `STARTED` rows, and any row in a terminal status (`COMPLETED`/`ERRORED`) regardless of age, are unaffected.
- Operation records (SDK-399 rows with no `pipeline_name`) are unaffected.
- The threshold is configurable via `PIPELINE_RUN_ABANDON_AFTER_SECONDS`, documented in `.env.template`, and malformed/out-of-range values fall back safely to the default instead of misbehaving or crashing the endpoint.
- All new and existing tests in `test_get_activity_router_pipeline_runs.py` pass.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation


cognee/modules/pipelines/methods/get_effective_pipeline_status.py
```diff
@@ -0,0 +1,59 @@
...
+    session records.
+
+    Only applies to pipeline rows (pipeline_name set) — SDK-399 operation
+    rows have no status column at all and are untouched.
+    """
+    if run.status is None:
+        return None
+    if run.pipeline_name is None:
+        return run.status.value
+    if run.status != PipelineRunStatus.DATASET_PROCESSING_STARTED:
```
NMZivkovic, line 161: **[Major] Every successfully completed run older than the threshold now shows an ABANDONED row**

The guard is per *row*, but a run is several rows. `/activity/pipeline-runs` applies no dedup — the query is a plain `select(PipelineRun) ... order_by(created_at desc)` with no `DISTINCT`, no `ROW_NUMBER` and no `group_by` — so a completed run's `STARTED` row is returned alongside its `COMPLETED` row, and that `STARTED` row is older than 30 minutes for **every run that happened more than 30 minutes ago**.

Ran the real helper against exactly that shape:

```
A cognify run that STARTED 2h ago and COMPLETED 10s later.
Both rows share pipeline_run_id and both are returned by /activity/pipeline-runs:

  STARTED   row -> 'ABANDONED'
  COMPLETED row -> 'DATASET_PROCESSING_COMPLETED'
```

So the feed gains an `ABANDONED` entry for a run that succeeded — and for essentially the whole run history, not an edge case. The acceptance criterion *"any row in a terminal status (COMPLETED/ERRORED) regardless of age is unaffected"* holds literally, but it is scoped to the row rather than to the run, which is probably why this slipped through.

Before this PR that row read `DATASET_PROCESSING_STARTED`, which is at least what the row actually says. `"ABANDONED"` is an active assertion of failure, so for completed runs this moves from confusing to wrong.

The fix has to know about the sibling rows. Either resolve staleness per `pipeline_run_id` (a `STARTED` row is abandoned only when no terminal row shares its id — the same test `get_unclosed_pipeline_runs` performs in #4966), or collapse to one row per run before mapping status. The first is a `NOT EXISTS` in the query; the second changes the response shape, so probably the first.

cognee/modules/pipelines/methods/get_effective_pipeline_status.py
```diff
@@ -0,0 +1,59 @@
...
+    if run.created_at is None:
+        return run.status.value
+    threshold = datetime.now(timezone.utc) - timedelta(
+        seconds=_pipeline_run_abandon_after_seconds()
+    )
+    created_at = run.created_at
+    if created_at.tzinfo is None:
+        created_at = created_at.replace(tzinfo=timezone.utc)
+    if created_at < threshold:
+        return "ABANDONED"
```
NMZivkovic: **[Minor] Casing diverges from the sessions equivalent this mirrors**

`SessionStatus.ABANDONED` is `"abandoned"` (lowercase) in `session_lifecycle/metrics.py:44`, and `/v1/sessions` emits it that way. This returns `"ABANDONED"`.

Both are read-time computed statuses on sibling endpoints, and the PR description frames this as the same approach as the sessions side, so a client handling both now needs two spellings of one concept. The stored `PipelineRunStatus` values are uppercase, so uppercase is defensible here — worth making it a deliberate call rather than an accident.

cognee/modules/pipelines/methods/get_effective_pipeline_status.py
```diff
@@ -0,0 +1,59 @@
...
+    """
+    if run.status is None:
+        return None
+    if run.pipeline_name is None:
+        return run.status.value
+    if run.status != PipelineRunStatus.DATASET_PROCESSING_STARTED:
+        return run.status.value
+    if run.created_at is None:
+        return run.status.value
+    threshold = datetime.now(timezone.utc) - timedelta(
```
NMZivkovic: **[Minor] `now` and the env var are re-read per row**

This helper is called once per row in the response mapping, so a 500-row page re-reads `os.environ`, re-parses the int, and takes a fresh `datetime.now()` 500 times. Rows in one response are therefore compared against slightly different clocks — harmless in practice, but it means the boundary is not consistent within a single response.

Hoisting the threshold to one value computed per request would remove both, and would let the frozen-clock test cover a whole page rather than one row at a time.

NMZivkovic (commented):
## Review

Read the whole diff and checked the load-bearing behaviour against the branch. The read-time approach is right — no `ALTER TYPE`, no writes, mirroring the sessions side — and the tests are unusually good: the strict `<` boundary with a frozen clock, the `OverflowError` case, and the operation-row case are all things most PRs would miss. The docstring caveats about this being a heuristic, and about writers always inserting rather than updating, are exactly the right things to have written down.

**One issue I think blocks it**, inline on `get_effective_pipeline_status`: because the endpoint does not dedupe by `pipeline_run_id`, the `STARTED` row of every *successfully completed* run older than the threshold now reports `"ABANDONED"`. I reproduced it against this branch. That is not the straggler case the docstring anticipates — it is the ordinary case, so the activity feed gains a false `ABANDONED` entry across most of the run history rather than only for genuinely stuck runs.

Everything else is minor.

### Two coordination questions

**Frontend contract.** `"ABANDONED"` is a value no client has seen before, and the bug this fixes was observed in `cognee-saas-frontend`. Is there a matching frontend change, and does anything switch exhaustively on status? Nothing in this repo outside the docstring references the new value.

**Overlap with #4966.** That PR makes startup recovery write a real `ERRORED` row for abandoned runs. Once both land, a recovered run has a `STARTED` row reading `ABANDONED` *and* an `ERRORED` row — two rows, two failure stories, one run. If the dedup fix above lands, the two compose cleanly: the `STARTED` row stops reading `ABANDONED` the moment recovery closes the run. Worth sequencing with that in mind.

---

cognee/modules/pipelines/methods/get_effective_pipeline_status.py
```diff
@@ -0,0 +1,59 @@
...
+_MAX_ABANDON_AFTER_SECONDS = 10**9  # ~31 years; timedelta stays well inside range
+
+
+def _pipeline_run_abandon_after_seconds() -> int:
+    raw = os.environ.get("PIPELINE_RUN_ABANDON_AFTER_SECONDS", "")
+    try:
+        value = int(raw) if raw else 1800
+    except ValueError:
+        return 1800
+    if value <= 0 or value > _MAX_ABANDON_AFTER_SECONDS:
```
NMZivkovic: **[Minor] Invalid config falls back silently**

A typo'd or out-of-range `PIPELINE_RUN_ABANDON_AFTER_SECONDS` silently becomes 1800 with nothing in the logs, so an operator who sets `-100` (or `30m`) gets default behaviour and no reason why.

Worth a `logger.warning` naming the rejected value, the way `_read_stale_run_min_age()` in #4966 does for the same class of problem — the two PRs currently handle bad config differently.

cognee/modules/pipelines/methods/get_effective_pipeline_status.py
goran-radonic: Fixed, though not with the suggested one-liner. get_pipeline_status also
feeds check_pipeline_run_qualification, which compares its result against
PipelineRunStatus members to decide whether a pipeline is already running,
so returning the effective status there makes both comparisons silently
false and the dedupe guard stops working. Reproducing that exact change
turns three tests red, one of them the pre-existing
test_reset_pipeline_status_allows_reexecution_with_cache.

Split it instead: get_pipeline_status and get_pipeline_progress keep the
raw enum for control flow, two reporting siblings return
EffectivePipelineRunStatus, and datasets.get_status/get_progress use those.
Response models needed updating too, ABANDONED would have failed validation
against the old PipelineRunStatus-typed model. Existing values are unchanged
on the wire

cognee/modules/pipelines/methods/get_effective_pipeline_status.py
goran-radonic: Fair, hoisted it. get_abandon_cutoff() computes the cutoff once and
get_effective_pipeline_status takes it as a required keyword arg, same
shape as run_has_terminal_row, so a call site can't quietly fall back to
its own clock.

Added a test for it, because the change is invisible from the outside:
per-row and per-page produce identical statuses, so nothing asserted on
the response body would catch it moving back. The test counts env var
reads across a multi-row page and wants exactly one. Putting the
computation back inside the helper turns that test red and nothing else.

You were right that it makes the frozen-clock test better too, the cutoff
is now one value for the whole page rather than per row.

cognee/modules/pipelines/methods/get_effective_pipeline_status.py
goran-radonic: Fair, hoisted it. get_abandon_cutoff() computes the cutoff once and
get_effective_pipeline_status takes it as a required keyword arg, same
shape as run_has_terminal_row, so a call site can't quietly fall back to
its own clock.

Added a test for it, because the change is invisible from the outside:
per-row and per-page produce identical statuses, so nothing asserted on
the response body would catch it moving back. The test counts env var
reads across a multi-row page and wants exactly one. Putting the
computation back inside the helper turns that test red and nothing else.

You were right that it makes the frozen-clock test better too, the cutoff
is now one value for the whole page rather than per row.

NMZivkovic (commented):
Re-reviewed at `cc24cb7a`. The per-run fix is right, and better than what I suggested: the correlated `EXISTS` over an aliased sibling holds across page boundaries (the `offset=1, limit=1` test proves it), the cutoff is computed once per page, and bad config is warned about once per distinct value. Keeping `get_pipeline_status` raw for `check_pipeline_run_qualification`, with a regression test for the one-liner the bot suggested, is exactly the right call. All four threads from my last review are addressed.

The part I'd hold is the follow-up that routes `/v1/datasets/status` and `/v1/datasets/status/progress` through the effective status. `/v1/datasets/status` is what every frontend status path polls, so widening its value set is a contract change, and two things break (inline):

1. **The frontend renders `ABANDONED` as `completed`** — in `cognee-frontend` here and in `cognee-saas-frontend`.
2. **Datasets queued in a background multi-dataset run read `ABANDONED` before they have started**, because their STARTED rows are written up front.

Together that is a dataset still waiting its turn showing up in the Brains list as completed. There is also a smaller SDK-facing change in what `cognee.datasets.get_status()` returns (inline).

My suggestion is to split it. The activity-feed half — per-run `EXISTS`, cutoff hoist, config warnings, raw/effective split — is ready and fixes the bug the ticket was about. The dataset-status half can follow once the frontend knows `ABANDONED` and staleness is measured from when a run actually starts executing.

cognee/api/v1/datasets/datasets.py
```diff
@@ -166,7 +167,9 @@ async def has_data(dataset_id: str, user: User | None = None) -> bool:
 
     @staticmethod
     async def get_status(dataset_ids: list[UUID], pipeline_names: list[str] | None = None) -> dict:
-        return await _fan_out_by_pipeline(dataset_ids, pipeline_names, get_pipeline_status)
+        return await _fan_out_by_pipeline(
+            dataset_ids, pipeline_names, get_effective_pipeline_status_by_datasets
```
NMZivkovic: **[Major] `/v1/datasets/status` can now return a value no frontend handles, and the UI shows it as completed**

Every status path in the frontend reads `/v1/datasets/status`, and none of them knows `ABANDONED`. In `cognee-frontend` in this repo:

- `src/app/(app)/datasets/brainsTypes.ts:36-49` — `mapProcessingStatus` checks the four stored values and falls through to `return docCount > 0 ? "completed" : "empty"`. `useBrainsData.ts:127/156` feeds it straight from `/v1/datasets/status?include_error_detail=true`, so an abandoned dataset with documents is listed as **completed**.
- `src/app/(app)/datasets/[id]/DatasetDetailPage.tsx:476-482` — the status effect has branches for COMPLETED, ERRORED and STARTED/INITIATED only; `ABANDONED` matches none, so the page status never updates.
- `src/modules/datasets/pollDatasetStatus.ts` — `ABANDONED` is in neither `IN_PROGRESS_STATUSES` nor `TERMINAL_STATUSES`, so upload polling runs to the 10-minute timeout and reports a `DatasetProcessingTimeoutError`.

`cognee-saas-frontend` on `main` has the same mapper and poller. Before this change a crashed run read as "running" forever, which was wrong but never claimed success; now it can read as completed.

Since `cognee-frontend` here is synced from `cognee-saas-frontend`, the fix has to start there (a status mapping for `ABANDONED` plus the `DatasetProcessingStatus` union) — an edit to `cognee-frontend` in this PR would be overwritten by the next sync. Until that lands I'd keep `/status` and `/status/progress` on the raw status.

cognee/modules/pipelines/operations/get_pipeline_status.py
```diff
@@ -30,3 +45,58 @@ async def get_pipeline_progress(dataset_ids: list[UUID], pipeline_name: str):
         }
         for dataset_id, run in runs.items()
     }
+
+
+async def get_effective_pipeline_status_by_datasets(dataset_ids: list[UUID], pipeline_name: str):
```
NMZivkovic, line 50: **[Major] Datasets waiting their turn in a background batch read ABANDONED before they start**

Staleness here is measured from the newest row's `created_at`, and for a background multi-dataset run that is when the batch was queued, not when the dataset started:

- `run_pipeline_as_background_process` (`layers/pipeline_execution_mode.py`) calls `anext()` on every dataset's pipeline in a loop before anything runs. That advances each one to its `yield PipelineRunStarted`, and `log_pipeline_run_start` sits before that yield (`run_tasks.py:72`) — so every dataset gets its STARTED row immediately.
- `handle_rest_of_the_run` then executes the pipelines **one at a time**.

So with `cognify(datasets=[a, b], run_in_background=True)`, if `a` takes 31 minutes, `b` reports `ABANDONED` for as long as `a` keeps running — and on the frontend (previous comment) it shows as completed. The docstring's heuristic caveat covers "a pipeline genuinely still running past the threshold"; this one is not running yet, it is just waiting.

#5018's heartbeat will not close this by itself either: a queued dataset never ticks, so its `last_heartbeat_at` stays NULL and the documented fallback is `created_at` again. The missing signal is when execution began — the point where `handle_rest_of_the_run` resumes the generator after that yield. I left the same note on #5018, since stamping it there would give both PRs the right clock.

cognee/modules/pipelines/methods/get_effective_pipeline_status.py
```diff
@@ -0,0 +1,172 @@
...
+from datetime import datetime, timedelta, timezone
+from functools import cache
+
+from cognee.modules.pipelines.models import PipelineRun, PipelineRunStatus
+from cognee.shared.logging_utils import get_logger
+
+logger = get_logger(__name__)
+
+
+class EffectivePipelineRunStatus(str, enum.Enum):
```
NMZivkovic, line 12: **[Minor] `cognee.datasets.get_status()` values no longer compare equal to `PipelineRunStatus`**

`datasets.get_status` / `get_progress` are public SDK methods — the `datasets` class docstring shows `get_status` as the way to check cognify status — and they now return `EffectivePipelineRunStatus` members instead of `PipelineRunStatus`. `PipelineRunStatus` is a plain `enum.Enum`, so the two never compare equal:

```python
EffectivePipelineRunStatus.DATASET_PROCESSING_COMPLETED == PipelineRunStatus.DATASET_PROCESSING_COMPLETED
# False (checked on 3.9 and 3.13)
```

An SDK user doing `status == PipelineRunStatus.DATASET_PROCESSING_COMPLETED` on a `get_status()` result silently gets `False` for every dataset. Nothing in the tree does that outside tests (the CLI only prints the values), so this is about external callers. Either keep the SDK methods returning the stored enum and apply the effective mapping in the router, or call the change out explicitly in the release notes.

cognee/modules/pipelines/operations/get_pipeline_status.py
goran-radonic: Confirmed, and thanks for tracing it through both loops.

Scope has shrunk since you wrote this: I moved /status and /status/progress
back to the stored status in the other thread, so the "shows as completed"
half is gone. What remains is /activity reporting ABANDONED for a dataset
that is only waiting its turn, which is still wrong.

I am not fixing it here. started_at can't be reused as the execution clock
since it is stamped in the same log_pipeline_run_start call and carried to
the terminal rows to define run duration, so this needs a new column, and
this branch has no migration. #5018 does, and you already left the note
there, so that is where the stamp goes and the reader follows.

Widening the docstring caveat here to name the queued case, since today it
only covers a pipeline genuinely still running past the threshold.

cognee/modules/pipelines/methods/get_effective_pipeline_status.py
goran-radonic: Confirmed, and it cuts both ways: EffectivePipelineRunStatus is a str enum
and PipelineRunStatus is not, so a caller comparing against the enum gets
False and one comparing against a bare string gets True only on the new
type.

Already closed by the change in the /status thread. datasets.get_status and
get_progress are back on get_pipeline_status / get_pipeline_progress, so the
SDK returns the stored enum again, which is your first option. The effective
mapping did not move to the router either, it stays on /activity only, for
the frontend reason in that thread.

dexters1 (changes requested):
Some pipelines currently take days to execute:

For example Nikolas 10 Gb ingestion took 8 days

We can't mark pipelines as abandoned if they took more than 30min, one thing we could do is adapt the existing function:

recover_stale_cognify_runs_on_startup

that only runs on Cognee startup to set pipeline status to STARTED again, we can change it to show and store in the DB that a run was abandoned

We just need to test that approach make sure it won't coflict with already existing systems for keeping track of pipeline state

cognee/modules/cognify/recovery.py
```diff
@@ -60,22 +99,23 @@ async def recover_stale_cognify_runs_on_startup() -> None:
         recovery_candidates = [
             run
             for run in latest_per_dataset.values()
+            # Both conditions are load-bearing. Without the status filter an
+            # already-closed run is re-selected and its rollback repeats on
+            # every boot; without the origin filter this deletes the graph of
+            # a run another process is still executing.
             if run.status == PipelineRunStatus.DATASET_PROCESSING_STARTED
+            and (run.origin in owned_origins or (run.origin is None and _RECOVER_UNATTRIBUTED))
```
NMZivkovic, line 107: The only thing standing between this sweep and a live run is the surface origin. There's no host, process or age check, and dev's 1h `COGNEE_STALE_RUN_RECOVERY_MIN_AGE_SECONDS` guard is gone. So any API process that boots while another API process on the same database is still running cognify rolls that run back and writes ERRORED: a second replica, or the Helm chart's default rolling update, where the new pod boots while the old one still serves. Once #5001's multiple workers land, a replaced worker would do it to its siblings too. The docstring names the rolling deploy, but the result is worse than a stale status: the live run keeps going and later inserts COMPLETED over a graph that lost everything extracted before the rollback. On a real database the rows end up as STARTED (api), ERRORED `AbandonedPipelineRunError`, COMPLETED (api).

Until there's a real liveness signal, I'd scope this to a process identity (a boot id, or host + pid + start time, stamped on the `STARTED` row, closing only rows whose owner is provably gone) or keep the age floor next to the origin check. Separately, a terminal writer shouldn't insert COMPLETED once a terminal row already exists for that `pipeline_run_id`.

cognee-mcp/src/server.py
```diff
@@ -919,6 +919,34 @@ async def main():
...
+        # Placed here deliberately. It rolls back graph data, so it must not
+        # run before the schema is migrated, and it must not run at all when
+        # --api-url or --serve-url says this process does not own the database.
+        # Guarded like the origin stamp above, since cognee comes from PyPI and
+        # may predate either piece.
+        try:
+            from cognee.modules.cognify.recovery import recover_stale_cognify_runs_on_startup
+            from cognee.modules.operations import ORIGIN_MCP
+
+            await recover_stale_cognify_runs_on_startup(owned_origins=frozenset({ORIGIN_MCP}))
```
NMZivkovic, line 938: stdio is the default transport, and stdio MCP is one process per client, so every IDE window or agent session starts its own cognee-mcp against the same local database, and every one of them stamps `mcp`. When a second session boots, this sweep rolls back the first session's in-flight `remember(background=True)` ingestion and writes ERRORED, and that run later writes COMPLETED over the partial graph, the same sequence as the API case. The test's premise ("Local mode owns its database") doesn't hold for stdio. I wouldn't run the sweep from stdio processes at all, or only with the same per-process identity as the API fix.

cognee/modules/cognify/recovery.py
```diff
@@ -1,57 +1,96 @@
...
+# server against one, and `cognee-cli` without --api-url executes in the
+# caller's own process against the same default SQLite file. So the API sweeps
+# rows stamped "api", MCP sweeps "mcp", and neither reaches for the other's or
+# for a user's own script.
+#
+# ORIGIN_BACKGROUND is deliberately not a default for anyone. "background" says
+# a continuation spawned this work, not which process did: remember()'s session
+# bridge stamps it around the whole of improve(), so an SDK script's bridge
+# carries it just as a server's does. It costs nothing to exclude today, since
+# that bridge runs improve() as memify_pipeline and this sweep only looks at
```
NMZivkovic, line 28: The session bridge does start cognify runs: `remember(session_id=...)` runs `improve()` under `ORIGIN_BACKGROUND`, and `cognify_session` calls `cognee.cognify(...)`, which writes `cognify_pipeline` rows stamped `background`. Neither the API nor the MCP sweep owns that origin, so if the process dies mid-bridge, the dataset's cognify status stays `STARTED` across every restart, which is the SDK-591 symptom this PR sets out to fix. A background-origin `STARTED` row survives both sweeps on a real database. Either keep the surface origin on bridge runs (the `background` flag already records that it was a continuation), or let each surface close the background rows it spawned.

cognee/api/v1/activity/routers/get_activity_router.py
```diff
@@ -59,6 +59,22 @@ async def get_pipeline_runs(
         - `"operation"` — a single-row operation record (`pipeline_name` and
           `status` are NULL, so these are invisible to status-based readers).
 
+        `status` for a `"pipeline"` row is one of `PipelineRunStatus`,
+        reported exactly as stored. A run whose process died without writing a
+        terminal row is closed as `DATASET_PROCESSING_ERRORED` by startup
```
NMZivkovic, line 64: This promises more than recovery does. It only closes `cognify_pipeline` runs, and only ones whose origin the booting surface owns, while this endpoint also returns `STARTED` rows for `add_pipeline`, `memify_pipeline`, `code_graph_pipeline`, the skills pipelines and `migration_import_pipeline`, plus cognify runs started by the SDK, the CLI, the session bridge or before the upgrade. Those stay `STARTED` forever. I'd narrow the wording to what's actually closed, so client authors don't build on it.

cognee/modules/cognify/recovery.py
```diff
@@ -1,57 +1,96 @@
...
+# cognify. If a background path ever starts a cognify run, it needs a stamp
+# that names the process rather than the reason.
+_DEFAULT_OWNED_ORIGINS = frozenset({ORIGIN_API})
+
+_RECOVER_UNATTRIBUTED = os.getenv("COGNEE_RECOVER_UNATTRIBUTED_RUNS", "false").lower() in (
```
NMZivkovic, line 33: Nit: `COGNEE_RECOVER_UNATTRIBUTED_RUNS` isn't in `.env.template` or the docs, though the docstring tells operators it's how to close runs stuck from before the upgrade. And `COGNEE_STALE_RUN_RECOVERY_MIN_AGE_SECONDS` is dropped silently, so someone who raised it to protect long runs loses that protection without noticing. A `.env.template` entry and a startup warning when the old variable is still set would cover both.

cognee-mcp/tests/test_startup_recovery.py
```diff
@@ -0,0 +1,106 @@
+"""The MCP server closes the runs it abandoned when it died (SDK-591).
+
+A cognify this server was executing when its container went down leaves a
+DATASET_PROCESSING_STARTED row. The API server will not close it: each surface
+sweeps only the origin it stamps, because neither can tell the other's dead run
+from a live one. So MCP has to close its own, or a dataset whose ingest died
+with this container reports "processing" forever.
+
+The call is destructive (it rolls back that run's graph), so where it sits in
+main() is part of the contract and is asserted here: after the arguments are
```
NMZivkovic, line 10: Nit: the docstring says the ordering after migrations and the `--serve-url` case are asserted, but the stubs don't record call order and only `--api-url` is exercised, so moving the sweep before `run_migrations()` would still pass. A shared call-order list in both stubs and a `--serve-url` case would make it true.

cognee/api/v1/activity/routers/get_activity_router.py
```diff
@@ -117,7 +133,7 @@ async def get_pipeline_runs(
 
         from cognee.infrastructure.databases.relational import get_relational_engine
         from cognee.modules.data.models.Dataset import Dataset
-        from cognee.modules.pipelines.models import PipelineRun
+        from cognee.modules.pipelines.models import PipelineRun, PipelineRunStatus
```
NMZivkovic, line 136: Nit: `PipelineRunStatus` is unused now that the response passes `run.status` through. Ruff won't flag it because F401 is ignored in this repo.

NMZivkovic (commented):
Re-reviewed at `48b7fc52`. Moving from read-time ABANDONED to closing runs at startup answers dexters1's point about multi-day ingestions, and the mechanics are careful: the rollback runs before the ERRORED row is written, so a failed rollback leaves `STARTED` as a retry token; origin is stamped on the `STARTED` row and tested against the real producer; the closing row copies `user_id`/`tenant_id` so the owner sees it; `log=False` keeps boots quiet; and the read-time pieces are cleanly gone.

What worries me is that the sweep can't tell a dead run from a live one in another process (inline), and that #4966 now implements the same thing with different rules.

**Overlap with #4966.** Both PRs replace dev's startup reset with the same end state: roll back cognify, then write `DATASET_PROCESSING_ERRORED` with `AbandonedPipelineRunError` on the dead run's ids. The rules are opposite:
- #4966 runs in the background after startup (what dexters1 asked for there), covers every unclosed run of every pipeline, and only touches runs older than 1h.
- This PR runs inline in the lifespan before the port opens, looks only at the newest cognify row per dataset, filters by origin, and has no age check.

They conflict in six files, and the conflicts are semantic, not just textual. The two `AbandonedPipelineRunError` constructors are incompatible, so #4966's `pipeline_name=..., rolled_back=...` call raises `TypeError` against this class after its rollback has already run. The integration tests assert opposite `graph_warmup` behavior for abandoned runs. And `cognee-mcp` imports `cognee.modules.cognify.recovery`, which #4966 deletes, so the MCP sweep would quietly degrade to a warning; if the merge keeps both files instead, the API and MCP end up sweeping with different rules. I'd make #4966 the single implementation and port the origin stamp, an `owned_origins` parameter and the MCP hook into it, keeping the age floor next to the origin check until #5018's heartbeat has a reader. Merge order would then be #4993, then #4966, with #5018 whenever (it merges cleanly with all of them).

The title and description still describe the read-time approach (`PIPELINE_RUN_ABANDON_AFTER_SECONDS`, "nothing is written back"). The PR now rolls back graphs and inserts rows at startup, adds a destructive step to MCP boot, and removes the 1h `COGNEE_STALE_RUN_RECOVERY_MIN_AGE_SECONDS` guard, which is what the release notes need to say.

CI is green at `48b7fc52`.

PR #4989, opened 2026-09-08
fix(api): map skill ingest failures to 400/500 instead of 409

### LINKED ISSUE #4281: `POST /v1/skills` returns generic 409 "Failed to ingest skill" for every exception, masking unrelated ingest bugs

csy20:
## Description

Fixes #4281.

`POST /api/v1/skills` wrapped the whole `remember()` call in `except Exception` and always returned HTTP 409 `{"error": "Failed to ingest skill"}`. Parse errors and graph-store failures therefore looked like duplicate conflicts, and callers could not tell them apart without server logs.

## Why This Change Was Made

- Re-raise `CogneeApiError` so typed errors keep their own status (including 409 for a genuine conflict).
- Return 400 for caller-controlled input errors (`ValueError`, `PermissionError`, `FileNotFoundError`).
- Return 500 for unexpected pipeline failures, with `exception_type` and `detail` in the body.

## User Impact

Clients of `POST /api/v1/skills` can distinguish malformed skill input from internal ingest failures instead of treating every error as a conflict.

## Evidence

```text
uv run pytest cognee/tests/unit/api/test_skills_router.py -v
======================= 10 passed, 15 warnings in 3.28s ========================
```

Covers success, missing dataset, malformed input → 400, `RuntimeError` / `KeyError` → 500 with type and message, and `CogneeApiError(status_code=409)` still mapping to 409.

### DIFF
--- a/cognee/api/v1/skills/routers/get_skills_router.py
+++ b/cognee/api/v1/skills/routers/get_skills_router.py
@@ -12,6 +12,7 @@ from fastapi.responses import JSONResponse
 from pydantic import BaseModel, Field
 
 from cognee import __version__ as cognee_version
+from cognee.exceptions import CogneeApiError
 from cognee.modules.data.constants import DEFAULT_DATASET_NAME
 from cognee.modules.data.methods import get_authorized_existing_datasets
 from cognee.modules.users.exceptions import PermissionDeniedError
@@ -73,6 +74,8 @@ class ErrorResponse(BaseModel):
     """Generic API error response."""
 
     error: str
+    detail: str | None = None
+    exception_type: str | None = None
 
 
 def get_skills_router() -> APIRouter:
@@ -88,7 +91,11 @@ def get_skills_router() -> APIRouter:
     @router.post(
         "",
         response_model=dict,
-        responses={400: {"model": ErrorResponse}, 409: {"model": ErrorResponse}},
+        responses={
+            400: {"model": ErrorResponse},
+            409: {"model": ErrorResponse},
+            500: {"model": ErrorResponse},
+        },
     )
     async def ingest_skill(
         payload: SkillIngestRequest,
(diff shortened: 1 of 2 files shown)


dexters1:
Thanks for the PR @csy20! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #4993, opened 2026-09-09
SDK-595 fix: Clear the pipeline reset marker

goran-radonic:
## Description

`pipeline_runs` is append-only. A status reset used to write its INITIATED row under a freshly minted `uuid4()`, so the row carried an id that no STARTED or terminal row would ever share. Nothing deduped it away and nothing moved it on, so it sat in the table forever describing work that had finished minutes earlier. `add()` resets `add_pipeline` and `cognify_pipeline` on every call and `forget()` resets cognify, which is why an ordinary workspace grew one or two of these per add and the activity feed filled up with queued runs that never finish. Reported from cloud production and traced here from CLO-754.

The row is only ever control state. It says "this pipeline may run again", and that question is answered per dataset and pipeline name. So it now carries the identity of the run it resets rather than a new one, and the run it unblocks deletes it when it starts. `reset_pipeline_run_status` takes the run being reset instead of rebuilding that identity from a user id, since a caller resetting somebody else's run had no way to rebuild it correctly.

The delete is keyed on `dataset_id` and `pipeline_name`, deliberately not on `pipeline_id`. `pipeline_id` folds in the acting user, so a marker written by one user for a dataset another user then runs would never be matched. Startup recovery hits that every time, since it acts as the dataset owner while the run was started by whoever called it, and shared datasets hit it whenever the writer is not the owner. The pair used instead is exactly what every status reader partitions on, and it rides an index that already exists.

One thing came along for the ride because the delete depends on it: `PipelineRun.__table_args__` was assigned twice, so the second assignment silently discarded the first and `ix_pipeline_runs_dataset_pipeline_created_at` never reached the model. Alembic-migrated databases have that index, databases built by `create_all` did not, which is every fresh install and every test database. Both declarations are now in one tuple. No new migration, both indexes already have one.

Behaviour deliberately left alone: `add()` still resets cognify without running it, so until a cognify follows, one cognify marker stands. That is the truthful reading of the state, since data is staged and uncognified, and it is bounded at one row because the reset skips a pipeline already at INITIATED. There is a test pinning that bound. Removing the reset instead would mean the next `cognify()` sees COMPLETED and never processes the newly added data, which is the bug the reset exists to prevent.

## Acceptance Criteria

* An INITIATED row written by a reset shares its `pipeline_run_id` with the run it resets.
* A repeated add on one dataset does not grow `pipeline_runs` by a row that no later row supersedes.
* `get_pipeline_status` and `get_pipeline_progress` report the same status for the same sequence of calls, so a COMPLETED run still does not skip execution.
* A test that adds twice to one dataset and asserts the row count and the surviving row per id.

Proof, driving the real `add()` three times against an isolated database rather than calling the run loggers directly:

| code | rows | distinct run ids for 3 runs | leftover INITIATED |
|---|---|---|---|
| dev | 8 | 5 | 2 (ids no other row shares) |
| this PR | 6 | 3 | 0 |

Adversarial checks against the fix, same real path:

| case | result |
|---|---|
| 6 concurrent `add()` on one dataset | 12 rows, 0 leftover, 6 started / 6 completed, 0 started rows without a terminal row |
| two datasets adding interleaved | 4 rows each, 0 leftover on both, no cross-dataset deletion |
| marker written under user A, next run started by user B | marker before 1, after 0 |
| add without cognify, repeated 4 times | markers after each add: 1, 1, 1, 1 |

The same checks with the delete keyed on `pipeline_id` leave the marker uncleared in the cross-user case and show two markers instead of one, which is what motivated the key change.

Six unit tests cover the reset shape, the add-shaped two-pipeline reset through the layer `add()` and `forget()` share, a reset by someone other than the runner, both status readers, and the accepted add-only residue. Five mutations were each caught by the matching test: dropping the delete, keying it on `pipeline_id`, minting a fresh id, deriving `pipeline_id` from the actor, and passing the wrong object in recovery.

Full unit suite: 5784 passed, 70 skipped.

Fixes SDK-595

## Type of Change
- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation


cognee/modules/pipelines/models/PipelineRun.py
```diff
@@ -28,10 +28,14 @@ class OperationOutcome(str, enum.Enum):
 
 class PipelineRun(Base):
     __tablename__ = "pipeline_runs"
+    # One tuple, not one per index: a second __table_args__ assignment silently
```
NMZivkovic, line 31: **[Praise] Good catch on the double `__table_args__` assignment**

This is the most valuable line in the PR and it is easy to miss in review, so worth naming: the second assignment silently replaced the first, which meant `ix_pipeline_runs_dataset_pipeline_created_at` never existed on any `create_all` database — every fresh install and every test database. Alembic-built databases had it, so the divergence would only ever have shown up as "slow in dev, fine in prod", or the reverse.

Folding both into one tuple with the comment explaining why is the right fix.

cognee/modules/pipelines/operations/log_pipeline_run_start.py
```diff
@@ -38,6 +40,26 @@ async def log_pipeline_run_start(
...
+        # exactly it, see get_pipeline_runs_by_dataset), so it is the unit the
+        # marker speaks for. pipeline_id folds in the acting user, and a marker
+        # written by one user for a dataset another user then cognifies would
+        # never be matched, leaving behind the phantom this clears. It also
+        # rides the existing (dataset_id, pipeline_name, created_at) index.
+        await session.execute(
+            delete(PipelineRun).where(
+                PipelineRun.dataset_id == dataset_id,
+                PipelineRun.pipeline_name == pipeline_name,
+                PipelineRun.status == PipelineRunStatus.DATASET_PROCESSING_INITIATED,
```
NMZivkovic, line 64: **[Minor] The delete is broader than the run that is starting**

The predicate is `(dataset_id, pipeline_name, INITIATED)`, so a starting run clears **every** marker for that pair, not just the one that unblocked it. The PR argues the key choice well and I agree with dropping `pipeline_id` — the cross-user case is real and the pair is what every status reader partitions on.

The residual case is two markers for the same pair from different resets, where the first run to start clears both, including one for a run that has not started yet. Today that is close to harmless: markers only feed `get_pipeline_status`, and `check_pipeline_run_qualification` (the one consumer that would skip on it) runs only with `use_pipeline_cache=True`, which add/cognify/memify all pass as False.

Not asking for a change — the alternative keys are all worse. Worth a sentence in the comment noting the marker is per (dataset, pipeline) rather than per run, so the next reader does not assume one marker maps to one run.

NMZivkovic (commented):
## Review

Read the whole diff and traced the call sites against the branch. This one I have very little to push back on — the diagnosis is precise, the key choice is argued from the actual readers rather than from taste, and the before/after row counts driving real `add()` calls are the right kind of evidence.

Things I checked rather than assumed:

- **All call sites of both changed signatures are updated.** `reset_pipeline_run_status` has exactly two non-test callers (`reset_dataset_pipeline_run_status`, `cognify/recovery.py`) and both moved to the new shape; `log_pipeline_run_initiated` has one, and it moved too. No stragglers.
- **The `__table_args__` fix is real and is the sleeper item here** — commented inline. It silently kept a composite index off every `create_all` database.
- The DB-backed tests in `cognee/tests/unit/modules/pipelines/` follow existing convention in that directory (six files there already call `create_db_and_tables`), so no objection there.

Everything I found is minor and is inline. Nothing blocks from my side.

### Merge ordering with #4966

`cognee/modules/cognify/recovery.py` and `cognee/tests/unit/modules/cognify/test_recovery.py` are edited here and **deleted** by #4966, so those two PRs conflict. #4966 is also currently `CONFLICTING` against `dev` and has changes requested, and this one is `MERGEABLE` — so landing this first and rebasing #4966 onto it looks like the cheaper order. Flagged on both PRs.

---

cognee/modules/pipelines/methods/reset_pipeline_run_status.py
```diff
@@ -1,18 +1,28 @@
...
+from cognee.modules.pipelines.models import PipelineRun
 from cognee.modules.pipelines.operations.log_pipeline_run_initiated import (
     log_pipeline_run_initiated,
 )
-from cognee.modules.pipelines.utils.generate_pipeline_id import generate_pipeline_id
 
 
-async def reset_pipeline_run_status(user_id: UUID, dataset_id: UUID, pipeline_name: str):
-    pipeline_id = generate_pipeline_id(user_id, dataset_id, pipeline_name)
+async def reset_pipeline_run_status(pipeline_run: PipelineRun, user_id: UUID | None = None):
```
NMZivkovic: **[Minor] `user_id` defaults to None, but both callers always pass it**

`user_id: UUID | None = None` makes the attribution optional, and `log_pipeline_run_initiated` accepts `None` too, so a call that omits it writes a marker with a NULL `user_id`. Both call sites in the tree pass it (`reset_dataset_pipeline_run_status` and `cognify/recovery.py`), so nothing hits that today.

Since `/activity/pipeline-runs` filters visibility on `user_id`, a NULL-attributed row is one nobody can see. Making the parameter required would remove the shape entirely, and costs nothing given both callers already supply it.

cognee/modules/cognify/recovery.py
```diff
@@ -95,11 +95,7 @@ async def recover_stale_cognify_runs_on_startup() -> None:
                 )
                 # Clear the lingering STARTED status so a re-run is not blocked by
                 # check_pipeline_run_qualification ("already being processed").
-                await reset_pipeline_run_status(
-                    user_id=dataset.owner_id,
-                    dataset_id=dataset.id,
-                    pipeline_name="cognify_pipeline",
-                )
+                await reset_pipeline_run_status(pipeline_run, user_id=dataset.owner_id)
```
NMZivkovic, line 98: **[Note] This call site is deleted by #4966**

Heads-up rather than a change request: #4966 deletes this file (and `cognee/tests/unit/modules/cognify/test_recovery.py`, which this PR also edits), replacing the reset with a terminal `ERRORED` row written by `cognee/modules/pipelines/recovery.py`.

So these two PRs conflict textually, and whichever merges second needs a manual resolution. The semantic half matters more: this PR's headline justification for keying the marker off the run rather than the acting user is *"startup recovery hits that every time, since it acts as the dataset owner"* — and #4966 removes recovery's use of `reset_pipeline_run_status` altogether. The fix stays correct either way (the shared-dataset case in `reset_dataset_pipeline_run_status` still needs it), but that argument stops applying once #4966 lands.

NMZivkovic (commented):
Re-reviewed at `d616a2b3`. Both follow-ups are in: `user_id` is required now, with the reason in the docstring, and the comment in `log_pipeline_run_start` spells out that a marker is per `(dataset, pipeline_name)`, not per run. Nothing else changed on the branch, and CI is green.

LGTM from my side.

One ordering note, same as on #4966: this PR edits `cognee/modules/cognify/recovery.py` and its unit test, and #4966 deletes both. This one is clear to merge while #4966 still has changes requested, so I'd land this first and let #4966 rebase onto it.

NMZivkovic (approved):

ArmagedonFlamer:
Thanks @goran-radonic for the precise diagnosis. I'm closing this without merging because #5091 (COG-6606) has since removed the INITIATED reset row entirely, together with the reset helpers and the reset calls in add, forget and startup recovery, since the cache gate they existed to defeat was already dead code behind the per-dataset lock.

PR #4994, opened 2026-09-09
SDK-601 fix(mcp): Guard SSE, apply --path, drop unreachable code

NMZivkovic:
## Description

Linear: [SDK-601](https://linear.app/cognee/issue/SDK-601/cognee-mcp-fix-transport-bugs-correct-release-metadata-drop)

The MCP server is **not** broken — it works end to end on cognee 1.5.4. What was broken was the ability to demonstrate that, plus a live security gap and two release blockers.

Baseline verified *before* any changes, on Python 3.11 (CI default) and 3.12: **80 tests pass**, live stdio `initialize` → `tools/list` → `tools/call` all succeed, HTTP transport serves correctly, and all four live tools match cognee's current signatures with no drift.

### Security — SSE dropped the DNS-rebinding guard

`http_app()` forwards `host_origin_protection` / `allowed_hosts` / `allowed_origins` on its streamable-http branch only. The `sse` branch calls `create_sse_app()` without them, so they were accepted and silently discarded — while the startup banner reported the guard as **enabled**. `docker-compose.yml` shipped `TRANSPORT_MODE=sse` on a `0.0.0.0` bind, so the documented stack was unguarded.

Fixed three ways: compose now uses `http`, the server warns explicitly when `sse` is chosen, and the README no longer recommends SSE. Verified with raw sockets so the `Host` header is exactly what was sent:

```
Host: 127.0.0.1:8952     -> 200 OK
Host: localhost:8952     -> 200 OK
Host: evil.example.com   -> 421 Misdirected Request
Host: 10.0.0.5:8952      -> 421 Misdirected Request

with MCP_ALLOWED_HOSTS="10.0.0.5:*":
Host: 10.0.0.5:8953      -> 200 OK
Host: evil.example.com   -> 421 Misdirected Request
```

### `--path` was parsed, logged, documented — never applied

It reached the startup banner but never `http_app()`, so the endpoint stayed at the default and the printed URL 404'd.

The naive fix breaks SSE: `--path` defaulted to `/mcp`, but the transports have **different** defaults (`/mcp` for http, `/sse` for sse), so forwarding it verbatim moves the SSE endpoint and breaks every existing SSE client. The default is now `None`, so each transport keeps its own unless overridden.

| case | expected live | expected 404 |
|---|---|---|
| http, no `--path` | `/mcp` → 200 | `/sse` → 404 |
| http, `--path /custom` | `/custom` → 200 | `/mcp` → 404 |
| sse, no `--path` | `/sse` → 405 | `/mcp` → 404 |
| sse, `--path /events` | `/events` → 405 | `/sse` → 404 |

(405 = route exists, SSE expects GET not POST. 404 = no route.)

### Release metadata

- Without `version=`, FastMCP reports **its own** version — every client displayed "Cognee v3.4.6". Now read from package metadata.
- `0.5.5` → `0.5.6`. `0.5.5` has been on PyPI since 2026-07-21 with ~40 commits (the entire FastMCP 3 migration) stranded behind an already-taken version.

### Python support

`requires-python = ">=3.10"` was false. `cognee[docs]` → `unstructured` → `spacy`, which has no cp314 wheel or sdist:

```
before: error: Distribution `spacy==3.8.14` can't be installed because it doesn't
        have a source distribution or wheel for the current platform
after:  error: The requested interpreter resolved to Python 3.14.4, which is
        incompatible with the project's Python requirement: `>=3.10, <3.14`
```

The trap: `uv lock --check` **passes** on 3.14 — resolution is fine, only installation fails — so CI could never catch it. Verified installing on 3.10–3.13. Dropping the unreachable 3.14 arm removed duplicate wheel sets for `lxml` / `onnxruntime` / `unstructured`, shrinking `uv.lock` by ~1.9k lines.

### Dead code

Only **4 of 14** tool-shaped functions were registered. The other ten carried `@log_usage` alone and were reachable by nothing — and had begun to rot: `save_interaction` called `add_rule_associations(context=…)` against a parameter cognee renamed to `ctx`.

Removed `cognify`, `search`, `save_interaction`, `get_document`, `get_chunk_neighbors`, `list_data`, `delete_dataset`, `delete`, `prune`, `improve`, plus everything they orphaned (`strip_vectors.py`, `src/codingagents/`, four `server_utils` helpers, three local helpers). `server.py`: **1872 → 917 lines**.

### Docs

`mcp dev src/server.py` cannot load a FastMCP 3 object (*"not a valid server object"*) — the whole Development section was dead; rewritten around `fastmcp dev` / `fastmcp inspect`, both verified. Also fixed a documented `chromadb` extra and a `gemini` extra that do not exist (all 16 listed extras checked against cognee 1.5.4's real metadata), and documented the Host/Origin guard, which was returning HTTP 421 on LAN deployments with no explanation anywhere.

## ⚠️ Breaking for compose users

The `cognee-mcp` compose service now serves **Streamable HTTP**, not SSE. The endpoint moves:

```
http://localhost:8001/sse   ->   http://localhost:8001/mcp
```

An unchanged client config returns **404**. Update the URL (and drop `"type": "sse"` if your client config sets it). To stay on the old transport, set `TRANSPORT_MODE=sse` in the service's `environment:` — it is still supported and, as of this PR, now actually guarded.

**Not affected:** `cognee-cli -ui`, which runs its own container and still serves `/sse`. Aligning that launcher and the onboarding page is a deliberate follow-up — they describe a different deployment, and moving only one of them would tell users to start an SSE server and then connect to `/mcp`.

## Test plan

- `cognee-mcp/tests/` — **68 passed** on Python 3.11 and 3.12 (was 80; 12 removed as they only covered deleted code). Coverage hook confirms all 4 registered tools still exercised.
- Live stdio handshake — same 5-tool surface as before this change.
- Transport matrix above — all four cases.
- Host/Origin guard + `MCP_ALLOWED_HOSTS` — raw-socket results above.
- `uv sync` on 3.10 / 3.11 / 3.12 / 3.13 succeeds; 3.14 rejected with a message naming the cause.
- `ruff format` clean.

**The advertised tool surface is unchanged.**

## ⚠️ Conflicts with #4940

[#4940](https://github.com/topoteretes/cognee/pull/4940) (SDK-572) removes `list_data` alone (−129). This PR removes `list_data` as one of ten (−955), so the two conflict in `server.py` and `README.md`. Only one should land — either close #4940 as subsumed, or land it first and I rebase.

## Review notes

- `uv.lock` is ~1.9k lines of the diff and is almost entirely deletion from the 3.14 cap. Worth reviewing `pyproject.toml` and reading the lock as a consequence.
- Heads-up for operators: the compose switch means clients connecting by **LAN IP or custom hostname now get 421** where SSE previously let them through unguarded. `MCP_ALLOWED_HOSTS` is documented in both compose and the README.
- Not addressed here: `cognee_client.get_document` / `get_chunk_neighbors` and `retrieval_utils.py` are now unreachable from any MCP tool but still have direct tests. Trimming the client abstraction felt like a separate call.

https://claude.ai/code/session_01BQqjBtUgx5wAJm9jbfyCyw


NMZivkovic:
## Security report received privately — now fixed here

A researcher reported this via SECURITY.md against 0.5.5, independently of the audit that produced this PR. Their finding matched what this branch had already identified, but **the earlier commits only mitigated it** — they switched compose to `http`, corrected the misleading startup log, and stopped the README recommending SSE. The vulnerability itself was untouched: anyone running `--transport sse` was still exposed.

`994fa3a` fixes it properly.

### Root cause (as reported, confirmed)

FastMCP installs `HostOriginGuardMiddleware` inside `create_streamable_http_app` only. `create_sse_app()` accepts no equivalent option in any released version — the reporter confirmed this still holds on fastmcp 4.0.3, so upgrading the dependency would not have helped. `http_app()` therefore accepted our allow-lists for `transport="sse"` and dropped them, while the startup log reported the guard as applied.

Fix: mount the same middleware on the SSE app ourselves, using the allow-lists the server already computes — the reporter's first suggested option.

### A second hole the report did not cover

Writing the regression test caught this: `fastmcp.settings.http_host_origin_protection` defaults to **`False`**, and `_transport_security_kwargs()` returned `{}` for a loopback bind, deferring to that default. So `--transport http --host 127.0.0.1` — the CLI default — was **also unguarded**. Loopback now requests `"auto"` explicitly. DNS rebinding is specifically an attack on loopback services, so that bind is the one that least deserves to be skipped.

### Verified

Reporter's exact reproduction, before and after:

| | foreign `Host` | foreign `Origin` |
|---|---|---|
| **before** — sse, any bind | 200 OK | 200 OK |
| **after** — sse `127.0.0.1` | 421 | 403 |
| **after** — sse `0.0.0.0` | 421 | 403 |
| **after** — http, both binds | 421 | 403 |

`MCP_ALLOWED_HOSTS` and `MCP_DISABLE_DNS_REBINDING_PROTECTION` still honoured on both transports.

**80 tests pass** — 12 new ones cover the matrix (2 transports × 2 binds × Host/Origin, plus both escape hatches and the `--path` defaults) over a real ASGI round trip, no sockets. A permitted SSE request opens an event stream and never completes, so the probe treats a timeout as "accepted" rather than hanging on a status code.

### Disclosure

The reporter has not published and offered to hold until a fix ships. They also offered a pytest reproduction and a PR. Worth replying to them directly — this PR is public, so the issue is effectively disclosed once it is visible; someone should confirm the timeline with them.

cc @Vasilije1990 — this also raises the priority of resolving the #4940 overlap.

cognee/tests/e2e/docker_compose/mcp_client.py
```diff
@@ -22,14 +26,14 @@ class McpToolCall:
     structured: Any
 
 
-async def _call_cognify_status(sse_url: str) -> McpToolCall:
+async def _call_cognify_status(mcp_url: str) -> McpToolCall:
     # Imported lazily so the rest of the suite still collects if `mcp` (the
     # client library) is not installed in the runner's environment.
     from mcp import ClientSession
-    from mcp.client.sse import sse_client
+    from mcp.client.streamable_http import streamablehttp_client
```
dexters1, line 33: This import is what fails the Docker Compose e2e job. The workflow installs the `mcp` client unpinned (`--with mcp` in `.github/workflows/docker_compose.yml`, line 141), which now resolves to mcp 2.2.0, and 2.x only ships `streamable_http_client`; the old name is gone. Pinning the runner to the same major the server uses, `--with "mcp<2"`, fixes it without touching this file.

Switching only the import name is not enough on its own: the 2.x function yields a different stream tuple, so the unpacking below would have to change as well.

cognee-mcp/tests/test_mcp_server_hardening.py
```diff
@@ -385,7 +308,9 @@ async def remember(self, **kwargs):
         # client deletes it as soon as remember() returns.
         if isinstance(data, str) and os.path.isfile(data):
             self.seen_paths.append(data)
-            with open(data, "rb") as handle:
+            # ASYNC230: a blocking read is fine in a test double — there is no
+            # event loop to starve, and the file is a few bytes on tmpfs.
+            with open(data, "rb") as handle:  # noqa: ASYNC230
```
dexters1: Stale after the dev merge: dev now ignores ASYNC230 in `[tool.ruff.lint]`, so ruff 0.16 reports this directive as unused (RUF100) and both the Code Quality job and the pre-commit hook fail on it. Drop the `# noqa` and the two comment lines explaining it.

dexters1 (commented):
Ran the branch locally: the MCP test suite passes (80, all 4 registered tools exercised), and the Host/Origin guard matrix, `--path` handling, version reporting, and the stdio and Streamable HTTP client round trips all behave as described. The inline comments cover the remaining CI failures and two follow-ups.

cognee-mcp/tests/test_mcp_server_hardening.py
```diff
@@ -1141,3 +970,112 @@ async def handler(request: httpx.Request) -> httpx.Response:
...
+
+    default_path, method = TRANSPORT_PATHS[transport]
+    target = path or default_path
+    result = []
+
+    def run():
+        try:
+            with TestClient(app, base_url="http://127.0.0.1:8000") as client:
+                result.append(client.request(method, target, headers=headers, json={}).status_code)
+        except Exception as exc:  # noqa: BLE001 - reported via the empty result
```
dexters1: The Code Quality job on dev now runs `ruff check --select BLE001,S110,S112,G201,TRY201,TRY002 --ignore-noqa`, so this `except Exception` fails the gate regardless of the noqa.

Suggestion: 

Logging the exception should be enough for BLE001, simplest fix.

Or

construct the client with `TestClient(app, base_url=..., raise_server_exceptions=False)` so an app error comes back as a 500 status instead of an exception, then catch only the transport failures, `except (httpx.HTTPError, OSError) as exc:`. That keeps the probe's contract (empty result means the stream opened) and removes the blind catch.

docker-compose.yml
```diff
@@ -77,7 +77,13 @@ services:
...
       - SYSTEM_ROOT_DIRECTORY=/cognee-storage/system
       - DATA_ROOT_DIRECTORY=/cognee-storage/data
-      - TRANSPORT_MODE=sse
+      # Streamable HTTP rather than the legacy sse transport. Endpoint: /mcp.
+      # The container binds 0.0.0.0, so the Host/Origin (DNS-rebinding) guard
+      # rejects requests arriving by LAN IP or a custom hostname with HTTP 421
+      # by design; allow them explicitly with
+      # MCP_ALLOWED_HOSTS=192.168.1.50:*,myhost.local:* (the ":*" port glob is
+      # required), or MCP_DISABLE_DNS_REBINDING_PROTECTION=true to turn it off.
+      - TRANSPORT_MODE=http
```
dexters1, line 86: This moves the endpoint on port 8001 from `/sse` to `/mcp` for everyone running the compose stack, so an existing client config gets a 404 after upgrading. That needs a release note next to the 421 heads-up already in the description.

Two places still point at the old endpoint: the onboarding step in `cognee-frontend/src/app/(app)/onboarding/ConnectionSteps.tsx` (line 389) tells users to connect to `http://localhost:8001/sse`, and the `cognee-cli -ui` launcher in `cognee/api/v1/ui/ui.py` (line 550) still starts the container with `TRANSPORT_MODE=sse`. Aligning the launcher can be a follow-up, but the PR should say which stack is canonical, and the onboarding text should match whichever it is.

cognee/tests/test_usage_logger_e2e.py
```diff
@@ -182,103 +182,3 @@ async def test_api_endpoint_logging(e2e_config, authenticated_client, cache_engi
     assert search_logs[0]["type"] == "api_endpoint"
     assert search_logs[0]["user_id"] == str(user.id)
     assert search_logs[0]["success"] is True
-
-
-@pytest.mark.asyncio
-async def test_mcp_tool_logging(e2e_config, cache_engine):
```
dexters1, line 188: This drops the only e2e check that MCP tool calls produce usage-log entries. The API test still runs, but nothing covers the MCP path now. The test only broke because it called tools that no longer exist, so a version that calls `remember` and `recall` through the registry and asserts the Redis entries would keep the coverage.

If that is deferred: the `mcp_data_setup` fixture (line 89) is now unused, and the job in `.github/workflows/e2e_tests.yml` (line 1715) is still named "Usage logger test (API/MCP)".

cognee/tests/e2e/docker_compose/mcp_client.py
NMZivkovic: Confirmed, and the CI log agrees: the failing run shows `ImportError: cannot import name 'streamablehttp_client'` from a **python3.10 runner** site-packages, while the server image build logs `mcp==1.28.1`. Pinned the runner to `mcp<2` and left `mcp_client.py` alone.

One addition, and one correction to the reasoning:

**Addition** — a rename-only fix fails *earlier* than the tuple unpacking. The 2.x signature is `(url, *, http_client=None, terminate_on_close=True)`; `timeout`, `headers`, `sse_read_timeout` and `auth` are all gone, so `streamable_http_client(mcp_url, timeout=30)` raises `TypeError` before it ever yields. Your conclusion holds, just via a different first failure.

**Correction** — `streamable_http_client` is not a 2.x-only name. I checked the 1.24.0, 1.28.1 and 1.30.0 wheels and all three define it alongside a `@deprecated` `streamablehttp_client` alias; 2.x dropped the alias and separately changed the yield. So a pin **plus** rename would also work and would drop the DeprecationWarning — but the pin alone is the smaller change and I've gone with your version.

I put the rationale in a comment above the `uv run` block rather than inline: a `#` line inside the `\`-continued command gets swallowed by the continuation and breaks the invocation.

cognee-mcp/tests/test_mcp_server_hardening.py
NMZivkovic: Right, and it went stale under me mid-review — `ASYNC230` moved into the ignore list on dev in SDK-583 after I branched, so ruff 0.16 reports RUF100 and both gates fail.

Dropped the directive. I kept the rationale as prose without naming the rule, because that ignore is explicitly temporary — pyproject says *"Re-enable when that lands"* for SDK-600 — and when it does, this line needs the directive back. A comment that merely mentions the reason isn't a directive, so RUF100 stays quiet either way.

cognee-mcp/tests/test_mcp_server_hardening.py
NMZivkovic: Fixed with your first option: `logging.getLogger(__name__).exception(...)`, no noqa. Verified against the exact gate command from `community_tests.yml` (`--select BLE001,S110,S112,G201,TRY201,TRY002 --ignore-noqa`) — clean.

Two notes on the alternatives:

**`logging.error(..., exc_info=True)` would not work** — it silences BLE001 but trips **G201**, which is in the same `--select` list. Only `.exception()` satisfies both.

**I did not take the narrow-catch option**, though it's the more elegant one. `raise_server_exceptions=False` does turn app errors into 500s and changes no current outcome, but it doesn't cover **lifespan startup** failures — and these apps do run a lifespan (the streamable-http app starts the FastMCP session manager there). A lifespan crash would still escape `(httpx.HTTPError, OSError)`, leave `result` empty, and read as `_ACCEPTED` — exactly the false pass the broad catch exists to prevent.

While in there I also fixed the comment, which was wrong: it claimed the exception was "reported via the empty result", but the handler appends a `repr`, so the result is non-empty. The contract is upheld by `join(timeout=...)`, not by the catch.

docker-compose.yml
NMZivkovic: Added the release note to the description, called out as breaking, with the `TRANSPORT_MODE=sse` escape hatch for anyone who wants the old endpoint.

I swept the repo for every `/sse` and `TRANSPORT_MODE=sse` reference rather than just the two you cited. Two were genuinely stale **in this PR** — they describe code this PR changed — and are fixed:

- `cognee/tests/e2e/docker_compose/README.md:14` — said the tool call goes over SSE
- `.claude/skills/cognee-integrations/SKILL.md:60` — said the compose MCP server is SSE

On the two you cited, I've **deliberately left them**, and I think the framing is slightly off: neither is broken by this change. `ConnectionSteps.tsx` never documents compose — it documents `cognee -ui` and a standalone `cognee-mcp --transport sse`, both of which really do serve `/sse`. And `ui.py` pulls the published image and runs its own container, so its `/sse` keeps working; it also points at `127.0.0.1:8001/sse`, which is inside the guard's default allow-list, so no 421 there either.

The risk is in changing them piecemeal: flipping only the URL at `ConnectionSteps.tsx:389` would tell users to start an SSE server and then connect to `/mcp`, which 404s. If we flip the `-ui` stack, `ui.py:550/586/481` and `ConnectionSteps.tsx:354/359/369/389` have to move together. Happy to do it here if you'd rather it not linger — otherwise it's a clean follow-up.

**Canonical:** compose is Streamable HTTP at `/mcp`; `cognee-cli -ui` stays SSE until that follow-up.

cognee/tests/test_usage_logger_e2e.py
NMZivkovic: You're right that coverage should exist, and I've restored it — but the diagnosis differs, and it corrects something **I** got wrong earlier in this PR.

`@registry.tool` **does** fold in `@log_usage` (`tool_registry.py:58`), so the four live tools have been logging as `MCP <name>` / `mcp_tool` all along. I'd previously claimed the opposite in a commit message here; that was wrong, and I've said so in the new commit.

What that means for the deleted test: it wasn't covering MCP tool calls. Since b52fcc335, `cognify`, `search`, `list_data`, `save_interaction` and `prune` have carried no registration at all — the test reached them as **module attributes**, which no MCP client can do. So it asserted logging for five unreachable functions and never touched a reachable one. Deleting it removed no coverage of the live surface.

Rather than restore that, I've written the test it looked like it was: it drives `recall` and `cognify_status` through `fastmcp.Client` — a real MCP client against the in-memory server — and asserts the `mcp_tool` entries in Redis. That covers the path a user actually exercises, which the old one never did. `mcp_data_setup` is back in use and the job name stays accurate.

Worth flagging separately: `log_usage` records `_extract_parameters(...)`, so `remember(data=...)` writes user memory content into usage logs whenever `usage_logging` is on. Pre-existing, out of scope here, but someone should decide whether that's intended.

dexters1 (approved):

PR #4995, opened 2026-09-09
fix(ui): skip MCP server instead of aborting when its port is busy

### LINKED ISSUE #4959: [Bug]: cognee-cli -ui aborts when optional MCP port 8001 is occupied
### Description

`cognee-cli -ui` starts the frontend, backend, and optional MCP server together. Its initial port preflight currently treats all three ports as mandatory. If the optional MCP port (8001 by default) is occupied, `start_ui` returns before starting the otherwise available frontend and backend.

This differs from the existing Docker preflight behavior, which skips MCP and continues launching the UI/backend when Docker is unavailable.

### Steps to reproduce

1. Leave ports 3000 and 8000 free.
2. Bind another local process to port 8001.
3. Run `uv run cognee-cli -ui`.

### Actual behavior

Startup aborts with messages similar to:

```text
Port 8001 is already in use for MCP Server
Cannot start cognee UI: The following services have ports already in use: MCP Server (port 8001)
Error: Failed to start UI server
```

Neither the frontend nor backend starts.

### Expected behavior

Frontend/backend port conflicts should remain fatal. An occupied MCP port should emit a warning, skip the optional MCP server, and continue starting the UI/backend, matching the current Docker-unavailable fallback.

### Environment

- Cognee runtime: 1.5.3
- macOS, local CLI/UI

I reproduced this with another local service bound to `127.0.0.1:8001` and confirmed the same preflight logic is present on `dev`.

kaikanor:
Fixes #4959.

`cognee-cli -ui` treats all three ports as mandatory today, so an occupied MCP port (8001) aborts the whole launch even though the UI and backend work fine without MCP. The Docker preflight already degrades gracefully; this makes a busy MCP port do the same.

What changed:
- The MCP port no longer participates in the required-ports preflight. Frontend/backend conflicts stay fatal.
- If the MCP port is taken, one warning explains the likely cause (a leftover cognee MCP container) and the launch continues without MCP.
- The CLI summary now reports the MCP server from the actual container state (`docker_container` set via the pid callback), not from the flag it passed in — so a skipped MCP server prints "skipped (port busy or Docker unavailable)" instead of a stale "available at :8001".

What I deliberately didn't change:
- No post-start container health verification. The bind happens inside dockerd (`docker run -p`), so a Python-side try/except around EADDRINUSE would never fire; if `docker run` succeeds but the container dies immediately, the CLI can still over-report. That's pre-existing behavior and feels like its own change if you want it.

Tests: `uv run pytest cognee/tests/unit/api/v1/ui/ -q` — 11 passed (8 existing + 3 new covering: busy MCP port skips the Docker preflight, free MCP port still runs it, busy frontend port remains fatal).

### DIFF
--- a/cognee/api/v1/ui/ui.py
+++ b/cognee/api/v1/ui/ui.py
@@ -478,13 +478,21 @@ def start_ui(
     """
     logger.info("Starting cognee UI...")
 
+    # The MCP server is optional: the UI and backend work fine without it, so a
+    # busy MCP port only disables the MCP server instead of aborting the whole
+    # launch — same graceful degradation as the Docker-unavailable path below.
     ports_to_check = [(port, "Frontend UI")]
 
     if start_backend:
         ports_to_check.append((backend_port, "Backend API"))
 
-    if start_mcp:
-        ports_to_check.append((mcp_port, "MCP Server"))
+    if start_mcp and not _is_port_available(mcp_port):
+        logger.warning(
+            f"Port {mcp_port} is already in use (possibly a leftover cognee MCP "
+            f"container), skipping the MCP server. The UI and backend will start "
+            f"without it. Stop whatever holds the port to get the MCP server back."
+        )
+        start_mcp = False
 
     logger.info("Checking port availability...")
     all_ports_available, unavailable_services = _check_required_ports(ports_to_check)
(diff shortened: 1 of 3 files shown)


kaikanor:
Closing - superseded by #5066 which landed the same fix (occupied optional MCP port no longer breaks the ui startup). Thanks for maintaining!

PR #5002, opened 2026-09-09
fix: Paginate the dataset data endpoint (SDK-607)

NMZivkovic:
## Description

`GET /api/v1/datasets/{dataset_id}/data` returned every row in the dataset, and callers had no way to ask for less: the route declared exactly two parameters (`dataset_id`, `user`), so FastAPI built its request model without `limit`/`offset` and **silently discarded** them.

Measured on a dataset of 171,828 documents (cognee 1.5.1, Aurora PG 16.9):

```
GET /api/v1/datasets/{id}/data
  → HTTP 200 in 408.3 s, 70,234,163 bytes, 171,828 rows
```

That is the dataset page's never-resolving spinner.

## Changes

**Bounded by default.** `limit` (default 100, 1–1000) and `offset` (default 0), threaded through to the query. The default is bounded rather than opt-in deliberately: it makes an un-updated frontend fast without a coordinated deploy.

**`get_dataset_data()` keeps its unbounded default.** Most of its ~20 callers are internal pipeline code that genuinely needs the whole set — cognify, sync, the estimator, permission fan-out. The HTTP route is the caller that must not do that, so the bound lives there.

**New `GET /{dataset_id}/data/count`.** Several callers only ever wanted a number and read `len()` off the full row set to get it — the dataset list did that *once per dataset* — so counting cost the same as listing. A page's length is a page size, not a total, so those callers now ask the question they actually have. Frontend: `useBrainsData` (list badges, post-upload refresh), the onboarding doc-count probe, and `waitForDatasetDocs` (which would otherwise never satisfy a `minCount` above the page size).

**One serialization pass per row.** `jsonable_encoder(data)` walked the whole ORM row — including the large JSON columns `DataDTO` does not expose (`node_set`, `pipeline_status`, `system_metadata`) — and `response_model=list[DataDTO]` then re-validated and re-encoded the result: three passes per row, most of the work discarded. Fields are now named explicitly, which also keeps the `dataset_id` override safe (the row's own column is nullable, so validating the row first would reject a NULL before we could substitute the requested id).

## ⚠️ API change

The default response goes from *all rows* to *the first 100*. Callers needing more rows must follow `limit`/`offset` pages; `limit` alone is capped at 1,000. Offsets above 1,000,000 return 422, while `/data/count` remains an uncapped total. The tail of datasets larger than 1,001,000 rows is unreachable through this endpoint. Offset traversal is not a snapshot: concurrent inserts/deletes can cause omissions. The Python and frontend traversal helpers deduplicate overlapping IDs and advance offsets by raw page length; errors are raised instead of returning an apparently complete partial result. These limits are documented in the changelog and pagination example.

## Tests

`cognee/tests/unit/api/test_dataset_data_pagination.py` — 9 tests: the default is bounded; paging repeats and skips nothing; out-of-range values are rejected with 422 rather than silently clamped; the requested dataset id wins over the nullable column; `/data/count` returns the total; and `/data/count` is not shadowed by a parameter route (asserted against the router's route table, not a status code, since a handler can 404 for its own reasons).

## Adjacent, not fixed here

`GET /{dataset_id}/data/{data_id}/raw` calls `get_dataset_data(dataset[0].id)` and filters in Python to find one row it has *already resolved by id* — another full-table read, on the same table, to serve one file. Left alone to keep this PR to one change; worth its own ticket.

Fixes SDK-607
Cloud port: CLO-767

Review follow-up validation: 22 Python pagination/remote-routing tests and 10 frontend pagination tests passed; Ruff checks passed. The requested future DatasetDetailPage refactor is deferred as requested.


cognee/api/v1/serve/cloud_client.py
```diff
@@ -391,14 +391,24 @@ async def update(
...
+                params={"limit": limit, "offset": len(rows)},
+            ) as resp:
+                if resp.status >= 400:
+                    body = await resp.text()
+                    raise RuntimeError(f"Remote list_data failed ({resp.status}): {body}")
+                page = await resp.json()
+            if not isinstance(page, list):
+                raise RuntimeError("Remote list_data returned an invalid document list")
+            rows.extend(page)
+            if len(page) < limit:
```
goran-radonic, line 415: and [getDatasetData.ts (line 61)]
If an older server ignores limit and offset and returns all rows, both clients will continue requesting pages indefinitely. Add a compatibility guard or detect non-paginated responses.

cognee-frontend/src/modules/datasets/waitForDatasetDocs.ts
```diff
@@ -26,13 +26,15 @@ export default async function waitForDatasetDocs<T = unknown>(
...
 
   for (;;) {
     try {
-      const data = await getDatasetData(datasetId, instance);
-      latest = Array.isArray(data) ? data : [];
-      if (latest.length >= minCount) return latest;
+      // Poll the count, not the rows: /data is paged, so its length caps at
+      // the page size and a minCount above it would never be reached.
+      if ((await getDatasetDataCount(datasetId, instance)) >= minCount) {
+        const data = await getDatasetData(datasetId, instance);
```
goran-radonic: After the count reaches minCount, this helper returns only the default 100-row page. This changes the previous contract, and the same issue exists in the background path at line 71.

cognee-frontend/src/modules/datasets/useDatasetDataPages.ts
```diff
@@ -0,0 +1,71 @@
...
+    setError(false);
+    if (!append && changedDataset) {
+      setData([]);
+      setHasMore(false);
+    }
+    try {
+      const page: T[] = await getDatasetData(id, instance, { offset });
+      if (request.current !== token) return;
+      const seen = new Set(previous.map(row => row.id));
+      setData([...previous, ...page.filter(item => !seen.has(item.id))]);
```
goran-radonic: previous is captured before the await. If a document is deleted while Load more is in progress, the response can reinsert the deleted document into the UI

cognee/api/v1/datasets/routers/get_datasets_router.py
```diff
@@ -369,18 +372,37 @@ async def get_dataset_data(
             description="Dataset UUID, the id field from GET /api/v1/datasets (not the name)",
             examples=["b8a7c3de-4f5a-4b6c-8d9e-0f1a2b3c4d5e"],
         ),
+        limit: int = Query(
+            100,
+            ge=1,
+            le=1000,
+            description="Maximum number of data items to return.",
+        ),
+        offset: int = Query(
```
goran-radonic, line 400: offset only validates ge=0. Extremely large values reach the database and can cause an OverflowError or expensive database scans.

cognee-frontend/src/app/(app)/datasets/[id]/DatasetDetailPage.tsx
Comment (goran-radonic) [inline cognee-frontend/src/app/(app)/datasets/[id]/DatasetDetailPage.tsx]: The detail page renders files.length, although only one page is now loaded. A dataset with 250 documents is displayed as having 100 documents.

cognee-frontend/src/app/(app)/datasets/useBrainsData.ts
goran-radonic: A count request failure is converted to documents: 0, so a non-empty dataset may incorrectly appear empty.

cognee-frontend/src/app/(app)/datasets/useBrainsData.ts
```diff
@@ -121,9 +125,10 @@ export function useBrainsData(): UseBrainsDataResult {
...
 
       for (const ds of list) {
-        getDatasetData(ds.id, cogniInstance)
-          .then((data) => {
-            const count = Array.isArray(data) ? data.length : 0;
+        // Only the count is wanted here. Fetching the rows to measure them
+        // downloaded every dataset in full, once per dataset on this page.
+        getDatasetDataCount(ds.id, cogniInstance)
+          .then((count) => {
             setDatasets((prev) => prev.map((d) => d.id === ds.id ? { ...d, documents: count, status: mapProcessingStatus(statusData[ds.id]?.status, count, statusData[ds.id]?.reason ?? null) } : d));
```
goran-radonic, line 143: Count requests from consecutive refreshes are not cancelled or generation-guarded. An older response can overwrite a newer count and status.

cognee-frontend/src/app/(app)/graph-models/[id]/GraphModelEditorPage.tsx
```diff
@@ -213,7 +213,7 @@ export default function GraphModelEditorPage({ modelId }: GraphModelEditorPagePr
     setRegenSelectedFiles(new Set());
     setRegenLoadingFiles(true);
     try {
-      const data = await getDatasetData(datasetId, cogniInstance);
+      const data = await getAllDatasetData(datasetId, cogniInstance);
```
Comment (goran-radonic) [inline cognee-frontend/src/app/(app)/graph-models/[id]/GraphModelEditorPage.tsx]: When switching datasets, a late getAllDatasetData() response from the previous dataset can overwrite the files and selection of the newly selected dataset.

cognee/api/v1/serve/cloud_client.py
```diff
@@ -391,14 +391,24 @@ async def update(
             return await resp.json()
 
     async def list_data(self, dataset_id: UUID) -> list:
```
goran-radonic, line 393: cognee.serve() points the SDK at a remote instance whose version you don't control. An older server ignores limit/offset and returns all N rows. If N ≥ 1000, page 1 gives N rows, offset becomes N, page 2 returns the same N rows again, rows grows unbounded. Cheap guard: if a page comes back at >= limit but the row ids repeat, or simply if len(page) > limit, bail with an error instead of continuing.

cognee-frontend/src/modules/datasets/getDatasetData.ts
```diff
@@ -1,6 +1,63 @@
...
+        const body = await response.json();
+        if (!Number.isSafeInteger(body?.count) || body.count < 0) {
+          throw new Error("Unexpected dataset count response");
+        }
+        return body.count;
+      });
+}
+
+/** Explicit full traversal for callers that need every document, such as schema regeneration. */
+export async function getAllDatasetData(datasetId: string, instance: CogneeInstance) {
```
goran-radonic, line 56: On the 171k-doc dataset in the PR body this is 172 sequential round trips. Fine for useDatasets (legacy ingestion page), questionable for GraphModelEditorPage's regeneration file picker, which is a picker, it wants a page, not the corpus. Worth a follow-up ticket at minimum.

cognee-frontend/src/app/(app)/datasets/partials/DocumentsPanel.tsx
```diff
@@ -88,7 +94,7 @@ export default function DocumentsPanel<T extends DocRow>({
               <span style={{ fontSize: 11, fontWeight: 700, color: "rgba(237,236,234,0.55)", letterSpacing: "0.08em", textTransform: "uppercase" }}>{selectedName}</span>
               <span style={{ fontSize: 11, color: "rgba(255,255,255,0.2)" }}>·</span>
               <span style={{ fontSize: 11, color: "rgba(237,236,234,0.35)", display: "inline-flex", alignItems: "center", gap: 4 }}>
-                {docsLoading ? <SkeletonBar width={36} height={8} /> : processing ? "processing" : <>{docs.length} doc{docs.length !== 1 ? "s" : ""}</>}
+                {docsLoading ? <SkeletonBar width={36} height={8} /> : processing ? "processing" : <>{total != null && total >= 0 ? total : docs.length} docs</>}
```
goran-radonic: {total != null && total >= 0 ? total : docs.length} docs dropped the doc{docs.length !== 1 ? "s" : ""} branch, so a one-document dataset now reads "1 docs".

cognee-frontend/src/modules/datasets/useDatasetDataPages.ts
```diff
@@ -0,0 +1,71 @@
...
+    active.current = true;
+    setLoading(true);
+    setError(false);
+    if (!append && changedDataset) {
+      setData([]);
+      setHasMore(false);
+    }
+    try {
+      const page: T[] = await getDatasetData(id, instance, { offset });
+      if (request.current !== token) return;
```
goran-radonic: When the total is an exact multiple of 100 the user gets a "Load more" button that fetches nothing. total is already available in DocumentsPanel; data.length < total would be exact where it's known.

cognee-frontend/src/modules/datasets/useDatasetDataPages.ts
```diff
@@ -0,0 +1,71 @@
...
+      if (request.current === token) setError(true);
+    } finally {
+      if (request.current === token) {
+        active.current = false;
+        setLoading(false);
+      }
+    }
+  }, [instance]);
+
+  return {
```
goran-radonic, line 129: New function identity every render, passed as a prop to DocumentsPanel. Everything else in the hook is useCallback'd.

goran-radonic (changes requested):
Note on the API change

The breaking-change call-out in the PR body is honest and limit is a real migration path. One thing not covered: GET /{dataset_id}/data had no X-Total-Count header before and doesn't now, so a caller that only reads the body has no way to detect it's been truncated except by counting to 100 the separate /count call is the only signal. Acceptable, but worth a line in the release notes next to the limit migration path.

cognee/api/v1/serve/cloud_client.py
```diff
@@ -391,14 +391,29 @@ async def update(
...
+                params={"limit": limit, "offset": len(rows)},
+            ) as resp:
+                if resp.status >= 400:
+                    body = await resp.text()
+                    raise RuntimeError(f"Remote list_data failed ({resp.status}): {body}")
+                page = await resp.json()
+            if not isinstance(page, list):
+                raise RuntimeError("Remote list_data returned an invalid document list")
+            page_ids = {str(row["id"]) for row in page if isinstance(row, dict) and "id" in row}
+            if len(page) > limit or (page and page_ids and page_ids <= seen_ids):
```
goran-radonic, line 411: he page_ids <= seen_ids guard catches a server that ignores the params entirely, but a partial overlap passes it and the row goes into rows twice.

I ran it against a fake server: 2000 rows, one insert at the head between page 1 and page 2, and got 2001 rows back with r999 duplicated and the new row missing. With the old data_size DESC an insert landed at some arbitrary offset so this was occasional. Once #5003 lands and the order is created_at DESC, every new row goes to position 0, so the shift is exactly +1 every time, and this loop runs precisely while ingest is active.

Cheapest fix is to filter against seen_ids before rows.extend(page). The real fix is keyset pagination on (created_at, id), since offset pagination over a moving head can't be made correct.

cognee/api/v1/datasets/routers/get_datasets_router.py
```diff
@@ -388,18 +391,38 @@ async def get_dataset_data(
...
+            0,
+            ge=0,
+            le=1_000_000,
```
goran-radonic, line 403: offset is capped at 1,000,000 here but /data/count has no cap, so for any dataset past roughly 1.001M rows the two endpoints contradict each other: count reports a total the listing can't page to, and the request that tries 422s. CloudClient.list_data surfaces that as RuntimeError: Remote list_data failed (422) rather than stopping on a short page.

The dataset in the description is 171,828 rows, one order of magnitude off. The 422 is deliberate and tested, and the CHANGELOG publishes the 0 to 1,000,000 bound, but neither mentions that it sits below what count will tell you. Either cap the count the same way, or document that past a million rows the listing can't reach the tail.

cognee-frontend/src/app/(app)/datasets/[id]/DatasetDetailPage.tsx
```diff
@@ -2,13 +2,13 @@
```
Comment (goran-radonic) [inline cognee-frontend/src/app/(app)/datasets/[id]/DatasetDetailPage.tsx:2]: We need to refactor this file, not now, but in the future definitely

goran-radonic (dismissed):

goran-radonic (approved):

PR #1454, opened 2025-09-23
fix: fixes distributed pipeline

hajdul88:
## Description
This PR fixes distributed pipeline + updates core changes in distr logic.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update
- [x] Code refactoring
- [x] Performance improvement
- [ ] Other (please specify):

## Changes Made
Fixes distributed pipeline:
-Changed spawning logic + adds incremental loading to run_tasks_diistributed
-Adds batching to consumer nodes
-Fixes consumer stopping criteria by adding stop signal + handling
-Changed edge embedding solution to avoid huge network load in a case of a multicontainer environment

## Testing
Tested it by running 1GB on modal + manually

## Screenshots/Videos (if applicable)
None

## Pre-submission Checklist

- [x] **I have tested my changes thoroughly before submitting this PR**
- [x] **This PR contains minimal changes necessary to address the issue/feature**
- [ ] My code follows the project's coding standards and style guidelines
- [ ] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if applicable)
- [ ] All new and existing tests pass
- [ ] I have searched existing PRs to ensure this change hasn't been submitted already
- [ ] I have linked any relevant issues in the description
- [ ] My commits have clear and descriptive messages

## Related Issues
None

## Additional Notes
None

## DCO Affirmation
I affirm that all code in every commit of this pull request conforms to the terms of the Topoteretes Developer Certificate of Origin.

PR #5003, opened 2026-09-09
fix: Order dataset listings by an indexed column (SDK-608)

NMZivkovic:
Dataset file listings previously sorted by unindexed `data_size`, requiring a full dataset sort even with pagination. HTTP listings, local SDK listings, and processing-status items now select `created_at DESC NULLS LAST, id`, backed by `ix_data_dataset_created`. Internal pipeline batches retain their largest-first default.

The model and migration use PostgreSQL `DESC NULLS LAST` and SQLite `DESC`. Like the sibling pipeline index revision, index creation stays inside Alembic's transaction instead of releasing its locks through `autocommit_block()`. The migration skips missing tables and valid indexes, and repairs invalid indexes. A regular index build can block writes on a large table; deployments needing an online build can pre-create the same index concurrently before upgrading.

`DataDTO` accepts legacy null timestamps, including on the final HTTP page. It exposes `dataSize`, and raw and mapped frontend rows display it while preserving the paginated loader and overlap deduplication from #5002.

Stacked on #5002; retarget to `dev` after that PR merges.

Validation:
- 70 Python data, pagination, ordering, processing-status, and SDK routing tests passed after integrating #5002.
- Real PostgreSQL 14 checks passed for missing-table guards, index creation, repeated upgrade, downgrade, and recreation. Concurrent full Alembic upgrades were not exercised locally.
- Ruff and whitespace checks passed.
- Earlier frontend validation: TypeScript passed; 505 tests passed initially, with one unrelated polling timeout passing on a focused rerun.

Fixes SDK-608. Cloud rollout: CLO-768.


cognee-frontend/src/app/(app)/datasets/[id]/DatasetDetailPage.tsx
```diff
@@ -425,12 +425,14 @@ export default function DatasetDetailPage({ datasetId }: { datasetId: string })
     if (!cogniInstance) return;
     try {
       const data = await getDatasetData(datasetId, cogniInstance);
```
Comment (goran-radonic) [inline cognee-frontend/src/app/(app)/datasets/[id]/DatasetDetailPage.tsx]: This branch is behind the current #5002 base and GitHub marks it DIRTY. Rebasing will conflict in the pagination query and detail page. Preserve #5002’s paginated loader and merge only the dataSize mapping, otherwise resolving with this branch’s version restores the old one-shot, max-100-document UI.

cognee/tests/unit/modules/data/test_dataset_data_ordering.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+@pytest.mark.asyncio
+async def test_method_issues_the_indexed_ordering(monkeypatch):
+    """Reads the ORDER BY off the statement get_dataset_data actually builds.
+
+    Compiled SQL, not the module source: a comment naming the old column would
+    satisfy a text search while the query still sorted the wrong way.
+    """
+    import uuid
```
goran-radonic, line 55: The function-local imports are not sorted. uv run ruff check . reports I001, so the repository lint workflow will fail. Move import importlib above import uuid.

goran-radonic (changes requested):
1. The branch is one commit behind its own base, so it will conflict on merge — blocker

fix/sdk-608 was cut from bd88f7c66. [#5002](https://github.com/topoteretes/cognee/issues/5002)'s head is 08d38f61f ("complete pagination across clients and fix CI"), which is the commit that introduced useDatasetDataPages, getAllDatasetData, the cloud-client paging loop and the /data/count wiring. git merge-tree against the real head produces two conflicts, and both matter for correctness, not just for mechanics.

cognee-frontend/src/app/(app)/datasets/[id]/DatasetDetailPage.tsx:424 — this PR edits the old inline loadFiles body. On [#5002](https://github.com/topoteretes/cognee/issues/5002)'s head that body no longer exists: the row mapping moved into a useMemo at line 109 and loadFiles at line 436 is now just await loadFilePage(datasetId). Git cannot reconcile those, and if someone resolves the conflict by taking "ours" the dataSize fix silently disappears while the rest of the PR still merges green. That is the bad outcome here, because nothing in CI checks that the size column renders.
cognee-frontend/src/app/(app)/datasets/[id]/DatasetDetailPage.tsx:435 — the size: d.dataSize ?? d.size ?? d.size_bytes ?? d.file_size line needs re-applying inside that useMemo after the rebase.
cognee/modules/data/methods/get_dataset_data.py:30 — second conflict. [#5002](https://github.com/topoteretes/cognee/issues/5002)'s head already rewrote the same statement at its line 25 to order_by(Data.data_size.desc(), Data.id), adding the id tiebreak. Both branches touch the identical expression.

One thing to check while rebasing rather than after: [#5002](https://github.com/topoteretes/cognee/issues/5002) changed DocumentsPanel to render raw API rows through useDatasetDataPages<FileEntry> (useBrainsData.ts:58), so on the brains page the doc.dataSize half of your doc.size ?? doc.dataSize fallback is the live branch, while DatasetDetailPage still maps to its own size. Two pages, two shapes, one component. The fallback covers both, but only one of them is exercised by the diff as written, so verify the other by hand.

goran-radonic:
2. The data_size DESC sort is probably deliberate, and this PR removes it from a caller it never mentions — main finding

The PR's argument is that the old ordering was invisible: DataDTO did not expose data_size, so rows arrived sorted by a value the API never returned. That reasoning is sound for the HTTP route and does not hold for the pipeline.

cognee/modules/data/methods/get_dataset_data.py:30 — the new order_by(Data.created_at.desc(), Data.id) applies to every caller, because the ordering lives in the shared method rather than in the route.
cognee/modules/pipelines/operations/pipeline.py:128 — body_data = data if data else await get_dataset_data(dataset_id=dataset.id), and that list goes straight into run_tasks(tasks, dataset.id, body_data, ..., data_per_batch=20). Batches are cut off the list in order, so the ordering of this method is the batching policy.

git log -S"data_size.desc" puts the sort's origin at faeca138d, "fix: fixes distributed pipeline ([#1454](https://github.com/topoteretes/cognee/issues/1454))", whose own description lists "Adds batching to consumer nodes" and "Changed spawning logic". Largest-first is the standard longest-processing-time-first heuristic: you schedule the big items first so the small ones fill the gaps and no worker is left holding one giant document after everyone else finished. Under created_at DESC the batches become chronological, which means a run of 20 large files uploaded together now lands in a single batch of 20 while another batch gets 20 tiny ones.

I am not certain it was intentional, and the commit message does not say so outright. But "it was added by a distributed-pipeline batching commit" is a strong enough signal that it should be settled before merge rather than after. The clean resolution is to stop making one function serve two different needs: keep the size ordering on the pipeline path and give the route its own created_at DESC ordering, either as an order_by parameter on get_dataset_data or as a separate listing method. That also makes the intent readable, which the current shared default never did. Failing that, a yes from @hajdul88 on the record is enough.

The other unbounded callers (sync.py:500, estimator.py:585, incremental.py:556, get_document_ids_for_user.py:53) genuinely do not care about order, so the blast radius is that one pipeline caller.

goran-radonic:
3. NULLS FIRST on Postgres puts the rows most likely to be broken on page 1 — low probability, bad failure mode

cognee/modules/data/models/Data.py:66 — created_at = Column(DateTime(timezone=True), default=lambda: ...). Nullable (no explicit nullable=False) and the default is Python-side only, so it is applied by the ORM and not by the database. Anything that writes this table outside the ORM can leave it NULL.
cognee/modules/data/methods/get_dataset_data.py:30 — in Postgres, ORDER BY x DESC defaults to NULLS FIRST. Those rows therefore sort to the very top of "newest first", ahead of every real document.
cognee/api/v1/datasets/dto.py:18 — created_at: datetime with no default, so it is required. A NULL row cannot be constructed into a DataDTO and the route raises, which the API surfaces as a 500 for the whole page.

Under the old data_size DESC ordering such a row could be anywhere in the dataset, so it would only break the specific page it happened to land on, and on a paginated endpoint most callers would never reach that page. After this change it is on page 1 of every listing, which converts a latent bug into the first thing the dataset page hits.

Probability is genuinely low. The one migration that reinserts rows into this table, d6e8f0a2b4c6, carries created_at across from the source row rather than reinserting without it, so it does not manufacture NULLs. But the guard costs nothing: Data.created_at.desc().nullslast() in the query, and the index is built exactly the same way, since Postgres treats created_at DESC as DESC NULLS FIRST for index ordering and a NULLS LAST query would need DESC NULLS LAST in the index to stay covered. Worth checking that pairing if you take the fix, otherwise you quietly lose the index scan that is the whole point of the PR.

goran-radonic:
4. Alembic already has three heads, and this revision picks one of them — pre-existing, not caused here

cognee/alembic/versions/e7f9a1c3d5b8_add_data_dataset_created_index.py:26 — down_revision = "a7c2e9f4b8d1"

Walking every revision file on this base, the heads are 6ee94db7381f (remove unique constraint tenant name), a7c2e9f4b8d1 (add session records agent id) and f1a2b3c4d5e6 (add importance weight to data). This PR chains onto the second one, so the head count stays at three rather than going to four. That is the right choice given the situation, and alembic upgrade head is already ambiguous in this repo regardless.

Flagging it only so that when someone eventually hits the multiple-h

goran-radonic:
5. Inspector.from_engine has been deprecated since SQLAlchemy 1.4 — trivial

cognee/alembic/versions/e7f9a1c3d5b8_add_data_dataset_created_index.py:35 — in upgrade()
cognee/alembic/versions/e7f9a1c3d5b8_add_data_dataset_created_index.py:63 — in downgrade()

It emits SADeprecationWarning: The from_engine() method on Inspector is deprecated and will be removed in a future release on every migration run. I saw it in the local test output, which incidentally is useful confirmation that the migration really executes during the unit suite rather than being dead code. sa.inspect(conn) is a drop-in replacement with the same return type.

The d1e2f3a4b5c6 precedent this file follows has the same call, so copying it is at least internally consistent and I would not block on it. Fixing both in one go is a two-line change if you want it gone.

goran-radonic:
6. One brittle assertion in an otherwise well-aimed test file — trivial

cognee/tests/unit/modules/data/test_dataset_data_ordering.py:103 — assert body.count("(dataset_id, created_at DESC, id)") == 2

This reads the migration file as text and counts a literal substring, so it breaks the moment someone reformats the f-string across two lines or ruff rewraps it, and the failure message ("both the postgres and non-postgres branches must build the same index") will not point at the real cause. Since the model-side index is already reachable as an object, comparing the migration's compiled DDL, or at minimum matching on a whitespace-normalized string the way test_index_covers_filter_and_sort_in_order does, would survive formatting.

By contrast cognee/tests/unit/modules/data/test_dataset_data_ordering.py:84 is exactly the right instinct: it reads the ORDER BY off the statement the method actually compiles, so a stale comment naming the old column cannot make it pass. The docstring even says why. Keep that one as the model for the rest.

goran-radonic:
What I verified rather than assumed

Checked out the branch and ran cognee/tests/unit/modules/data/ plus test_dataset_data_pagination.py: 39 passed. Built the index by hand on both dialects: it compiles to CREATE INDEX ix_data_dataset_created ON data (dataset_id, created_at DESC, id) identically on Postgres and SQLite, creates cleanly under create_all, and reflects back under its own name, so the existing_indexes guard in the migration will actually see it. The column order does match ORDER BY created_at DESC, id ASC for a forward index scan, and the claim in the PR body that a LIMIT cannot avoid the sort on an unindexed column is correct.

cognee/alembic/versions/e7f9a1c3d5b8_add_data_dataset_created_index.py
```diff
@@ -0,0 +1,70 @@
...
+
+def upgrade() -> None:
+    conn = op.get_bind()
+    if conn.dialect.name == "postgresql":
+        # Reflection and IF NOT EXISTS only establish existence, not validity.
+        # A failed concurrent build leaves an index that the planner cannot use.
+        valid = conn.execute(
+            text(
+                "SELECT i.indisvalid FROM pg_index i "
+                "JOIN pg_class c ON c.oid = i.indexrelid "
```
goran-radonic, line 41: autocommit_block() commits the migration transaction, which drops the alembic_version row lock that normally serializes concurrent upgraders. I ran two alembic upgrade head at the same time from a7c2e9f4b8d1 against PG 17 and both reached the CREATE: one died with DeadlockDetectedError on CREATE INDEX CONCURRENTLY ix_data_dataset_created, exit 1, the other exited 0. The index came out valid, so the database is fine, but the process is dead.

That matters because SqlAlchemyAdapter.create_database() runs this chain on adapter init, so any multi-process start (uvicorn workers, or api + worker + mcp in compose) has one container failing to boot. The DROP kept IF NOT EXISTS but the CREATE lost it in the rewrite, which is part of why the retry path doesn't save you here.

cognee/alembic/versions/e7f9a1c3d5b8_add_data_dataset_created_index.py
```diff
@@ -0,0 +1,70 @@
...
+
+revision: str = "e7f9a1c3d5b8"
+down_revision: str | None = "a7c2e9f4b8d1"
+branch_labels: str | Sequence[str] | None = None
+depends_on: str | Sequence[str] | None = None
+
+INDEX_NAME = "ix_data_dataset_created"
+
+
+def upgrade() -> None:
```
goran-radonic, line 33: c4e8a1f6b3d7_add_pipeline_runs_created_at_index.py is the same problem on the same chain and its docstring argues the other way: plain CREATE INDEX, not CONCURRENTLY, because CONCURRENTLY can't run inside Alembic's transaction and migrations run before the app serves traffic, so the brief write lock is uncontended.

If data is different enough that the reasoning no longer applies, worth saying so here. As it stands the two revisions disagree and neither acknowledges the other.

cognee/alembic/versions/e7f9a1c3d5b8_add_data_dataset_created_index.py
```diff
@@ -0,0 +1,70 @@
...
+    conn = op.get_bind()
+    if conn.dialect.name == "postgresql":
+        # Reflection and IF NOT EXISTS only establish existence, not validity.
+        # A failed concurrent build leaves an index that the planner cannot use.
+        valid = conn.execute(
```
goran-radonic, line 38: 'data'::regclass raises UndefinedTableError and aborts the whole chain if the table isn't there. I dropped data and re-ran to check, and it blows up on this line rather than skipping.

SqlAlchemyAdapter.create_database's docstring states the contract as "every revision is existence-guarded", and c4e8a1f6b3d7 opens with if TABLE_NAME not in insp.get_table_names(): return. Not reachable through the normal chain since the frozen baseline creates data, so this is a convention gap rather than a live bug, but it's the one guard the sibling revision spends a paragraph on.

cognee/api/v1/datasets/routers/get_datasets_router.py
```diff
@@ -466,7 +466,9 @@ async def get_dataset_data(
 
         dataset_id = dataset[0].id
 
-        dataset_data = await get_dataset_data(dataset_id=dataset_id, limit=limit, offset=offset)
+        dataset_data = await get_dataset_data(
+            dataset_id=dataset_id, limit=limit, offset=offset, order_by="created_at"
```
goran-radonic, line 470: The order_by="created_at" split is the right call and it does preserve the pipeline's largest-first batching. One thing it breaks that isn't in this diff: get_dataset_processing_status.py:30 still calls get_dataset_data(dataset_id) with the default, while DatasetProcessingStatusDTO.items is documented at line 98 of this file as "One entry per data item, in the same order as GET /datasets/{id}/data". That description ships in the OpenAPI schema, and the same claim is repeated in the route docstring and the method docstring.

After this change the two orders are unrelated, so anyone lining the responses up by index gets the wrong rows. Either pass order_by="created_at" from the status method too, or drop the claim from all three places.

cognee/api/v1/datasets/dto.py
goran-radonic, line 1: This docstring says one definition serving both sides means callers read the same thing in local and remote mode. That was true for attribute names and it still is, but it no longer covers order.

datasets.list_data() returns get_dataset_data(dataset.id) locally, which is now size DESC, and parses CloudClient.list_data remotely, which reads the paginated route and gets created_at DESC. Same SDK call, two orderings depending on whether cognee.serve() is connected. examples/python/dataset_data_pagination.py:6 repeats the parity promise as well.

Either pass order_by="created_at" from datasets.py:156 so both paths agree, or narrow the promise here to attribute names.

cognee/tests/unit/modules/data/test_dataset_data_ordering.py
```diff
@@ -0,0 +1,241 @@
...
+            session.add(
+                Data(id=UUID(int=base + 5), dataset_id=UUID(int=base + 101), created_at=now)
+            )
+            await session.commit()
+        # A legacy writer may omit the ORM's timestamp default.
+        async with sessions() as session:
+            from sqlalchemy import update
+
+            await session.execute(
+                update(Data).where(Data.id == UUID(int=base + 4)).values(created_at=None)
```
goran-radonic, line 225: Good that this covers the NULL case at all, and nullslast() plus DESC NULLS LAST in the index is the correct pairing.

The catch is that this test only calls get_dataset_data directly, never the route. DataDTO.created_at is a required datetime, so a NULL row still 500s GET /datasets/{id}/data, and with NULLS LAST it now lands on the last page instead of the first. The comment here says a legacy writer may omit the ORM default, so the scenario is considered real, but the green test makes it look covered when the user-facing path isn't. cognee/tests/unit/api/test_dataset_data_pagination.py is the place for that one and has no NULL rows.

goran-radonic (approved):

PR #5004, opened 2026-09-09
fix(datasets): bound document scrolling and rendering (SDK-609)

NMZivkovic:
The dataset tables could freeze the browser by rendering 171,828 document rows. Both views now load pages as the footer approaches the viewport, with a 2,000-row limit on stored and rendered rows, an explicit limit message, and a manual loading fallback. Search applies to loaded files and pauses automatic loading.

Stacked on #5003 (SDK-608), which is stacked on #5002 (SDK-607). This incorporates the API file-size field and its display in both tables alongside the rendering cap, avoiding conflicting sibling changes. Retarget to `dev` after #5003 merges.

The shared loader cancels superseded requests, deduplicates rows, guards concurrent appends, and retries failed refreshes from the first page. Deletion completion only changes the still-selected dataset. Trimming rows to the render cap does not decrement server totals. Counts load independently; selected-dataset refreshes reuse the loader's count request. Known totals suppress unnecessary empty-page requests, while short/duplicate-only pages also stop loading.

The paging hook owns the live cap. The table components retain a defensive cap for independent consumers. Documents beyond 2,000 require a future server-side search or windowed view.

Validation:
- 74 targeted frontend tests passed, including cross-dataset deletion, cap accounting, actual DOM limits, and raw/mapped file-size display.
- TypeScript and ESLint on changed frontend files passed.
- Final full frontend run: 56 suites, 531 tests passed.
- The previous branch's production build passed; the final revision was checked with TypeScript and tests.

Fixes SDK-609. Cloud port: CLO-769.


cognee-frontend/src/app/(app)/datasets/useBrainsData.ts
goran-radonic: A deletion started in dataset A can resolve after the user selects dataset B. This calls the mutation-aware setSelectedDocs setter, which invalidates and aborts B’s in-flight document request, leaving B blank until a manual refresh. Capture the dataset ID before awaiting the deletion, then only update documents when selectedIdRef.current still matches that ID

goran-radonic:
1. It conflicts with #5003, and it half-implements #5003's frontend fix — needs a decision before either merges

[#5003](https://github.com/topoteretes/cognee/issues/5003) and [#5004](https://github.com/topoteretes/cognee/issues/5004) are siblings off the same base and both touch the same two frontend files. git merge-tree between the two branches confirms two conflicts, and this is not theoretical: both branches edit the same lines.

cognee-frontend/src/app/(app)/datasets/partials/DocumentList.tsx:23 — this PR inserts const { visible, hidden } = capRows(docs) and rewrites the .map() and the borderBottom index comparison. [#5003](https://github.com/topoteretes/cognee/issues/5003) rewrites the same component to add dataSize?: number to DocRow and render formatFileSize(doc.size ?? doc.dataSize). Whoever merges second resolves by hand, and the losing half disappears silently because nothing in CI covers the size column.
cognee-frontend/src/app/(app)/datasets/[id]/DatasetDetailPage.tsx:114 — this PR adds dataSize?: number to the generic and size: d.dataSize ?? d.size ?? ... to the useMemo, which is exactly the change #5003 makes. So the same fix now exists in two open PRs with slightly different shapes.

The asymmetry is what makes this worth stopping on. This PR took #5003's DatasetDetailPage half and not its DocumentList half. The brains panel renders raw API rows straight through DocumentList (useBrainsData.ts:58 types the hook as useDatasetDataPages<FileEntry>), so on that page the doc.dataSize branch is the one that matters, and this PR leaves DocumentList rendering formatFileSize(doc.size). Net effect if #5004 merges after [#5003](https://github.com/topoteretes/cognee/issues/5003) and the conflict is resolved toward this branch: the detail page shows sizes, the brains panel silently goes back to dashes.

Also worth noting that dataSize is only ever populated if #5003's backend change (cognee/api/v1/datasets/dto.py:29) lands. Until then the d.dataSize ?? added here is dead code that reads as if it does something. Either pull the DataDTO.data_size field into this PR, or drop the dataSize references and let #5003 own that change end to end. Two PRs owning one field is how it gets lost.

goran-radonic:
2. The count is now fetched twice per refresh on the brains page

The hook learned to fetch its own total, but the two call sites in useBrainsData that were fetching it explicitly were not updated, so both fire.

cognee-frontend/src/modules/datasets/useDatasetDataPages.ts:96 — every non-append load() now issues getDatasetDataCount(id, instance, signal) when maxRows is finite, which it is for both consumers.
cognee-frontend/src/app/(app)/datasets/useBrainsData.ts:176 — Promise.all([loadDocs(completedSelectedId), getDatasetDataCount(completedSelectedId, cogniInstance)]). The first element already counted. This runs on every statusDetails change, which is a polled value.
cognee-frontend/src/app/(app)/datasets/useBrainsData.ts:249 — Promise.all([fetchSelectedDocs(), getDatasetDataCount(ds.id, cogniInstance).catch(() => null)]). Conditionally duplicated: fetchSelectedDocs only calls loadDocs when selectedIdRef.current === ds.id, so when the dataset is selected this is a second count, and when it is not it is the only one and is genuinely needed for the badge.

Not a correctness bug, the counts agree. But /data/count is a fresh SELECT count(*) per call and the whole point of the stack is to stop asking the database for things nobody reads. The second site needs to stay because of the not-selected case; the first can just read total off the hook, which DatasetsPage is already doing at DatasetsPage.tsx:176 via docsTotal ?? selectedDataset?.documents.

goran-radonic:
3. Two independent caps, and one of them is unreachable in production

cognee-frontend/src/modules/datasets/useDatasetDataPages.ts:64 — setData slices to maxRows, and load breaks out of the append loop at maxRows, so data can never exceed 2000.
cognee-frontend/src/app/(app)/datasets/partials/DocumentList.tsx:49 — hidden > 0 therefore never fires here, because both consumers pass MAX_RENDERED_ROWS to the hook and the panel renders docs straight from it.
cognee-frontend/src/app/(app)/datasets/[id]/partials/FilesTable.tsx:112 — same, and more so: FilesTable receives filtered (DatasetDetailPage.tsx:686), which is a filter over an already-capped list, so it can only ever be smaller.

I am not asking you to remove it. Defense in depth on a component that froze a browser tab is reasonable, and RenderedRows.test.tsx is a genuinely good test: it renders 2001 real rows and counts real button[title="Delete file"] nodes rather than trusting a prop, which is the only way to test a DOM-size claim. The thing to be aware of is that it makes two components able to print a "Showing 2,000 of N" message, and ScrollLoader prints a third at ScrollLoader.tsx:117. If the hook cap were ever raised or a consumer forgot to pass maxRows, a reader would see the table's message and the loader's message stacked, disagreeing about the number. Worth at least a comment saying which cap is the live one.

cognee-frontend/src/app/(app)/datasets/partials/ScrollLoader.tsx
```diff
@@ -0,0 +1,144 @@
...
+        <span>
+          {loaded.toLocaleString()}{total === null ? " loaded" : ` of ${Math.max(total, loaded).toLocaleString()}`} {noun}
+        </span>
+      )}
+
+      {atCap && more && (total === null || loaded < total) && (
+        <span style={{ color: "rgba(237,236,234,0.3)", textAlign: "center" }}>
+          Showing the first {maxLoaded.toLocaleString()} {noun}. This view is limited to {maxLoaded.toLocaleString()} {noun}.
+        </span>
+      )}
```
goran-radonic, line 117: Showing the first {maxLoaded} {noun}. This view is limited to {maxLoaded} {noun}.

Both halves state the same fact with the same number. The second sentence is the one carrying information the reader does not already have from the first, so keep that one, or merge them into something that says what to do instead ("use search on the server" once that exists). Cosmetic, but it is the one string a user hits when they cannot find their document, so it is worth being the clearest line in the file.

cognee-frontend/src/modules/datasets/useDatasetDataPages.ts
```diff
@@ -1,71 +1,127 @@
...
+      rows.current = next;
+      nextOffset.current = offset + page.length;
+      more.current = page.length >= limit && next.length > previous.length;
+      updateData(next);
+      setHasMore(more.current);
     } catch {
-      if (request.current === token) setError(true);
+      if (generation.current === token) {
+        failedAppend.current = append;
+        setError(true);
```
goran-radonic, line 115: I raised this on [#5002](https://github.com/topoteretes/cognee/issues/5002) when hasMore only controlled a button; here it controls an IntersectionObserver. On a dataset of exactly 300 documents, the third page returns 100 rows, more stays true, the sentinel renders, scrolling into it fires a fourth request that comes back empty, and only then does the loader disappear. One wasted request and a flicker of "Loading more…" at the bottom of a fully-loaded list.

This PR now has total in hand, which #5002 did not, so the exact form is available: next.length < (total ?? Infinity) as an additional condition. Cheap, and it removes the flicker. Note the hook deliberately does not let the count drive paging (the comment at useDatasetDataPages.ts:94 says a stale count must never cause an endless retry), which is the right instinct. Using total only to stop earlier, never to keep going, preserves that.

cognee-frontend/src/modules/datasets/useDatasetDataPages.ts
```diff
@@ -1,71 +1,127 @@
...
+    setTotal(value => value === null ? null : Math.max(0, value - removed));
+    setLoading(false);
+  }, [invalidate, maxRows]);
 
   const load = useCallback(async (id: string, append = false): Promise<void> => {
-    if (!instance || (append && active.current)) return;
-    const token = ++request.current;
-    const offset = append ? rows.current.length : 0;
+    if (!instance || (append && (active.current || !more.current || rows.current.length >= maxRows))) return;
+    if (!append) invalidate();
```
goran-radonic, line 67: removed is computed as rows present before and absent after, which includes rows the .slice(0, maxRows) on the line above threw away, and that number then decrements both nextOffset and total.

Harmless today: I checked every caller, and they only filter or clear (useBrainsData.ts:200, useBrainsData.ts:358, DatasetDetailPage.tsx:617). The first row prepend into a list sitting at 2000 rows would report a phantom deletion and quietly decrement the server total. Computing removed against the pre-slice list closes it in one line.

goran-radonic (commented):
What I checked and found fine

The generation counter plus AbortController pairing in invalidate is correct, and the active.current synchronous guard genuinely does prevent the double-append that a loading-state guard would have missed, which the "ignores a duplicate request before loading state has rendered" test pins down properly. The failedAppend ref reads odd initialised to true, but it is only consulted when error is set, which can only happen after the catch has written it, so the initial value is unreachable. The loadMore.current ref in ScrollLoader correctly avoids tearing down the observer every render, which matters because loadMore is still a fresh closure out of the hook. The early return null sits after the useEffect, so no conditional hook. And passing rootRef for the brains panel is the detail most implementations get wrong, since a sentinel inside an inner scroller never intersects the viewport root.

The jest ESM allowlist change at cognee-frontend/jest.config.mjs:43 is correct: the generated negative lookahead requires a following /, so adding uuid cannot accidentally capture a uuid-prefixed package.

PR #5005, opened 2026-09-09
fix(improve): persist only new agent trace steps

0x90000:
## Description

Fixes #4996.

Agent trace persistence now tracks a per-session count watermark and extracts only the append-only suffix that has not been persisted yet. The watermark advances after `add` and `cognify` succeed; exceptions and errored pipeline results leave the window pending for retry. Trace steps with no persistable feedback advance without invoking graph ingestion.

The existing non-incremental extractor mode remains available for direct task callers, while the persistence pipeline and `improve()` use incremental windows by default.

## Acceptance Criteria

- The first persistence run processes all existing trace steps.
- A repeated run with no new trace steps performs no graph ingestion.
- Later runs process only newly appended trace steps.
- Failed graph ingestion does not advance the watermark.
- A cleared and rebuilt session recovers from a stale count watermark.

## Type of Change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots

CLI test results:

```text
30 passed, 12 warnings in 8.44s
61 passed, 12 warnings in 11.08s
All Ruff checks passed
```

## Tests

```text
python -m pytest cognee/tests/unit/tasks/memify/test_agent_trace_persist_watermark.py cognee/tests/unit/modules/memify_tasks/test_extract_agent_trace_feedbacks.py cognee/tests/unit/modules/memify_tasks/test_cognify_agent_trace_feedback.py cognee/tests/unit/memify_pipelines/test_persist_agent_trace_feedbacks_pipeline.py cognee/tests/unit/api/v1/improve/test_improve_agent_context.py -q
python -m pytest cognee/tests/unit/modules/agent_memory/test_agent_memory.py -q
python -m ruff check <changed Python files>
```

## Pre-submission Checklist

- [x] **I have tested my changes thoroughly before submitting this PR** (See `CONTRIBUTING.md`)
- [x] **This PR contains minimal changes necessary to address the issue/feature**
- [x] My code follows the project's coding standards and style guidelines
- [x] I have added tests that prove my fix is effective or that my feature works
- [x] I have added necessary documentation (if applicable)
- [x] All new and affected existing tests pass
- [x] I have searched existing PRs to ensure this change hasn't been submitted already
- [x] I have linked any relevant issues in the description
- [x] My commits have clear and descriptive messages

## DCO Affirmation

[...]

PR #5006, opened 2026-09-09
fix: Bound graph-view seed ranking (SDK-610)

NMZivkovic:
Opening the graph view without a query or explicit seeds previously loaded every node and edge into Python to select ten IDs, exhausting memory on the reported 5.59M-node / 35.64M-edge graph.

Seed selection now delegates to `get_top_degree_node_ids(top_k)`. PostgreSQL, Neo4j, Ladybug, Neptune, and Turso rank both endpoints from at most 200,000 sampled edges in the database and return only seed IDs. PostgreSQL and Turso use one materialized edge sample for both directions. Limited ID-only queries fill sparse samples and preserve isolated nodes.

Community adapters retain the full-graph fallback, including registered adapters that do not subclass the interface. Its memory-cost warning appears once per adapter type per process. All implementations reject nonpositive limits before reading the graph.

Sampling is approximate and uses a physical prefix, not random sampling. It can systematically favor early ingestion and miss recent hubs as the graph grows. The bound limits aggregation and returned data; it does not guarantee bounded query-planning cost or latency. The original exact PostgreSQL aggregate measured 57 seconds and about 8.5 GB of temporary spill. No new large-scale timing claim is made for these revised queries or the other backends.

Validation:
- 71 adapter tests passed, including real PostgreSQL 14, Ladybug, and Turso databases. One optional Neptune import test skipped because botocore was unavailable; shared Cypher query coverage passed. Live Neo4j and Neptune services were not exercised locally.
- 148 visualization unit/API tests passed.
- Real PostgreSQL seed cases now live in the adapter test file already run by the graph database CI workflow, covering empty graphs, isolated nodes, both edge directions, and duplicate-free sample completion.
- Ruff lint/format and whitespace checks passed on changed files.

This addresses seed selection only. Neighborhood expansion still applies `max_nodes` after reading into Python, so large hub neighborhoods can still exhaust memory. Bounding that read remains separate SDK-610 work. No MCP protocol or frontend code changes.

Related: [SDK-610](https://linear.app/cognee/issue/SDK-610). Cloud port: CLO-777.

### DIFF
--- a/cognee/infrastructure/databases/graph/graph_db_interface.py
+++ b/cognee/infrastructure/databases/graph/graph_db_interface.py
@@ -543,6 +543,49 @@ class GraphDBInterface(ABC):
         """
         raise NotImplementedError
 
+    async def get_top_degree_node_ids(self, top_k: int) -> list[str]:
+        """Ids of up to ``top_k`` well-connected nodes, to seed a graph view.
+
+        **Approximate by contract.** An adapter may sample rather than count
+        exactly, and the order of near-equal nodes may differ between calls, so
+        callers must not treat the result as a ranking — only as "some nodes
+        worth starting from". Ranking exactly is what made this unusable: see
+        below.
+
+        Deliberately NOT abstract: every adapter inherits this working
+        implementation, so a community adapter keeps loading. But the default
+        is the expensive one — it reads the whole graph and counts degree in
+        Python, which on a 5.59M-node / 35.6M-edge graph means tens of
+        gigabytes of Python objects to produce ten ids, and got the worker
+        OOM-killed at ~20.8 GB RSS instead of answering.
+
+        This is the seed source for the default (no query, no explicit seed)
+        graph visualization, so it is a hot path, not a corner — which is why
+        the exactness is what gives, not the feature. Note that an exact SQL
+        aggregate is not the answer either: measured on that graph it took 57 s
+        and spilled ~8.5 GB to temp, because it must group 71M endpoint rows
+        into 5.59M distinct ids. Overriding adapters should bound the work,
+        not just move it into the database.
+        """
+        logger.warning(
+            "%s has no native get_top_degree_node_ids; falling back to a full "
+            "graph read to rank %d seeds. This is O(graph) in memory.",
+            type(self).__name__,
+            top_k,
+        )
+        nodes, edges = await self.get_graph_data()
+        if not nodes:
+            return []
+
+        degree: dict[str, int] = {str(node_id): 0 for node_id, _ in nodes}
+        for edge in edges:
+            for endpoint in (str(edge[0]), str(edge[1])):
+                if endpoint in degree:
+                    degree[endpoint] += 1
+
+        ranked = sorted(degree.items(), key=lambda item: item[1], reverse=True)
+        return [node_id for node_id, _ in ranked[:top_k]]
+
     @abstractmethod
     async def get_graph_metrics(self, include_optional: bool = False) -> dict[str, Any]:
         """
--- a/cognee/infrastructure/databases/graph/ladybug/adapter.py
+++ b/cognee/infrastructure/databases/graph/ladybug/adapter.py
@@ -2849,6 +2849,26 @@ class LadybugAdapter(GraphDBInterface):
 
     # Graph-wide Operations
 
+    async def get_top_degree_node_ids(self, top_k: int) -> list[str]:
+        """Rank seeds in the query engine rather than in Python.
+
+        The inherited default reads the whole graph to keep ten ids, which is
+        what makes the default visualization unopenable on a large graph.
+        """
+        if top_k < 1:
+            raise ValueError("top_k must be >= 1")
+
+        rows = await self.query(
+            """
+            MATCH (n:Node)-[e:EDGE]-()
+            RETURN n.id AS id, count(e) AS degree
+            ORDER BY degree DESC, id
+            LIMIT $top_k
+            """,
+            {"top_k": top_k},
+        )
+        return [str(row[0]) for row in rows if row and row[0] is not None]
+
     async def get_graph_data(
         self,
     ) -> tuple[list[tuple[str, dict[str, Any]]], list[tuple[str, str, str, dict[str, Any]]]]:
--- a/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
+++ b/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
@@ -1829,6 +1829,27 @@ class Neo4jAdapter(GraphDBInterface):
 
         return (nodes, edges)
 
+    async def get_top_degree_node_ids(self, top_k: int) -> list[str]:
+        """Rank seeds in Cypher rather than by reading the graph into Python.
+
+        The inherited default calls get_graph_data() and counts degree in
+        memory, which on a large graph exhausts the worker to produce ten ids.
+        """
+        if top_k < 1:
+            raise ValueError("top_k must be >= 1")
+
+        rows = await self.query(
+            """
+            MATCH (n)
+            WHERE n.id IS NOT NULL
+            RETURN n.id AS id, COUNT { (n)--() } AS degree
+            ORDER BY degree DESC, id
+            LIMIT $top_k
+            """,
+            {"top_k": top_k},
+        )
+        return [str(row["id"]) for row in rows if row.get("id") is not None]
+
     async def get_graph_data(self):
         """
         Retrieve comprehensive data about nodes and relationships within the graph.
--- a/cognee/infrastructure/databases/graph/postgres_demo/adapter.py
+++ b/cognee/infrastructure/databases/graph/postgres_demo/adapter.py
@@ -702,6 +702,60 @@ class PostgresDemoAdapter(GraphDBInterface):
             for row in result.mappings().all()
         ]
 
+    # Endpoint rows sampled per side when ranking seeds. Measured on a
+    # 5.59M-node / 35.6M-edge graph: the exact aggregate over all 71M endpoint
+    # rows took 57 s and spilled ~8.5 GB to temp, while sampling 200k per side
+    # took 1.14 s and returned the same top five hubs. Raising it buys nothing
+    # here — TABLESAMPLE SYSTEM(0.2%) cost 3.2 s for an identical answer.
+    _SEED_SAMPLE_ROWS = 200_000
+
+    async def get_top_degree_node_ids(self, top_k: int) -> list[str]:
+        """Approximately highest-degree node ids, from a bounded edge sample.
+
+        Ranking exactly is the wrong trade here. The inherited default reads
+        every node and edge into Python and gets the worker OOM-killed at this
+        scale; an exact SQL aggregate avoids the OOM but has to group all 71M
+        endpoint rows into 5.59M distinct ids, which measured 57 s with an
+        external sort spilling ~8.5 GB to temp. Neither is usable for a view
+        that opens on page load.
+
+        These ids are seeds for a default 500-node view, so "a genuinely
+        well-connected node, cheaply" is the actual requirement, not the
+        mathematically-top-k. Counting endpoints within a bounded slice of
+        graph_edge satisfies it: hub nodes appear in any slice precisely
+        because they touch so many edges, and on the graph above this
+        recovered all five of the true top five in 1.14 s.
+
+        Two honest caveats, in exchange for that:
+          * The result is approximate, and the *order* of near-equal hubs can
+            differ between calls. Callers must not treat it as a ranking.
+          * LIMIT without ORDER BY takes a physically-contiguous slice, not a
+            random one, so this leans on hubs being spread through the table.
+            TABLESAMPLE would remove that assumption; it was measured 3x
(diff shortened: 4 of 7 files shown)


NMZivkovic:
PR number in the ticket comment reads #5005; the actual PR is #5006. Noting here rather than editing the ticket, so the correction is visible from both sides.

cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
```diff
@@ -1829,6 +1829,27 @@ async def get_model_independent_graph_data(self):
...
+
+        The inherited default calls get_graph_data() and counts degree in
+        memory, which on a large graph exhausts the worker to produce ten ids.
+        """
+        if top_k < 1:
+            raise ValueError("top_k must be >= 1")
+
+        rows = await self.query(
+            """
+            MATCH (n)
```
goran-radonic, line 1843: This still performs an unbounded database-side aggregation: MATCH (n) scans every eligible node, COUNT { (n)--() } traverses their relationships, and only then does LIMIT $top_k apply. That improves worker memory usage, but it conflicts with the new bounded-work contract. Please use a bounded candidate sample or another bounded-degree strategy for Neo4j.

cognee/infrastructure/databases/graph/ladybug/adapter.py
```diff
@@ -2849,6 +2849,27 @@ async def remove_connection_to_successors_of(
...
 
+    async def get_top_degree_node_ids(self, top_k: int) -> list[str]:
+        """Rank seeds in the query engine rather than in Python.
+
+        The inherited default reads the whole graph to keep ten ids, which is
+        what makes the default visualization unopenable on a large graph.
+        """
+        if top_k < 1:
+            raise ValueError("top_k must be >= 1")
+
```
goran-radonic, line 2860: Both have to compute a degree for every node before ORDER BY can pick ten, so they remain O(V+E) inside the store. The LIMIT cannot prune anything, which is exactly the argument #5003 makes in the opposite direction for its own ORDER BY. That is still a large improvement over building 35.6M Python tuples, and the PR says so honestly in the body ("their improvement is avoiding full-graph materialization in Python").

The gap is between that honesty and the PR title. "Bound graph-view seed ranking" is true of one adapter out of three. Every number in the description, and every number in the docstrings, is Postgres: 57 s versus 1.14 s, 8.5 GB spill, 200k sample. Ladybug is the default graph backend per the project's own docs, so the backend most users are running has no measurement attached to the claim that the view becomes openable at 5.6M nodes, and the only test behind it is a four-node graph. Either measure Ladybug at something resembling that scale, or narrow the title and say plainly in the body that Ladybug and Neo4j get a memory fix rather than a latency fix.

Checked and fine: COUNT { ... } is Neo4j 5 syntax, the extra pins neo4j>=5.28.0,<6 (pyproject.toml:124), and the adapter already uses that form elsewhere (neo4j_driver/adapter.py:2431), so there is no 4.x compatibility problem. The return shapes are also right in both: Neo4j's query() is typed list[dict[str, Any]] so row["id"] / row.get("id") is correct, and Ladybug's is list[tuple] so row[0] is correct. Easy to get backwards, and it is not.



+ cognee/infrastructure/databases/graph/neo4j_driver/adapter.py:1843 — MATCH (n) WHERE n.id IS NOT NULL RETURN n.id, COUNT { (n)--() } ... ORDER BY degree DESC, id LIMIT $top_k

cognee/infrastructure/databases/graph/postgres_demo/adapter.py
```diff
@@ -702,6 +702,74 @@ async def _fetch_edges_within(
...
+            random one, so this leans on hubs being spread through the table.
+            TABLESAMPLE would remove that assumption; it was measured 3x
+            slower for the same answer, so the assumption is kept knowingly.
+        """
+        if top_k < 1:
+            raise ValueError("top_k must be >= 1")
+
+        async with self.sessionmaker() as session:
+            result = await session.execute(
+                text(
```
goran-radonic, line 745: Neither subquery has an ORDER BY, so both take the physically-first rows of graph_edge and the query is counting both endpoints of the same roughly 200k edges rather than sampling each column independently. The PR body's "counts endpoints from a bounded sample of 200,000 edge rows per side" implies otherwise.

Not a bug, and arguably the better behaviour, since both endpoints of one coherent subgraph is a more meaningful degree count than two disjoint slices. But it should be described as what it is, and it is also not guaranteed: nothing in the SQL forces the two subqueries onto the same rows, so a plan change (a parallel seq scan, say) could silently make them disjoint. If the coherent-subgraph reading is the intent, one subquery selecting both columns would state it and remove the ambiguity.

cognee/infrastructure/databases/graph/graph_db_interface.py
```diff
@@ -543,6 +543,55 @@ async def get_graph_data(self) -> tuple[list[Node], list[EdgeData]]:
...
+        This is the seed source for the default (no query, no explicit seed)
+        graph visualization, so it is a hot path, not a corner — which is why
+        the exactness is what gives, not the feature. Note that an exact SQL
+        aggregate is not the answer either: measured on that graph it took 57 s
+        and spilled ~8.5 GB to temp, because it must group 71M endpoint rows
+        into 5.59M distinct ids. Overriding adapters should bound the work,
+        not just move it into the database.
+        """
+        if top_k < 1:
+            raise ValueError("top_k must be >= 1")
```
goran-radonic, line 574: Correct to warn. But with Turso and Neptune inheriting it, this is one warning per default graph-view open, immediately before a read that on a large graph will not return anyway. A module-level _warned set keyed on type(self).__name__ gives the same signal once per process without adding a line to the log for every page load on a small graph where the fallback is perfectly fine.

goran-radonic (commented):
2. Turso and Neptune keep the OOM path, and the test meant to catch that excludes them

cognee/infrastructure/databases/graph/turso/adapter.py — no get_top_degree_node_ids, inherits the full-graph default
cognee/infrastructure/databases/graph/neptune_driver/adapter.py — same
cognee/tests/unit/infrastructure/databases/graph/test_top_degree_node_ids.py:118 — test_in_tree_adapters_override_the_expensive_default is parameterized over postgres_demo, neo4j_driver and ladybug only

The docstring on that test reads "These are the adapters cognee ships; none should be on the full-read path." cognee ships five in-tree GraphDBInterface subclasses, and the two omitted from the list are precisely the two that fail the property the test is named after. So the test passes by not asking about them, and a reader scanning the file comes away believing the opposite of what is true.

This matters beyond tidiness because Neptune is the AWS managed graph service, which is the backend most likely in practice to be holding 5.59M nodes. The OOM this PR is fixing is more probable there than on the Postgres demo adapter, which the project's own docs describe as not production-ready.

Two acceptable resolutions, and I would not insist on which. Implement both overrides and add them to the parametrize list. Or add them to the list with an explicit known-gap marker naming the follow-up ticket, so the test documents the real state of the codebase instead of a flattering subset of it.

3. The sampling bias is temporal and systematic, not the generic "not random" the docstring describes

cognee/infrastructure/databases/graph/postgres_demo/adapter.py:748 — (SELECT target_id AS node_id FROM graph_edge LIMIT :sample)
cognee/infrastructure/databases/graph/postgres_demo/adapter.py:740 — the caveat, which says "LIMIT without ORDER BY takes a physically-contiguous slice, not a random one, so this leans on hubs being spread through the table"

The caveat is there and it is honest, which is more than most PRs manage. But it understates the shape of the problem. graph_edge is written by cognify as documents are ingested, so the physically-first rows are not an arbitrary contiguous slice, they are the oldest edges. That has two consequences the docstring does not name:

A hub that formed entirely out of recently ingested documents is invisible to the sample no matter how large its degree, because none of its edges are in the first 200k rows. And on a graph that keeps growing, the sampled window stops moving, so the default view's seeds freeze on the earliest ingestion and the visualization quietly stops reflecting anything the user has added since. For a feature whose whole purpose is "show me my graph", seeds that never update is a user-visible failure mode, not just an approximation.

The PR rejected TABLESAMPLE SYSTEM(0.2%) on the basis that 3.2 s is 3x slower than 1.14 s for an identical answer on the one graph measured. That comparison was made against an assumed random-slice risk. Against a systematic recency blind spot, a 2 s penalty on a view that opens on page load looks like the better trade, and it removes the freeze entirely. At minimum the docstring should name the recency bias, because "leans on hubs being spread through the table" reads as a probabilistic caveat and this one is deterministic.


5. The Postgres path has no real-database coverage in CI, and it breaks a local convention

cognee/tests/unit/infrastructure/databases/graph/test_top_degree_node_ids.py:150 — captures SQL off a hand-rolled _Session
cognee/tests/unit/infrastructure/databases/graph/test_top_degree_node_ids.py:215 — the sparse-fill test drives a MagicMock sessionmaker

Every Postgres assertion is against a mock. The PR says the adapter was exercised against a temporary local instance covering the empty graph, isolated nodes, both edge directions and sample completion, which I believe, but a manual run is not repeatable and Ladybug got a real integration test while Postgres did not.

The specific thing that made me look is a convention break:

cognee/infrastructure/databases/graph/postgres_demo/adapter.py:762 — SELECT id FROM graph_node WHERE NOT (id = ANY(:seed_ids)) LIMIT :remaining, with seed_ids possibly []
cognee/infrastructure/databases/graph/postgres_demo/adapter.py:417 — delete_nodes guards if not node_ids: return before its ANY(:ids)
cognee/infrastructure/databases/graph/postgres_demo/adapter.py:442 — get_nodes guards the same way

Every other ANY(:ids) in the file guards against an empty list, and empty-array binding is a classic asyncpg type-inference failure, so this looked like the bug. I tested it rather than reporting it: against a live Postgres 17 through the same SQLAlchemy plus asyncpg path, WHERE NOT (id = ANY(:seed_ids)) with [] binds cleanly and returns every row, which is the correct behaviour here. So it works. But since it reads exactly like the oversight the two neighbours guard against, a one-line comment saying the empty case is intentional and correct would stop the next reader from "fixing" it into a guard that would then skip isolated nodes on an edgeless graph.



What is genuinely good here

The tests are the best in this stack. test_default_ignores_edges_pointing_outside_the_node_set (test_top_degree_node_ids.py:104) covers a dangling-edge case that would have invented a seed with no node, which is the kind of thing that only surfaces in production. test_postgres_ranking_is_bounded_not_an_exact_aggregate asserts on emitted SQL and bound parameters rather than module source, with a docstring explaining that a comment could satisfy a text search, which is the same discipline #5003 showed and is right both times. The Ladybug integration test (test_ladybug_degree_seeds.py:17) stubs get_graph_data to raise and asserts it is never awaited, so the test fails loudly if someone reintroduces the full read, rather than just getting a slower pass.

The choice to leave the expensive implementation on the interface instead of deleting it, and to test that it is not abstract, is the right one for a project with out-of-tree adapters.

cognee/tests/integration/infrastructure/graph/test_ladybug_degree_seeds.py
```diff
@@ -0,0 +1,48 @@
...
+
+        for source, target in [("incoming", "hub"), ("hub", "outgoing")]:
+            await adapter.query(
+                "MATCH (a:Node), (b:Node) WHERE a.id = $source AND b.id = $target "
+                "CREATE (a)-[:EDGE {relationship_name: 'rel'}]->(b)",
+                {"source": source, "target": target},
+            )
+
+        assert await adapter.get_top_degree_node_ids(1) == ["hub"]
+        assert await adapter.get_top_degree_node_ids(10) == [
```
goran-radonic, line 40: This asserts the full exact ordering, but GraphDBInterface.get_top_degree_node_ids's own docstring says the order of near-equal nodes may differ between calls and callers must not treat the result as a ranking. An adapter that honours that contract breaks this test for no real reason.

Asserting the set plus result[0] == "hub" keeps everything the test actually cares about.

Separately, stubbing get_graph_data to raise and asserting it's never awaited is a nice touch. That's the bit that will catch a regression rather than just get slower.

cognee/infrastructure/databases/graph/graph_db_interface.py
```diff
@@ -543,6 +543,55 @@ async def get_graph_data(self) -> tuple[list[Node], list[EdgeData]]:
...
+        **Approximate by contract.** An adapter may sample rather than count
+        exactly, and the order of near-equal nodes may differ between calls, so
+        callers must not treat the result as a ranking — only as "some nodes
+        worth starting from". Ranking exactly is what made this unusable: see
+        below.
+
+        ``top_k`` must be positive. Isolated nodes are valid seeds; a graph
+        without edges should still produce a nonempty view if it has nodes.
+
+        Deliberately NOT abstract: every adapter inherits this working
```
goran-radonic, line 558: "every adapter inherits this working implementation, so a community adapter keeps loading" holds for adapters that subclass GraphDBInterface. use_graph_adapter.py:5 is a plain dict assignment with no subclass check, and resolve_seeds_by_degree(graph_engine: Any, ...) duck-types, so a registered adapter that doesn't subclass gets an AttributeError on the default graph view. Before this, the call was get_graph_data(), which any adapter has to have.

I couldn't check the community repo from here so this may be moot, but if any of those adapters register without subclassing, a getattr fallback to the old in-Python count would keep the promise the docstring makes.

goran-radonic (approved):

PR #5009, opened 2026-09-09
fix: Make session improve idempotent and cheap

siillee:
## Summary

Follow-up to #4920 (improve as an explicit orchestrator) closing what is left of
SDK-593 on the server side. The plugin work landed in SDK-594.

Cloud testing of the Claude Code plugin showed one prompt producing dozens of improve
operations and busy answers retried in a loop. #4920 removed most of that cost: a lost
lock claim now records a `noop` row and returns `lock_held` instead of a succeeded
improve, every session stage is watermark-gated, distillation and trace persistence
no longer redo their work, and background mode returns at once. This PR adds the one
piece that was still missing for the plugin's session end.

Fixes SDK-593. Companion: SDK-594. Follow-up: SDK-613 (distillation batch trigger).

## The gap

A plugin session's last minute: the idle improve fires and takes the lock; the user
exits; the final sync flushes its buffered entries and calls improve; that call gets
`lock_held` and returns. The idle improve only bridged what existed when it started,
and no further trigger ever fires for that session. The entries the final sync flushed
never reach the graph.

`remember()` already handles the same collision for the SDK path by re-arming its
debounce so the next call retries. The plugin's final sync has no next call, and
retrying busy answers is exactly the storm SDK-594 removed.

## What changes

**A rerun request on the improve lock** (`cognee/infrastructure/locks/session_lock.py`)

- `request_improve_rerun_many(keys)`: a run that lost its claim flags the holder of
  its session keys. Returns `True` only when a key is actually held — a request with
  nobody to fulfil it is not recorded.
- `release_or_rerun_improve_lock_many(keys, rerun_keys=...)`: the holder's release
  checks the flag in the same critical section as the release. A pending request is
  consumed, every key stays claimed, and the caller runs the stages once more.
  Nothing can land between "checked" and "released" and be lost.
- A fresh claim clears stale requests: a new holder starts with a full watermark pass,
  which is what the request asked for.
- Same scope as the claim itself: in-process, per worker.

**Bounded rerun loop in the orchestrator** (`cognee/api/v1/improve/improve.py`)

- After the stages run, a session-keyed run releases through the new primitive. On
  `False` it runs the stages again; every stage is watermark-gated, so the extra pass
  touches only what arrived meanwhile. `IMPROVE_MAX_RERUN_PASSES` (3) bounds it; past
  the bound the lock is released and a leftover request is the next claimant's.
- **Session keys only.** Dataset-only runs are untouched: a burst of `remember()` calls
  on one dataset still collapses into one enrichment. A session-keyed loser colliding
  with a dataset-only holder gets `rerun_requested=False`: no promise is made that
  nobody can keep.
- A lost claim on session keys sets `rerun_requested=True` on the `lock_held` result.

**Result shape** (`cognee/modules/improve/result.py`)

- `ImproveResult.rerun_requested: bool` and `rerun_passes: list[list[StageResult]]`.
  `stages` stays the first pass. `status` covers rerun passes, so an errored stage in
  an extra pass still marks the run `errored` and its operation row `failed`.
  `stage_summary()` appends `rerun_passes=N` when any ran. Additive; existing fields
  and their meaning are unchanged.

**Surfaces**

- `record_operation.merge_run_info(..., allow_overwrite=True)`: a rerun pass refreshing
  its own stage's enrichment stamp is expected, not a clobber to warn about.
- `remember()`: when the lock-held bridge reports `rerun_requested`, the holder covers
  this entry, so the debounce window is not refunded.
- CLI: the `lock_held` headline says the running improve will cover the newer entries
  when a rerun was requested, and a completed run prints how many extra passes it ran.
- CLAUDE.md: one sentence in the orchestrator section.

## Merge notes

This branch merged `dev` after #4920. Everything SDK-593 originally added that #4920
superseded was dropped in dev's favour: the pending-work probe, the time-based
distillation watermark, the separate trace-persist watermark, the status-body
responses, the CLI status renderer, the DB-backed lock with TTL, and the session end
endpoint. The last two were dropped by decision (multi-worker exclusion is not needed
on our infra; the end endpoint waits on the cloud team), not because dev covers them.

## Testing

- `cognee/tests/unit/infrastructure/locks/test_improve_rerun_lock.py`: request lands
  only on held keys, release consumes exactly once and keeps every key, dataset-key
  requests are ignored by a session-scoped release, a fresh claim clears stale flags.
- `cognee/tests/unit/api/v1/improve/test_improve_rerun.py` (dev's `FakeStage` harness):
  session loser flags the holder and the holder runs one extra pass; dataset-only
  collisions make no request and no extra pass; session loser against a dataset-only
  holder gets no promise; the pass bound holds and the leftover request is cleared by
  the next claim; an errored rerun pass marks the run `errored` and the row `failed`;
  the stamp refresh logs no clobber warning; both new fields serialize.
- Full unit suite: 6559 passed, 86 skipped. Collection errors only from `plotly` /
  `neo4j` not installed locally. CLI improve and dispatch suites pass. Ruff clean.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/api/v1/improve/improve.py
+++ b/cognee/api/v1/improve/improve.py
@@ -32,7 +32,13 @@ logger = get_logger("improve")
 IMPROVE_MAX_RERUN_PASSES = 3
 
 # Strong refs for background improve tasks: the event loop keeps only weak
-# references, so an un-anchored task can be garbage-collected mid-run.
+# references, so an un-anchored task can be garbage-collected mid-run. Tasks
+# discard themselves on completion. The set is not capped: single-session
+# improves are already bounded to one in-flight task per session by the
+# improve lock (the only path the plugin uses), and multi-session improves
+# serialize on the dataset lock inside memify. Unbounded but self-draining,
+# the same trade-off as ``_BACKGROUND_PIPELINE_TASKS`` and
+# ``_BACKGROUND_REMEMBER_TASKS``.
 _BACKGROUND_IMPROVE_TASKS: set[asyncio.Task] = set()
 
 
--- a/cognee/api/v1/improve/improve.py
+++ b/cognee/api/v1/improve/improve.py
@@ -1,3 +1,5 @@
+import asyncio
+from dataclasses import dataclass, field
 from typing import Any
 from uuid import UUID
 
@@ -18,9 +20,19 @@ from cognee.modules.operations import record_operation
 from cognee.modules.pipelines.layers.resolve_authorized_user_datasets import (
     resolve_authorized_user_datasets,
 )
+from cognee.modules.session_bridge import SessionPendingWork, probe_sessions_pending_work
 from cognee.shared.logging_utils import get_logger
 
 logger = get_logger("improve")
+
+# Upper bound on "one more pass" reruns a single lock hold performs when other
+# callers keep finding the lock busy. Whatever is still above the watermarks
+# after that is picked up by the next trigger.
+IMPROVE_MAX_RERUN_PASSES = 3
+
+# Strong refs for background improve tasks: the event loop keeps only weak
+# references, so an un-anchored task can be garbage-collected mid-run.
+_BACKGROUND_IMPROVE_TASKS: set[asyncio.Task] = set()
 
 
 class ImproveKwargs(TypedDict, total=False):
@@ -34,6 +46,43 @@ class ImproveKwargs(TypedDict, total=False):
     vector_db_config: dict
     graph_db_config: dict
     feedback_alpha: float
+
+
+@dataclass
+class _SessionImprovePlan:
+    """Everything one session-bridging improve run needs, resolved up front."""
+
+    dataset: str | UUID
+    write_dataset_ref: Any  # the resolved dataset UUID
+    session_ids: list[str]
+    user: Any
+    feedback_alpha: float
+    node_name: list[str] | None
+    build_global_context_index: bool
+    build_truth_subspace: bool
+    forced: bool
+    memify_kwargs: dict = field(default_factory=dict)
+    # Set when this run holds the single-session improve lock.
+    lock_session_id: str | None = None
+    lock_token: str | None = None
+    background: bool = False
+
+
+def _status_payload(status: str, *, dataset_id: Any, session_ids: list[str], **extra: Any) -> dict:
+    """The non-pipeline improve answers: ``busy``, ``no_op`` and ``accepted``."""
+    return {
+        "status": status,
+        "dataset_id": str(dataset_id),
+        "session_ids": list(session_ids),
+        **extra,
+    }
+
+
+def _pending_stage_names(pending: dict[str, SessionPendingWork]) -> list[str]:
+    names: set[str] = set()
+    for work in pending.values():
+        names.update(work.stages())
+    return sorted(names)
 
 
 async def improve(
@@ -76,24 +125,46 @@ async def improve(
 
     Without ``session_ids``, only stage 3 runs by default.
 
+    Every session stage is guarded by a per-session watermark or applied
+    marker, so repeating an improve is safe. Since SDK-593 it is also cheap:
+    before any stage starts, the session cache is probed for work above those
+    watermarks. When nothing is pending the call returns
+    ``{"status": "no_op", ...}`` at once — no LLM call, no pipeline run, no
+    operation record. Stages with nothing to do are skipped individually.
+
+    A single-session improve holds a cross-worker per-session lock (TTL
+    ``IMPROVE_LOCK_TTL_SECONDS``) for the whole run. A caller that finds it
+    held gets ``{"status": "busy", "holder_age_seconds": ..., "rerun_requested":
+    True}`` immediately — again without an operation record — and the holder
+    runs one more watermark-driven pass before releasing, so the caller's newer
+    tail is covered without any retry.
+
     Args:
         dataset: Dataset name or UUID to process.
-        run_in_background: Run processing asynchronously.
+        run_in_background: Return as soon as the run is accepted. With
+            ``session_ids`` the whole stage sequence (including agent-context
+            extraction, distillation and enrichment) runs in one ordered
+            background task that holds the lock until it finishes; the
+            answer is ``{"status": "accepted", "background": True, ...}`` and
+            the operation record is written when the run completes. Without
+            ``session_ids`` the enrichment pipeline itself runs in the
+            background, as before.
         node_name: Filter graph to specific named entities.
         session_ids: Session IDs whose feedback and Q&A content
             should be bridged into the permanent graph.
         build_global_context_index: Opt-in flag for building the global
-            context index after default enrichment. Skipped in background
-            mode because ordered background pipeline chaining is not
-            supported yet.
+            context index after default enrichment. Forces a run even when
+            no session work is pending.
         build_truth_subspace: Opt-in flag (default ``False``) for building the
             truth subspace from distilled session learnings after distillation
             and before enrichment. Only runs when ``session_ids`` is provided.
-            Off by default = no behaviour change.
+            Forces a run even when no session work is pending.
         **kwargs: Additional options -- see ``ImproveKwargs``.
 
     Returns:
-        Pipeline run info (same as ``cognee.memify()``).
+        Pipeline run info (same as ``cognee.memify()``) for a completed
+        blocking run, or one of the status dicts described above
+        (``status`` is ``"busy"``, ``"no_op"`` or ``"accepted"``).
 
     Example::
 
@@ -105,8 +176,6 @@ async def improve(
     """
     from cognee import __version__ as cognee_version
     from cognee.shared.utils import send_telemetry
-
-    stages_run = []
 
     send_telemetry(
         "cognee.improve",
@@ -135,12 +204,14 @@ async def improve(
 
         from cognee.modules.users.methods import get_default_user
 
-        async with record_operation("improve") as operation_context:
-            user = kwargs.pop("user", None)
-            if user is None:
-                user = await get_default_user()
-            operation_context.set_user(user)
+        user = kwargs.pop("user", None)
+        if user is None:
+            user = await get_default_user()
 
+        # Preconditions run before the operation record is opened, so a busy or
+        # no-op answer below writes no row. A precondition failure is still
+        # recorded as a failed improve, exactly as before.
+        try:
             # The pipeline-run log writers INSERT the operation-record columns
             # (user_id, outcome, tokens, ...), so an existing database must be
             # at the current Alembic head before the first write — same gate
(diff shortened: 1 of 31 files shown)


cognee/infrastructure/locks/session_lock.py
```diff
@@ -98,40 +107,280 @@ async def session_turn_lock(user_id: Any, session_id: Any) -> AsyncGenerator[Non
...
+            if _rowcount(result) != 1:
+                return None  # lost the race to a concurrent claimant
+            if expired:
+                logger.warning(
+                    "improve lock: took over an expired lock on session '%s' "
+                    "(held for %s s, TTL %d s)",
+                    session_id,
+                    "unknown" if age is None else int(age),
+                    improve_lock_ttl_seconds(),
+                )
```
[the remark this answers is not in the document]
siillee: The warning already sits where the reviewer wants it. In try_acquire_improve_lock the order is: UPDATE, commit, if _rowcount(result) != 1: return None, then if expired: logger.warning(...). The code block in the comment is character for character what is on disk. The reviewer is most likely reading an earlier revision of the diff. Nothing to change.

cognee/infrastructure/locks/session_lock.py
```diff
@@ -98,40 +107,280 @@ async def session_turn_lock(user_id: Any, session_id: Any) -> AsyncGenerator[Non
...
+            update(SessionRecord)
+            .where(
+                and_(
+                    SessionRecord.session_id == session_id,
+                    SessionRecord.user_id == user_uuid,
+                    SessionRecord.improve_lock_token.is_not(None),
+                )
+            )
+            .values(improve_rerun_requested=True)
+        )
```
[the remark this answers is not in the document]
siillee: The interleaving is real: a caller observes holder A, A expires and B takes over, then the caller's flag lands on B's row. What follows is fine. B's acquire started a full watermark pass, and the caller's entries were already in the cache before the caller ever called improve, so B's first pass covers them. When B's release consumes the flag, it re-probes, finds nothing above the watermarks, and the loop exits without running any stage. The "redundant work" is one cache probe. A gets LOST at its release, which is correct because A no longer holds anything, and the promise made to the caller is kept by B. The proposed fix makes things worse: matching on the observed acquired_at would return busy=False after a takeover, the caller would retry the acquire against B, fail, and report busy with rerun_requested=False, which is a weaker promise than today for a scenario that is already covered.

cognee/modules/session_bridge/pending_work.py
```diff
@@ -0,0 +1,167 @@
...
+            exc_info=True,
+        )
+        return SessionPendingWork(
+            session_id=session_id,
+            new_qa=True,
+            feedback_qas=True,
+            new_traces_to_persist=True,
+            new_traces_for_agent_context=True,
+            distillable_entries=True,
+        )
```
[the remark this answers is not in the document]
siillee: Fail-closed here does not hide the failure: the probe logs with traceback and the stages then fail loudly on the same cache, recorded as errored pipeline runs. Raising from the probe would produce one unrecorded exception instead, since it runs before the operation record opens. A disabled cache is already reported as no_op with reason=session_cache_unavailable. Keeping fail-closed.

cognee/api/v1/improve/improve.py
```diff
@@ -18,10 +21,20 @@
...
 logger = get_logger("improve")
 
+# Upper bound on "one more pass" reruns a single lock hold performs when other
+# callers keep finding the lock busy. Whatever is still above the watermarks
+# after that is picked up by the next trigger.
+IMPROVE_MAX_RERUN_PASSES = 3
+
+# Strong refs for background improve tasks: the event loop keeps only weak
+# references, so an un-anchored task can be garbage-collected mid-run.
+_BACKGROUND_IMPROVE_TASKS: set[asyncio.Task] = set()
```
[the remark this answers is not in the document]
siillee: Single-session background improves are capped at one per session by the lock, which is the only path the plugin uses. Multi-session calls are unbounded but serialize on the dataset lock and self-drain, same as _BACKGROUND_PIPELINE_TASKS. Adding a comment stating that.

NMZivkovic (commented):
Reviewed mostly from the plugin side. The direction is right, and moving the lock and the probe ahead of `record_operation` is the change that actually stops busy skips from showing up as improve rows.

Four things inline. The two I'd want fixed before merge are in the probe path, because they decide whether "a redundant improve costs nothing" holds: a rated QA whose weights failed to apply keeps a session from ever reaching `no_op`, and distillation advances its watermark when the curator call failed, not only when it judged the entries. The busy/dead-holder one matters for the plugin's final sync, and I'm happy to take the client half. I skipped the lock races the bot already raised.

cognee/modules/session_bridge/pending_work.py
```diff
@@ -0,0 +1,167 @@
...
+            except (TypeError, ValueError):
+                return 0
+    return 0
+
+
+def _has_eligible_feedback(entries: list) -> bool:
+    # Imported lazily: the eligibility rule lives with the task that consumes it.
+    from cognee.tasks.memify.extract_feedback_qas import _is_eligible
+
+    return any(isinstance(entry, SessionQAEntry) and _is_eligible(entry) for entry in entries)
```
NMZivkovic, line 92: This can stay true forever, so the session never gets a `no_op`.

`_is_eligible` only drops an entry once `memify_metadata[feedback_weights_applied] is True`. `_process_feedback_item` writes `False` whenever a referenced node or edge isn't found, and the getters return only found ids, so one element deleted since the recall (forget, prune) is enough. On kuzu, neptune, postgres_demo and turso `get_node_feedback_weights` isn't implemented at all: it raises before anything is marked, `_bridge_sessions` logs it as non-fatal, and the entry stays unmarked.

Either way a single rated QA keeps `feedback_qas=True` on every probe. On ladybug/neo4j it's more than cost: every improve re-applies the update to the elements that were found, so that one rating keeps pulling their weights.

I'd treat a written marker as done here and in `_is_eligible` (key present, `True` or `False`), or cap the retries, and only report feedback work when the graph adapter implements the weight getters.

cognee/modules/session_distillation/distill.py
```diff
@@ -396,11 +441,17 @@ async def distill_session(
 
     proposed = await propose_lessons(qa_rows, context_entries)
     if not proposed:
+        await advance_distillation_watermark(scope, qa_rows, context_entries)
```
NMZivkovic, line 444: This also advances when the LLM call failed, not only when the curator looked and dropped everything.

`curate_batch` catches every exception (and a missing prompt file) and returns `[]`, so a provider outage, a 429 or a bad key arrives here as `no_proposed_lessons` and the watermark moves past the whole window. Same at 449: `write_or_reject` fails open to `None`, so `accept_proposed_lessons` coming back empty can mean every writer call failed. Those entries are never curated after that. The module docstring says an exception leaves the watermark in place, but nothing propagates far enough for that to happen.

`test_distill_session_advances_on_no_proposed_lessons` mocks `propose_lessons` to `[]`, so it can't tell the two cases apart. Could `propose_lessons` / `accept_proposed_lessons` report whether at least one LLM call actually completed, and advance only then?

cognee/api/v1/improve/routers/get_improve_router.py
```diff
@@ -61,6 +62,18 @@ async def improve(payload: ImprovePayloadDTO, user: User = Depends(get_authentic
...
 
+        ## Response
+        A completed blocking run answers with the pipeline run info keyed by
+        dataset id (the memify shape). With ``sessionIds`` the answer can instead
+        be a status object — always HTTP 200, so clients need not special-case it:
+        - ``{"status": "no_op", "reason": ...}``: nothing above the session
+          watermarks; no LLM call, no pipeline run, no operation record.
+        - ``{"status": "busy", "holder_age_seconds": ..., "rerun_requested": true}``:
+          another improve of that session is in flight; it will run one more
+          pass over the newer tail before releasing, so do not retry.
```
NMZivkovic, line 73: "Do not retry" only holds while the holder is alive, and nothing acts on the flag otherwise.

`busy` is only returned for a holder younger than the TTL. If that holder died (a deploy or OOM mid-improve, the case the description calls out), `improve_rerun_requested` stays set, and outside `session_lock.py` nothing picks it up. The promised pass only happens if another improve for that session id arrives after the TTL.

For the Claude Code plugin the final sync at SessionEnd is now the last trigger, since SDK-594 dropped the idle-watcher shutdown improve. If that final improve gets `busy` from a dead holder and treats it as landed, as the description proposes, the session's tail is never bridged.

Two cheap options: include `lock_expires_in_seconds` (TTL minus holder age) in the busy body so a final sync can make one retry after it, or have `POST /sessions/{id}/end` return `improve_rerun_requested` / `improve_lock_held` (already computed in `to_dict`) so the client can see the promise is still outstanding. I'll do the plugin side either way.

cognee/api/v1/improve/improve.py
```diff
@@ -158,162 +236,437 @@ async def improve(
...
-                ):
-                    kwargs["data"] = [{}]
-
-                result = await memify(
-                    dataset=dataset,
-                    node_name=node_name,
-                    user=user,
-                    run_in_background=run_in_background,
-                    **kwargs,
+                pending = await _probe_pending(plan)
```
NMZivkovic, line 335: If this re-probe still finds nothing, we fall through into a real run anyway.

That's what `test_rerun_landing_during_no_op_with_still_nothing_pending_ends_cleanly` pins: an improve operation row with no stages, and `{}` as the result. It's the one path where a no-op still writes a row, and `{}` is exactly what the pre-PR busy answer looked like, so a blocking caller written against the old contract reads it as busy. In background mode it becomes `accepted` with `pending_stages: []` plus a `background=True` improve row that ran nothing, which the Activity page shows as running indefinitely.

Could this loop back into the release instead? Re-probe, and if it's still empty release again (honouring another RERUN the same way), and return `no_op` once the release goes through.

cognee/api/v1/improve/improve.py
```diff
@@ -144,18 +168,53 @@ async def execute_stages(
...
+                    break
+                if passes >= IMPROVE_MAX_RERUN_PASSES:
+                    logger.info(
+                        "improve: %d-pass bound reached for %s; a further rerun request "
+                        "is left to the next claimant",
+                        passes,
+                        ", ".join(session_keys),
+                    )
+                    break
+                if await release_or_rerun_improve_lock_many(lock_keys, rerun_keys=session_keys):
```
NMZivkovic, line 232: **Blocker — double release.** When this returns `True` the keys are already released; the loop breaks and the `finally` (line 231) releases them a second time, unconditionally.

That is only harmless when uncontended. If another run is queued on `_improve_registry_lock`, the release here wakes it, our re-acquire in the `finally` queues behind it, the contender claims the freed keys, and our `difference_update` then drops *its* claim. I reproduced that exact interleaving on this head: contender claimed, keys ended up unheld.

Suggest a `released = True` flag set here and `if not released:` guarding the `finally` release — the bound-break, `not session_keys`, and exception paths still need it.

cognee/api/v1/improve/improve.py
```diff
@@ -144,18 +168,53 @@ async def execute_stages(
...
+                        # A rerun pass refreshing its own stage's stamp is expected.
+                        operation.merge_run_info(
+                            stage_result.run_info_stamp, allow_overwrite=passes > 1
+                        )
+                    if stage.fatal and stage_result.status == "errored":
+                        raise _abort_run(result, stages[index + 1 :], stage, stage_result)
+
+                if not session_keys:
+                    break
+                if passes >= IMPROVE_MAX_RERUN_PASSES:
```
NMZivkovic, line 209: Narrow gap worth a line in the description: breaking here skips the flag check, so a loser that set `_rerun_requested` moments earlier was already told `rerun_requested=True` and will not retry, while this run releases and leaves the flag for a claimant that may never arrive (the plugin's session has ended — the exact case SDK-593 is about).

It needs three passes to reach, so I would not restructure for it, but right now the bound silently converts a kept promise into a broken one. Either note it under Merge notes, or log it at `warning` rather than `info` so it is visible if it ever happens.

NMZivkovic (commented):
Good design work — I went looking for the hole I expected in the rerun protocol and it's genuinely closed. `improve.py:280` requests a rerun only on `_session_lock_keys(lock_keys)`, so a flag is recorded only when the holder actually shares that session key and will therefore check it on release. A session loser colliding with a dataset-only holder correctly gets `False`. That's the right invariant and it's worth keeping the comment that explains it.

One blocker, one narrow gap, one latent nit.

## Blocker: the lock is released twice, and the second release can free another run's claim

`dev` has exactly one release in `execute_stages` — the `finally`. This PR adds a second: `release_or_rerun_improve_lock_many` (line 211) releases and returns `True`, the loop breaks, and the `finally` at line 231 then calls `release_improve_lock_many(lock_keys)` unconditionally.

That second call is not a no-op when another run is waiting on `_improve_registry_lock`. The holder's release inside `release_or_rerun_...` wakes the waiter, so the holder's own re-acquire fails `asyncio.Lock`'s uncontended fast path and queues *behind* it. The contender then claims the keys we just freed, and the holder's `difference_update` removes them.

Reproduced deterministically against this head:

```
contender claimed the lock : True
keys still marked held     : []
```

The keys are free while an improve is running, so a third run can claim and improve the same session concurrently — the storm this PR exists to prevent. Contention is the normal case here, not an exotic one.

Minimal fix — the bound-break, `not session_keys`, and exception paths all still need the `finally`, so make it conditional rather than removing it:

```python
released = False
...
            if await release_or_rerun_improve_lock_many(lock_keys, rerun_keys=session_keys):
                released = True
                break
...
        finally:
            ...
            if not released:
                await release_improve_lock_many(lock_keys)
```

Worth a regression test: holder takes the clean-release path while a contender is queued on the registry lock; assert the contender's keys are still held afterwards.

## Latent nit

`ImproveResult.stage(name)` (result.py:271) still walks only `self.stages`, so after a rerun it reports the first pass's outcome for that stage. No non-test callers today, so nothing is broken — but now that `all_stage_results()` exists, `stage()` looking at one pass is a trap for whoever calls it next. Either search all passes or say in the docstring that it is first-pass-only.

## Housekeeping

The two unresolved bot threads look stale against the current code: the `skipped` check already uses `all_stage_results()`, and the "DB-backed locks" blocker contradicts the description, which says that was dropped by decision. Worth resolving them so the real ones stand out.

NMZivkovic (approved):

PR #5011, opened 2026-09-09
fix(graph): return Neo4j get_edges in true edge direction

### LINKED ISSUE #4967: Neo4jAdapter.get_edges returns incoming edges with source/target inverted (traversal-side order, not true edge direction)

kaikanor:
Fixes #4967.

`get_edges` matched relationships undirected and always put the queried node first in the tuple, so a stored edge `(other) -[:REL]-> (queried)` came back as `('queried', 'other', ...)` — inverted. The query now projects `startNode(r).id` / `endNode(r).id` and builds the tuple from those, so direction is real regardless of traversal side.

Call-site audit: `get_edges` has exactly two consumers — the Postgres hybrid adapter (pure forward) and `memify_pipelines/consolidate_entity_descriptions.py` (`format_edges`), which renders these tuples as prompt material for entity-description consolidation and is where the swapped direction actually hurts. Nothing relied on the old inverted order.

Notes:
- `RETURN` trimmed to `r, startNode(r).id, endNode(r).id` — the node columns were dead weight once the tuple stopped using them.
- `str()` on both endpoint ids matches the neighboring methods in this adapter (`has_edges`, `add_edges` already `str()` what they return).
- `result["r"][1]` for the relationship type is the established pattern in this file (`get_predecessors_successors` uses `relation[1]` in production).
- Self-loop behavior is unchanged by this PR (source == target, nothing to invert).
- CI has no live Neo4j, so the new tests mock `query()` and pin the serialization contract; worth a run against your integration suite if you want the live check.

Tests: `uv run pytest cognee/tests/unit/infrastructure/databases/graph/ -q` — 6 passed (3 new: outgoing edge keeps direction, incoming edge is not inverted, cypher projects startNode/endNode).

### DIFF
--- a/cognee/api/v1/ui/ui.py
+++ b/cognee/api/v1/ui/ui.py
@@ -478,13 +478,21 @@ def start_ui(
     """
     logger.info("Starting cognee UI...")
 
+    # The MCP server is optional: the UI and backend work fine without it, so a
+    # busy MCP port only disables the MCP server instead of aborting the whole
+    # launch — same graceful degradation as the Docker-unavailable path below.
     ports_to_check = [(port, "Frontend UI")]
 
     if start_backend:
         ports_to_check.append((backend_port, "Backend API"))
 
-    if start_mcp:
-        ports_to_check.append((mcp_port, "MCP Server"))
+    if start_mcp and not _is_port_available(mcp_port):
+        logger.warning(
+            f"Port {mcp_port} is already in use (possibly a leftover cognee MCP "
+            f"container), skipping the MCP server. The UI and backend will start "
+            f"without it. Stop whatever holds the port to get the MCP server back."
+        )
+        start_mcp = False
 
     logger.info("Checking port availability...")
     all_ports_available, unavailable_services = _check_required_ports(ports_to_check)
--- a/cognee/cli/_cognee.py
+++ b/cognee/cli/_cognee.py
@@ -367,8 +367,10 @@ def main() -> int:
                 fmt.echo(f"The interface is available at: http://localhost:{frontend_port}")
                 if start_backend:
                     fmt.echo(f"The API backend is available at: http://localhost:{backend_port}")
-                if start_mcp:
+                if start_mcp and docker_container:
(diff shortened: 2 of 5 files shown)


kaikanor:
Closing this PR - no longer able to maintain it. Feel free to reuse the fix and tests if the underlying issue (Neo4j get_edges direction #4967) is still relevant. Thanks!

PR #4441, opened 2026-08-12
Feature/cog 6154 resolve issue 4326

conorlazarparman:
## Description
As issue #4326 describes, the only way that Cognee checks if a background task is active or reapable is by timing it (expires after 3600secs, that is 1 hour.) So a long run that's making progress could be unfairly shut down. This necessitates a better system of tracking a job's progress.
#4326 proposed a heartbeat tracker, which basically records the time that certain checkpoints in the task are hit. However, some tasks (e.g. long LLM inference with a local LLM) have extremely long "heartbeats" that may exceed one hour even though they are making good progress. So putting a limit on the length of each step doesn't reliably work either (3 test runs with war and piece yielded a few pretty different figures of checkpoint time.) 
The best way is just to check whether the process is still alive or not. This is accomplished using an ownership record (which machine) and process check (check the PID) that ensures we're checking if the correct process is alive.
 
## Acceptance Criteria
This updated branch does not kill jobs early in practice. 
It accounts for a variety of different possible job lengths and fairly assesses whether they're done or not.
All relevant tests pass. 

## Type of Change

- [X] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots
Pretty much every test related to LLM/ingestion pipelines, which this PR affects. 
(big command used:
.venv/bin/python -m pytest \
  cognee/tests/unit/modules/pipelines/ \
  cognee/tests/unit/modules/cognify/ \
  cognee/tests/unit/infrastructure/llm/test_llm_gateway_heartbeat.py \
  -q --timeout=60
)
Screenshot 2026-08-12 at 2 57 29 PM

## Pre-submission Checklist

- [X] **I have tested my changes thoroughly before submitting this PR** (See `CONTRIBUTING.md`)
- [X] **This PR contains minimal changes necessary to address the issue/feature**
- [X] My code follows the project's coding standards and style guidelines
- [X] I have added tests that prove my fix is effective or that my feature works
- [X] I have added necessary documentation (if applicable)
- [X] All new and existing tests pass
- [X] I have searched existing PRs to ensure this change hasn't been submitted already
- [X] I have linked any relevant issues in the description
- [X] My commits have clear and descriptive messages

## DCO Affirmation
I affirm that all code in every commit of this pull request conforms to the terms of the Topoteretes Developer Certificate of Origin.

Issue #4326, opened 2026-08-03
[Feature]: Record pipeline run liveness so stalled background runs are detectable


chiruu12:
Follow-up to #4197, splitting out the background-pipeline half as discussed
there.

@dexters1 raised the blocking question: with local LLMs a run can legitimately
take days, so how do we know when a background task is still active versus
when it should be cleaned up?

Right now we cannot know. `PipelineRun` records `created_at` and a status, but
nothing that advances while the run is working. A run stuck for six hours and a
run healthily chewing through a large document look identical to anything
outside the task.

So the supervisor is the wrong first step. The first step is a liveness signal:

- Add a heartbeat column to `PipelineRun` that the pipeline touches as it
  completes each task, not on a timer.
- Progress, not wall clock, becomes the signal. A local-LLM run that takes days
  keeps advancing, so it is never a cleanup candidate. A run whose heartbeat
  has not moved while its peers advance is the interesting case.

With that in place a cleanup policy becomes a separate, much smaller decision,
and it can start as reporting only rather than anything that terminates work.

I am not proposing to build the supervisor here. Filing this so the liveness
question has a home outside #4197, and to check the framing before anyone
writes code.

dexters1:
@chiruu12 I would avoid adding systems like this, liveness is mostly an issue for local LLM models and to check if Cognify is running just checking if the LLM model is processing requests still.

Most of the other time in the Cognee pipelines is pretty quick and I have personally never had a stuck Cognee pipeline

PR #5018, opened 2026-09-10
SDK-578 feat: Record pipeline run liveness

goran-radonic:
## Description

Right now nothing can tell a pipeline run that is still working from one whose process
died. Both look the same in `pipeline_runs`: a `DATASET_PROCESSING_STARTED` row with no
terminal row after it. So everything that needs to know falls back to the row's age, and
picks a different number for it. Startup recovery waits an hour before rolling a run back;
the dataset status endpoints call it abandoned after half that. Neither number measures
anything about the run, so a long cognify gets rolled back while it is running, and a run
that died a minute after a restart sits as "processing" until the next boot.

The signal we need is already being produced and thrown away. `log_pipeline_run_progress`
updates the run's STARTED row in place every time items complete, recording how far along
the run is but never when that was true. This adds a nullable `last_heartbeat_at` and
stamps it on that same UPDATE, so a reader can ask "when was this last alive" instead of
"how old is this row". No extra write, no extra row, no background timer.

Nothing reads the column in this PR. Recovery still decides from `created_at`, and the
read-time consumer lands separately. That split is on purpose: recovery is the destructive
reader, and a wrong verdict there runs `cognify_rollback_handler`, which deletes graph
nodes, edges and their embeddings while the live worker is still writing. I would rather
the value be visible in status for a release before anything acts on it.

Two things worth being upfront about.

First, the column is nullable with no backfill, and NULL means "no evidence either way,
judge by age", never "dead". Several paths legitimately hold a STARTED row without ever
ticking: an `update()` run driven outside `run_tasks`, or a dataset still sitting in the
queue behind others in a background multi-dataset run. Backfilling `created_at` would have
been easy and wrong, since it fabricates a tick the run never produced.

Second, the resolution is one item completion, throttled to about twenty per run. That
means it advances for a multi-item run and does not advance while a run is inside a single
long item. A one-document cognify stamps once, when that document finishes. I had a second
mechanism in here to cover that case and removed it: `PipelineRunYield` turns out to be
emitted once per terminal result of the last task rather than per stage, and with the
default `chunks_per_batch` of 2000 a whole document is one batch, so it fired exactly once
at the very end and bought nothing. Closing that gap needs an event source inside the task
itself, which is a separate decision and not attempted here. The limitation is written into
the column comment rather than left for someone to discover.

Related prior art, so reviewers have it: #4326 and #4441 proposed a fuller version of this
(heartbeat plus host and PID ownership) and were closed on the grounds that liveness is
mostly a local-LLM concern. This is deliberately much smaller: no new writer, no new
mechanism, no ownership, no reaper. It puts a timestamp on a write that already happens.

## Acceptance Criteria

* A run doing work records when it last did so, on the row a reader already selects.
* Rows written before this migration behave exactly as they do today, with no backfill.
* Write volume against `pipeline_runs` is unchanged: same number of writes, same number of
  rows, one more column in the SET.
* No behaviour change on its own, since nothing reads the column yet.

Proof:

* `test_progress_write_stamps_a_real_utc_instant` asserts the stored value lands inside a
  window taken around the call, so a frozen stamp or one taken off the local clock fails.
* `test_each_progress_write_advances_the_stamp` pins that repeated ticks move it forward.
  A value that is merely non-NULL says nothing.
* `test_the_insert_fallback_stamps_the_row_it_creates` covers the branch that inserts a row
  when none exists.
* The same tests assert `log_pipeline_run_start` leaves the column NULL: starting is not
  evidence of progress.
* All four are mutation checked. Freezing the stamp reddens all three; deleting only the
  UPDATE stamp reddens two; deleting only the INSERT stamp reddens exactly one. That last
  one is the point, it shows the fallback branch has real coverage and not incidental
  coverage.
* Migration tests follow the existing per-migration convention: idempotent upgrade, the
  create_all case where the column is already present, downgrade, and a no-op against a
  database with no `pipeline_runs` table.

## Type of Change
- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots

    $ pytest cognee/tests/unit/modules/pipelines/ \
        cognee/tests/unit/test_add_pipeline_runs_last_heartbeat_at_migration.py \
        cognee/tests/unit/test_frozen_schema_seal.py \
        cognee/tests/e2e/migrations/test_migration_model_lockstep.py

    97 passed, 98 warnings in 15.95s

Frozen schema seal and the migration/model lockstep guard are in that run. `alembic heads`
reports a single head, `b2d6f4a8c150`, on parent `a7c2e9f4b8d1`.

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/modules/pipelines/models/PipelineRun.py
+++ b/cognee/modules/pipelines/models/PipelineRun.py
@@ -59,6 +59,32 @@ class PipelineRun(Base):
     dataset_id = Column(UUID, index=True)
     run_info = Column(JSON)
 
+    # When this run last showed a sign of life. Written by
+    # log_pipeline_run_progress, which already UPDATEs the STARTED row in
+    # place, so recording it costs no extra write and no extra row.
+    #
+    # Its resolution is one item completion, throttled in run_tasks.py to ~20
+    # per run. So it advances for a multi-item run and does NOT advance while
+    # a run is inside a single long item: a one-item cognify stamps exactly
+    # once, when that item finishes, and nothing before it — the run's start
+    # is only created_at, since log_pipeline_run_start records no liveness.
+    # Treating a frozen value as proof of death is therefore wrong for that
+    # shape; bounding the gap inside one item needs an event source inside
+    # the task, which this column does not provide.
+    #
+    # Nullable with no backfill: rows written before this column existed stay
+    # NULL, and readers fall back to created_at for them. NULL therefore means
+    # "no evidence either way, judge by age", never "dead" — several paths
+    # legitimately hold a STARTED row without ever ticking (an update() run
+    # driven outside run_tasks, a dataset still queued behind others in a
+    # background multi-dataset run).
+    #
+    # Readers must normalize before comparing: SQLite drops the tzinfo on a
+    # DateTime(timezone=True) round trip and returns naive UTC, while Postgres
+    # returns it aware, so `value < datetime.now(timezone.utc)` raises on the
+    # default backend. recovery.py does this for created_at already.
+    last_heartbeat_at = Column(DateTime(timezone=True))
+
     # Operation-record columns (SDK-399). All nullable: rows written before
     # this change stay NULL — no backfill. Non-pipeline operations (search,
     # recall, forget, remember, delete, prune) write exactly one row with
--- a/cognee/alembic/versions/b2d6f4a8c150_add_pipeline_runs_last_heartbeat_at.py
+++ b/cognee/alembic/versions/b2d6f4a8c150_add_pipeline_runs_last_heartbeat_at.py
@@ -1,0 +1,67 @@
+"""Add last_heartbeat_at to pipeline_runs so staleness stops being a guess
+
+Revision ID: b2d6f4a8c150
+Revises: a7c2e9f4b8d1
+Create Date: 2026-09-10
+
+PipelineRun gained a nullable ``last_heartbeat_at`` column
+(cognee/modules/pipelines/models/PipelineRun.py) recording when a run last
+showed a sign of life, so readers can tell a run that is still working from
+one whose process died instead of guessing from the row's age.
+
+Deliberately no backfill. Existing rows keep NULL, and readers fall back to
+``created_at`` for them, which is exactly today's behaviour — so this
+migration changes nothing on its own. Backfilling ``created_at`` into the
+column would be indistinguishable from a genuine tick and would claim
+evidence the run never produced.
+
+No index: the column is only ever read alongside a row already selected by
+the existing dataset_id/pipeline_name/created_at index, and this row is
+updated repeatedly while a run is in flight, where a second index would cost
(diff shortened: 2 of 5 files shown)


cognee/modules/pipelines/models/PipelineRun.py
```diff
@@ -59,6 +59,32 @@ class PipelineRun(Base):
...
+    # Treating a frozen value as proof of death is therefore wrong for that
+    # shape; bounding the gap inside one item needs an event source inside
+    # the task, which this column does not provide.
+    #
+    # Nullable with no backfill: rows written before this column existed stay
+    # NULL, and readers fall back to created_at for them. NULL therefore means
+    # "no evidence either way, judge by age", never "dead" — several paths
+    # legitimately hold a STARTED row without ever ticking (an update() run
+    # driven outside run_tasks, a dataset still queued behind others in a
+    # background multi-dataset run).
```
NMZivkovic, line 80: **[Suggestion] For a queued dataset the `created_at` fallback gives the wrong verdict — stamp once when execution begins**

This names "a dataset still queued behind others in a background multi-dataset run" as a legitimate NULL, to be judged by `created_at`. For that case `created_at` is the wrong clock. `run_pipeline_as_background_process` calls `anext()` on every dataset's pipeline up front, which runs each `run_tasks` to `yield PipelineRunStarted` and so writes every STARTED row immediately; `handle_rest_of_the_run` then executes the pipelines one at a time. A queued dataset's `created_at` is when the batch was queued, so judging it by age marks it stale while it waits its turn — #4983's `ABANDONED` on `/v1/datasets/status` already does exactly that.

Suggestion: stamp `last_heartbeat_at` once when the run actually starts executing — the first thing `run_tasks` does after resuming from `yield PipelineRunStarted`, just before `set_database_global_context_variables`. That is one UPDATE per run next to the ~20 progress ticks, and it changes what NULL means from "no evidence, judge by age" to "has not started executing", which a reader can treat as queued rather than stale.

It does not cut against the "starting is not evidence of progress" reasoning in the tests: that is about `log_pipeline_run_start`, which fires at *enqueue* time. A stamp at execution start is true when it is written, and a run that dies right after it still ages out against the threshold like any other. As a bonus, a one-item cognify gets a tick when it begins instead of only when it finishes.

NMZivkovic (commented):
Nice — small, and scoped honestly: a timestamp on a write that already happens, NULL left meaning NULL, and no reader until the value has been visible for a release. The note in the description about `PipelineRunYield` firing once per document is exactly the kind of dead end worth writing down.

Checked: the migration sits on dev's only head (`a7c2e9f4b8d1`), so the chain stays single-headed; the per-migration test file follows the existing convention (four `test_add_*_migration.py` files already sit next to it); the fallback INSERT branch has its own test; CI is green.

One design point before the reader lands, inline on the column comment: for the queued case the comment itself names, falling back to `created_at` gives the wrong answer, and one extra stamp at execution start would fix it.

NMZivkovic (approved):

PR #5022, opened 2026-09-10
fix(mcp): send the requested pipeline name to the status endpoint

### LINKED ISSUE #5012: [Bug]: cognify_status is unreliable in API mode
### Bug Description

`cognee-mcp/src/server.py:1537-1549` (cognify_status) passes the requested pipeline name to `CogneeClient.get_pipeline_status()`, but the API-mode branch of the method (cognee-mcp/src/cognee_client.py:477-485) only sends ``dataset=<id>`` to ``GET /api/v1/datasets/status`` and never the pipeline name. The server endpoint (`cognee/api/v1/datasets/routers/get_datasets_router.py: 448-451`, `StatusPipelineNamesQuery alias="pipeline"`) defaults to `cognify_pipeline` the param is omitted. So ``cognify_status(pipelines=["add_pipeline"])`` or ``code_graph_pipeline"]`` returns cognify_pipeline's status, and the multi-pipeline branch labels that same value under every requested pipeline name.

### Steps to Reproduce

```python
import sys
import pathlib

# Resolve the cognee-mcp package root relative to this file (cognee-mcp/tests/).
MCP_ROOT = pathlib.Path(__file__).resolve().parent.parent
sys.path.insert(0, str(MCP_ROOT))

import asyncio  # noqa: E402

import httpx  # noqa: E402
import pytest  # noqa: E402

import src.server as server  # noqa: E402
from src.cognee_client import CogneeClient  # noqa: E402

DATASET_ID = "11111111-1111-1111-1111-111111111111"


def _api_client(calls):
    def handler(request):
        calls.append(request)
        if request.url.path == "/api/v1/datasets/":
            return httpx.Response(200, json=[{"id": DATASET_ID, "name": "main_dataset"}])
        if request.url.path == "/api/v1/datasets/status":
            return httpx.Response(200, json={DATASET_ID: "DATASET_PROCESSING_COMPLETED"})
        return httpx.Response(404, json={})

    client = CogneeClient(api_url="http://cognee.test")
    client.client = httpx.AsyncClient(
        transport=httpx.MockTransport(handler), base_url="http://cognee.test"
    )
    return client


def _status_requests(calls):
    return [r for r in calls if r.url.path == "/api/v1/datasets/status"]


@pytest.mark.asyncio
async def test_single_pipeline_name_is_sent_to_status_endpoint(monkeypatch):
    calls = []
    monkeypatch.setattr(server, "cognee_client", _api_client(calls))
    monkeypatch.setattr(server, "_agent_scoped_default_dataset", lambda: "main_dataset")

    await server.cognify_status(pipelines=["add_pipeline"])

    status_calls = _status_requests(calls)
    assert len(status_calls) == 1
    params = status_calls[0].url.params
    assert params.get("dataset") == DATASET_ID
    assert params.get_list("pipeline") == ["add_pipeline"], (
        f"pipeline name was dropped; query was {status_calls[0].url.query!r}"
    )


@pytest.mark.asyncio
async def test_multi_pipeline_query_asks_for_each_requested_pipeline(monkeypatch):
    calls = []
    monkeypatch.setattr(server, "cognee_client", _api_client(calls))
    monkeypatch.setattr(server, "_agent_scoped_default_dataset", lambda: "main_dataset")

    await server.cognify_status(pipelines=["add_pipeline", "code_graph_pipeline"])

    requested = set()
    for req in _status_requests(calls):
        requested.update(req.url.params.get_list("pipeline"))
    assert requested == {"add_pipeline", "code_graph_pipeline"}, (
        f"server was only ever asked for the default pipeline; pipelines sent = {requested}"
    )
```

### Expected Behavior

Expected: each requested pipeline name reaches the server as a ``pipeline`` query parameter (the endpoint's documented contract), so the returned status really is the requested pipeline's.                                                                                                                                                                 

### Actual Behavior

should pass the pipeline name.

### Environment

- OS: Ubuntu 26.04 LTS (kernel 7.0.0-29-generic)
- Python version: 3.12.14
- Cognee version: 1.5.4 (main @ e93a4f0), cognee-mcp 0.5.5, fastmcp 3.4.7, mcp 1.29.1, httpx 0.28.1
- LLM Provider: not involved (no LLM call on this path)
- Database: not involved (reproduced with an in-process mock of the HTTP API; the request is captured before any DB)

### Logs/Error Messages

```shell
-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
============================================================================================= short test summary info ==============================================================================================
FAILED ../claude-bugs/tests/test_memory_tools_status_pipeline_dropped_in_api_mode.py::test_single_pipeline_name_is_sent_to_status_endpoint - AssertionError: pipeline name was dropped; query was b'dataset=11111111-1111-1111-1111-111111111111'
FAILED ../claude-bugs/tests/test_memory_tools_status_pipeline_dropped_in_api_mode.py::test_multi_pipeline_query_asks_for_each_requested_pipeline - AssertionError: server was only ever asked for the default pipeline; pipelines sent = set()
========================================================================================== 2 failed, 11 warnings in 5.35s ==========================================================================================
```

### Additional Context

_No response_

### Pre-submission Checklist

linhongyu510:
Fixes #5012.

## The bug

`CogneeClient.get_pipeline_status()` accepts a `pipeline_name`, but the
API-mode branch built its query as only `dataset=<id>` and never sent the name:

```python
params = [("dataset", str(d)) for d in dataset_ids]
```

`GET /api/v1/datasets/status` documents `pipeline` as optional and **defaults to
`cognify_pipeline` when it is omitted**, so the request does not fail. It
returns a well-formed status — just for the wrong pipeline.

The direct-mode branch of the same method passes `pipeline_name` through
correctly, so the two modes disagree.

## Impact

`cognify_status(pipelines=["add_pipeline"])` returns `cognify_pipeline`'s
status. The multi-pipeline branch in `server.py` calls this once per requested
name and labels each result with the name it asked for, so a call for
`["add_pipeline", "code_graph_pipeline"]` reports one pipeline's status under
both names. Nothing raises, and the value looks plausible.

## Reproduction on `c0d18c80e`

```python
client = CogneeClient(api_url="http://cognee.test")
client.client = httpx.AsyncClient(transport=httpx.MockTransport(handler))
await client.get_pipeline_status([dataset_id], "add_pipeline")
```

Before: `dataset=11111111-1111-1111-1111-111111111111`
After: `dataset=11111111-1111-1111-1111-111111111111&pipeline=add_pipeline`

## The change

Append `pipeline` when a name is given. Two behaviors are deliberately left
alone:

- an empty name still omits the parameter, so the server keeps applying its own
  documented default and existing callers that pass nothing are unaffected;
- the direct-mode branch is untouched — it was already correct.

## Tests

Three tests in `cognee-mcp/tests/test_mcp_server_hardening.py`, reusing the
existing `_mock_api_client` helper. They assert the **query string** rather than
the returned value, because the failure mode here is a well-formed response for
the wrong pipeline — asserting on the response would pass against the bug.

- the requested name reaches the server
- two different names produce two different queries, so the multi-pipeline
  branch cannot relabel one status under every name
- an empty name still omits the parameter

Reverting `cognee_client.py` alone turns the first two red; the third passes
either way by design, since it pins the back-compat path.

## Verification

- `uv run python -m pytest tests/` in `cognee-mcp/` — **83 passed**
- The suite's own tool-coverage reporter now says *"all 4 registered tools
  exercised"*; `cognify_status` was previously reported as having no coverage.
- `ruff check` on both changed files — 68 findings, identical to the count on
  `main` before this change (all pre-existing `import src.server` style
  findings elsewhere in the file); **0 new**.
- `ruff format --check` — both files already formatted.


Vasilije1990 (dismissed):

Vasilije1990:
@linhongyu510 please rebase to dev

NMZivkovic:
Opened #5061 as a clean replacement for this fix, based on current `dev`. It carries the two-file change and all three regression tests, with attribution to the original contributor.

Local MCP suite: 84 passed. Repository-wide Ruff 0.16.6 lint, formatting, strict exception rules, and lockfile validation pass. The replacement now runs the full Internal Test Suites from the main repository, so its jobs can access the required integration-test secrets.

This PR's community Code Quality failure came from the broader main-based diff; its DCO check also lacked the required PR-body statement. Several internal jobs failed before running tests because fork runs do not receive the required secrets. The replacement addresses those setup issues and avoids the merge conflicts. Tracking: SDK-679.

linhongyu510:
@NMZivkovic Thanks — closing this, #5061 is merged (2026-09-17) and I confirmed it carries the fix rather than assuming it did.

On `dev`, the merged commit is the same two-file change plus the regression tests:

```
cognee-mcp/src/cognee_client.py                  7 +-
cognee-mcp/tests/test_mcp_server_hardening.py   66 ++++++++++
```

So the behaviour this PR was for — the MCP client sending the requested pipeline name to the status endpoint instead of the default — is shipped, and the regressions that lock it are in the tree. Thanks for carrying the attribution across.

Your diagnosis of why this one could not land as-is was also correct and worth recording for anyone who hits the same wall: the Code Quality failure came from the broader `main`-based diff rather than from the two-file change, the DCO check needed a statement in the PR body, and several internal jobs failed before running anything because fork runs do not receive the integration-test secrets. That last one is not something a contributor can fix from their side, which is exactly why a maintainer-owned replacement was the right call here.

Nothing outstanding on my end.

PR #5024, opened 2026-09-11
fix(retrieval): anchor the temporal prompt's "current date" to UTC

linhongyu510:
## The bug

`TemporalRetriever.extract_time_from_query` built the prompt's current-date
anchor from a naive `datetime.now()` — the **server's local date**:

```python
time_now = datetime.now().strftime("%d-%m-%Y")
```

Event timestamps, though, are written in UTC:

- `cognee/tasks/temporal_awareness/build_graph_with_temporal_awareness.py:38` — `reference_time=datetime.now(timezone.utc)`
- `cognee/modules/graph/utils/get_graph_from_model.py:63` — `datetime.now(timezone.utc)`

So the date the LLM resolves "now" / "today" / "current" against disagreed with
the graph it was resolving them for.

`extract_query_time.txt` makes that anchor load-bearing — rule 1 sends
unqualified queries to "now", and rule 5 pins `"now"/"current"/"today"` to both
interval ends. The resulting interval goes straight to `collect_time_ids`, so a
one-day skew silently returns events from the wrong calendar day. No error, no
warning, just a plausible answer about a different day.

## Why it is not an edge case

Local and UTC dates differ for part of **every** day in any non-UTC zone, in
both directions:

| TZ | UTC instant | graph (UTC) | prompt (naive) | |
|---|---|---|---|---|
| `Asia/Tokyo` (UTC+9) | 23:30 | 11-09-2026 | 12-09-2026 | a day **ahead** |
| `Asia/Shanghai` (UTC+8) | 22:10 | 11-09-2026 | 12-09-2026 | a day **ahead** |
| `America/Los_Angeles` (UTC-7) | 02:30 | 11-09-2026 | 10-09-2026 | a day **behind** |
| `UTC` | any | 11-09-2026 | 11-09-2026 | agrees |

A "what happened today" query lands on the wrong day for hours at a time,
depending only on where the server happens to run.

## The change

One line: `datetime.now(timezone.utc)`, so both sides of the comparison share a
clock.

This is also the established convention in this repo — 166 UTC-aware call sites
against 12 naive ones. I checked the other 11: they all format log lines,
filenames or console output, where local time is intentional. This was the only
naive site feeding a semantic comparison.

## Tests

Two tests in `cognee/tests/unit/modules/retrieval/temporal_retriever_test.py`.
They assert against a **real `TZ`** rather than a mocked clock, because a mocked
`datetime` cannot show the drift this pins — the existing tests in this file
mock `datetime` and so pass either way.

- `test_extract_time_from_query_anchors_now_to_utc`, parametrized across
  `UTC`, `Asia/Tokyo`, `Asia/Shanghai`, `America/Los_Angeles`,
  `Pacific/Kiritimati` (UTC+14) and `Pacific/Midway` (UTC-11)
- `test_extract_time_from_query_now_is_timezone_aware` — parses the anchor and
  asserts it equals UTC's date, not merely that it is well-formed

Two details worth flagging, both found while verifying rather than assumed:

**The zone spread is deliberate.** Whether a zone's local date differs from UTC
depends on the hour the suite runs. My first list (UTC through UTC+14, plus
`America/Los_Angeles`) had a three-hour window — 07:00–09:00 UTC — where *no*
zone in the set disagreed with UTC, so the tests would have passed with the bug
still in place. Adding `Pacific/Midway` (UTC-11) closes it; I verified every
half-hour across 24 hours that at least one zone now disagrees.

**`tzset` is handled carefully.** The test is skipped where `time.tzset` is
unavailable, since CI runs `windows-latest` and `tzset` is POSIX-only. And
because `tzset()` mutates process-global state that `monkeypatch`'s env
restoration does not undo, it is re-applied via `request.addfinalizer` so later
tests in the same shard do not inherit the zone — this file's own module
docstring documents a past cross-test pollution incident, so that seemed worth
being careful about.

Reverting `temporal_retriever.py` alone turns the zone cases red (2 of 6 at the
hour I ran it; which ones depends on the clock, by design).

## Verification

- `pytest cognee/tests/unit/modules/retrieval/temporal_retriever_test.py` — **29 passed**
- `pytest cognee/tests/unit/modules/retrieval/` — **552 passed, 1 skipped**
- `pytest cognee/tests/unit/` — **5752 passed, 75 skipped, 2 failed**. Both
  failures (`test_init_signal_killed_worker_surfaces_signal_name`,
  `test_directory_with_project_resolves_to_repo_item_plus_documents`) reproduce
  identically on `main` with this change stashed, and neither touches temporal
  retrieval. Three further files fail to collect locally for missing optional
  deps (`neo4j`, `plotly`) and were excluded.
- `ruff check` on both changed files — 20 findings, the same count as `main`
  before this change; **0 new**.
- `ruff format --check` — both files already formatted.

## AI disclosure

This pull request was prepared with AI assistance. The change was reviewed,
tested, and verified locally by the contributor.


Vasilije1990:
@linhongyu510 please don't open PRs against main, it won't get merged

linhongyu510:
@Vasilije1990 The base has been `dev` since before your note, so I went looking for what actually triggered it — the PR was 472 commits behind and showing `DIRTY`, which reads the same way from the PR page. That is fixed: rebased onto current `dev` (`45208c50e`), now `MERGEABLE`, new head `e1bc356`.

Rebasing turned up something worth reporting rather than quietly resolving: **the source fix is already upstream.** `4367ed753` ("Use UTC for the display timestamps as well", SDK-583) changed the same line to `datetime.now(timezone.utc)` independently. So the production change this PR proposed is redundant, and I dropped it rather than re-litigate it in a conflict resolution.

What `4367ed753` did *not* add is a regression test, so that is all this branch carries now — **1 file, `+97/-1`, no source change.**

The coverage is deliberately shaped around the way this bug hides. It runs against a real `TZ` via `tzset()` rather than a mocked clock, because a mocked `datetime` cannot show local-vs-UTC drift at all. The zone list spans UTC-11 to UTC+14 (`Pacific/Midway` … `Pacific/Kiritimati`) because whether any single zone's local date differs from UTC depends on what hour the suite happens to run — a narrower list passes silently for part of every day. With this spread at least one zone disagrees with UTC at every instant.

Verified it is actually protective rather than merely passing — reverting the line to naive `datetime.now()`:

```
TZ=Asia/Shanghai  →  4 failed, 3 passed
```

With the current `dev` source: `29 passed` for the whole file. `ruff check` and `ruff format --check` clean.

One incidental catch: `dev`'s DTZ ruleset (arriving with the same SDK-583 work) flagged my own `strptime` as naive. Fixed properly with an explicit `tzinfo=timezone.utc` rather than a noqa, since a naive parse would reintroduce the exact ambiguity the test exists to pin.

Happy to close this if you would rather not carry test-only PRs — but since the fix shipped without a guard, this is the piece that keeps `4367ed753` from silently regressing.

linhongyu510:
One note on the red checks, in case it factors into whether this is worth carrying:

The failures are all in the Integration / E2E suites, and they fail before any assertion — the embedding engine is invoked with an empty model and `api_key=None`:

```
litellm.exceptions.BadRequestError: LLM Provider NOT provided. You passed model=
→ LiteLLMEmbeddingEngine.embed_text retries 6× and gives up
→ RuntimeError: Graph edge indexing error  (index_graph_edges.py:90)
```

That is the LLM/embedding secrets not being present on a fork PR run, not anything this branch touches — the change here is a single test file, `+97/-1`, no source. The checks that don't need those secrets are green (Integration – Web, Infrastructure 1/3, Community Tests Status), and current `dev` itself is green because it runs with the secrets available.

Nothing to fix on my side here — it should go green on a maintainer re-run in a trusted context. Flagging it only so the red isn't read as a code failure.

PR #5025, opened 2026-09-11
COG-6477: Fix budget exhaustion escaping as InstructorRetryException

goran-radonic:
## Description

Budget-exhaustion errors from a LiteLLM proxy were reaching cognee wrapped in
`InstructorRetryException` instead of surfacing as our typed
`LLMPaymentRequiredError` (402). Seven call sites across five instructor
adapters (openai, azure_openai, bedrock, gemini, generic_llm_api) had a bare
`raise` ahead of the classification that converts the error, so any caller
not going through `LLMGateway` had to string-match the message itself to
tell a budget rejection apart from any other instructor failure.

While reviewing the fix I found gemini had a second, sharper bug: it
classified budget exhaustion at the very top of its shared except clause,
before checking whether a fallback model was configured. That meant a
rejection that was both policy-worded and budget-worded (the model's partial
completion gets rendered into the exception message) got converted and
raised immediately, skipping the fallback attempt entirely - even though a
differently-keyed fallback could plausibly have succeeded. Moved the
classification to each actual exit point of that method instead, matching
how openai and azure_openai already handle it, and added a test pinning
every exit point individually (including the no-fallback branch, which
stayed unpinned after my first pass and only turned up under mutation
testing - deliberately reverting the fix and confirming the test goes red).

Also brought mistral and ollama in line with the other five adapters: they
were converting budget errors with a bare `LLMPaymentRequiredError()`
instead of `raise_if_budget_exhausted()`, so their 402s lost the provider's
detail message that every other adapter now carries. And added test
coverage for azure's managed-identity fallback-also-capped path, which had
none before this.

Separately, the transcription/image-transcription decorators on six of
these adapters used a private retry-exclusion tuple that never checked for
budget/quota errors, so a rejection that could never succeed still burned a
few retries. Switched them to the shared `llm_retry_condition` used
everywhere else in the LLM layer.

Linear: COG-6477

## Acceptance Criteria

- A budget-exhausted call through any of the seven instructor adapters
  raises `LLMPaymentRequiredError`, never a raw `InstructorRetryException`,
  regardless of whether a fallback is configured.
- Gemini still attempts a configured fallback before converting, even when
  the primary rejection is both policy-worded and budget-worded.
- Mistral and ollama's 402s carry the provider's own budget message, not a
  generic one.
- Transcription/image decorators on the affected adapters do not retry a
  budget-exhausted call.
- All of the above covered by tests that fail against the pre-fix code
  (verified by hand via `git stash` on the production files, keeping the
  tests fixed, and confirming the specific tests go red).

## Type of Change
- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation

### DIFF
--- a/cognee/infrastructure/llm/structured_output_framework/litellm_instructor/llm/anthropic/adapter.py
+++ b/cognee/infrastructure/llm/structured_output_framework/litellm_instructor/llm/anthropic/adapter.py
@@ -12,7 +12,7 @@ from tenacity import (
 )
 
 from cognee.infrastructure.llm.config import get_llm_config
-from cognee.infrastructure.llm.exceptions import LLMPaymentRequiredError, is_budget_exhausted_error
+from cognee.infrastructure.llm.exceptions import raise_if_budget_exhausted
 from cognee.infrastructure.llm.retry_config import (
     llm_retry_condition,
     llm_retry_stop_condition,
@@ -122,6 +122,6 @@ class AnthropicAdapter(GenericAPIAdapter):
                     **merged_kwargs,
                 )
         except Exception as e:
-            if is_budget_exhausted_error(e):
-                raise LLMPaymentRequiredError() from e
+            # Same detail-carrying message as the other adapters.
+            raise_if_budget_exhausted(e)
             raise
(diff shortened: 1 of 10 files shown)


dexters1 (approved):

dexters1:
LLM tests need to be removed from unit test folder and into integrations or release_test, but this will be handled in a future PR

PR #5027, opened 2026-09-11
fix(neo4j): bind temporal ids as a query parameter

linhongyu510:
## The bug

`Neo4jAdapter.collect_events` declares `ids: List[str]`, but builds its query by
interpolating that value into the Cypher text:

```python
event_collection_cypher = """UNWIND [{quoted}] AS uid
    MATCH (start {{id: uid}})
    ...
"""
query = event_collection_cypher.format(quoted=ids)
return await self.query(query)
```

`.format()` only produces valid Cypher here if `ids` is already a pre-quoted,
comma-joined string. Passing the `List[str]` the signature promises renders the
list's Python repr:

```
UNWIND [['timestamp-1', 'timestamp-2']] AS uid
```

That is a list containing one list. `uid` binds to the whole list instead of to
each id, so `MATCH (start {id: uid})` matches nothing. **Nothing raises** — the
temporal retrieval path just returns zero events, and
`TemporalRetriever.get_context` reports "No events identified".

## Reproduced on `main` (`c0d18c80e`)

```python
adapter = Neo4jAdapter.__new__(Neo4jAdapter)
adapter.query = AsyncMock(return_value=[])

await adapter.collect_events(ids=["timestamp-1", "timestamp-2"])
# -> UNWIND [['timestamp-1', 'timestamp-2']] AS uid     (nested; matches nothing)

await adapter.collect_events(ids="'timestamp-1', 'timestamp-2'")
# -> UNWIND ['timestamp-1', 'timestamp-2'] AS uid       (the only shape that works)
```

An empty list — what `collect_time_ids` returns from its `else` branch — yields
`UNWIND [[]] AS uid`.

## The two halves disagreed on the shape they exchange

`collect_time_ids` was annotated `-> str` and returned
`", ".join(f"'{uid}'" for uid in ...)`, while its `else` branch returned `ids`,
an empty **list**. So the declared type was wrong for both paths.

The ladybug adapter had already moved to the other convention:
`collect_time_ids` returns `List[str]`, and `collect_events` accepts
`Union[List[str], str]`, normalizes via `_normalize_temporal_ids`, and binds
`UNWIND $ids`. Its tests pin exactly this:

- `test_collect_time_ids_returns_list_for_unwind_params`
- `test_collect_events_binds_ids_as_list`
- `test_collect_events_accepts_legacy_quoted_id_string`

So the ambiguity in this interface was already known and already handled — on one
adapter. The neo4j side had neither the normalization nor any test.

`temporal_retriever.py:145` calls `collect_events(ids=ids)` with the value from
`collect_time_ids` (`:132`/`:134`/`:136`), so both adapters serve the same
call site through the same shared shape.

## The change

Align neo4j with ladybug:

- `collect_time_ids` returns `List[str]`, and its docstring no longer advertises
  the quoted-string format.
- `collect_events` accepts `Union[List[str], str]` and binds the ids as a query
  parameter (`UNWIND $ids`) after normalizing them.

The legacy pre-quoted string keeps working, so any caller outside this repo
passing the old shape is unaffected. Binding also means ids are no longer
interpolated into query text.

## Tests

New `cognee/tests/unit/infrastructure/databases/graph/test_neo4j_temporal.py`,
mirroring the ladybug file:

- `test_collect_time_ids_returns_list_for_unwind_params`
- `test_collect_events_binds_ids_as_list` — asserts on the bound params, and that
  the nested-list rendering is absent
- `test_collect_events_accepts_legacy_quoted_id_string`
- `test_collect_time_ids_output_feeds_collect_events` — runs the composition
  `collect_time_ids -> collect_events`, which is the path `temporal_retriever`
  takes and where the mismatch lived

The tests assert on the bound parameters rather than on the return value: with
the bug present, the query is well-formed Cypher that simply matches nothing, so
asserting on results would pass either way.

Reverting `adapter.py` alone turns all 4 red. The 3 existing ladybug tests pass
unchanged.

## Verification

- `pytest .../test_neo4j_temporal.py .../test_ladybug_temporal.py` — **7 passed**
- `pytest cognee/tests/unit/infrastructure/databases/graph/` — **126 passed**
- `pytest cognee/tests/unit/modules/retrieval/` — **545 passed, 1 skipped**
- `ruff check` / `ruff format --check` at v0.15.11 (the version pinned in
  `.pre-commit-config.yaml`) — clean on both files

`neo4j_deadlock_test.py` was excluded from the graph run: it needs a live Neo4j,
which is not available here.

I installed the `neo4j` package locally to exercise `Neo4jAdapter` in unit
tests; no dependency changes are included in this PR.

### DIFF
--- a/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
+++ b/cognee/infrastructure/databases/graph/neo4j_driver/adapter.py
@@ -8,7 +8,7 @@ from neo4j import AsyncSession
 from neo4j import AsyncGraphDatabase
 from neo4j.exceptions import Neo4jError
 from contextlib import asynccontextmanager, nullcontext
-from typing import Optional, Any, List, Dict, Type, Tuple, Coroutine, Set
+from typing import Optional, Any, List, Dict, Type, Tuple, Coroutine, Set, Union
 from cognee.modules.observability import OtelStatusCode as StatusCode
 from cognee.infrastructure.engine import DataPoint
 from cognee.modules.engine.utils.generate_timestamp_datapoint import date_to_int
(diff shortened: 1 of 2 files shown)


linhongyu510:
Rebased onto `dev` as requested on #5022.

The branch was rebuilt from current `dev` rather than rebased, so the PR is now a single commit with no unrelated main-based history. The fix was re-applied on top of the refactored adapter (modern `list[str] | str` annotations instead of `Union`, no new imports needed), and all numbers were re-measured on the new base:

- `test_neo4j_temporal.py` + `test_ladybug_temporal.py` — 7 passed (my 4, ladybug's 3)
- `cognee/tests/unit/infrastructure/databases/graph/` — 175 passed
- Ruff 0.16.6 (the version now pinned in `.pre-commit-config.yaml`) — check and format clean

Reverse-verified on `dev`: reverting the adapter alone turns all 4 new tests red.

The defect is unchanged on `dev` — `collect_events` still calls `.format(quoted=ids)` at adapter.py:2464 while declaring `ids: list[str]`.

PR #5032, opened 2026-09-11
fix(forget): raise typed 404 for unknown dataset name instead of NoneType 500

### LINKED ISSUE #5013: fix(forget): Raise 404 for unknown dataset name
## Problem

`cognee.forget(dataset="<name>")` for a name the caller does not own crashes with
`AttributeError: 'NoneType' object has no attribute 'id'` instead of a typed
not-found error. Over HTTP the forget router's catch-all turns that crash, and
every other typed cognee error (e.g. the 403 `PermissionDeniedError` raised for an
unknown `dataset_id`), into `500 {"error": "An error occurred during deletion."}`.
CLI and MCP inherit the raw `'NoneType' object has no attribute 'id'` text.

Verified on `dev` @ `d3909dd26` (v1.5.4) and `main` @ `c0d18c80e` (the two differ only in
import order / `X | None` typing in the affected files), SQLite/LanceDB/Ladybug defaults,
no LLM call involved. The same crash hits a user with zero datasets for any `dataset` name,
since the name lookup scans an empty permitted-dataset list and returns `None`.

## Current vs expected

| Surface | Input | Current | Expected |
|---|---|---|---|
| SDK | `forget(dataset="does-not-exist")` | `AttributeError: 'NoneType' object has no attribute 'id'` | `DatasetNotFoundError` (or at least `ValueError`, like the `dataset_id` branch intends) |
| SDK | `forget(dataset_id=<unknown uuid>)` | `PermissionDeniedError` (403) | same (no-leak semantics are fine) |
| HTTP | `POST /v1/forget {"dataset": "does-not-exist"}` | `500 An error occurred during deletion.` | `404` |
| HTTP | `POST /v1/forget {"datasetId": "<unknown uuid>"}` | `500 An error occurred during deletion.` | `403` (the error already carries `status_code=403`) |
| CLI | `cognee-cli forget --dataset does-not-exist` | `Error during forget: Failed to forget: 'NoneType' object has no attribute 'id'` | `Failed to forget: DatasetNotFoundError: Dataset 'does-not-exist' not found or not accessible. (Status code: 404)` |
| MCP | `forget(dataset="does-not-exist")` | `Error: Forget failed: 'NoneType' object has no attribute 'id'` | `Error: Forget failed: DatasetNotFoundError: Dataset 'does-not-exist' not found or not accessible. (Status code: 404)` |

CLI/MCP expected text follows `str(CogneeApiError)`, which renders as `<Name>: <message> (Status code: N)`.

## Evidence

Repro script (SDK + in-process FastAPI `TestClient`, auth dependency overridden with the default user):

```python
import asyncio, os, uuid
os.environ.update({"TELEMETRY_DISABLED": "1", "LLM_API_KEY": "dummy",
                   "ENABLE_BACKEND_ACCESS_CONTROL": "false"})
import cognee

async def sdk():
    for kwargs in ({"dataset": "does-not-exist"}, {"dataset_id": uuid.uuid4()}):
        try:
            print("SDK", kwargs, "->", await cognee.forget(**kwargs))
        except Exception as e:
            print("SDK", kwargs, "->", type(e).__name__, e)
asyncio.run(sdk())

from fastapi.testclient import TestClient
from cognee.api.client import app
from cognee.modules.users.methods import get_authenticated_user, get_default_user
app.dependency_overrides[get_authenticated_user] = get_default_user
with TestClient(app) as c:
    for body in ({"dataset": "does-not-exist"}, {"datasetId": str(uuid.uuid4())},
                 {"dataset": "does-not-exist", "dataId": str(uuid.uuid4())}):
        r = c.post("/api/v1/forget", json=body)
        print("HTTP", body, "->", r.status_code, r.text)
```

Output:

```
SDK {'dataset': 'does-not-exist'} -> AttributeError 'NoneType' object has no attribute 'id'
SDK {'dataset_id': UUID('b294aa8f-...')} -> PermissionDeniedError Request owner does not have necessary permission: [delete] for all datasets requested. (Status code: 403)
HTTP {'dataset': 'does-not-exist'} -> 500 {"error":"An error occurred during deletion."}
HTTP {'datasetId': '0a8dfe3b-...'} -> 500 {"error":"An error occurred during deletion."}
HTTP {'dataset': 'does-not-exist', 'dataId': 'cc6ad940-...'} -> 500 {"error":"An error occurred during deletion."}
```

Server log for the first HTTP call: `Forget endpoint error ... exception_type=AttributeError`.

## Root cause

1. `cognee/api/v1/forget/forget.py:477-490` (`_resolve_dataset_id`): the UUID branch guards
   `if not dataset: raise ValueError(...)`, the name branch does not:

   ```python
   dataset = await get_authorized_dataset_by_name(dataset_ref, user, "delete")
   return dataset.id
   ```

   `get_authorized_dataset_by_name` (`cognee/modules/data/methods/get_authorized_dataset_by_name.py:9-25`)
   is typed `-> Dataset | None` and returns `next(..., None)`. `forget.py` is its only
   production caller. The UUID-branch guard is dead code: `get_authorized_dataset` goes
   through `get_specific_user_permission_datasets`, which raises `PermissionDeniedError`
   before `None` can be returned. The resolution runs before any dataset lock or database
   context is entered, so the crash provisions nothing.

2. `cognee/api/v1/forget/routers/get_forget_router.py:109-123`: the router catches
   `ValueError` (422) and then `Exception` (500). Because the catch-all runs before the
   app-level `CogneeApiError` handler (`cognee/api/client.py:219-235`), typed cognee
   errors that already carry a status code (`PermissionDeniedError` 403,
   `DatasetNotFoundError` 404) are downgraded to 500. The cognify router already handles
   this correctly (`get_cognify_router.py:406`, `except DatasetNotFoundError`).

## Proposed direction

- `_resolve_dataset_id`: raise `DatasetNotFoundError(f"Dataset '{dataset_ref}' not found or not accessible.")`
  (from `cognee.modules.data.exceptions`, `status_code=404`) when the name lookup returns `None`.
  The dead `if not dataset` guard in the UUID branch can be aligned to the same exception.
- Forget router: re-raise `CogneeApiError` ahead of the `except Exception` catch-all so
  the existing app-level handler maps 403/404 (smallest diff), or map
  `DatasetNotFoundError -> 404` explicitly like the cognify router. Update the docstring's
  "Error Codes" list accordingly.
- CLI and MCP need no change; they surface the exception message.
- Both halves are needed: fixing only `_resolve_dataset_id` still yields 500 over HTTP
  (`DatasetNotFoundError` is not a `ValueError`); fixing only the router still yields the
  `AttributeError` → 500.

Known side effect of the router change: other `CogneeApiError`s raised deeper now surface
with their own status instead of 500. Concretely `datasets.delete_data` raises
`UnauthorizedDataAccessError` (401) when `dataId` belongs to a different dataset
(`cognee/api/v1/datasets/datasets.py:269`). 401 there is pre-existing and semantically
odd, but strictly better than 500; re-typing it is out of scope. Response body for
403/404 will be the app handler's `{"detail": ...}` while the router's 422/500 keep
`{"error": ...}`, the same shape mismatch every other router has.

Open question for maintainers: should an unknown `dataset_id` stay 403 (current, no-leak)
or become 404 like the name path? The proposal keeps 403: it originates in
`get_specific_user_permission_datasets`, shared by every UUID path, so changing it only here
would be inconsistent. Name lookups only scan the caller's own permitted datasets, so 404
there leaks nothing.

## Tests

- `cognee/tests/unit/api/v1/forget/test_forget_dataset_args.py`:
  `test_forget_unknown_dataset_name_raises_not_found` (patch `get_authorized_dataset_by_name` to return `None`, assert `DatasetNotFoundError`, assert `dataset_lock` / `set_database_global_context_variables` never entered). Covers the zero-datasets case too, since that path also returns `None`.
- `cognee/tests/unit/api/v1/forget/test_forget_endpoint.py`:
  `test_forget_endpoint_unknown_dataset_returns_404` and
  `test_forget_endpoint_permission_denied_returns_403` (patch `forget` to raise the typed errors);
  regression guards that a raised `RuntimeError` still maps to 500 and `ValueError` still to 422.
- `cognee/tests/cli_tests/cli_unit_tests/test_cli_commands.py::TestForgetCommand`:
  `asyncio.run` raising `DatasetNotFoundError` exits 1 with `not found or not accessible` in the message (assert on substring, not the full `str()`).

## Related

- #4304 (`/datasets/{id}/data` 500 instead of 404), #4305 / #4306 (closed: activity/export and visualize returned 500/409 instead of 403), #4976 (closed: permissions endpoint 500). Same class of "typed error swallowed as 500".
- Community PR #3863 "standardize error responses in improve/forget/recall" (stale, unreviewed) overlaps in spirit; not verified to cover this path.
- Draft core PR #4597 (serve/cloud client) does not touch `forget.py` or the forget router.

Happy to open a PR against `dev` with the change and the tests above.

ptimizeroracle:
## Description

`cognee.forget(dataset="<name>")` for a name the caller cannot resolve crashed with `AttributeError: 'NoneType' object has no attribute 'id'`, and over HTTP the forget router's `except Exception` catch-all turned that crash, and every other typed error, into `500 {"error": "An error occurred during deletion."}`

two changes:

- `_resolve_dataset_id` in `cognee/api/v1/forget/forget.py`: when `get_authorized_dataset_by_name` returns None, raise `DatasetNotFoundError` (404) with a message that deliberately covers both missing and unauthorized, so dataset names are not leaked. the UUID branch already had a guard; the name branch indexed `dataset.id` unconditionally
- `get_forget_router.py`: re-raise `CogneeApiError` before the catch-all so the global handler at `cognee/api/client.py` renders the error's own status code. `DatasetNotFoundError` now returns 404 and `PermissionDeniedError` returns 403 instead of both becoming 500

## Reasoning

i considered fixing this only in the router by catching AttributeError, but that would leave the SDK, CLI and MCP surfaces with the raw crash. raising at the lookup site fixes all four surfaces at once, and the DatasetNotFoundError + global handler pairing is already the pattern the rest of the API uses (cognify, validate). the no-leak wording matches the existing UUID-path message ("not found or not accessible") so both branches behave the same way

## Acceptance Criteria

- `forget(dataset="does-not-exist")` raises `DatasetNotFoundError` (404) on SDK, CLI, MCP; HTTP returns 404
- `forget(dataset_id=<unknown uuid>)` still raises `PermissionDeniedError` (unchanged no-leak semantics); HTTP returns 403 instead of 500
- existing 422 parameter-validation behavior unchanged

## Type of Change

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature
- [ ] Breaking change
- [ ] Documentation update

## Screenshots (local tests passing)

```
$ uv run pytest cognee/tests/unit/api/v1/forget/ -q
18 passed, 14 warnings in 1.04s

$ uv run pytest cognee/tests/unit/api/ cognee/tests/unit/exceptions/ -q
688 passed, 2 skipped, 80 warnings in 60.64s

$ uv run pytest cognee/tests/unit/modules/users/ cognee/tests/unit/shared/ -q
121 passed, 13 warnings in 1.21s
```

SDK behavior after the fix (issue #5013 repro):

```
SDK {'dataset': 'does-not-exist'} -> DatasetNotFoundError | DatasetNotFoundError: Dataset 'does-not-exist' not found or not accessible. (Status code: 404)
SDK {'dataset_id': UUID('...')} -> PermissionDeniedError | PermissionDeniedError: Request owner does not have necessary permission: [delete]
```

new regression tests in `cognee/tests/unit/api/v1/forget/test_forget_endpoint_not_found.py`: unknown name returns 404 over HTTP, PermissionDeniedError is no longer masked as 500, and `_resolve_dataset_id` itself raises the typed error

## Pre-submission Checklist

- [x] I have tested these changes thoroughly locally
- [x] My changes are minimal and focused on the described issue
- [x] Code follows the repository's style guidelines (ruff check / ruff format clean)
- [x] I have updated documentation where applicable (endpoint docstring error codes)
- [x] All local tests pass
- [x] I have searched existing PRs to ensure this is not a duplicate
- [x] I have linked the relevant issue (Fixes #5013)
- [x] My commits have clear messages

## DCO Affirmation

### DIFF
--- a/cognee/api/v1/forget/forget.py
+++ b/cognee/api/v1/forget/forget.py
@@ -484,7 +484,12 @@ async def _resolve_dataset_id(dataset_ref: str | UUID, user: Any) -> UUID:
             raise ValueError(f"Dataset {dataset_ref} not found or not accessible.")
         return dataset.id
 
+    from cognee.modules.data.exceptions import DatasetNotFoundError
     from cognee.modules.data.methods import get_authorized_dataset_by_name
 
     dataset = await get_authorized_dataset_by_name(dataset_ref, user, "delete")
+    if dataset is None:
+        # Same message for missing and unauthorized: the name lookup returns None
+        # for both, and distinguishing them would leak which dataset names exist.
+        raise DatasetNotFoundError(message=f"Dataset '{dataset_ref}' not found or not accessible.")
     return dataset.id
--- a/cognee/api/v1/forget/routers/get_forget_router.py
+++ b/cognee/api/v1/forget/routers/get_forget_router.py
@@ -6,6 +6,7 @@ from pydantic import ConfigDict, Field
 
 from cognee import __version__ as cognee_version
 from cognee.api.DTO import InDTO
+from cognee.exceptions import CogneeApiError
 from cognee.modules.users.methods import get_authenticated_user
 from cognee.modules.users.models import User
 from cognee.shared.logging_utils import get_logger
@@ -83,6 +84,8 @@ def get_forget_router() -> APIRouter:
         accepted.
 
         ## Error Codes
+        - **404 Not Found**: Dataset name does not exist or is not accessible
+        - **403 Forbidden**: Caller lacks delete permission on the dataset
         - **422 Unprocessable Entity**: Invalid parameter combination (e.g. both `dataset`
           and `datasetId`, `dataId` without a dataset, or `memoryOnly` without a dataset)
         - **500 Internal Server Error**: Error during deletion
@@ -115,6 +118,11 @@ def get_forget_router() -> APIRouter:
                     "error": "Invalid request parameters. Specify dataset or dataset_id, data_id+dataset, or everything=True."
                 },
             )
+        except CogneeApiError:
+            # Typed errors carry their own status code (DatasetNotFoundError -> 404,
+            # PermissionDeniedError -> 403) and are rendered by the global
+            # CogneeApiError handler; swallowing them here used to turn both into 500s.
+            raise
         except Exception:
             logger = get_logger()
             logger.exception("Forget endpoint error")
--- a/cognee/tests/unit/api/v1/forget/test_forget_endpoint_not_found.py
+++ b/cognee/tests/unit/api/v1/forget/test_forget_endpoint_not_found.py
@@ -1,0 +1,97 @@
+"""Regression tests for typed errors from POST /v1/forget (issue #5013).
+
+Before the fix, an unknown dataset name crashed forget() with
+`AttributeError: 'NoneType' object has no attribute 'id'` and the router's
+catch-all turned every typed error (including PermissionDeniedError, which
+carries status_code=403) into `500 {"error": "An error occurred during deletion."}`.
+Now `_resolve_dataset_id` raises DatasetNotFoundError for an unresolvable name
+and the router re-raises CogneeApiError so the global handler renders its code.
+"""
+
+import importlib
+import uuid
+from types import SimpleNamespace
+from unittest.mock import AsyncMock, patch
+
+import pytest
+from fastapi.testclient import TestClient
+
+from cognee.api.client import app
+from cognee.modules.data.exceptions import DatasetNotFoundError
+from cognee.modules.users.exceptions import PermissionDeniedError
+from cognee.modules.users.methods import get_authenticated_user
+
+forget_pkg = importlib.import_module("cognee.api.v1.forget")
+
+
+@pytest.fixture(scope="module")
+def test_client():
+    with TestClient(app) as c:
+        yield c
+
+
+@pytest.fixture
+def client(test_client):
+    async def override_get_authenticated_user():
+        return SimpleNamespace(
+            id=str(uuid.uuid4()),
+            email="test@example.com",
+            is_active=True,
+            tenant_id=str(uuid.uuid4()),
+        )
+
+    app.dependency_overrides[get_authenticated_user] = override_get_authenticated_user
+    yield test_client
+    app.dependency_overrides.pop(get_authenticated_user, None)
+
+
+def test_unknown_dataset_name_returns_404(client):
+    """forget(dataset=<unknown name>) must surface as 404, not a 500 NoneType crash."""
+    with (
+        patch.object(forget_pkg, "forget", new_callable=AsyncMock) as mock_forget,
+        patch("cognee.api.v1.forget.routers.get_forget_router.send_telemetry"),
+    ):
+        mock_forget.side_effect = DatasetNotFoundError(
+            message="Dataset 'does-not-exist' not found or not accessible."
+        )
+
+        response = client.post("/api/v1/forget", json={"dataset": "does-not-exist"})
+
+    assert response.status_code == 404
+    assert "does-not-exist" in response.text
+    assert "DatasetNotFoundError" in response.text
+
+
+def test_permission_denied_no_longer_masked_as_500(client):
+    """PermissionDeniedError carries status_code=403 and must reach the client as 403."""
+    with (
+        patch.object(forget_pkg, "forget", new_callable=AsyncMock) as mock_forget,
+        patch("cognee.api.v1.forget.routers.get_forget_router.send_telemetry"),
+    ):
+        mock_forget.side_effect = PermissionDeniedError(
+            message="You do not have permission to delete this dataset."
+        )
+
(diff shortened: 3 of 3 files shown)


hardness1020:
Thanks for picking this up. I had been planning to implement this after working through the root cause and proposed solution in #5013 .

ptimizeroracle:
thanks. the repro and the current-vs-expected table in #5013 did most of the work here, the fix just follows your expected-behavior column. if you had a different shape in mind for the router mapping i'm happy to adjust, otherwise this is yours to review too.

Vasilije1990 (approved):

PR #5035, opened 2026-09-13
fix(permissions): scope user role lookups to tenant (SDK-638)

Vasilije1990:
## Description

A manager authorized for tenant A could request a user's roles and receive roles from tenant B. `get_user_roles` now verifies that the target belongs to the authorized tenant and returns only roles from that tenant. Missing or out-of-tenant users receive `UserNotFoundError` (404); members with no roles receive an empty list.

Integrates #3281 onto current `dev` for internal CI, preserving @fallintoplace's original commit and authorship. Conflict resolution retains the newer `get_users_in_role` implementation, which already validates the role's tenant and allows role members to see their co-members. The new SQLite integration tests run the real permission checks and are placed in the infrastructure suite that CI executes.

Linear: https://linear.app/cognee/issue/SDK-638

## Validation

- `pytest cognee/tests/unit/users/ cognee/tests/integration/infrastructure/test_tenant_role_lookups.py -q --timeout=60` — 39 passed, using the existing Python 3.11 environment.
- `pre-commit run --files cognee/modules/users/tenants/methods/get_user_roles.py cognee/tests/unit/users/tenants/test_role_lookup_methods.py cognee/tests/integration/infrastructure/test_tenant_role_lookups.py` — passed.
- `git diff --check` — passed.

Coverage includes users in multiple tenants, foreign and missing users, members with no local roles, unauthorized callers, and existing role-member visibility. No MCP or UI changes.

## Type of Change

- [x] Bug fix

## DCO Affirmation

### DIFF
--- a/cognee/modules/users/tenants/methods/get_user_roles.py
+++ b/cognee/modules/users/tenants/methods/get_user_roles.py
@@ -1,7 +1,8 @@
 from sqlalchemy import select
 from uuid import UUID
 
-from cognee.modules.users.models import User
+from cognee.modules.users.exceptions import UserNotFoundError
+from cognee.modules.users.models import Role, User, UserRole, UserTenant
 from cognee.infrastructure.databases.relational import get_relational_engine
 from cognee.modules.users.permissions.methods.has_user_management_permission import (
     has_user_management_permission,
@@ -14,10 +15,21 @@ async def get_user_roles(tenant_id: UUID, user_id: UUID, user: User):
 
     db_engine = get_relational_engine()
     async with db_engine.get_async_session() as session:
-        from cognee.modules.users.models import Role
+        tenant_membership_result = await session.execute(
+            select(UserTenant).where(
+                UserTenant.user_id == user_id,
+                UserTenant.tenant_id == tenant_id,
+            )
+        )
+        tenant_membership = tenant_membership_result.scalars().first()
+
+        if tenant_membership is None:
+            raise UserNotFoundError(message="User not found in tenant.")
 
         role_results = await session.execute(
-            select(Role).join(Role.users).where(User.id == user_id)
+            select(Role)
+            .join(UserRole, Role.id == UserRole.role_id)
+            .where(UserRole.user_id == user_id, Role.tenant_id == tenant_id)
         )
         roles = role_results.scalars().all()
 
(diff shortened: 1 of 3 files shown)


cognee/modules/users/tenants/methods/get_user_roles.py
```diff
@@ -15,10 +16,19 @@ async def get_user_roles(tenant_id: UUID, user_id: UUID, user: User):
...
     async with db_engine.get_async_session() as session:
-        from cognee.modules.users.models import Role
+        tenant_membership_result = await session.execute(
+            select(UserTenant).where(
+                UserTenant.user_id == user_id,
+                UserTenant.tenant_id == tenant_id,
+            )
+        )
+        if tenant_membership_result.scalars().first() is None:
+            raise UserNotFoundError(message="User not found in tenant.")
```
NMZivkovic, line 26: This adds a 404 to `GET /tenants/{tenant_id}/roles/users/{user_id}`, but that route's docstring in `get_permissions_router.py` still lists only 403. The neighboring routes document their 404s, so I'd add something like `- **404 Not Found**: User is not a member of this tenant` there, so API clients know to expect it.

NMZivkovic (commented):
Looks good to me. The old query joined `Role.users` with no tenant filter, so a manager in tenant A could read a user's roles from tenant B. The membership check plus `Role.tenant_id == tenant_id` closes that, and it's consistent with `get_users_in_tenant`, which already loads `User.roles.and_(Role.tenant_id == tenant_id)`.

I checked that the new membership check can't lock out real members: `create_tenant` and `add_user_to_tenant` both insert the `user_tenants` row, `remove_user_from_tenant` deletes it, and the multi-tenant migration backfilled existing memberships. The permission check still runs first, so a caller without user-management rights gets a 403 and can't use the 404 to probe who is in a tenant. Running the real permission checks against SQLite in the integration test is a good call.

One small docs follow-up inline, otherwise good to go once it's out of draft.

NMZivkovic (dismissed):

cognee/modules/users/tenants/methods/get_user_roles.py
Vasilije1990: Done in b8de5ba — added `- **404 Not Found**: User does not exist or is not a member of this tenant` to the route docstring, plus a note that the returned roles are tenant-scoped and a member with no roles gets an empty list. Also updated the route row in the permissions skill table.

PR #5052, opened 2026-09-14
SDK-270 fix: No LLM on empty context, 404 on empty graph

### LINKED ISSUE #3728: [Bug]: GRAPH_COMPLETION search returns a phantom LLM non-answer instead of an empty result when the knowledge graph is empty

dexters1:
## Description

Fixes #3728 · Linear: [SDK-270](https://linear.app/cognee/issue/SDK-270/graph-completion-returns-phantom-llm-non-answer-on-empty-graph-gh-3728)

Completion searches on an empty graph (fresh install, pre-cognify, or after memory was dropped) sent the empty context to the LLM and returned its polite deflection — *"There is no context provided…"* — as a normal, non-empty search result. Callers checking `if results:` could not tell "no memory found" from a real answer.

**The contract, after this PR** — two emptinesses, two behaviors:

| State | Before | After |
|---|---|---|
| Knowledge graph **empty** (dataset exists) | phantom LLM answer | **`NoDataError` → HTTP 404**: "The knowledge graph is empty. Add data and run cognify before searching." |
| Graph populated, retrieval **matched nothing** | phantom LLM answer | **`[]`** — quiet miss, no LLM call |

The 404 is the maintainer-chosen improvement over the issue's original `[]` ask: API callers get an explicit, actionable signal for the setup-level state, exactly symmetric with the RAG retriever's existing `NoDataError` on a missing vector collection. A mere query miss stays a normal empty result — raising there would make narrow queries and `node_type` filters explode on healthy systems, and empty results compose across the multi-dataset fan-out where exceptions do not.

### How

- **`BaseRetriever.skip_completion_on_empty_context`** (new, documented beside the existing `supports_*` flags): `True` for the graph-completion family (GRAPH_COMPLETION, COT, context-extension, summary, decomposition) **and — search is not an LLM gateway — for `CompletionRetriever` (RAG), `HybridRetriever` (the default type), and `TripletRetriever`**. Explicitly `False` on `AgenticRetriever`, whose skills/tools legitimately answer without memory context.
- **Guards at every completion site**: each retriever's `get_completion_from_context` (sessionless + sequential-session paths) and the concurrent session lane before `complete_turn` (which otherwise bypasses those methods). A skipped turn returns a `None` sentinel translated to `[]` before `commit_turn`, so **no QA pair is recorded for an answerless turn** in either session mode.
- **`NoDataError` on `is_empty()`** in `GraphCompletionRetriever.get_retrieved_objects` and the decomposition retriever — the state error, raised before any retrieval or LLM work.
- Hybrid's global-context prelude counts as grounding: its guard fires only when every channel is empty. Batch completions keep their per-query arity contract and are not skipped.
- Side benefit: `recall()`'s `on_empty` tools fallback can now actually fire for query misses — phantom answers previously made results look non-empty.

## Test plan

- **Three-state e2e** (real OpenAI, isolated storage): `add` without `cognify` → `NoDataError (404)`; after `cognify` → real answer; `forget(memory_only=True)` (dataset kept, graph dropped) → `NoDataError (404)`. No phantom anywhere. (Note: full `forget(dataset=…)` deletes the dataset itself, so that path returns `[]` via zero-datasets before any retriever runs — the `memory_only` variant is the true empty-graph state.)
- Unit: rewrote the seven tests across four files that encoded the phantom contract (empty/None context expecting a completion) to assert **no LLM call**; added session-path skip, agentic opt-out, flag-contract across all six retrievers, concurrent-lane skip (`complete_turn` never runs, `commit_turn` not awaited), and empty-graph `NoDataError` tests for both graph retrievers.
- Full retrieval unit suite: **562 passed**; search + session suites green; ruff check/format and the `--ignore-noqa` gate clean.


cognee/modules/retrieval/graph_completion_retriever.py
```diff
@@ -131,8 +137,14 @@ async def get_retrieved_objects(
...
         if is_empty:
-            logger.warning("Search attempt on an empty knowledge graph")
-            return []
+            # An empty graph is a state problem, not a query miss: surface it
+            # loudly (404 over the API) the same way the RAG retriever raises
+            # on a missing vector collection, instead of quietly returning
+            # nothing. A populated graph with no matching triplets still
+            # yields an empty result below — that is a normal miss.
+            raise NoDataError(
+                "The knowledge graph is empty. Add data and run cognify before searching."
```
siillee:
```suggestion
                "The knowledge graph is empty. Ingest data through Cognee before searching."
```

cognee/modules/retrieval/graph_completion_decomposition_retriever.py
```diff
@@ -162,12 +162,11 @@ async def get_retrieved_objects(
...
 
         if is_empty:
-            logger.warning("Search attempt on an empty knowledge graph")
-            self._decomposition_state = DecompositionRunState(
-                original_query=query,
-                subqueries=[SubqueryRunState(query=query)],
+            # Same contract as GraphCompletionRetriever.get_retrieved_objects:
+            # an empty graph is a loud state error, not a quiet miss.
+            raise NoDataError(
+                "The knowledge graph is empty. Add data and run cognify before searching."
```
siillee:
```suggestion
                "The knowledge graph is empty. Ingest data through Cognee before searching."
```

siillee (approved):

PR #5056, opened 2026-09-15
fix(mcp): Scope unscoped recall by dataset ids, not names

hardness1020:
## Description

**Problem.** In API mode, the MCP `recall` (and `search`) tools fall back to "every dataset the caller can see" when called without `datasets`/`session_id`: `CogneeClient.recall` lists `GET /api/v1/datasets/` and sends the returned **names** as `datasets` (`cognee-mcp/src/cognee_client.py`; same in `search()`). That listing includes datasets other users shared with the caller (`cognee/api/v1/datasets/routers/get_datasets_router.py` via `get_all_user_permission_datasets`), but the server resolves names only against datasets the caller **owns** (`cognee/modules/data/methods/get_dataset_ids.py` via `get_datasets.py`). So shared datasets silently drop out of an unscoped recall, and a caller who owns no datasets (a tenant member with read grants only) gets `DatasetNotFoundError("No datasets found.")` and the tool returns an error. The search router already documents the rule: "Names only resolve to datasets owned by the caller; use dataset_ids for datasets shared with you."

**Reproduction.** Multi-tenant server; user B has read permission on user A's `team_memory` and owns nothing. MCP in API mode with B's token: `recall("what do we know about X?")` -> listing returns `team_memory`, payload sends `datasets=["team_memory"]`, server resolves zero datasets -> 404 "No datasets found". With one owned dataset plus the shared one, the shared one is silently ignored.

**Expected.** Unscoped recall/search should query every dataset the listing returned.

**Root cause.** The fallback uses names, which are owner-scoped, instead of the ids the same listing already provides.

**Fix.** In both fallbacks send `dataset_ids` from the listing (`DatasetDTO.id`), and keep sending `datasets` names alongside so an older server that ignores `dataset_ids` behaves as before. Ids remain ACL-checked server-side (`cognee/modules/data/methods/get_authorized_existing_datasets.py`), so this widens nothing beyond what the caller can already read. Both routers give `dataset_ids` precedence over names when both are present.

**Tests.** `cognee-mcp/tests/test_mcp_server_hardening.py` using the existing `httpx.MockTransport` pattern, parametrized over `recall` and `search`: mock `/api/v1/datasets/` returning an owned and a shared dataset, assert the query payload carries both ids in `dataset_ids` (names alongside); assert an explicit `datasets=` argument is forwarded untouched with no `dataset_ids` and no listing call.

**Related work.** No open PR/issue. Fallback introduced in 79bb25666 (COG-5929). Team-memory work (#4942, #4943) makes the shared-only case more common.

## Acceptance Criteria
- Unscoped MCP recall/search in API mode reaches shared datasets; shared-only callers no longer 404.
- Explicit `datasets` still forwarded as names.
- MCP hardening tests pass (`cd cognee-mcp && ../.venv/bin/python -m pytest tests -q`): 85 passed.

## Type of Change

## DCO Affirmation

https://claude.ai/code/session_018tKVErRWRRWzmBqEWsW5wA


dexters1:
Thanks for the PR @hardness1020! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #5057, opened 2026-09-15
fix(frontend): Detect active agent connections

### LINKED ISSUE #4957: [Bug]: Integration cards ignore active agent connections until a session exists
### Description

The dashboard and Integrations page infer whether Claude Code or Codex is connected only from rows returned by the sessions endpoint. A freshly registered agent connection can be active before it has produced a Q&A/session row, so the UI still renders **Connect agent** even though `GET /api/v1/agents/connections?active_only=true` reports an active connection.

### Steps to reproduce

1. Start the local Cognee backend and frontend.
2. Register an agent with `POST /api/v1/agents/register` and a recognized `session_id`, for example `codex_<id>`.
3. Confirm that `GET /api/v1/agents/connections?active_only=true` returns that active connection.
4. Open the dashboard before the agent has created a session/Q&A record.

### Actual behavior

The Codex card displays **Connect agent**. It changes to **Connected** only after a matching row appears in the sessions response.

### Expected behavior

An active registered connection with a recognized `session_id` prefix should immediately mark the corresponding integration as connected. Existing session-derived and persisted detection should remain as a fallback.

### Environment

- Cognee runtime: 1.5.3
- Cognee frontend: 1.0.0
- macOS, local backend/UI

I reproduced the behavior locally and verified that the current `dev` implementation still derives the badge only from session rows.

Ishubhammohole:
## Summary

- Detect active Claude Code and Codex connections from the agent connection registry, before the first session row exists.
- Keep session-derived and tenant-scoped persisted detection as fallbacks.
- Prevent active connection state from leaking across tenant switches.

## Testing

- `npx jest --runTestsByPath 'src/modules/agents/__tests__/getActiveAgentConnections.test.ts' 'src/app/(app)/dashboard/hooks/__tests__/useConnectedIntegrations.test.ts' --runInBand` — passed (6 tests)
- `npx eslint <all six changed files>` — passed
- `npx tsc --noEmit` — passed
- `npm run build` — passed
- `npm test -- --runInBand` — 50 suites passed; one pre-existing suite failed to start because Jest cannot parse the ESM `uuid` package. Reproduced unchanged on upstream `dev` (48 suites passed, same one failure).
- `npm run lint` — reports 58 pre-existing errors and 8 warnings; reproduced unchanged on upstream `dev`.

Fixes #4957


dexters1:
Thanks for the PR @Ishubhammohole! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #5060, opened 2026-09-15
feat(mcp): summarize recall hits and empty memory states (SDK-438)

NMZivkovic:
## Description

MCP recall results now start with a memory-hit summary, so users can see whether memory contributed and distinguish empty results from ingestion still in progress. For example, two returned session entries produce `2 memories found (2 from sessions)` above the unchanged result body. System status markers do not count as memory hits.

Empty results use dataset-scoped pipeline progress and graph summaries to distinguish indexing, an empty graph, an unindexed dataset, indexing failure, and no match. Diagnostics have a two-second timeout and report unavailable status instead of guessing. Successful hits skip these checks. Queued background ingestion is recognized before its pipeline status is recorded.

The response remains one text-content block, with additive `count` and `state` fields under `content[0]._meta["cognee/memory"]`. Text consumers can separate the summary from the previous body at the first newline. Source hints use returned metadata; progress counts describe data items, not inferred chunks.

Fixes [SDK-438](https://linear.app/cognee/issue/SDK-438).

## Acceptance Criteria

- [x] Recall results have a first-line summary counting the returned memory entries.
- [x] Empty results distinguish known indexing, empty-graph, and no-match states; unavailable diagnostics are explicit.
- [x] Source hints use existing metadata without additional lookups.
- [x] The existing result body and text-content block type are preserved; machine-readable metadata is additive.

## Type of Change

- [x] New feature

## Validation

Using the existing MCP virtual environment with FastMCP 3.4.6 and `ENV=test TELEMETRY_DISABLED=1`:

- `python -m pytest cognee-mcp/tests/test_recall_summary.py cognee-mcp/tests/test_mcp_server_hardening.py cognee-mcp/tests/test_tool_search.py -q --disable-warnings --tb=short` — 105 passed.
- After the last two edge-case tests and classifier adjustment, `python -m pytest cognee-mcp/tests/test_recall_summary.py -q --disable-warnings --tb=short` — 34 passed.
- Ruff 0.16.6 lint and format checks passed for all changed Python files.
- `git diff --check` passed.

Coverage includes API and direct-mode routing, dataset authorization, real MCP wire content, marker exclusion, single-line source hints, diagnostic cancellation, queued ingestion, unavailable graph counts, and progress aggregation.

MCP changes only; no frontend changes. No live-provider integration suite was run. This is a draft; local test screenshots are not attached.

## Pre-submission Checklist

- [x] Targeted regression tests pass.
- [x] Code follows the project's lint and formatting rules.
- [x] Tests and MCP README documentation added.
- [x] Checked for an existing PR on this branch.
- [x] Related Linear issue linked.
- [ ] Full integration validation and test screenshots before marking ready for review.

## DCO Affirmation


cognee/tests/test_conversation_history.py
```diff
@@ -444,16 +462,12 @@ async def main():
...
-        "Feedback-only turn must store the short acknowledgement, not the generated "
-        f"answer; got {len(stored_answer_autofeedback)} chars: "
-        f"{stored_answer_autofeedback[:200]}"
-    )
-    assert "TechCorp" not in stored_answer_autofeedback, (
-        "Feedback-only turn stored a generated answer about the retrieved subject"
+    # An acknowledgement may mention the subject or exceed an arbitrary length.
+    # Compare against the actual analysis output: storing the independently
+    # generated answer instead must still fail this regression check.
+    expected_ack = acknowledgement_for_turn(feedback_analysis.response_to_user)
```
siillee, line 486: This file (and `test_concurrent_turn_dispatch.py`) is also modified in #5061, with an incompatible implementation — both PRs additionally parametrize the same concurrent-turn test with a different acknowledgement string. Whichever merges second gets a conflict, and neither PR's stated scope covers these tests.

For what it's worth I think **this** version is the right one. #5061 drops the `len < 120` / `"TechCorp" not in` assertions for `used_graph_element_ids is None` / `used_session_context_ids is None` — provenance checks rather than answer-text checks, so a generated answer stored with no provenance recorded would still pass. The comment being deleted flags exactly that:

> Truthiness alone is not enough: before this fix the concurrent path stored the generated answer here, which is also truthy.

Comparing against `acknowledgement_for_turn(feedback_analysis.response_to_user)` pins the actual contract.

Could we pull these two files into a standalone test-only PR using this version, and drop them from both #5060 and #5061? #5061 then becomes what it claims to be — a 6-line fix plus three tests.

cognee-mcp/src/server_utils.py
```diff
@@ -206,3 +204,127 @@ def format_recall_results(
...
+    for item in results if isinstance(results, list) else [results]:
+        if (
+            _get_field(item, "_source", "source") == "system"
+            and _get_field(item, "status") == "build_failed"
+        ):
+            return RecallState("build_failed")
+    return None
+
+
+def classify_recall_state(progress: dict, graphs: list[dict]) -> RecallState:
```
siillee: Seven states — `found / empty / not_indexed / no_match / indexing / build_failed / unknown` — for a consumer with three actionable outcomes: got results, retry shortly, nothing. The four "nothing" variants produce four near-identical sentences, and distinguishing them is the *entire* reason for the `/datasets/graph-summary` round-trip and the `pipeline_run_id set + computed_at null → unknown` analysis below.

Collapsing to `found / indexing / none` drops ~60 lines here and one network call per empty recall, keeping `completed`/`total` on the indexing state — that's the genuinely useful part.

I'd rather do this before merge than as follow-up: the README documents all seven values as a public `content[0]._meta["cognee/memory"]["state"]` contract. Nothing consumes it yet, so it's free to change today and breaking to change after release.

cognee-mcp/src/cognee_client.py
```diff
@@ -694,6 +694,73 @@ async def remember(
                     "session_id": session_id,
                 }
 
+    async def get_recall_state(self, datasets: list[str] | None = None) -> RecallState:
```
siillee: An empty recall costs three round-trips here. API mode: `list_datasets()` → `/datasets/status/progress` → `/datasets/graph-summary`. Local mode calls `get_pipeline_progress(ids, pipeline)` once per pipeline — the signature takes a single name (`cognee/modules/pipelines/operations/get_pipeline_status.py:12`) — so three sequential DB queries for `add_pipeline`, `cognify_pipeline`, `code_graph_pipeline`.

Collapsing the state machine removes the graph-summary hop. For the rest, either batch the pipeline lookup or narrow to `cognify_pipeline` — `add_pipeline` progress isn't really what "is my memory ready" means.

cognee-mcp/src/server.py
```diff
@@ -507,10 +531,39 @@ async def recall(
                 system_prompt=system_prompt,
                 top_k=normalized_top_k,
             )
+            empty_state = recall_marker_state(results)
```
siillee, line 585: These two values get recomputed inside `format_recall_results` below — `empty_state or recall_marker_state(results)` and `items = recall_items(results)` — and `_unwrap_results` ends up running four times over the same object.

Passing the already-computed `items` and `empty_state` into the formatter removes the duplication. Same shape in `get_recall_state`, where `classify_recall_state` is called twice on the same `progress` (once with `[]` purely to early-exit) — an explicit `graphs: list | None` would let it do one pass.

cognee-mcp/src/server_utils.py
```diff
@@ -206,3 +204,127 @@ def format_recall_results(
...
+        tick = (active[0].get("progress") or {}) if len(active) == 1 else {}
+        completed, total = tick.get("completed_items"), tick.get("total_items")
+        if type(completed) is int and type(total) is int and 0 <= completed <= total and total > 0:
+            return RecallState("indexing", completed, total)
+        return RecallState("indexing")
+    if any(run.get("status") == "DATASET_PROCESSING_ERRORED" for run in runs):
+        return RecallState("build_failed")
+    if not graphs:
+        return RecallState("unknown")
+    if any(graph.get("num_nodes", graph.get("numNodes", 0)) > 0 for graph in graphs):
```
siillee: The `num_nodes`/`numNodes`, `computed_at`/`computedAt`, `pipeline_run_id`/`pipelineRunId` duck-typing is correct today — `DatasetGraphSummaryDTO` extends `OutDTO`, which sets `alias_generator=to_camel`, so REST responses are camelCase while local mode's `asdict(DatasetGraphCounts)` is snake_case.

But it puts serialization knowledge inside what should be a pure classifier, and `get_recall_state` already knows which mode it's in. Normalizing to snake_case in the API branch there would let this function handle one shape.

cognee-mcp/src/cognee_client.py
```diff
@@ -694,6 +694,73 @@ async def remember(
...
+            return classify_recall_state(progress, graphs)
+
+        from dataclasses import asdict
+
+        from cognee.modules.data.methods import (
+            get_authorized_existing_datasets,
+            get_datasets_graph_counts,
+        )
+        from cognee.modules.pipelines.operations.get_pipeline_status import get_pipeline_progress
+        from cognee.modules.users.methods import get_default_user
```
siillee, line 807: This is the fourth in-function `get_default_user` import in this file — existing ones at lines 421 and 534, plus one added in #5063. Worth a single module-level `_default_user()` helper before a fifth lands.

Unrelated to the code, but flagging here since it's the file that collides: this PR, #5063 and #5065 all amend `CogneeClient.remember()` and `server.remember()` (plus its `_track_background(...)` call) in overlapping hunks, so a three-way conflict is certain. Suggest merging **#5065 → #5063 → #5060** with a rebase each, rather than in review order.

cognee-mcp/src/server_utils.py
siillee: @NMZivkovic can you just double check this comment, seems to me like it makes sense, but maybe I'm wrong, otherwise the PR is good to go.

NMZivkovic: Done — collapsed before merge, on your reasoning that the README publishes these as a public `_meta["cognee/memory"]["state"]` contract, so it is free today and breaking later.

Four states rather than three. `found` / `indexing` / `build_failed` / `none`, one per action a caller has. I kept `build_failed` out of `none`: it is the one no-results case with a different action attached (go look at `cognify_status`), and it falls out of the pipeline status that is already being read, so it costs nothing. `indexing` keeps `completed`/`total`.

What went with the four-way split:

- the `/datasets/graph-summary` round trip on every empty recall
- the local-mode per-dataset graph traversal (`get_datasets_graph_counts`) — the probe is now **one** HTTP call in API mode and one status lookup locally, with no graph read in either
- `classify_recall_state`'s `graphs` parameter, which also removes the double call you flagged in the other thread (once with `[]` purely to early-exit)
- net −33 lines across the three source files

Worth noting your point about reliability was the strongest part of the case: a zero node count from an unavailable graph store was never proof of emptiness, so several branches fell back to `unknown` regardless. The distinction was costing a round trip to produce an answer it could not always give.

README now documents the four as a table. 126 MCP tests pass; the test file lost the cases that only existed to separate the four, and gained one asserting the graph traversal is no longer awaited.

siillee (approved):

PR #5061, opened 2026-09-15
fix(mcp): forward the requested pipeline in API status

NMZivkovic:
## Summary

Fixes SDK-679. In MCP API mode, `cognify_status(pipelines=["add_pipeline"])` requested dataset status without forwarding the pipeline name, so the API returned the default `cognify_pipeline` status. Multi-pipeline requests consequently showed the same status under different pipeline names.

Forward the requested `pipeline` query parameter. Preserve the API default when the name is empty. Add three regression tests covering named pipelines, distinct pipeline queries, and the empty-name fallback.

Supersedes #5022 by carrying its focused fix onto current `dev`; credits the original contributor. This avoids the original PR's unrelated main-based history and merge conflicts.

## CI follow-up

The first internal run exposed an unrelated fragile conversation-history assertion: a live LLM acknowledgement was rejected if it mentioned `TechCorp` or exceeded 119 characters, although neither restriction is part of the acknowledgement contract. A separate test-only commit replaces those wording heuristics with checks that a feedback-only turn records no discarded graph or session guidance. The deterministic concurrent-turn test now also covers an acknowledgement mentioning TechCorp, while still asserting that the generated answer is discarded and the acknowledgement is stored and returned exactly.

The first run also hit the Windows runner's existing silent-exit/no-JUnit failure pattern. The workflow's test-execution guard remains intact; this must pass on the final run before merge.

## Validation

- MCP suite: **84 passed**, one existing Starlette warning (Python 3.12).
- Ruff **0.16.6**: full-repository lint, strict exception rules, and formatting checks passed.
- `uv lock --check` and `git diff --check` passed.
- Dataset status/progress router tests: **4 passed** with `REQUIRE_AUTHENTICATION=true ENABLE_BACKEND_ACCESS_CONTROL=true`. An initial run in the reused MCP environment failed its authentication assertion because authentication was disabled; the explicit test configuration resolves it.
- Session concurrent-turn and dispatch unit tests: **35 passed**, including the entity-bearing acknowledgement case.
- First CI run: all Community Tests, MCP unit/live integration tests, and MCP Docker HTTP e2e passed. Final CI is rerunning after the test-only follow-up.
- Final commit: conversation-session e2e passes on **SQLite, Postgres, Redis, and filesystem** backends; MCP unit/live integration and Docker HTTP e2e also pass.
- Internal Test Suites and Community Tests will run on this same-repository PR, with repository secrets available to internal integration jobs.

Local MCP command (with the existing MCP virtual environment and `PYTHONPATH` pointing at this checkout):
```sh
cd cognee-mcp
ENV=test TELEMETRY_DISABLED=1 python -m pytest tests/ -q
```

Linear: https://linear.app/cognee/issue/SDK-679

## DCO


NMZivkovic (commented):
Reviewed `cbfc00372` against `dev`.

No blocking code findings in the final diff:

- MCP API mode now forwards the requested `pipeline` query parameter, matching the dataset-status route's alias and retaining the default when the name is empty.
- The three regression tests fail on unmodified dev in the two named-pipeline cases and pass with the fix. The complete MCP suite passes locally (84 tests), and the first CI run also passed MCP unit/live integration and Docker HTTP e2e.
- The session test correction replaces unsupported assumptions about LLM acknowledgement wording with the no-answer turn's graph/session provenance contract. The deterministic dispatch test continues to assert exact stored/returned acknowledgement text and discarded generated-answer context, including a TechCorp acknowledgement. All 35 focused session tests pass locally.

Merge remains pending final-commit CI and the required teammate approval. The preceding run exposed a Windows silent-exit/no-JUnit runner failure; the test-execution guard is preserved, and that job still needs a successful final run. This comment is a code-review report, not an approving review by a separate maintainer.

cognee/tests/test_conversation_history.py
```diff
@@ -444,16 +444,14 @@ async def main():
...
-        "Feedback-only turn must store the short acknowledgement, not the generated "
-        f"answer; got {len(stored_answer_autofeedback)} chars: "
-        f"{stored_answer_autofeedback[:200]}"
-    )
-    assert "TechCorp" not in stored_answer_autofeedback, (
-        "Feedback-only turn stored a generated answer about the retrieved subject"
+    # The LLM's acknowledgement may mention TechCorp or vary in length. Check the
+    # no-answer turn's recorded provenance instead of guessing from its wording:
+    # discarded retrieval and session guidance must not be recorded as served.
+    assert second_entry_autofeedback.used_graph_element_ids is None, (
```
siillee, line 456: The actual fix in this PR is right and minimal — good tests, and the comment explaining why the failure is silent is exactly the sort of thing that should be in the diff. No notes on it.

This file and `test_concurrent_turn_dispatch.py` are the issue: they're also modified in #5060, with a different and incompatible implementation, and both PRs parametrize the same concurrent-turn test with a different acknowledgement string. Guaranteed conflict.

I'd also argue this version weakens the test. Swapping the `len < 120` / `"TechCorp" not in` assertions for `used_graph_element_ids is None` / `used_session_context_ids is None` checks provenance rather than the stored answer — a generated answer recorded with no provenance would still pass. The comment being deleted flags this directly ("Truthiness alone is not enough..."). #5060's variant asserts `stored_answer == acknowledgement_for_turn(...)`, which pins the contract.

Proposal: extract the test fix to its own PR using #5060's version and drop it from both. This PR is then a clean two-file change and can merge first.

siillee (approved):

PR #5063, opened 2026-09-15
feat(mcp): select uploaded ontologies per remember write (SDK-681)

### LINKED ISSUE #5026: cognee-mcp: let remember() select an uploaded ontology (ontology_key) in API mode
## Summary

The Cognee REST API supports per-request ontology grounding: `POST /api/v1/ontologies` uploads an OWL file under an `ontology_key`, and `POST /api/v1/remember` / `POST /api/v1/cognify` accept `ontology_key` to ground extraction with it. `cognee-mcp` (0.5.5, and `main`) exposes `remember(data, filename, content_base64, dataset_name, session_id, custom_prompt, background)` without any ontology parameter, and `CogneeClient.remember` never sends `ontology_key`.

Today the only way to ground MCP writes is the server-global `ONTOLOGY_FILE_PATH` environment variable, which is static, applies to every dataset, and requires a file mount on the server.

## Use case

Several agent workspaces share one self-hosted Cognee server. Each workspace owns its own (evolving) ontology alongside its agent instructions and writes to its own dataset (`dataset_name`). We want to upload each ontology once via `/api/v1/ontologies` and have the MCP `remember` call pick it per write, without touching the server configuration.

## Proposal

Add an optional `ontology_key: str | list[str] | None` parameter to the `remember` tool and forward it as the `ontology_key` form field in API mode (and as `config={"ontology_config": ...}` in local mode, resolved through `OntologyService`). Optionally a `COGNEE_MCP_DEFAULT_ONTOLOGY_KEY` env var for clients that cannot pass it.

## Environment

cognee-mcp 0.5.5 (PyPI), Cognee server 1.5.4 (Docker), API mode (`--api-url`), stdio transport.

NMZivkovic:
## Description

MCP callers can now select uploaded ontologies for each permanent `remember` write with `ontology_key="workspace_v2"` or `ontology_key=["organizations", "software"]`.

API mode sends repeated `ontology_key` multipart form fields. Local mode resolves the keys through `OntologyService` for the default user and passes the resolver into the extraction configuration. Both foreground and background writes forward the selection. Session-only writes reject nonempty ontology keys because they do not perform extraction. Omitting the argument preserves the existing ontology configuration.

Fixes [SDK-681](https://linear.app/cognee/issue/SDK-681).
Closes #5026.

Scope: MCP client, tool schema, documentation, and regression tests. No UI changes. The optional default-ontology environment variable is not included.

## Acceptance Criteria

- [x] MCP `remember` accepts a single key or a list of keys.
- [x] API mode sends the expected multipart fields for text and file uploads.
- [x] Local mode resolves ontologies for the default user and fails before ingestion if a key is missing.
- [x] Foreground and background tool calls forward ontology selection.
- [x] Existing behavior is preserved when the argument is omitted.
- [x] Usage and session-mode limitations are documented.

## Type of Change

## Validation

```text
ENV=test TELEMETRY_DISABLED=1 python -m pytest \
  cognee-mcp/tests/test_remember_ontology.py \
  cognee-mcp/tests/test_mcp_server_hardening.py \
  cognee-mcp/tests/test_tool_search.py -q --disable-warnings --tb=short
90 passed, 1 warning
```

Ruff check and format check passed for all changed Python files. `git diff --check` passed. Tests use mocked HTTP transport and local dependencies; live-provider integration was not run.

## Screenshots

Not attached; local test results are recorded above.

## Pre-submission Checklist

- [x] I have added necessary documentation
- [ ] All new and existing tests pass (selected suites pass; full suite not run)
- [x] I checked for an existing PR on this branch
- [x] I have linked relevant issues in the description

## DCO Affirmation


cognee-mcp/README.md
```diff
@@ -670,3 +670,25 @@ We are committed to making open source an enjoyable and respectful experience fo
 ## Star History
 
 [![Star History Chart](https://api.star-history.com/svg?repos=topoteretes/cognee&type=Date)](https://star-history.com/#topoteretes/cognee&Date)
+
+### Select an uploaded ontology for a write
```
siillee: This lands below the Star History chart — the README ends Contributors → Star History, so this is an H3 orphaned under no H2, at the very bottom of the file. #5060 and #5065 both insert near the tool docs (~line 521/585).

As written the feature ships effectively undocumented. Could it move under the `remember` tool docs, next to #5065's `self_improvement` block?

cognee-mcp/src/cognee_client.py
```diff
@@ -670,6 +700,9 @@ async def remember(
                     "data": remember_data,
                     "dataset_name": dataset_name,
                 }
+                if ontology_config is not None:
+                    kwargs["config"] = {"ontology_config": ontology_config}
+                    kwargs["user"] = user
```
siillee: This is set only on the ontology branch. Without `ontology_key`, local `remember()` resolves its own user; with it, MCP resolves `get_default_user()` and pins it — two user-resolution paths for the same tool call, differing on an unrelated argument. The ontology also ends up resolved against a user that may not be the one `remember()` would have used.

Either pass `user` unconditionally, or don't pass it and resolve the ontology against whatever user `remember()` picks.

cognee-mcp/src/cognee_client.py
```diff
@@ -656,6 +668,24 @@ async def remember(
             return response.json()
         else:
             with redirect_stdout(sys.stderr):
+                ontology_config = None
+                if ontology_keys:
+                    from io import StringIO
+
+                    from cognee.api.v1.ontologies.ontologies import OntologyService
```
siillee, line 684: Four local imports in one branch (`StringIO`, `OntologyService`, `RDFLibOntologyResolver`, `get_default_user`) — and `get_default_user` is already lazy-imported inside functions at lines 421 and 534, so this makes three, with #5060 adding a fourth. A single module-level `_default_user()` helper would stop the pattern here.

Also flagging on this file since it's where they collide: this PR, #5060 and #5065 all amend `CogneeClient.remember()` and `server.remember()` in overlapping hunks. Suggest **#5065 → #5063 → #5060**, rebasing each.

cognee-mcp/README.md
NMZivkovic: Fixed. Moved up to sit with the other tool-behaviour H3s under `## Basic Usage`, just before `## Development and Debugging` — it was an H3 under no H2 at the very bottom.

While moving it I found what was actually failing **Code Quality**: `ruff format` lints python blocks inside markdown, and the `remember(...)` example was **101 characters** against the 100 limit:

```
1 file would be reformatted --> cognee-mcp/README.md:681:10
```

That one line also took **Community Tests Status** down with it — that job is `needs: [code-quality, ...]` and its entire log body is `echo "One or more community tests failed."`. Two red checks, one line break. Both gates verified locally with the exact CI commands.

cognee-mcp/src/cognee_client.py
NMZivkovic: Took the second half of your either/or — not passing it — and I think it's the better one.

`remember()` already does `if user is None: user = await get_default_user()` (remember.py:401-404), so resolving it for the lookup and *not* pinning it on the call means both agree by construction, with no behaviour change. I tried the other half first and it broke two tests that pin the exact kwargs (`test_cognee_client_local_remember_passes_text_through_untouched` and `test_local_omission_preserves_config`) — a fair signal the call should stay minimal.

The PR's own `test_local_resolves_for_default_user` asserted `kwargs["user"] is user`, so it now asserts `"user" not in kwargs`; the `lookup.assert_called_once_with([...], user)` it already had is what does the real work.

One extra thing while in there: `get_default_user()` queries the DB, and `remember()` is deliberate about running `setup()` first (comment at remember.py:1591). The ontology branch resolved a user before any of that, so it now calls `setup()` itself. 98 MCP tests pass.

siillee (dismissed):

cognee-mcp/src/cognee_client.py
```diff
@@ -605,12 +606,21 @@ async def remember(
...
             raise ValueError("Pass either `data` or `filename` + `content_base64`, not both.")
         if content_base64 and session_id:
             raise ValueError("File uploads (content_base64) do not support session_id.")
 
+        ontology_keys = [ontology_key] if isinstance(ontology_key, str) else ontology_key
+        ontology_keys = [key.strip() for key in (ontology_keys or []) if key.strip()]
+        if ontology_keys and session_id:
+            raise ValueError(
+                "ontology_key is only supported for permanent memory; omit session_id."
+            )
```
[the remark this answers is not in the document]
NMZivkovic: Fixed server-side. `POST /v1/remember` now refuses `ontology_key` together with `session_id`, added beside the existing `labels`/`external_metadata` + `session_id` rule you pointed at — same shape, same 400.

Placement mattered and cost a round to find: raising inside the `try` that wraps the ontology handling gets rewrapped as `409 "An error occurred during remember: 400: ..."`. The guard sits outside that block so a direct HTTP caller gets a real 400.

cognee-mcp/tests/test_remember_ontology.py
```diff
@@ -0,0 +1,132 @@
...
+        await client.remember(**kwargs, ontology_key=keys)
+
+
+@pytest.mark.asyncio
+@pytest.mark.parametrize("use_api", [False, True])
+async def test_session_rejects_ontology(use_api):
+    client = CogneeClient.__new__(CogneeClient)
+    client.use_api = use_api
+    with pytest.raises(ValueError, match="permanent memory"):
+        await client.remember("hello", session_id="session", ontology_key="one")
```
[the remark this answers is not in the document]
NMZivkovic: Agreed — the existing test only proved the client refuses to send it. Added `cognee/tests/unit/api/v1/remember/test_remember_router_ontology_session.py`: four tests driving the endpoint over HTTP via `TestClient`.

It covers the rejection (400, and `remember` never awaited, so it fails before any work is dispatched), each parameter working alone, and one case your sketch didn't — an empty `ontology_key`, which Swagger UI submits for untouched array items, must not block a session write.

siillee (approved):

Issue #3229, opened 2026-06-21
[Bug]: mcp_process Docker container leaked when UI startup fails in start_ui


aniruddhaadak80:
### Bug Description
In [cognee/api/v1/ui/ui.py](https://github.com/topoteretes/cognee/blob/dev/cognee/api/v1/ui/ui.py#L753-L768), the `start_ui` function starts the MCP server inside a Docker container using `subprocess.Popen` (stored in `mcp_process`) at lines 565-570.
However, if the frontend server fails to start later, the exception handler gracefully terminates and kills `backend_process` (if started), but does not terminate or kill `mcp_process`. This leaks the subprocess and the running Docker container on the host machine.

### Source Code Reference
- File: [cognee/api/v1/ui/ui.py](https://github.com/topoteretes/cognee/blob/dev/cognee/api/v1/ui/ui.py#L753-L768)

```python
    except Exception as e:
        logger.error(f"Failed to start frontend server: {str(e)}")
        # Clean up backend process if it was started
        if backend_process:
            logger.info("Cleaning up backend process due to frontend failure...")
            try:
                backend_process.terminate()
                backend_process.wait(timeout=5)
            except (subprocess.TimeoutExpired, OSError, ProcessLookupError):
                try:
                    backend_process.kill()
                    backend_process.wait()
                except (OSError, ProcessLookupError):
                    pass
        return None
```

### Steps to Reproduce
1. Invoke `start_ui(..., start_mcp=True)`.
2. Cause the frontend to fail during startup (for example, by forcing an exception during dependency installation).
3. Observe that while `backend_process` is cleaned up, the Docker container for the MCP server remains running and must be manually stopped.

### Expected Behavior
`mcp_process` should be terminated and the Docker container cleaned up when the UI startup fails.

### Actual Behavior
The MCP Docker container is leaked and left running in the background.

aniruddhaadak80:
Here is a minimum viable code script to reproduce this bug programmatically:

```python
import subprocess
from unittest.mock import patch
from cognee.api.v1.ui.ui import start_ui

def dummy_callback(pid_or_tuple):
    print("Spawned PID/Container:", pid_or_tuple)

# Force a failure during frontend startup by simulating Node.js missing
with patch("cognee.api.v1.ui.ui.check_node_npm", return_value=(False, "Simulated Node.js not installed")):
    start_ui(pid_callback=dummy_callback, start_mcp=True)

# Check if the Docker container is still running
result = subprocess.run(
[...]

PR #5064, opened 2026-09-15
fix(ui): clean up owned services when startup fails

NMZivkovic:
## Summary

Fixes SDK-685 and #3229. `cognee-cli -ui` could start MCP and the backend, then return on missing Node.js, a dependency-install failure, or an early frontend exit without stopping those resources. The old exception handler also used an unbounded backend shutdown wait.

Track every created process/container in a startup cleanup scope. Failure or interruption cleans up owned resources in reverse order; successful startup transfers ownership to the caller. Partial optional MCP startup is cleaned up even if the UI continues without MCP. Shutdown uses bounded waits, Unix process groups or Windows process-tree termination, and explicit removal of the uniquely named Docker container. Existing services and the separate optional-port behavior are preserved.

## Validation

- UI startup/preflight unit tests: **31 passed** (early returns, exceptions, Ctrl-C, exited processes, timeout escalation, Windows cleanup, partial MCP setup, and success).
- Real Docker lifecycle smoke: **1 passed**. Simulated missing Node.js after a real container started; verified that container was removed and a separate container remained running. The smoke uses a lightweight sleeping container instead of the MCP application and requires no LLM credentials.
- Ruff 0.16.6 lint and formatting; `git diff --check` passed.
- Combined UI and CLI unit checks: **243 passed, 1 skipped**. Final UI-specific rerun: **31 passed**.
- Community and internal PR CI run automatically on this branch.

Run the Docker smoke with a pre-pulled lightweight image:
```sh
COGNEE_TEST_DOCKER_CLEANUP=1 COGNEE_CLEANUP_TEST_IMAGE=alpine:3.22 \
  uv run pytest cognee/tests/integration/api/test_ui_startup_cleanup_docker.py -q
```

## Release note

Failed UI startup now cleans up MCP containers and backend/frontend processes created by that launch instead of leaving them running.

Linear: https://linear.app/cognee/issue/SDK-685

## DCO


cognee/api/v1/ui/ui.py
```diff
@@ -427,8 +428,67 @@ def prompt_user_for_download() -> bool:
         return False
 
 
+def _stop_ui_process(process: subprocess.Popen) -> None:
```
siillee: The `ExitStack` ownership model is the right shape for this, and replacing the unbounded backend wait is a clear win. But this adds a *second* implementation of teardown rather than replacing the first.

`cognee/cli/_cognee.py:275-329` already does all of this: `docker stop` → `docker rm -f` on timeout, `os.killpg(SIGTERM)` on Unix, `taskkill /F /T /PID` on Windows.

They also disagree about the correct behaviour. This function uses `os.killpg(process.pid, ...)` with a comment explaining why — the original group id survives npm's parent exiting. The CLI resolves `os.getpgid(pid)` first, which is the variant that breaks in exactly that case. So the correct implementation is the new one, and the incorrect one stays in the tree.

There's a concrete consequence on the failure path: `start_ui` now cleans up and returns `None`, but the CLI still calls `signal_handler(SIGTERM, None)` on the `None` branch (`_cognee.py:388`). Double teardown. `os.getpgid()` on a dead PID raises `ProcessLookupError` and is caught, but if the OS recycled that PID between the two cleanups, the CLI signals an unrelated process group. Low probability, bad outcome, avoidable.

Suggested shape:
1. Export `stop_ui_process` / `remove_ui_container` from `ui.py` as the single teardown API.
2. Have `signal_handler` call them instead of reimplementing.
3. Skip cleanup entirely on the `server_process is None` branch — `start_ui` already owns it, which is what this PR's new docstring promises.

Worth noting (3) can't just be deleting the CLI handler: it also handles Ctrl-C *after* successful startup, which `pop_all()` deliberately releases. It has to call into the shared helpers.

Minor, while here: this runs three `process.wait(timeout=5)` calls across two kill phases. On Windows `taskkill /T /F` is already forceful, so the follow-up `process.kill()` is unreachable in practice.

cognee/api/v1/ui/ui.py
```diff
@@ -479,9 +540,42 @@ def start_ui(
         >>> server = cognee.start_ui(dummy_callback, start_mcp=True)
         >>> # UI will be available at http://localhost:3000
         >>> # MCP server will be available at http://127.0.0.1:8001/sse
-        >>> # To stop all servers later:
-        >>> server.terminate()
+        >>> # Track the callback's PIDs/container name to stop all servers later.
     """
+    # Keep ownership until the entire startup succeeds. ExitStack also handles
+    # early returns and BaseException (for example Ctrl-C during npm install).
+    with ExitStack() as resources:
```
siillee, line 571: Rebase needed before this lands. This branch is on current dev (`ui.py` blob `3e154fb7e5`), but #5066 is on `5cafe7967d` and rewrites the same ~20 lines of `start_ui`'s port preflight (`ports_to_check` → `required_ports`, MCP pulled out of the required set).

Suggest landing #5066 first since it's the smaller change, then rebasing this one onto it.

siillee (approved):

PR #5065, opened 2026-09-15
fix(mcp): expose remember self-improvement control

NMZivkovic:
## Summary

Fixes the remaining scope of SDK-683: MCP callers can now pass `self_improvement=False` to `remember`.

Expose an optional boolean and forward explicit values through the MCP tool, `CogneeClient`, and REST handlers. Permanent text/file ingestion and queued background writes preserve the flag; `False` skips automatic improvement while add/cognify still run. Omission or null leaves the core default unchanged (currently true).

Session behavior is documented and tested:
- Direct mode: `False` disables the session-to-graph bridge while keeping session storage.
- API mode: typed QA entries remain cache-only for every flag value. This flag does not control separately requested skill improvement.

Scope is parameter forwarding, documentation, and regression coverage. The historical projection/re-embedding work Igor marked resolved remains completed; no improve/memify performance implementation or default is changed.

Related earlier attempt: #4246 (closed without merge). Original community report: #4233. This PR addresses only its missing MCP control.

## Validation

- Full MCP suite: **118 passed**.
- Related REST/core/session, COGX and code-router tests: **93 passed**.
- Coverage includes true/false/omission, direct/API modes, text/uploads, foreground/background writes, schema exposure, REST boolean parsing, retained ingestion, and session semantics.
- Repository-wide Ruff 0.16.6 lint and formatting, `uv lock --check`, and `git diff --check` passed.

```python
remember(data="A new fact", self_improvement=False)
remember(data="A new fact", self_improvement=False, background=True)
```

Linear: https://linear.app/cognee/issue/SDK-683

## DCO


cognee-mcp/src/cognee_client.py
```diff
@@ -595,10 +595,13 @@ async def remember(
         custom_prompt: str | None = None,
         filename: str | None = None,
         content_base64: str | None = None,
+        self_improvement: bool | None = None,
```
siillee, line 598: Forwarding looks clean and the session-mode semantics are well documented — particularly the direct-vs-API distinction, which is easy to get wrong.

One coordination note: this PR, #5063 and #5060 all amend `CogneeClient.remember()` and `server.remember()` (plus its `_track_background(...)` call) in overlapping hunks. Three-way conflict is certain, and the last one merged gets a manual re-resolution of a behaviour-sensitive call site.

This is the smallest of the three at that call site — suggest merging it first, then **#5063 → #5060**, rebasing each.

cognee/api/v1/remember/routers/get_remember_router.py
```diff
@@ -189,6 +189,14 @@ async def remember(
                 "to skip tagging."
             ),
         ),
+        self_improvement: bool | None = Form(
```
siillee, line 192: Optional follow-up, not blocking: the `**({"self_improvement": x} if x is not None else {})` idiom appears five times across this PR — twice in this router, once as `improvement_kwargs` in `server.py`, twice in `cognee_client.py` — purely because core `remember()` declares `self_improvement: bool = True` and can't accept `None`.

Changing core to `self_improvement: bool | None = None`, with the default applied at the top of the function, would collapse all five to plain forwarding. Wider blast radius, so reasonable to defer to its own PR.

cognee-mcp/src/cognee_client.py
NMZivkovic: Noted on the ordering — agreed, and the triage landed on the same sequence.

Separately: the three Ubuntu 3.10 shard failures on this PR were **not** flaky. All three were the same new test failing on a mock target:

```
AttributeError: <function get_authorized_dataset ...> does not have
the attribute 'get_authorized_dataset'
```

`methods/__init__.py` re-exports the function under its module's own name, so the package attribute shadows the submodule and mock's getattr-first `_dot_lookup` resolves the prefix to the function. Three shards rather than one because pytest-split scattered the parametrizations — a flake hits one shard, a broken patch target hits all of them. This file already documents that trap and importlib-imports four modules to dodge it; this target just hadn't followed the convention. Fixed and pushed; the three parametrizations pass.

siillee (dismissed):
Merge conflicts left to resolve, but otherwise lgtm

siillee (dismissed):

NMZivkovic:
### Dev landed the same control while this was in review — reverted to its contract

`MCP Tests` failed on two tests I didn't write:

```
FAILED tests/test_mcp_improve.py::test_mcp_remember_forwards_self_improvement - KeyError: 'self_improvement'
FAILED tests/test_mcp_improve.py::test_mcp_remember_advertises_self_improvement - assert None is True
```

They came in with the dev merge. #4920 (SDK-416, *Make improve() an explicit orchestrator*) added `self_improvement` to `server.remember()` and `CogneeClient.remember()` independently, with the opposite contract:

| | this branch (before) | dev (#4920) |
|---|---|---|
| MCP signature | `bool \| None = None` | `bool = True` |
| forwarded to client | only when not `None` | always |
| client → core / wire | whenever not `None` | only when `False` |
| advertised schema default | `null` | `true` |

The merge resolved `server.py` in favour of this branch but took dev's tests verbatim, so the two halves disagreed.

**I've reverted the MCP layer to dev's shape.** Its contract is already merged and tested, and the two agree wherever it matters — `False` is the only value that changes behaviour, and dev's client drops `True` so the core default stays authoritative either way. The tri-state only bought the ability to force `True` past a core default that doesn't exist. Rewriting a contract someone else just shipped isn't worth that.

Kept from this branch:

- the docstring and README wording separating **direct-mode** session bridging from **API-mode** typed entries;
- the opt-out on the JSON session path (`/remember/entry`) — dev forwards it on the multipart path only;
- the router-level tri-state and its core tests: omitting the form field now leaves core's default in charge instead of pinning `True` at the HTTP boundary.

`test_remember_self_improvement_end_to_end` (6 call shapes × API/local × 3 flag values) stays and now asserts the "only `False` travels" rule directly. Dropped `test_remember_advertises_optional_self_improvement`, which dev's advertise test supersedes.

@siillee — re: the five `**({"self_improvement": x} if x is not None else {})` sites, this takes it down to **two**, both in the router. Your suggested fix (core accepting `bool | None = None` and applying the default itself) would clear those last two; still worth its own PR, and the collision above is a decent argument for doing it centrally rather than per-caller.

Local: `cognee-mcp` **181 passed, 1 skipped**; the touched core unit tests **105 passed**.

---

The other red check, `Unit tests 3.13.x on windows-latest (2/3)`, is **not a test failure** — the job hung and hit the 15-minute action timeout with no `junit-unit.xml` written. `pytest-timeout` (`--timeout=300`) never fired and py-spy couldn't attach to either pytest pid, so the stall is below the plugin's reach. Same intermittent Windows hang dev shows on its own (1, 0, 2, 1, 0 failures across five recent dev runs); unrelated to this diff.

Needs a re-approval — the push dismissed the previous one.

siillee (approved):

PR #5066, opened 2026-09-15
fix(ui): handle occupied optional MCP port (SDK-682)

NMZivkovic:
## Description

Fixes [SDK-682](https://linear.app/cognee/issue/SDK-682).

`cognee-cli -ui` now treats MCP as optional during port preflight. An occupied MCP port logs a warning and skips MCP while the frontend and backend continue to start. Frontend and backend port conflicts remain fatal. The CLI summary reports MCP as skipped unless a container was actually launched.

## Validation

- Full unit suite: 5,773 passed, 75 skipped.
- Full CLI unit suite: 205 passed, 1 skipped.
- Focused UI/startup tests: passed.
- Real loopback socket smoke test: occupied MCP port detected; UI/backend launch continued; Docker was not invoked; existing listener remained reachable.
- Ruff and `git diff --check`: passed.

The broader integration suite was not fully green because this environment lacks the optional `unstructured` dependency and contains unrelated agent-memory fixture errors.

## DCO

### DIFF
--- a/cognee/api/v1/ui/ui.py
+++ b/cognee/api/v1/ui/ui.py
@@ -478,16 +478,13 @@ def start_ui(
     """
     logger.info("Starting cognee UI...")
 
-    ports_to_check = [(port, "Frontend UI")]
+    required_ports = [(port, "Frontend UI")]
 
     if start_backend:
-        ports_to_check.append((backend_port, "Backend API"))
-
-    if start_mcp:
-        ports_to_check.append((mcp_port, "MCP Server"))
+        required_ports.append((backend_port, "Backend API"))
 
     logger.info("Checking port availability...")
-    all_ports_available, unavailable_services = _check_required_ports(ports_to_check)
+    all_ports_available, unavailable_services = _check_required_ports(required_ports)
 
     if not all_ports_available:
         error_msg = f"Cannot start cognee UI: The following services have ports already in use: {', '.join(unavailable_services)}"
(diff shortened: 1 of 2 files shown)


cognee/api/v1/ui/ui.py
```diff
@@ -478,24 +478,28 @@ def start_ui(
...
+    all_ports_available, unavailable_services = _check_required_ports(required_ports)
 
     if not all_ports_available:
         error_msg = f"Cannot start cognee UI: The following services have ports already in use: {', '.join(unavailable_services)}"
         logger.error(error_msg)
         logger.error("Please stop the conflicting services or change the port configuration.")
         return None
 
-    logger.info("✓ All required ports are available")
+    if start_mcp and not _is_port_available(mcp_port):
```
siillee, line 501: Behaviour change is right — an occupied optional port shouldn't be fatal.

Note that `start_mcp` is now progressively falsified at three separate sites: dev already has the Docker-preflight `start_mcp = False` followed *immediately* by a second `if start_mcp:` block, this adds a third, and #5064 adds a fourth failure mode (`mcp_resources.close()` in its except). The CLI then re-derives "did MCP actually start" from whether the `docker_container` sentinel got set.

A single `_try_start_mcp(...) -> str | None` returning the container name would make `start_mcp` non-mutable and let the CLI report MCP as started iff the return is non-None, instead of inferring it. It would also fold away the `_check_required_ports`-vs-direct-`_is_port_available` split this introduces — `_check_required_ports` is now a 6-line loop serving at most two ports while the third is checked inline.

Happy for that to be a follow-up if you'd rather keep this PR focused.

Separately: this branch is on `ui.py` blob `5cafe7967d` while dev is at `3e154fb7e5`, and #5064 restructures the same block from current dev. Suggest rebasing and landing this one first since it's smaller, then rebasing #5064 onto it.

cognee/cli/_cognee.py
```diff
@@ -367,8 +367,12 @@ def pid_callback(pid_or_tuple):
                 fmt.echo(f"The interface is available at: http://localhost:{frontend_port}")
                 if start_backend:
                     fmt.echo(f"The API backend is available at: http://localhost:{backend_port}")
-                if start_mcp:
+                if start_mcp and docker_container:
```
siillee, line 372: This is the CLI inferring MCP's real state from a sentinel that `start_ui` happens to set — which works, but it's the third place that has to agree about what "MCP started" means.

If `start_ui` grew a `_try_start_mcp(...) -> str | None` (see my note on `ui.py`), this becomes a direct read of that return value rather than a correlation between two variables.

siillee (approved):

PR #5071, opened 2026-09-15
fix(recall-coverage): Close a run no process will ever finish (COG-6598)

LStromann:
Fixes COG-6598. Base is `feat/recall-coverage`, where the feature lives — it is not on `dev` yet.

## The defect

A run's status is written by the process executing it, so a process that dies writes nothing. `config.py` already documented the consequence — "a pod rescheduled mid-run leaves the row at 'running' for ever" — and answered it by bounding the 409 guard: `runs_in_flight(..., stale_after_seconds=...)` ignores rows past an hour, so a new run is never blocked.

Only the guard learned that. The stored status did not, and every reader trusts the status:

- The Memory Coverage page reports **"Scoring…"** against a dataset where nothing is running, to a user who started nothing, and polls it every 4 seconds for the life of the tab.
- It disables "Run analysis" for that dataset, so the page offers no way out of a state that will never change by itself.
- `GET /v1/coverage/runs` reports it pending/running for ever.

There is no cancel or delete route, so the only recovery was manual SQL. This is what it looks like in the product — it is why the bug was found:

> "why do you for some say 'Scoring…'? Is there something running right now for scoring? because I did not trigger anything […] Because of this bug I cannot rerun the test"

## The change

`expire_stale_runs(owner_ids, stale_after_seconds)` fails pending/running rows older than the window and writes `summary.error`:

> Run abandoned: no ending was reported within 3600s, so the process that owned it is gone.

Deliberately *abandoned*, not *failed*: the run may well have done the work, and the only thing known is that it stopped reporting. A reader deciding whether to start another one needs those two to read differently.

Called lazily, where these rows are already touched:

| Call site | Why there |
|---|---|
| `start_recall_coverage_run`, before the guard | So the row a user sees and the row the guard sees agree — the guard already skipped it |
| `GET /v1/coverage/runs` | Reading the trend is what a UI does while it waits for a run; until something looks, nothing else will ever close it |

No scheduler, no new column, no migration. The window is the one the guard already applies, so this adds no second rule about liveness — it makes an existing judgement visible.

## Why failing a live run is safe

`persist_run_results` writes `complete` unconditionally, so a run that was merely slow still lands its rows and its summary afterwards. The only cost of calling a slow run abandoned is that it reads as abandoned until it finishes. There is a test for exactly that.

This matters because a coverage run has **no mid-run heartbeat** — nothing writes the row between `mark_run_running` and completion — so `created_at` is the only evidence available. SDK-596 makes the case that `created_at` alone is too blunt for `PipelineRun`, and the same caveat applies here: an hour-plus run that is genuinely working gets marked abandoned and then corrects itself on completion. Adding a heartbeat (SDK-596 option 1) would let the window shrink and is the natural follow-up; it is out of scope for this fix.

## Tests

`386 passed` across `cognee/tests/unit/modules/recall_coverage` and `cognee/tests/unit/api/v1/recall_coverage`; `ruff check` and `ruff format` clean.

New coverage: an abandoned row is failed and says so; a row inside the window is untouched; a completed run is never rewritten; only the owner scope passed in is affected; `stale_after_seconds=0` disables it as it does for the guard; a run marked abandoned still persists its results; starting a run closes the row it steps over; the listing route runs the sweep.

The router tests stub the sweep in the shared app fixture — it is the one repository call the routes make that no per-test fake covers, and it wants a real table. Recorded rather than silenced, so one test asserts the route actually runs it.

## Reviewer notes

- The comment in `config.py` that justified leaving these rows alone is updated, since this PR overturns that decision.
- The frontend half is cognee-saas #554, which stops the UI believing such a row after the same hour. Either side fixes the symptom alone; this one fixes it for every reader.


LStromann:
Closing: wrong repo — filed against cognee while the recall-coverage code may belong in cognee-internal-plugins. Will re-file where the module actually lives.

dexters1:
@LStromann closing PR as a run can take days to finish and that is perfectly normal and expected especially with local models

LStromann:
@dexters1, but stuck runs never self-heal and users cannot restart the run

dexters1:
> @dexters1, but stuck runs never self-heal and users cannot restart the run
> 
> 

Fixed in v1.6.0 try it out @LStromann

PR #5085, opened 2026-09-16
fix: Make only_context return the full LLM input

siillee:
## Description

`only_context=True` promises "what the LLM would have received", but it returned only the
bare retrieval context. The real completion also sends the session guidance block, the
conversation history, and the question and context rendered through the retriever's
templates. #4654 added the missing layers behind a `context_format="prompt"` knob that
returned a dict, so the fix was opt-in and changed the type when opted in.

This PR removes `context_format` (shipped in 1.5.4, unused) and makes `only_context`
return one string for completion types: the system prompt and the rendered user prompt
under `=== SYSTEM PROMPT ===` / `=== USER PROMPT ===` lines, assembled by the same
`build_session_prompt` and `build_completion_prompts` the completion uses. No LLM call,
no session write; one embedding call for the history recall, shared across the fan-out.

- Retrieval-only types (CHUNKS, SUMMARIES, CODE, ...) and the Cypher/Agentic opt-outs keep
  returning the bare context. An empty retrieval returns the empty context, never a prompt
  around nothing.
- `verbose=True` exposes the string as `prompt_result` next to the bare `context_result`.
- `@agent_memory(memory_only_context=True)` reads `context_result`, so cognee's answer
  instructions never land in an agent's prompt.
- Fixed along the way: the cloud client never sent `session_id` on search; recall counted
  an empty-string context as a hit, so the `on_empty` tools fallback did not fire.

Behaviour change for existing `only_context` callers on completion types: the type is
unchanged but the content is now the full prompt. Callers that need the bare context use
`verbose=True` and `context_result`.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation


cognee/modules/agent_memory/runtime.py
```diff
@@ -407,15 +407,25 @@ async def retrieve_cognee_memory_context(context: AgentMemoryContext) -> str:
...
                 dataset_ids=[context.scope.dataset_id],
                 system_prompt=context.config.memory_system_prompt,
                 top_k=context.config.memory_top_k,
-                only_context=context.config.memory_only_context,
+                only_context=only_context,
+                # An only_context search returns the full LLM input — cognee's own answer
+                # instructions included — which must not be pasted into the agent's
+                # prompt as "memory". The verbose payload keeps the bare retrieval
+                # context under context_result, so read that instead.
+                verbose=only_context,
```
NMZivkovic, line 423: This is the right fix, and it is the one the Claude Code plugin also needs but cannot get: it lives in `cognee-integrations`, calls us over HTTP at `/api/v1/recall` with `only_context: true` and `search_type: HYBRID_COMPLETION` (`session-context-lookup.py:255`), and never sends `verbose`.

So after this PR it receives the full LLM input — cognee's answer instructions included — and injects it straight into `additionalContext` for every Claude Code prompt. Same hazard this comment describes, one repo over.

Worth referencing that consumer here, and landing the matching `verbose: true` + `context_result` change in the plugin before this ships.

cognee/modules/retrieval/only_context_prompt.py
```diff
@@ -79,14 +85,35 @@ def render_context_for_prompt(context: Any) -> Any:
...
+    if context is None:
+        return False
+    if isinstance(context, str):
+        return bool(context.strip())
+    if isinstance(context, (list, tuple)):
+        return any(has_context(entry) for entry in context)
+    return True
+
+
+def render_llm_input(system_prompt: str, user_prompt: str) -> str:
```
NMZivkovic: Ordering has a consequence worth a comment: system prompt first means any consumer that truncates the string keeps the boilerplate and drops the retrieved content. The Claude Code plugin truncates at 1500 chars (`TRUNCATE_GRAPH_CTX`), so in practice it would inject the task instructions and cut the memory.

Not arguing for a different order — system-then-user is the order the messages are sent in, and this is documented as a rendering rather than a protocol. But a line here saying "truncating consumers lose the context end first" would have saved me the trace.

NMZivkovic (commented):
The design here is right, and the parts that are easy to get wrong are the parts that are done well: the string is assembled through the *same* `build_session_prompt` / `build_completion_prompts` the real completion uses, so it cannot drift from the real call; `stamp_served=False` plus no QA turn keeps it genuinely read-only; `SharedSessionHistory` makes the one billed embedding call once per search instead of once per dataset; and `has_context()` correctly refuses to wrap a prompt around an empty retrieval, which is what keeps the `on_empty` tools fallback honest. The module docstring's "knowingly unfaithful in one place" paragraph is the kind of thing I want to find in this code a year from now.

I also checked the `context_format` removal rather than take it on trust: it appears nowhere in cloud-backend, cognee-saas, cognee-integrations, cognee-internal-plugins, infra or qira-demo-app. Retiring it as *ignored rather than rejected* is the graceful call — a stale caller sending `context_format="prompt"` now simply gets what it was asking for.

One blocker, and it is outside this repo.

## Blocker: this silently breaks the Claude Code plugin

`cognee-integrations/integrations/claude-code/` is a live `only_context` consumer, and it asks for exactly the combination this PR changes:

- `session-context-lookup.py:255` — the `graph` scope is `(["graph"], "HYBRID_COMPLETION", None)`
- `_recall_http.py:117` and `_plugin_common.py:2120` — `only_context: True`, and `verbose` is never sent
- `session-context-lookup.py:85` — reads `entry.get("content") or entry.get("text")`, truncates to `TRUNCATE_GRAPH_CTX = 1500`, and injects the result as `hookSpecificOutput.additionalContext`

I confirmed `retriever_sends_one_prompt(HybridRetriever())` is `True` (against `False` for `ChunksRetriever`), so that path now returns the prompt rather than the context. Rendering it shows what would be injected into every Claude Code prompt:

```
=== SYSTEM PROMPT ===
Answer the question using the provided context. Be as brief as possible.

=== USER PROMPT ===
The question is: `what did we decide about the egress spike?`
Answer using this sectioned context. Keep the answer brief an…
```

Because `render_llm_input` puts the system prompt first, the 1500-char truncation now spends its front on cognee's own answer instructions and cuts the retrieved memory off the end. This is precisely the hazard this PR already fixes for `@agent_memory` — but that fix is `verbose=True` plus reading `context_result`, which a separate repo talking to us over HTTP cannot inherit, and the plugin never sends `verbose`.

Three ways out, in my order of preference:

1. Land a `cognee-integrations` change first that sends `verbose: true` and reads `context_result`, and note the version dependency here. It is a handful of lines and it mirrors the fix already in this PR.
2. Keep the bare context as the default and put the full prompt behind an explicit opt-in — which is roughly what `context_format` already was.
3. Accept it and treat the plugin update as a release-blocking companion.

## Smaller

"Behaviour change for existing `only_context` callers" in the description understates the risk: the type is unchanged, so this lands silently at runtime rather than failing loudly at the call site. Worth naming the plugin explicitly so whoever cuts the release knows there is a companion.

Nothing else blocking — once the consumer story is settled this is good to go.

NMZivkovic (approved):

NMZivkovic (dismissed):

PR #5090, opened 2026-09-16
SDK-764 fix(pipelines): close abandoned pipeline runs as ERRORED at startup

dexters1:
## Summary

A pipeline run whose process dies leaves its newest `pipeline_runs` row at `STARTED`. Startup recovery on dev found such rows for cognify only, rolled their graph work back, and reset the dataset's status to `INITIATED`, which reopened the run gate but erased the run from the activity feed.

This PR makes startup recovery cover every pipeline and close the run instead of erasing it.

## Behaviour

- **Every run of every pipeline.** Candidates are the runs whose newest row is still `STARTED`, keyed by `pipeline_run_id`, across all pipeline names. A run left `STARTED` by add, memify, code graph, migration or a custom pipeline is closed the same way a cognify run is, and an abandoned run is still found when a newer run of the same pipeline on the same dataset completed after it (a newest-row-per-dataset query would hide it forever; see the review thread).
- **Only runs without a terminal row are candidates.** A run whose newest row is already `ERRORED` stays `ERRORED` (its inline rollback already ran when it failed); `COMPLETED` and `INITIATED` rows are not touched.
- **Rollback per pipeline.** `ROLLBACK_HANDLERS` in the recovery module maps a pipeline name to the rollback it supplies for its own failed runs, the same policy `run_tasks` applies inline. cognify is the only entry today; pipelines without one are closed without a rollback, exactly as they are when they fail while running. Rollback runs before the ERRORED row is written; if it raises, no row is written and the run stays `STARTED` so the next boot retries it.
- **The ERRORED row describes the run that died.** It carries the STARTED row's user, tenant, start time, input summary, origin and parent operation, plus the new `AbandonedPipelineRunError` as its error, so `error_class` tells readers the run was abandoned rather than failed.
- **Startup only.** No periodic sweep and no liveness inference inside a running server. A booting process has no runs of its own in flight, which is what makes the pass safe. The existing one-hour floor (`COGNEE_STALE_RUN_RECOVERY_MIN_AGE_SECONDS`, unchanged from dev) still leaves a STARTED row younger than an hour alone, in case a sibling replica is running it during a rolling deploy.
- No new status value, so no enum migration.

## Changes

- `cognee/modules/cognify/recovery.py`: rewritten as above; entry point renamed to `recover_stale_pipeline_runs_on_startup` (call site in `cognee/api/client.py` updated).
- `cognee/modules/pipelines/methods/get_pipeline_run_by_dataset.py`: `get_unterminated_pipeline_runs`, one window query partitioned by `pipeline_run_id` returning each run's newest row when it is `STARTED`; rows without a pipeline name (operation records) are excluded. The existing per-pipeline helpers are unchanged.
- `cognee/modules/pipelines/operations/log_pipeline_run_error.py`: optional `data_info`, `origin`, `parent_operation_id`; defaults are the context-derived values every other caller already gets.
- `cognee/modules/pipelines/exceptions`: `AbandonedPipelineRunError`.

## Tests

- Unit (`cognee/tests/unit/modules/cognify/test_recovery.py`, 11 cases): cognify is rolled back then closed; the ERRORED call carries user, tenant, start time, input summary, origin and parent operation from the STARTED row; a pipeline without a rollback is only closed; three pipelines in one pass; ERRORED, COMPLETED and INITIATED rows untouched; user-less rows; failed rollback writes nothing; missing dataset skipped; a minutes-old STARTED row is treated as live; one failure does not stop the others.
- Integration (`cognee/tests/integration/tasks/test_cognify_rollback_recovery.py`): cognify run rolled back and its newest row `ERRORED` with `error_class == "AbandonedPipelineRunError"`; a stale `add_pipeline` row is closed with the user, start time and input summary copied, and a second recovery pass leaves it alone; an abandoned run buried under a newer completed run of the same pipeline on the same dataset is closed while the completed run is untouched. Run locally with fastembed embeddings, no LLM.
- Also green: exception contract suite, background-tasks suite, run_tasks rollback suite, parent-run chain suite, and the pipelines and activity-router unit suites (114 tests) that share the latest-run query.

Related: #4983 covers similar ground with origin stamping and a periodic sweep; this is the startup-only approach.

https://claude.ai/code/session_01XeGaj3RcudyfVVRdhhdU94


cognee/modules/pipelines/methods/get_pipeline_run_by_dataset.py
```diff
@@ -58,3 +66,16 @@ async def get_latest_pipeline_runs_by_datasets(
         runs = (await session.execute(query)).scalars().all()
 
     return {run.dataset_id: run for run in runs}
+
+
+async def get_latest_pipeline_runs_for_all_pipelines() -> list[PipelineRun]:
```
goran-radonic: This keys recovery off the newest row per (dataset, pipeline) instead of per run, and that reopens the bug this is supposed to fix. add() writes a fresh INITIATED row for cognify_pipeline before run_pipeline takes the lock (add.py:319). So if add() runs while a cognify is still going, its STARTED row gets buried under that INITIATED row and this query never sees it again, on any future boot. Tested it locally against the real reset, happens exactly like that. Same problem if two runs of one pipeline get abandoned back to back, only the newer one ever gets picked up.

Needs to key off runs with no terminal row of their own instead of latest-per-pair, otherwise this doesn't cover every pipeline the way the PR claims.

dexters1: Agreed, and fixed in 78147d559. The INITIATED row from add() is gone since #5091 merged, but the mechanism you describe stands without it: run ids are a fresh uuid4 per execution, so any newer run of the same pipeline on the same dataset hides an older abandoned one from a newest-per-pair query. The realistic single-process case is a run under the one-hour floor at boot, skipped as possibly live, then a later run completes and the abandoned one is never looked at again.

Selection is now `get_unterminated_pipeline_runs`: newest row per `pipeline_run_id`, kept when that row is STARTED. The per-pipeline helper is back to what dev has. New integration test inserts an abandoned run and a newer completed run of the same pipeline on the same dataset and checks the abandoned one is closed and the completed one untouched.

PR #5092, opened 2026-09-16
SDK-562 feat(code-graph): install enola from PyPI through the codegraph extra

dexters1:
## Summary

enola, the code-graph extractor, was only ever obtained by a runtime download: on first use cognee fetched the pinned release from GitHub into `~/.cognee/bin`. Upstream now publishes the same binary as platform wheels on PyPI under the name `enola-cli` (`enola` was taken). This PR makes that the only route: `enola-cli` ships through the `codegraph` extra, the runtime downloader is removed, and without the extra the CODE route raises `EnolaNotInstalledError` telling the user to install `cognee[codegraph]`.

Why an extra and not a base dependency: PyPI has `enola-cli` wheels only for macOS arm64/x86_64, Linux x86_64/aarch64 (glibc) and Windows amd64, with no sdist, and the Linux binaries are glibc-linked (verified on Alpine 3.20: `pip install enola-cli` finds no distribution, and the release binary only runs with `gcompat`). PEP 508 markers cannot distinguish musl from glibc, so a base dependency would make `pip install cognee` fail on Alpine. Once upstream publishes a musllinux wheel, enola can move to the base dependency set.

## Publisher verification

- PyPI holds a PEP 740 provenance attestation for every `enola-cli` wheel. For 0.4.20 (and the same for 0.4.19) it names the publisher as GitHub repository `enola-labs/enola`, workflow `release.yml`, environment `pypi`, via Trusted Publishing. That is the same repository whose GitHub releases the auto-installer downloads today, and the same tag-push workflow builds both the release tarballs and the wheels.
- Upstream's `release.yml` publishes with `pypa/gh-action-pypi-publish` under an OIDC `id-token` permission; no long-lived API token.
- The binary inside the wheel is not byte-identical to the one in the release tarball for the same version. Upstream's workflow says why: the wheel gets a "pip-stamped" build so they can tell install channels apart in their usage counters. Both report the same version and produce the same snapshot on cognee's fixture (23 nodes, 54 edges either way).

## Changes

- `pyproject.toml`: `enola-cli==0.4.19` in the `codegraph` extra. This is the only enola version pin. `uv.lock`: the package added, nothing else changes.
- `cognee/tasks/code_graph/enola.py`: `find_enola_binary` now checks the interpreter's scripts directory (`sysconfig.get_path("scripts")`) before PATH. The wheel installs the binary there, and that directory is not on PATH when Python is run by path, which is how container entrypoints and service units start it. Order is still `ENOLA_PATH`, then scripts dir, then PATH.
- `cognee/tasks/code_graph/install_enola.py` and its unit tests: removed. `run_enola_generate` no longer downloads anything; a missing binary raises `EnolaNotInstalledError` with the hint `The code graph feature needs the `enola-cli` package. Install it with: pip install "cognee[codegraph]"`, the same shape as the GLiNER missing-package error. `ENOLA_PATH` still overrides the lookup. `ENOLA_AUTO_INSTALL` is gone.
- `CLAUDE.md`: the enola note now describes the extra.
- Tests: two new unit tests for the lookup order; the existing test that simulates "binary missing" by emptying PATH now also hides the scripts directory; the e2e test reads the pinned version from the installed `enola-cli` wheel metadata; end-to-end known answers updated for enola's output change since 0.4.14 (see below).

## Why 0.4.19 and not 0.4.20

The lockfile's `exclude-newer = "2 days"` window still covered 0.4.20 (published 15 September 08:19 UTC) when this was written, so `uv lock` refused it. 0.4.20 clears the window on 17 September; bumping is a two-line change plus re-pinned checksums.

## Verification

- `pip install cognee[codegraph]` puts `enola` 0.4.19 in the environment's bin; `find_enola_binary()` returns that path from a bare interpreter with the venv **not** on PATH.
- End-to-end (`cognee/tests/test_code_graph_e2e.py`) with no `ENOLA_PATH` and the CI environment (`ENV=dev`, default access control): passes in full. The log shows enola running from `.venv/bin/enola`, snapshot provenance `version=0.4.19 format_version=1`; typed graph, known symbols, modules, dependencies, calls, imports, the `SearchType.CODE` known-answer queries through `search()` and `recall()`, idempotent re-ingest, the `add(<dir>) + cognify()` CODE_REPO route and `forget(everything=True)` all pass. Final line: `enola 0.4.19; graph: 23 nodes, 54 edges`.
- Unit: `cognee/tests/unit/tasks/code_graph`: 162 passed. One failure, `test_code_repo.py::test_directory_with_project_resolves_to_repo_item_plus_documents`, fails identically on dev and does not involve enola.

## One upstream change, followed here

enola's output changed between 0.4.12 and 0.4.14: the resolved package import moved to a module-level `module-edge:` dependency fact, and file-level import facts now carry the imported file path with no resolved module id. The e2e known answers follow that. Whether cognee should also resolve file-level imports itself is a separate question for the code-graph owners.

## Docker and CI

- The Docker image's two `uv sync` lines include `--extra codegraph`, so the image gets enola from the wheel at build time instead of downloading it into the container's home directory on first use.
- The code-graph e2e job in `e2e_tests.yml` installs the `codegraph` extra, so CI proves the wheel route, the one users and the image get. The enola binary cache step is removed with it.

https://claude.ai/code/session_01XeGaj3RcudyfVVRdhhdU94


Vasilije1990:
It needs to be a part of the main package or?
It is a base functionality

dexters1:
> It needs to be a part of the main package or? It is a base functionality

Moving it to default dependency, but we will lose support for some platforms:

PyPI has enola wheels for five platforms and no source distribution. On musl-based images like Alpine, or Windows ARM, pip install cognee will now fail to resolve rather than just lacking the code graph.

PR #5101, opened 2026-09-17
fix: fail recall request when any dataset name doesn't exist

siillee:
## Description

Description: dataset names were resolved by a membership filter, so a name matching no dataset was silently dropped from the search scope. A recall over ["real", "typo"] answered 200 from real alone, with nothing in the response saying typo was never searched — while the same request by UUID has always failed outright. Names now get the same all-or-nothing contract: any name that does not resolve raises DatasetNotFoundError (404) naming the misses, before any search or warm-up probe runs.

404 rather than the id path's 403 because the two misses differ: an unknown id may belong to another tenant and can only be answered with "denied", while names resolve solely within the caller's own namespace, so the failure can name them. It also matches the existing error for a request in which every name missed.

Scoped to the read paths (search(), recall's graph and code lanes). get_authorized_existing_datasets is unchanged — add() and the memify pipelines rely on unmatched names resolving to nothing so they can create the dataset.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation


dexters1:
Change was handled in this PR instead:

https://github.com/topoteretes/cognee/pull/5109/changes

PR #5108, opened 2026-09-17
fix: don't show llm key error when gliner is used for file uploads (SDK-748)

siillee:
## Description

Ingesting a file with no LLM_API_KEY failed with LLMAPIKeyNotSetError (422) before ingestion started, even on a keyless setup where the graph is extracted with the local GLiNER model and embedded locally. add() marks its pipeline as needing the LLM for any input that is not a bare string — a conservative guess, because the loader is not resolved yet and the media loaders do call the LLM — and that guess triggered the first-run connection probe. Inline text worked; the same content as a .txt upload did not.

The probe exists to catch a key that is set but broken. With no key configured it can only restate what the config already says, so it is now skipped in that case, along with the provider-config warning that claims extraction needs an LLM (untrue for a keyless run). Behaviour with a key configured is unchanged.

Media is the one ingestion step that still needs a key, so each media loader now says so itself: require_llm_for_media raises before the loader reads or writes anything, naming the file kind, what the LLM was needed for, and that text, document and code files need no key.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation


cognee/api/v1/add/add.py
```diff
@@ -32,7 +32,30 @@
...
+    (``require_llm_for_media``) — a moment later than the probe, naming the
+    file kind and what still works without a key.
+
+    *Otherwise, only known plain-text inputs can safely skip the check*: a file
+    whose loader is not resolved yet could still be an image or a recording,
+    whose loaders transcribe through the LLM.
+    """
+    from cognee.modules.preflight import llm_available
+
+    if not llm_available():
```
dexters1: Image

Check if this function would be a better fit here, so we only skip LLM checks here if and only if Gliner is used and the rest should stay as is in terms of the checks

cognee/api/v1/add/add.py
```diff
@@ -32,7 +32,43 @@
...
+    whose loader is not resolved yet could still be an image or a recording,
+    whose loaders transcribe through the LLM.
+    """
+    from cognee.modules.cognify.config import (
+        GLINER_DEMO_EXTRACTOR,
+        get_cognify_config,
+        resolve_extractor,
+    )
+
+    if resolve_extractor(None, get_cognify_config()) == GLINER_DEMO_EXTRACTOR:
```
NMZivkovic: Two problems with using `resolve_extractor` as the predicate here.

**It raises.** On `auto` (the default) with no key configured, `resolve_extractor` raises `KeylessExtractorNotInstalledError` when `gliner2` is missing, and `gliner2` ships only in the `gliner` extra. So on a plain `pip install cognee` with no `LLM_API_KEY`, every `add()` now fails before ingesting anything, `add("plain text")` included, which returned `False` here and worked before. `KeylessExtractorNotInstalledError` is a `CogneeConfigurationError`, so `POST /add` answers **500** where the old path gave the 422 you describe. The docstring argues this is the right answer because nothing downstream could build a graph, but `add()` on its own is a documented staging step (custom pipelines, `cognee-cli add`, DLT ingestion), and `remember()` already made the opposite call for the same reason: its code route skips `resolve_extractor` with the comment that keyless installs without gliner2 would otherwise fail the extractor gate for a pipeline that never uses it. The add pipeline never uses the extractor either. CI will not show this, because `COGNEE_SKIP_PREFLIGHT` / `MOCK_EMBEDDING` make `keyless_local_defaults_apply()` false, so `auto` resolves to `llm` and never reaches the raise.

**It skips the probe even when a key is configured.** With `GRAPH_EXTRACTOR=gliner_demo` pinned and a key that is set but broken, this returns `False`, so the first-run connection probe never runs, and that probe is the one thing that catches a broken key. A media upload then passes `require_llm_for_media` (a key is set, so `llm_available()` is true) and fails deep in the vision or transcription call instead.

Both point the same way: resolve defensively, and ask whether an LLM is configured at all.

```python
try:
    extractor = resolve_extractor(None, cognify_config)
except KeylessExtractorNotInstalledError:
    return False  # no key and no local extractor: cognify() reports that, not add()
if extractor == GLINER_DEMO_EXTRACTOR and not llm_available():
    return False
```

That keeps dexters1's "only when GLiNER is used" condition, keeps the probe when a key exists, keeps `GRAPH_EXTRACTOR=llm` with no key failing fast, and never 500s a keyless bare install. It also stops `add()` from emitting the enterprise GLiNER notice that `resolve_extractor` logs once per process for users who never cognify.

cognee/api/v1/add/add.py
```diff
@@ -32,7 +32,43 @@
...
+    cognify argument and never reaches ``add()`` — and raises
+    ``KeylessExtractorNotInstalledError`` when the keyless path is selected
+    without ``gliner2`` installed, which is the right answer at ingestion time:
+    nothing downstream could build a graph.
+
+    *Otherwise, only known plain-text inputs can safely skip the check*: a file
+    whose loader is not resolved yet could still be an image or a recording,
+    whose loaders transcribe through the LLM.
+    """
+    from cognee.modules.cognify.config import (
```
NMZivkovic: `default_pipeline_needs_llm(extractor, config)` already answers this question, and its docstring says it serves the early provider preflight in `add()`/`remember()`; `remember()` calls it right next to `validate_provider_config`. Importing `GLINER_DEMO_EXTRACTOR` and comparing here gives `add()` a second, narrower rule, and the two disagree in a real case.

`default_pipeline_needs_llm` is `extractor == LLM_EXTRACTOR or config.contradiction_detection`, and contradiction detection is an LLM task on the GLiNER path too: `cognify()` passes `check_contradictions=cognify_config.contradiction_detection` into `get_gliner_demo_tasks`, which appends `Task(detect_contradictions, ...)`, and that task list's own docstring says optional contradiction detection still uses the LLM. So with GLiNER plus `CONTRADICTION_DETECTION=true`, `remember()` still validates and probes while `add()` now skips both, for a pipeline that will make LLM calls, and the opaque mid-cognify auth error this preflight exists to prevent comes back.

Using `default_pipeline_needs_llm(extractor, cognify_config)` instead of the `== GLINER_DEMO_EXTRACTOR` comparison keeps one authority for the question and makes the two entry points agree.

cognee/infrastructure/loaders/core/image_loader.py
```diff
@@ -109,6 +110,10 @@ async def load(self, file_path: str, **kwargs: Any) -> "str | LoaderResult":
         if not os.path.exists(file_path):
             raise FileNotFoundError(f"File not found: {file_path}")
 
+        require_llm_for_media(
```
NMZivkovic, line 113: This is the right place for the message, but the cost is worth naming in the description. By the time a loader raises, `add()` has already run `setup()` and resolved (creating if needed) the dataset, and `ingest_data` has copied the file into cognee storage: `save_data_item_to_storage_detailed` runs before `data_item_to_text_file`. So a keyless media upload now leaves a stored copy and an errored `add_pipeline` run behind, where before it failed before any of that existed. The derived text is what the guard protects, and the tests cover that; the raw copy is not cleaned up.

If that matters, checking the known media extensions in `_add_pipeline_needs_llm` when no LLM is configured would keep media failing before the pipeline starts, with the same message. Otherwise I would just say so in the description, since it changes what a failed keyless upload leaves in the dataset.

cognee/tests/unit/infrastructure/loaders/test_media_loaders_require_llm.py
```diff
@@ -0,0 +1,84 @@
...
+    loader_class, filename, media, tmp_path, no_llm
+):
+    """The key check must not mask a plain typo in the path."""
+    with pytest.raises(FileNotFoundError):
+        await loader_class().load(str(tmp_path / f"absent-{filename}"))
+
+
+@pytest.mark.parametrize(("loader_class", "filename", "media"), MEDIA)
+@pytest.mark.asyncio
+async def test_nothing_is_written_before_the_guard(loader_class, filename, media, tmp_path, no_llm):
```
NMZivkovic: Nit: this one cannot fail. The loaders write their derived text through `store_derived_text` into cognee's storage root, never next to the input, so `tmp_path` is unchanged whether the guard runs before or after that write. Patching `store_derived_text` (or `LLMGateway`) and asserting it was never awaited would pin what the docstring claims.

NMZivkovic (commented):
Good fix, and the right shape. The three media loaders really are the only loaders that call the LLM (nothing else under `cognee/infrastructure/loaders/` touches `LLMGateway`), the guard sits after the existence check so a typo still reports `FileNotFoundError`, and the message names the file kind plus what still works without a key. It also does what dexters1 asked for: only the GLiNER case skips, the rest of the rule is untouched. Both new test files pass here (22 passed), and the branch merges cleanly into today's dev.

Three things I would change, all in the new predicate:

1. **It raises, so a bare install now gets a 500 from every `add()`** — including plain text, which worked before this PR (inline).
2. **It keys on the extractor rather than on whether an LLM is configured**, so pinned GLiNER with a set-but-broken key also loses the probe that exists for exactly that case (same comment).
3. **It duplicates `default_pipeline_needs_llm` and disagrees with `remember()`** when contradiction detection is on (inline).

Also inline: what the loader-level guard leaves behind on a keyless media upload, and a nit on the test that checks nothing is written.

dexters1 (approved):

PR #5111, opened 2026-09-18
COG-6610 fix: Keep credential metadata on reconnect

goran-radonic:
## Description

While planning a Google Drive connector I went looking at how the existing integrations store their credentials, and found that `upsert_credential` assigns `workspace_id` and `provider_metadata` unconditionally. Any caller that leaves one of them out therefore clears whatever was already stored. `connect.py` is exactly such a caller, and it runs on every OAuth connect and reconnect.

The consequence is visible on Slack. `allowed_channel_ids` is written only through `update_provider_metadata` from the channels router, and it appears in no token response, so it cannot be rebuilt on reconnect. `handle_slack_command.py` reads an empty allowlist as "every channel allowed". A workspace that had deliberately restricted `/cognee-ask` to two channels quietly went back to all channels the next time somebody reconnected the app, with nothing in the logs saying so.

The two fields need different treatment, which is the part that took the longest to settle. Nobody passes `workspace_id`, so a sentinel is enough there: an omitted argument now means "leave it alone", while passing `None` explicitly still clears it. But `connect.py` does pass `provider_metadata`, filled from the fresh token response, so a sentinel would not help at all. That one has to merge into what is stored instead of replacing it, which is the same behaviour `update_provider_metadata` already had. Both changes stay inside `upsert_credential` so there is one write path rather than two that can drift.

Merging is safe for the other providers because all three adapters build their whole metadata dict on every install through `.get()` calls, so every key they own is present each time and gets overwritten. I checked that against responses with every optional field stripped out, including the GitHub and Linear keys that decide a dataset name.

I left `account_label`, `auth_type`, `scopes` and `token_expires_at` assigned unconditionally. Extending the same treatment to them is defensible but `auth_type` has a default that a new row genuinely needs, so it deserves its own change with its own tests. The docstring now states that boundary instead of leaving the next reader to guess.

## Acceptance Criteria

* A reconnect no longer clears a Slack workspace's channel allowlist, so a restricted workspace stays restricted instead of silently reopening.
* An omitted `workspace_id` leaves the stored value alone, while an explicit `None` still clears it.
* GitHub and Linear are unaffected: every metadata key they own is still refreshed on each install, including the ones that determine a dataset name.
* New tests cover the reconnect merge, both omission cases, the insert path, the fact that metadata cannot be cleared through this function, and the revoked-row takeover where a new owner inherits the previous allowlist.

Proof: 295 tests pass in `cognee/tests/unit/modules/integrations/`, of which four of the new ones go red against the previous implementation. The other two guard boundaries the new code could lose rather than old behaviour, confirmed by mutating the fix and watching each one fail alone.

## Type of Change

- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation


cognee/modules/integrations/credentials.py
```diff
@@ -103,11 +152,20 @@ async def upsert_credential(
...
         credential.user_id = user_id
-        credential.workspace_id = workspace_id
+        if not isinstance(workspace_id, _Unset):
+            credential.workspace_id = workspace_id
         credential.account_label = account_label
         credential.auth_type = auth_type
         credential.scopes = scopes
-        credential.provider_metadata = provider_metadata
+        if not isinstance(provider_metadata, _Unset):
+            # Best-effort under concurrency: this read-modify-write and
```
NMZivkovic, line 161: Accurate disclosure, and the race is strictly *better* than before — the old blind write clobbered whatever `update_provider_metadata` had just written, where the worst case now is losing one merge. So this is not a regression.

But it is already solved next door, including the mistake not to repeat. Cloud closed the same race under CLO-637 with `SELECT ... FOR UPDATE` inside the write's own transaction, after an earlier advisory-lock approach (`lock_account`, keyed on a sha256 of `provider:account_id` rather than `hash()`) deadlocked the pool under load by holding two or three connections per logical operation. See `cloud-backend/cognee-saas-backend/modules/integrations/credentials.py`.

Not asking you to grow this PR — a follow-up ticket pointing at that prior art is the right size.

cognee/modules/integrations/credentials.py
```diff
@@ -62,13 +78,46 @@ async def upsert_credential(
...
-    conflict key instead, while ``user_id`` still records who connected it.
+    ``workspace_id`` is optional. Omitted, it reproduces the original
+    single-user contract exactly: ``user_id`` is the owner, conflicts are
+    compared on it, and whatever the row already carries is left untouched.
+    Pass ``workspace_id`` when several users share one connection (a
+    cognee-hosted multi-user workspace, or a downstream layer's own tenant
+    concept) and it becomes the owner/conflict key instead, while ``user_id``
+    still records who connected it.
+
+    ``provider_metadata`` is **merged** into what is stored rather than
```
NMZivkovic, line 89: "Providers rebuild their whole metadata dict from each token response" is the load-bearing sentence of this design, and right now nothing enforces it — the day someone writes `if x: metadata["k"] = x` in an adapter, that key goes stale forever and no test notices.

Cheap guard, and it is the check I ran by hand to convince myself the claim is true today: for each adapter, assert `parse_installation` returns the same key set for a minimal token response as for a full one. It fails the moment a key becomes conditional, and it reads as the executable form of the rule you just wrote into `base.py`.

NMZivkovic (commented):
Good fix, and the bug behind it is worse than the description lets on. No blockers.

`handle_slack_command.py:52` reads an empty allowlist as unrestricted, so clearing it silently reopens a workspace that had deliberately narrowed `/cognee-ask` — a security-relevant regression with nothing in the logs. That part is exactly as you describe it.

## It is not only reconnects

`connect.py` is not the only caller that trips this. `slack/adapter.py:166` calls `upsert_credential` from the **automatic token-refresh path**, passing freshly-parsed metadata — so the allowlist was being wiped on an unattended refresh, with no human in the loop at all. And `handle_slack_link.py:119` omits both fields, so under the old code every member re-link cleared them too. Both are worth naming in the description; they make the case stronger than "somebody reconnected the app".

## What I checked rather than took on trust

**The claim the whole design rests on** — that all three adapters build their metadata dict unconditionally, so merging can never strand a stale key. I read all three (flat dict literals, unconditional `.get()`) and then exercised each `parse_installation` against a minimal token response versus a full one: Slack 3 keys, GitHub 5, Linear 5, **identical key sets both times**. It holds.

**The blast radius of the takeover case.** The only non-OAuth writer of metadata in core is `get_slack_channels_router.py:118`, writing `allowed_channel_ids` and nothing else — so the allowlist is the *only* key a new owner inherits. That makes your "restrictive, never widened" argument complete rather than merely plausible. And since `installed_by_slack_user_id` is adapter-owned it is correctly overwritten on takeover, so no stale authorization identity rides along. I would put both facts in the docstring; they are what make the inheritance defensible.

**The tests.** I reverted `credentials.py` to dev's version and ran the new file against it: **5 of the 7 new tests fail**, 17 pass restored. The description says four — you are underselling it. 295 pass in the full integrations suite, exactly as claimed. Ruff clean.

## The thing I would most like you to look at

Cloud already solved this, differently, and the two implementations are drifting apart. `cloud-backend/cognee-saas-backend/modules/integrations/` is an independent copy of this module that hit the identical bug and fixed it with caller-side selective preservation (`framework/selection.py:preserve_across_reconnect`), leaving `upsert_credential` replacing wholesale. Its docstring names your failure mode from the other side: *"any key this function does not explicitly preserve is gone."*

Both designs are defensible and they fail differently. Blanket merge protects any externally-written key for free but lets an adapter-owned key go stale forever if someone later makes it conditional. Selective preserve cannot go stale but silently reintroduces this exact bug the first time a new externally-written key is added and forgotten. With one such key in core today I think your call is right and simpler — but a line in the description acknowledging the sibling implementation would mean the two converge deliberately instead of drifting.

Inline: the race, and a cheap test for the invariant `base.py` now documents.

cognee/modules/integrations/base.py
```diff
@@ -102,6 +102,13 @@ def parse_installation(self, token_response: dict[str, Any]) -> OAuthInstallatio
         else (display/routing data such as a bot user id) goes in
         ``provider_metadata``, which is stored in the clear. Never put secret
         material in ``provider_metadata``.
+
+        ``provider_metadata`` is **merged** into whatever the row already
+        holds, not swapped for it, so that settings written outside the OAuth
+        flow (Slack's channel allowlist) survive a reconnect. Two rules follow
```
NMZivkovic, line 108: Right place for this — the contract belongs where someone writing a new adapter will actually meet it, not only in `credentials.py` where they have no reason to look. "Write a key with a null value rather than leaving it out" is the whole rule in one line.

Worth adding the consequence as well: an omitted key keeps whatever the previous install stored, so a conditional key is not merely stale, it is silently wrong for that tenant until the next write that happens to include it.

ArmagedonFlamer:
Good work guys! This seems already covered by #5163 so I'll close it.

PR #5117, opened 2026-09-18
fix(vector): Record the embedding model per dataset and check it once (SDK-756)

dexters1:
## Description

Alternative to #5116, targeting `dev` directly. Same goal — a clear error when the embedding model changes under a dataset — with the check moved from the vector adapters to the dataset registry, so it costs nothing on the happy path. **No schema change.** Reuses #5116's `EmbeddingDimensionMismatchError` (reshaped to dataset level); if this lands, #5116 can be closed.

**Why move it**

#5116 reads the collection schema (LanceDB) or the `pg_attribute` catalog (pgvector) before *every* `create_data_points` and *every* `search`. One `GRAPH_COMPLETION` query searches four collections, so that is four extra round trips per user question, on every deployment, forever, to catch a state that only exists right after someone changes `EMBEDDING_MODEL`.

The dataset's `dataset_database` row already describes its vector store in the `vector_database_connection_info` JSON (host, port, schema…). The embedding model and width go in there too, and are compared once when the dataset context is entered — against a row that is already loaded.

**What changes** (vs `dev`: 10 files, ~250 lines of code + 236 of tests)

| Where | What |
|---|---|
| `get_or_create_dataset_database` | records `embedding_model` / `embedding_dimensions` in the row's connection-info JSON, from the config that will embed this dataset |
| `_bind_dataset_databases` | after the dataset's vector config is bound: `ensure_embedding_model_matches(row, config)` — equal → pass; different → `EmbeddingDimensionMismatchError` (409). `config` is `get_embedding_context_config()`, the same resolver the vector engine uses (per-call ContextVar override, else global), so the record and the embedder cannot disagree |
| **rows from `dev`** (no keys) | completed **once** from the store: `get_stored_vector_size()` reads the width of the existing `vector` column (LanceDB: table schema; pgvector: `format_type` from the catalog), records it, then checks. Empty store → adopts the configured model. Providers without the reader → left alone |
| `_record_embedding_model` | re-reads the row and merges only the two keys — pgvector injects credentials into that dict after loading and they must never be persisted (tested) |
| `prune_system` | a vector prune keeps the rows, so it drops the two keys; nothing else in the JSON is touched |
| `EmbeddingDimensionMismatchError` | dataset-level: dataset id, both models, both widths; "an earlier embedding model (N dimensions)" when the builder is unknown |
| adapters | back to `dev` + one reader each (`get_stored_vector_size`, +18/+25 lines) |

**A remediation fix.** #5116's message suggests `forget(dataset=..., memory_only=True)` then `remember()`. That path deletes vectors **by id** and leaves the collections in place, so their width is unchanged and re-ingesting with the new model fails the same way. The new message says: set `EMBEDDING_MODEL` back, or delete the dataset (`forget(dataset=...)` / `cognee-cli forget`) and ingest again.

**Scope.** Only with `ENABLE_BACKEND_ACCESS_CONTROL=true` (the default), where datasets have registry rows. Single-tenant mode has one shared store and no row; unchanged from `dev`.

## Verification

Real stack (SQLite + LanceDB + Ladybug), one dataset ingested at 384 dims:

| Step | Row before | Result | Row after |
|---|---|---|---|
| ingest at 384 | — | works | `model='BAAI/bge-small-en-v1.5', dims=384` |
| **legacy row** (keys stripped), same model | no keys | search + add work | `dims=384` read from LanceDB, recorded |
| **legacy row**, switched to 3072 | no keys | `409: Dataset … was embedded with an earlier embedding model (384 dimensions), but the configured embedding model 'openai/text-embedding-3-large' produces 3072.` | `dims=384` recorded |
| recorded row, switched to 3072 | `dims=384` | same 409, no store call | unchanged |

On `dev` the switched search fails with `ValueError: Invalid input, No vector column found to match with the query vector dimension: 3072`, and `add` succeeds (fails later in cognify).

- `test_dataset_embedding_model_registry.py` (8): pass without touching the store / refuse with both models named / legacy row completed from store then checked / legacy over empty store adopts / provider without reader left alone / LanceDB reader / pgvector reader (parametrized) / real-registry round trip incl. prune clear and the no-credential-leak check
- `test_migration_model_lockstep.py`, `test_cognee_error_contract.py`, `test_dataset_database_provisioning_race.py`: pass
- `unit/infrastructure/databases`, `unit/modules/data`, `unit/api`, `unit/exceptions`, `unit/modules/pipelines`: 1795 passed (the one failure, `test_lancedb_schema_migration`, is pre-existing on `dev` in this environment)
- Pinned ruff 0.16.6 check + format clean on the whole tree

Note: the branch carries no Linear key yet, so `linear-issue-check` will stay red until one is attached to the title.


siillee:
One thing to fix before this lands: the check reads `get_embedding_context_config()`, but that is not what the engine embeds with. `get_embedding_engine()` applies `resolve_embedding_defaults(config, llm_config)` on top of it (`get_embedding_engine.py:28-30`), which reroutes embeddings to fastembed on a keyless install. Measured on `8d2f8ba`:

```
engine embeds with: ('fastembed', 'BAAI/bge-small-en-v1.5', 384)
check compares    : openai/text-embedding-3-large 3072
```

The 3072 is the stock default (`DEFAULT_EMBEDDING_MODEL`), not a user choice, so on a keyless install it is a width that nothing ever embeds with. That gives four cases:

| Registry row | State | Expected | Actual |
|---|---|---|---|
| legacy (dev-era, no keys) | still keyless | pass | **409 — false positive on a working deployment** |
| legacy (dev-era, no keys) | LLM key added | 409 | 409, but by coincidence — the inputs are identical to the row above |
| new (written by this PR) | still keyless | pass | pass (record says 3072 against a 384 store — wrong, but unused) |
| new (written by this PR) | LLM key added | **409** | **passes, then fails with the raw Arrow error** |

Row 1 is the regression: an existing keyless deployment that changed nothing gets a hard 409 on upgrade, on every operation for that dataset. Row 4 is the scenario named in the commit message ("a keyless install gaining an LLM key and leaving the local embedder") — recorded == configured == 3072 forever, so the check passes and the store error comes back unchanged.

Adding a key changes what `resolve_embedding_defaults` returns but not `config.embedding_dimensions`, which is the only thing the check reads — which is why rows 1 and 2 are indistinguishable to it.

Consequently the comment in `_bind_dataset_databases` ("the same resolver `get_embedding_engine` uses, so the recorded model and the check can never disagree with the embedder") does not hold, and the same claim is now repeated in `get_or_create_dataset_database`.

**Fix:** record and compare `resolve_embedding_defaults(config, get_llm_context_config())`, or better the bound engine's `embedding_engine.get_vector_size()` — that is by construction the width `create_collection` uses (`LanceDBAdapter.py:393`), so the record cannot drift from the store. All four rows then behave, and the message names the real models ("was embedded with `BAAI/bge-small-en-v1.5` (384 dimensions), but `openai/text-embedding-3-large` produces 3072") instead of "an earlier embedding model" versus a model the user never configured.

Worth one test that constructs a real `EmbeddingConfig` with no LLM key — the current suite passes `SimpleNamespace` stand-ins, which cannot catch this class of divergence.

siillee (approved):

PR #5118, opened 2026-09-18
fix(improve): Let the session bridge run without an LLM key (SDK-753)

dexters1:
Fixes [SDK-753](https://linear.app/cognee/issue/SDK-753/keyless-improve-session-bridge-fails-on-the-pipelines-llm-connection). Follow-up to SDK-691 (#5086): `remember()` works with no LLM key, but `remember(session_id=...)` did not, because the improve bridge behind it failed before running a single task. SDK-691 also asked for the LLM-dependent improve stages to be skipped with a reason when no LLM is configured; that is the second half of this PR.

## What was wrong

Found by running the keyless flow end to end on dev, in a clean environment built from `cognee[gliner]` with empty model caches and no key. Permanent `remember()` is fine: GLiNER extracts the graph, fastembed embeds, the automatic `improve()` skips every stage. With a `session_id` the entry lands in the session cache, then:

- the background bridge fails at once with `LLMAPIKeyNotSetError` (logged as a warning by `remember()`, so nothing surfaces to the caller);
- an explicit `improve(dataset, session_ids=[...])` raises the same error;
- session content never reaches the permanent graph.

The cause is not an LLM task. `Task` defaults to `needs_llm=True`, and the pipeline setup layer runs the LLM connection probe whenever any task in a pipeline claims to need one. The default memify tasks set `needs_llm=False`; the three improve-stage pipelines build their own tasks and never did:

| pipeline | tasks | LLM call? |
|---|---|---|
| `apply_feedback_weights` | `extract_feedback_qas`, `apply_feedback_weights` | no |
| `persist_sessions_in_knowledge_graph` | `extract_user_sessions`, `cognify_session` | no; `cognify_session` runs `cognify()`, which resolves its own extractor and its own probe |
| `persist_agent_trace_feedbacks_in_knowledge_graph` | `extract_agent_trace_feedbacks`, `cognify_agent_trace_feedback` | no |

The probe fired before any task ran, `persist_session_qa` is the one fatal stage, and the run stopped.

Beyond the bridge, three places still built an LLM client on a keyless install and let it raise: the stages that draft text (`extract_agent_context`, `distill_sessions`, `global_context_index`) ended `errored` with a traceback per session, the per-turn analysis on every `recall(session_id)` logged an error-level traceback before failing open, and the trace-step summary did the same on every `remember(TraceEntry)`.

## What changes

- **The bridge.** `needs_llm=False` on the six tasks above. Nothing else about them changes; with a key configured, `cognify()` inside stage 2 and 3 still runs its own probe.
- **The LLM-only stages decline instead of erroring.** `extract_agent_context`, `distill_sessions` and `global_context_index` gate on `llm_available()` and report `skipped: no_llm_configured`, the same way every other gate reports, with zero LLM calls. Earlier gates keep precedence (`opt_in_disabled`, `auto_feedback_disabled`).
- **No LLM client where there is no LLM.** The per-turn analysis and the trace-step summary check `llm_available()` first and take their existing fail-open path directly: empty analysis, deterministic trace summary, one debug line.
- **Keyless e2e.** `cognee/tests/e2e/keyless/keyless_ingest_check.py` gains the session path its TODO named: `remember(session_id)`, await the bridge, assert `persist_session_qa` completed and nothing errored, search the bridged content, explicit `improve()`, then a `TraceEntry` through the bridge with `persist_agent_traces` completed and the two LLM-only stages `skipped: no_llm_configured`. The check no longer forces `AUTO_FEEDBACK=false`; it runs on defaults, which is the point.

## Verification

Live, keyless, GLiNER demo + fastembed, models downloaded fresh into empty caches:

| step | before | after |
|---|---|---|
| `remember()` permanent | 8 entities, improve skipped every stage | same |
| `remember(session_id)` x2, bridge awaited | `feedback_weights` and `persist_session_qa` failed on `LLMAPIKeyNotSetError`, run stopped | both completed; graph 8 → 15 entities incl. Niels Bohr, Copenhagen, uncertainty principle |
| `remember(TraceEntry)`, then `improve()` | not reachable | `persist_agent_traces` completed (1 session), `extract_agent_context` and `distill_sessions` `skipped: no_llm_configured`, graph 15 → 16 entities, three documents |
| `recall(session_id)` | session hits + graph chunk, error traceback logged | same hits, no traceback |
| explicit `improve(session_ids)` | raised | completed |
| `LLMAPIKeyNotSetError` in the log | 6 | 0 |

The two concurrent bridges started by two `remember(session_id)` calls resolve as designed: the second loses the improve lock and the holder reruns for it, so both entries are bridged.

- `keyless_ingest_check.py` with the session and trace steps, on default settings: PASS in the GLiNER environment.
- Unit: `test_improve_stage_pipelines_need_no_llm.py` (each pipeline's tasks declare no LLM need, `pipeline_needs_llm` false), `test_llm_stage_gates.py` (the three stages decline without an LLM, run with one, earlier gates keep precedence, trace summary builds no client), and the keyless skip in `test_feedback_detection.py`.
- The existing turn-analysis tests mock the LLM call itself and count on the analysis running. CI has no key, so the new gate skipped it and two dispatch tests failed on the first push (one LLM call instead of two; the no-answer acknowledgement not produced). One autouse fixture in `cognee/tests/unit/conftest.py` now pins `llm_available` to true for unit tests, and the gate's own tests patch it back to false inside. The full unit suite passes with no LLM key in the environment, which is how CI runs it.

## Type of Change


cognee/infrastructure/session/feedback_detection.py
```diff
@@ -74,6 +75,12 @@ async def analyze_turn_for_session_context(
     """
     if not (user_message and str(user_message).strip()):
         return SessionTurnAnalysis()
+    if not llm_available():
+        # Keyless setups (the GLiNER + fastembed defaults) answer with CHUNKS
+        # and cannot analyze a turn; skip quietly instead of building an LLM
+        # client that raises and logs a traceback on every recall.
+        logger.debug("Feedback detection: no usable LLM configured, skipping turn analysis")
+        return SessionTurnAnalysis()
```
NMZivkovic, line 83: Not great not terrible

cognee/modules/improve/stages.py
```diff
@@ -181,6 +186,8 @@ def gate(self, inputs: ImproveRunInputs) -> str | None:
             return REASON_SESSION_MANAGER_UNAVAILABLE
         if not session_manager.is_auto_feedback_enabled():
             return REASON_AUTO_FEEDBACK_DISABLED
+        if not llm_available():
+            return REASON_NO_LLM_CONFIGURED
```
NMZivkovic, line 190: This gates the improve-side caller of the agent-context extraction. The **trace-write** side calls the same LLM drafting and is still ungated:

`SessionManager.add_agent_trace_step` → `_maybe_extract_agent_context` (`session_manager.py:266`) → `extract_pending_agent_context(...)` with `min_new_traces` defaulted to `TRACE_EXTRACTION_INTERVAL` (10) → `_extract_batch_from_traces` → `LLMGateway.acreate_structured_output` (`agent_context_extraction.py:253`). There is no `llm_available()` check on that path, and `extract_pending_agent_context`'s docstring is explicit that failures propagate to the caller — the caller here being the blanket `except Exception` that logs `Agent-context extraction skipped: %s` with `exc_info=True`.

So on a keyless install every 10th trace step still builds an LLM client and leaves a warning-level traceback in the log — one more of the entries the 6 → 0 row counts. It needs 10 trace steps to show up, and both the live verification and the new e2e store one, which is why it reads as clean.

`extract_live_agent_context` on the line above it is deterministic ("no-LLM candidates"), so that one genuinely needs nothing.

The fix that matches what you did for the other two operations is the check inside the operation — an early `if not llm_available(): return []` in `extract_pending_agent_context` — keeping this gate for the reporting. It also covers `distill_session`, which is exported from `cognee/api/v1/session/__init__.py` and so reachable without `DistillSessionsStage.gate` in the path at all.

cognee/tests/unit/conftest.py
```diff
@@ -46,3 +47,21 @@ async def _run():
     )
 
     create_relational_engine.cache_clear()
+
+
+@pytest.fixture(autouse=True)
+def _keyless_gates_see_a_usable_llm():
```
NMZivkovic, line 56: Worth recording that this fixture is load-bearing, not just convenient: `Unit Tests (Mocked, No Secrets)` in `community_tests.yml` sets no LLM key and none of `COGNEE_SKIP_PREFLIGHT` / `COGNEE_SKIP_CONNECTION_TEST` / `MOCK_EMBEDDING`, so `llm_available()` really is false there and without this the mocked suites would skip the calls they count.

My concern is the three hard-coded targets. There are four module-level `from cognee.modules.preflight import llm_available` sites now — these three plus `recall.py:653` — and the fourth is deliberately left alone, because the CHUNKS default *is* the behaviour under test in the recall suites. That distinction is only in your head and this docstring, so the next gate (session distillation, say) gets no patch, and the failure mode is quiet rather than loud: the test takes the skip branch and asserts against an empty `SessionTurnAnalysis()` or a `skipped` stage instead of failing.

Two ways to make it survive: call it module-qualified in the gates (`preflight.llm_available()`) so one `patch("cognee.modules.preflight.llm_available")` covers every present and future call site, or keep the list and add a small meta-test that walks the modules importing `llm_available` at module level and fails when one is neither pinned here nor in a declared exclusion list. Either way the drift becomes an error instead of a silently skipped assertion.

cognee/tests/e2e/keyless/keyless_ingest_check.py
```diff
@@ -113,6 +116,64 @@ async def main() -> None:
...
+            origin_function="search_docs",
+            status="success",
+            method_params={"query": "Bohr model"},
+            method_return_value={"hits": 3},
+            generate_feedback_with_llm=True,
+        ),
+        dataset_name="keyless_remember",
+        session_id="keyless_session",
+    )
+    await trace_result
```
NMZivkovic: One trace step is below `TRACE_EXTRACTION_INTERVAL`, so this exercises the improve-side gate but not the periodic extraction the trace-write path triggers (see my comment on `stages.py`). Storing `TRACE_EXTRACTION_INTERVAL` steps here instead of one would cover both, at the cost of ten cheap session writes.

While you're in this file: the PR's "`LLMAPIKeyNotSetError` in the log: 6 → 0" row is the real acceptance criterion and it's currently verified by hand. This check is keyless by construction, so attaching a handler to the root logger at the top and asserting at the end that nothing warned about a missing LLM key would turn that table row into the regression test — it is the thing that quietly comes back, since every one of these paths fails open.

NMZivkovic (commented):
Read the whole change and ran the new tests locally: `test_llm_stage_gates.py`, `test_improve_stage_pipelines_need_no_llm.py` and the feedback-detection addition pass, and so does the recall dispatch suite the first push broke. The shape is right — the check goes inside the operation for the two places that actually draft text, `needs_llm=False` for the six tasks that never touch an LLM, and a gate for the three LLM-only stages so they report a reason instead of a traceback.

What I checked beyond the description:

- none of the six tasks reaches an LLM; `cognify_session` / `cognify_agent_trace_feedback` delegate to `cognify()`, which probes on its own resolved extractor, so `needs_llm=False` on the outer task is right rather than merely convenient;
- no other stage is missing a gate — `update.py`, `build.py` and `get_triplet_datapoints.py` carry no LLM references at all;
- gate precedence is preserved in all three stages and covered by the new tests;
- the new e2e assertions really run in CI: `run-keyless-e2e` is the one job that pins `COGNEE_SKIP_CONNECTION_TEST: 'false'`, so the keyless rule is live there;
- `RememberResult.__await__` exists, so the `await session_result` the e2e now does is valid.

Two things I'd like your read on before merge, inline: the **trace-write** caller of the agent-context extraction is still ungated (the caller this PR gated is the improve-side one), and the new autouse fixture's three patch targets are a hand-maintained list that will drift.

Doc nit: three stages silently skipping on every keyless `improve()` is user-visible behaviour, and the LLM-free-mode bullet in CLAUDE.md is where someone would go looking for it — `no_llm_configured` isn't mentioned there. (The skip-reason list a few lines up ends in an ellipsis, so that one doesn't need touching.)

NMZivkovic (approved):

NMZivkovic (approved):

PR #5146, opened 2026-09-19
fix(mcp): honour runtime args and the real HTTP port in the MCP container

### LINKED ISSUE #5145: [Bug]: cognee-mcp: entrypoint.sh overrides runtime --transport and Dockerfile healthcheck ignores HTTP_PORT
### Bug Description

Two related defects in the cognee-mcp container packaging (cognee-mcp/entrypoint.sh and cognee-mcp/Dockerfile), both about the container not honouring the configuration it is given.

1. The entrypoint says it forwards any args passed to the container at runtime, but it builds the argument list as user args first, then appends --transport $TRANSPORT_MODE (and --host/--port/--api-url/--api-token) after them. argparse takes the last occurrence, so any --transport the user passes on the docker run command line is silently replaced by the env-derived default (stdio when TRANSPORT_MODE is unset). Because --host 0.0.0.0 --port are only added when TRANSPORT_MODE is not stdio, the user's HTTP transport also ends up with no bind args.
2. The HEALTHCHECK probes the literal URL http://localhost:8000/health. The entrypoint honours an HTTP_PORT env var and passes it as --port, so a container started with HTTP_PORT=9000 serves on 9000 while the health probe keeps hitting 8000 and the container is marked unhealthy.

### Steps to Reproduce

1 (transport override):
```
docker run --rm cognee/cognee-mcp:main --transport http
```
or, with a stub on PATH that echoes its argv, run `bash entrypoint.sh --transport http` and inspect the argv passed to `cognee-mcp`.

2 (healthcheck):
```
docker run -d --name mcp -e TRANSPORT_MODE=http -e HTTP_PORT=9000 -p 9000:9000 cognee/cognee-mcp:main
sleep 90
docker inspect --format '{{.State.Health.Status}}' mcp
curl -s http://localhost:9000/health
```

### Expected Behavior

1. A `--transport` (or `--api-url`/`--api-token`) passed at runtime wins over the env default, as the entrypoint's own comment promises. The server starts with the HTTP transport bound to `0.0.0.0` on `HTTP_PORT`.
2. The health probe targets the port the server was actually started on (`${HTTP_PORT:-8000}`); the container reports `healthy` and `curl` returns `{"status":"ok"}`.

### Actual Behavior

1. argv becomes `--transport http --transport stdio`; the server starts on stdio with no log line explaining that the user's argument was discarded.
2. The server answers on 9000 (`curl` returns `{"status":"ok"}`), but the probe fails against 8000 every 30 s and the container reports `unhealthy` after the retries.

### Environment

- OS: Ubuntu 26.04 LTS (kernel 7.0.0-29-generic), Docker 
- Python version: 3.12.14 (inside the image: whatever the Dockerfile pins)
- Cognee version: cognee-mcp 0.5.5 / main as of 2026-09-20 (both defects still present after #4994)
- LLM Provider: not involved
- Database: not involved

### Logs/Error Messages

```shell

```

### Additional Context

**Suggested fix**

- entrypoint.sh: build the env defaults first and append `"$@"` last, or only append `--transport` when `"$@"` does not already contain it.
- Dockerfile: `urllib.request.urlopen('http://localhost:' + os.environ.get('HTTP_PORT', '8000') + '/health', timeout=5)` (or `${HTTP_PORT:-8000}` inside the shell-form CMD).

### Pre-submission Checklist

Madin-H23:
## Related issue

Fixes #5145

## Summary

Two related defects in the cognee-mcp container packaging, both about the container not honouring the configuration it is given:

**1. Runtime flags silently replaced.** `entrypoint.sh` built `ARGS=("$@")` first and appended the env-derived flags after them — argparse keeps the **last** occurrence of a flag, so a runtime `--transport http` was replaced by the env default (`stdio`), and because `--host/--port` were only added when `TRANSPORT_MODE != stdio`, the HTTP transport also ended up with no bind args.

Restructured: env-derived defaults are built **first** (transport, `--host/--port` following the *effective* transport — a runtime `--transport http` gets bind defaults even when `TRANSPORT_MODE` is unset — plus `--api-url/--api-token`), and the runtime args are appended **last**, so they win while the defaults still fill anything not passed. The effective transport is derived by scanning the runtime args for `--transport`.

**2. HEALTHCHECK probed the wrong port and used a build-time literal.** The probe hit the hardcoded `http://localhost:8000/health` while the entrypoint binds `HTTP_PORT` (a container on 9000 answered fine but reported `unhealthy`), and the exec-form `${TRANSPORT_MODE:-stdio}` check substitutes at build time, so a runtime `-e TRANSPORT_MODE=stdio` could never skip the probe. The probe is now a single python one-liner that reads both `TRANSPORT_MODE` and `HTTP_PORT` from `os.environ` at container runtime.

## Verification

- `pytest cognee-mcp/tests/test_entrypoint_args.py` → **3 passed**: runs the real `entrypoint.sh` with a fake `cognee-mcp` on PATH that dumps its final argv
  - runtime `--transport http` (env unset) → last `--transport` is `http`, bind defaults `0.0.0.0:8000` present
  - runtime `--transport http --port 9000` with `TRANSPORT_MODE=http` → `http` on port `9000`
  - env `TRANSPORT_MODE=http`, no runtime args → argv unchanged from the previous behaviour (regression guard)
  - all three fail against the previous script (red-check via `git stash`)
- `bash -n entrypoint.sh` clean; `ruff check` / `ruff format --check` (0.16.6) clean on the new test file.

## Additional context

Manual end-to-end check of the argv assembly (`bash entrypoint.sh --transport http`) shows the launch command `cognee-mcp --transport http --host 0.0.0.0 --port 8000` as the issue's "Expected behavior" describes.


dexters1:
Thanks for the PR @Madin-H23! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

Madin-H23:
Done — rebased onto `dev` (branch now carries a single commit, no unrelated changes) and switched the PR base to `dev`. Thanks for the pointer!

Vasilije1990 (approved):

PR #5152, opened 2026-09-20
fix(chunking): give CsvChunker document-scoped chunk ids

joaquinhuigomez:
## Description

CsvChunker takes its chunk id straight from chunk_by_row, which derives it as uuid5 of the row text alone. Two identical rows in one CSV therefore get the same id — and both dedup passes (get_graph_from_model and add_data_points) then silently drop one of them, so a chunk of the document just disappears. The same row appearing in two different CSVs collapses to a single shared node, attributed to whichever document got there first.

chunk_id.py documents the intended identity — content_chunk_id(document_id, content_hash, occurrence), with the occurrence counter keeping equal texts in one document distinct — and TextChunker has used it since d73092675. CsvChunker was never migrated. This applies the same pattern: an occurrence dict in read(), the derived id, and content_hash / max_chunk_tokens stamped the way text chunks get them (which also moves CSV chunks onto the non-fallback branch in incremental.py).

One caveat worth stating: chunk ids for existing CSV documents change, so a re-ingest produces new nodes rather than matching old ones — the same one-time break the text-chunker migration took.

## Acceptance Criteria

* Two identical rows in one document produce two distinct chunk ids, and both survive deduplicate_nodes_and_edges.
* The same row content in two different documents produces different ids.
* Ids are derived as content_chunk_id(document_id, content_hash, occurrence) and are stable across re-runs.

All three are pinned by tests in cognee/tests/unit/modules/chunking/test_csv_chunker.py; all three fail on dev without the change.

## Type of Change
- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots

Local run of the chunking suites (text output rather than an image — happy to attach a capture if preferred):

```
$ pytest cognee/tests/unit/modules/chunking/
86 passed, 2 skipped, 12 warnings in 1.19s
```

## Pre-submission Checklist

## DCO Affirmation


dexters1:
Thanks for the PR @joaquinhuigomez! We're working through a big review backlog right now, so we can't give a date, but this is on our list and a maintainer will review it and respond here. We appreciate the contribution.

PR #5160, opened 2026-09-21
fix: return structured context evidence from HYBRID_COMPLETION (SDK-698)

vkovac2:
## Description

Closes SDK-698.

`get_retriever_output` fills `SearchResultPayload.evidence` from the retriever's `get_context_evidence(retrieved_objects, dataset_id)` hook when the caller passes `include_references=True`. `CompletionRetriever` (RAG) and `GraphCompletionRetriever` override it. `HybridRetriever` did not, so it inherited the base default that returns `[]`. A `HYBRID_COMPLETION` search with `include_references=True` therefore got the answer-grounded "Evidence:" text footer but an empty structured `evidence` list, and HYBRID_COMPLETION is the default search type.

This adds `hybrid_context_evidence` in `cognee/modules/retrieval/utils/evidence.py`, beside `chunk_context_evidence` and `graph_context_evidence`, and a three-line override on `HybridRetriever` that delegates to it. It describes exactly what the hybrid prompt rendered:

- `chunks` become `segment` items, reusing `chunk_context_evidence` unchanged.
- each entity block becomes a `graph_node` labelled with the entity name.
- each entity edge bullet becomes a `graph_edge` between its endpoints, and the endpoints become `graph_node` items, as the graph builder does for every edge it renders. The id is the `edge_object_id` the graph stamped, else the deterministic id `graph_context_evidence` derives with `generate_edge_object_id`. A bullet missing an endpoint or its relationship is not cited rather than given an invented id.
- `facts` are not cited. They are `EdgeType` vector rows keyed by relationship *type*, not graph edges, so they carry no stable identity to point at. `extract_context_object_ids` excludes them for the same reason.
- `chunk_summaries` are not rendered as passages and are not cited.
- batch results (one dict per query) flatten in order; every `(kind, artifact_id)` appears once; ranks run chunks, then nodes, then edges.

Nothing else changes: the hybrid lanes, the text footer (`cite_hybrid_completions`), `get_retriever_output`, `base_retriever`, and the search factory are untouched. The hook is synchronous and does no graph or vector I/O. With `include_references=False` it is never called, so the default path is byte-identical.

One thing for review: the assembly tail of `hybrid_context_evidence` (accumulators, edge-id derivation, and the node-then-edge reference build) is close to a copy of the same block in `graph_context_evidence`. Only the traversal genuinely differs — graph reads `Edge` objects, hybrid reads dicts. Extracting a shared assembly helper would drop roughly 25 lines from each, but that means editing the existing graph builder, which this ticket's scope puts outside the change (the new function is specified as a sibling, added *beside* the existing ones). Left as-is deliberately; happy to extract it if you'd rather.

One consequence to be aware of: once hybrid emits `graph_edge` items, a dataset-scoped search with the flag on reaches the existing `graph_source_evidence` sidecar lookup in `get_retriever_output`, exactly as graph completion does today, and its `append_source_evidence_text` block. A hybrid answer can then carry two "Evidence:" blocks: the existing answer-grounded chunk block and the sidecar source block. They cite different things. Merging them is left as a follow-up if wanted.

## Acceptance Criteria

- [x] `HybridRetriever().get_context_evidence(result, dataset_id)` returns `segment` items for chunks, `graph_node` items for entities and edge endpoints, and `graph_edge` items for bullets, with the same field conventions as the RAG and graph builders. Tests: `test_context_evidence.py` (9 new cases), `hybrid_retriever_test.py::test_context_evidence_describes_what_the_retriever_actually_rendered`, which drives the real entity lane, and `test_get_retriever_output.py::test_get_retriever_output_returns_structured_evidence_for_hybrid_completion`, which goes through the public consumer with the real `HybridRetriever` and asserts the evidence on `SearchResultPayload`.
- [x] With `include_references=False` nothing changes and the hook is not called. Existing test `test_get_retriever_output_skips_evidence_hook_when_references_disabled`.
- [x] No graph or vector I/O in the hook. Measured 0.19 ms on a worst-case result of 10 chunks and 50 edge bullets (115 evidence items), against 0.02 ms for the RAG builder on 10 chunks; both are far below the vector search and LLM call in the same query.
- [x] Facts and chunk summaries produce no evidence items.
- [x] Batch queries produce a flattened, deduplicated list with continuing ranks.

Verification run locally, rebased onto `dev` at `45208c50e`: the three touched test files (90 passed), the retrieval and search unit sweep (719 passed), ruff check and ruff format clean.

## Type of Change
- [ ] Code refactoring
- [ ] Other (please specify):

## Screenshots

```
90 passed   (test_context_evidence.py, hybrid_retriever_test.py, test_get_retriever_output.py)
719 passed  (cognee/tests/unit/modules/retrieval cognee/tests/unit/modules/search)
```

## Pre-submission Checklist

## DCO Affirmation


cognee/modules/retrieval/hybrid_retriever.py
```diff
@@ -300,6 +301,10 @@ async def append_references(self, completions: list[Any], retrieved_objects: Any
             enabled=self.include_references and self.response_model is str,
         )
 
+    def get_context_evidence(self, retrieved_objects: Any, dataset_id: Any = None):
+        """Return the exact chunks, entities and edge bullets rendered into hybrid context."""
+        return hybrid_context_evidence(retrieved_objects, dataset_id=dataset_id)
```
lxobr, line 306: This now triggers the graph-source footer, so the same chunk can appear under two `Evidence:` headings. Maybe we could deduplicate the text citations by document/chunk ID, keep both structured roles, and cover the overlap with a completion test?

lxobr (commented):
Nice work! Left one comment. Thanks for doing this.

PR #5161, opened 2026-09-21
fix(pgvector): refuse unsafe shared-database prune (COG-6550, #4956)

dexters1:
## Summary

- refuse vector-only PGVector pruning when the adapter borrows the relational PostgreSQL engine
- let `prune_system(metadata=True)` perform the single authoritative relational database deletion
- honor a `VECTOR_DB_URL` with a different database name as a dedicated, independently prunable vector database
- avoid ownership markers and table-shape inference

Alternative, fail-closed fix for #4956 and #4977.

## Behavior

- **Dedicated PGVector database:** `prune()` keeps its existing whole-database behavior.
- **Shared relational/PGVector database, `metadata=False`:** raises `SharedDatabasePruneError` before clearing metadata or dropping any table.
- **Shared database, `metadata=True`:** the expected refusal is handled and relational metadata pruning deletes the shared database once.

## Tests

- `uv run pytest cognee/tests/unit/infrastructure/databases/vector/test_pgvector_prune.py cognee/tests/unit/modules/data/deletion/test_prune_system.py -v` — **6 passed**
- `uv run pytest cognee/tests/unit/infrastructure/databases/vector/ cognee/tests/unit/modules/data/deletion/ -q -k 'not test_pgvector_update_payload'` — **79 passed, 15 skipped, 1 deselected**
- targeted `ruff check` and `ruff format --check` — passed
- pre-commit hooks — passed

The full broader command was also attempted. Its only failure was the existing real-Postgres `test_pgvector_update_payload`, because the local PostgreSQL instance could not create a throwaway database: `No space left on device`. No assertion in the changed behavior failed.


cognee/infrastructure/databases/vector/pgvector/PGVectorAdapter.py
```diff
@@ -148,8 +148,9 @@ def __init__(
                 pool_args=effective_pool_args,
             )
             self._owns_engine = True
-        elif backend_access_control_enabled() and (db_name1 != db_name2):
-            # If backend access control create new instances of engine and sessionmaker
+        elif db_name1 != db_name2:
```
siillee, line 151: **Upgrade impact that needs a release note (or a guard).**

This fixes a real bug, but it changes where data lives for existing deployments. Before this PR, with `ENABLE_BACKEND_ACCESS_CONTROL=false` and a `VECTOR_DB_NAME` different from `DB_NAME`, the `postgresql` branch below reused the relational engine. The configured vector database was silently ignored, and all vectors were written to the relational DB.

After upgrading, the same config connects to `VECTOR_DB_NAME` instead:

| State of `VECTOR_DB_NAME` | What happens after upgrade |
|---|---|
| database doesn't exist | connection errors: nothing creates it when access control is off (`pgvector/create_db_and_tables.py` only runs `CREATE EXTENSION`) |
| database exists | existing vectors stay orphaned in the relational DB, and search quietly returns nothing until the user re-cognifies |

Suggestions:
- Add a changelog / release-note entry.
- Optionally, log a one-time warning when the relational DB still holds vector collection tables and this branch is taken.

One small correction to the PR description: pgvector doesn't read `VECTOR_DB_URL`. The connection string is built from the `VECTOR_DB_HOST/PORT/USERNAME/PASSWORD/NAME` fields (`create_vector_engine.py`), so the "honor a `VECTOR_DB_URL`" wording is a bit misleading.

cognee/modules/data/deletion/prune_system.py
```diff
@@ -77,7 +78,15 @@ async def _prune():
 
         if vector and not backend_access_control_enabled():
             vector_engine = await get_vector_engine_async()
-            await vector_engine.prune()
+            try:
+                await vector_engine.prune()
```
siillee, line 82: **A refused prune still leaves things half-pruned.**

`_prune()` runs the graph step first, so on a shared DB `cognee.prune.prune_system()` (SDK default `metadata=False`) does this:
1. `graph_engine.delete_graph()` runs and the whole graph is gone.
2. `vector_engine.prune()` raises `SharedDatabasePruneError`.
3. The cache prune never runs.

The relational tables survive, which was the point. But the caller gets an error after an irreversible graph wipe. With `GRAPH_DATABASE_PROVIDER=postgres_demo` in the same database, the graph tables are dropped too. So "raises before clearing metadata or dropping any table" isn't quite true end to end.

Suggestion: make the refusal happen before anything is deleted. Two options:
- expose something like `vector_engine.can_prune_independently` (from `_owns_engine`) and check it at the top of `_prune()` when `vector and not metadata`; or
- run the vector step before the graph step when `metadata=False`.

A unit test asserting `delete_graph` is not awaited when the refusal propagates would pin this down.

siillee (commented):
The core fix looks right to me. Refusing a vector-only prune when the adapter reuses the relational engine closes the #4956 data loss without guessing which tables are vector-shaped, and the `metadata=True` path drops the vector tables once, through the relational `delete_database()`. I ran the new unit tests locally and all 6 pass.

I've left two inline comments on things I think need addressing before merge:
1. The change to how the vector connection is chosen (`db_name1 != db_name2`) affects existing deployments on upgrade, and that isn't documented.
2. A refused prune leaves things half-pruned: the graph is already deleted when the error is raised.

I've left out a few low-priority nits: the base class of `SharedDatabasePruneError` (it's `RuntimeError`, not the Cognee exception hierarchy), the error text suggesting `metadata=True` without saying that it also deletes users/datasets/ACLs, deciding whether this PR or #4977 goes in, and a real-Postgres test for the acceptance criterion in `test_pgvector.py`. Happy to add any of them as follow-ups.

PR #5163, opened 2026-09-21
feat(integrations): add Drive and Gmail sync with progress (COG-6620)

NMZivkovic:
## Acceptance Criteria

- [x] Bundle Google Drive and Gmail connectors in the SDK with OAuth, explicit resource selection, incremental sync, and deletion-aware ingestion.
- [x] Put Google Drive and Gmail in the SDK UI's Data sources, with Slack-style dialogs and controls, inline provider icons, and separate reconnect/disconnect actions.
- [x] Preserve resource selection on reconnect. Connecting authorizes the account; Refresh starts an import. Handle cancelled/closed/timed-out OAuth popups without leaving the controls locked.
- [x] Show actual stored-item counts and running/degraded sync state. A processing failure does not automatically demand reconnection.
- [x] Display persisted ready/remaining counts, progress bars, and per-document completion in Brains and brain details. Polling survives closing the connection dialog and reopening the page; totals cover imported items, not files still downloading from Google.
- [x] Fix Drive shared-drive resource listing and retry transient Gmail API failures.
- [x] Reconcile the current document resource's staged rows on a zero-change sync, so an empty local brain or failed processing can recover without modifying the remote files. Preserve folder isolation and deletion propagation.
- [x] Document local OAuth setup, dataset isolation, incremental behavior, and cloud deployment prerequisites.
- [x] Verify real local Drive and Gmail authorization/imports and restore the missing Drive batch after the zero-change sync fix.
- [ ] Complete the remaining live shared-drive, remote edit/delete propagation, and disconnect/delete-data checks with disposable test data.
- [ ] Add the author-written description and local test screenshots before marking ready.

## Validation

Latest verification for `bcc6a8274`:

- **442 passed, no skips:** integration modules, Google connector sources, DLT document mode and cleanup scope, zero-change staged replay, and Drive ingestion/forget/example integration tests. Includes `test_add_foreground_orphan_cleanup.py`; the previously reported failure now passes. Google/LLM calls in the automated integration tests are mocked; DLT and local storage are real.
- **41 passed:** Google UI/API behavior, processing counts, dataset switching, stale-count handling, and document rendering.
- TypeScript `tsc --noEmit`, targeted ESLint, Ruff lint/format, pre-commit hooks, and staged diff checks passed.
- Backend tests run with `REQUIRE_AUTHENTICATION=true`, which is required by the unauthenticated-request rejection test. An initial run with optional authentication failed that test; the configured rerun passed the full selection.
- Earlier SDK packaging verification passed `uv build --wheel` and the missing-extra/import checks; packaging was not changed by this update.

## Review dependencies and deployment limits

- Tracking: COG-6620; related reconnect work: COG-6610.
- This branch incorporates Drive OAuth work from #5156 and the reconnect fix from #5111. Coordinate consolidation of those overlapping PRs before merge.
- The SDK bundles the connectors; community draft https://github.com/topoteretes/cognee-community/pull/201 is superseded. No community connector package is required.
- The Google cards currently target the SDK server and are restricted to SDK mode. Cloud consumption requires the SDK dependency update, OAuth/secret configuration, and route/deployment wiring.
- There is no periodic sync scheduler. Background sync and its duplicate-run guard are process-local; multiple workers and scheduled operation require a durable queue and distributed locking.
- Legacy documents without staging-table ownership metadata are retained rather than guessed at during cleanup.

## Type of Change

- [x] New feature
- [x] Bug fix
- [x] Code refactoring

## Screenshots

Local test screenshots still pending, as required by CONTRIBUTING.md. Keep this PR in draft.

## DCO Affirmation


NMZivkovic:
Fixed the CI failures in 566008a4f: status DTO tests now expect syncCounts and verify metadata secrets stay private; Google sync/cleanup tests patch explicitly imported modules to avoid Python 3.10 function/module shadowing; shared OAuth JSON parsing catches specific client/parse/timeout errors instead of a broad exception suppressed with noqa. Added five sanitized-error regression cases. Local verification: 422 tests passed on Python 3.10 and 422 on Python 3.11 (all integration-module unit tests plus connector/document cleanup tests); repository-wide Ruff, strict exception lint with --ignore-noqa, format check, and ty all passed. Cross-platform CI is pending; this is not an all-green claim.

NMZivkovic:
Addressed the orphan-cleanup blocker in 64a18bbb8 (review: https://github.com/topoteretes/cognee/pull/5163#issuecomment-5773837714).

- Took the explicitly suggested conservative option: background add() skips deferred orphan cleanup entirely and logs a warning. Existing documents are not deleted before detached ingestion commits, including when startup, stream materialization, cancellation, or a later pipeline failure occurs.
- Foreground cleanup now requires PipelineRunCompleted or PipelineRunAlreadyCompleted after unwrapping the dataset result. A returned PipelineRunErrored is not success and cannot trigger cleanup.
- Documented the foreground-sync requirement in add() and both SDK connector guides. A no-change incremental retry after a background run does not guarantee reconciliation: the affected source tables must be loaded again in the foreground.
- Corrected the failing foreground integration test to exercise document-mode ingestion, the path used by Gmail/Drive. Its old relational-manifest fixture assumed re-add updates a stable manifest, which conflicts with the existing idempotent-add contract. It now verifies real Data-record deletion, including the final document, with isolated DLT/DB state.

Validation:
- 20 new safety regression cases pass on Python 3.10 and 3.11.
- All 6 real-storage orphan-cleanup/Drive integration tests pass, including graph-content deletion and last-file deletion (25 passed in the combined run with the initial 19 safety cases).
- Broader Python 3.10 run covering integration modules, connector/DLT units, API v1, and pipelines: 1,280 passed, 1 skipped, 1 dataset-lock timeout. The unchanged test_delete_data_waits_for_pipeline_lock hit its 1-second timeout in session invalidation; isolated rerun of its module plus the safety tests passed all 22 cases.
- Repository ruff check, changed-file format check, and git diff --check pass.
- Optional ty check of add.py still reports two pre-existing diagnostics: user: User = None and preferred_loaders reassignment; neither was introduced here.

No cloud, UI, or MCP files changed in this fix. Fresh CI/review is pending.

goran-radonic:
## PR #5163 review notes (final, all passes combined)

general bugs, cloud-backend fit, and a security-focused pass. All findings below are either probe-confirmed or explicitly marked as reasoned; refuted claims are listed separately so nothing gets overstated.

### Blockers

**Cross-account cursor collision (Gmail)**
`cognee/tasks/ingestion/ingest_dlt_source.py` builds `dlt.pipeline(pipeline_name="ingest_dlt_source", ...)` with one hardcoded literal shared by every dlt ingestion in the system. dlt's `resource_state()` is keyed only by pipeline_name + source + resource name, never by `dataset_name`. Gmail's connector additionally hardcodes `name="gmail_messages"` with no per-account suffix (Drive avoids this by accident, since its resource name is hashed from the folder id). Reproduced directly:
```
RUN acctA dataset=gmail_a_aaaa   -> state at start: {}
RUN acctB dataset=gmail_b_bbbb   -> state at start: {}
RUN acctA2 dataset=gmail_a_aaaa  -> state at start: {'last_history_id': 'hist-acctB'}
```
Sync account A, then B, then A again: A's incremental run resumes from B's Gmail historyId. Reproduces in a single process, zero concurrency required. Fix: put `provider_account_id` into Gmail's dlt resource name, the way Drive already does with its folder-id hash.

**Credential resurrection after disconnect**
`upsert_credential` unconditionally sets `status=ACTIVE, revoked_at=None` on every call, with no precondition on the row's current status. `refresh()` (called from `access_token_for()` at the start of every sync or resource listing) goes through this same function. If a refresh that's already in flight completes after a `DELETE /connection` call, it silently reactivates the row. `revoke_token` itself is best-effort (a failed remote revoke is swallowed with a warning), so the resurrection can be durable, not just transient. Reproduced against a real DB:
```
after revoke:         status=revoked
after refresh-upsert: status=active, revoked_at=None
visible to UI:        True
```
Companion bug, not timing-dependent: nothing in the sync loop re-checks `credential.status` per iteration, so a sync spawned before a disconnect keeps ingesting to completion after the user disconnected.

**Deletions never propagate outside the happy path (three separate instances, all reproduced)**
All three share one root cause: both connectors use `write_disposition="merge"`, which makes the `_deleted` tombstone row the only deletion channel, and three code paths never emit one:
- Drive full-listing mode: one permanently-unparseable file blocks the cursor forever, and that branch has no tombstone logic at all.
- Gmail `full_backfill`: no tombstone logic; deletions during a label-scope narrowing or a historyId-expiry gap are lost permanently.
- Drive folder deselection: removing a folder in the UI just stops that folder's sync from ever running again, no cleanup, files stay forever.

**Revoked credential reuse leaks the previous owner's configuration**
Reproduced end to end: a new owner connecting the same underlying Google account inherits the previous owner's folder/label selection and stale "synced, healthy" status, with no opt-in. Not a raw data-access leak (the new owner already holds a valid token for that account), but it breaks the documented invariant that a new connection must opt into scope.

### Real, lower priority

- **Drive query-language injection via `folder_id`** (`PUT /{provider}/resources`): confirmed a working malicious payload that produces a valid, unbounded Drive query. However, verification found this is not a privilege-boundary bypass: only the account owner can set their own folder selection, and the OAuth grant already covers the entire Drive regardless of the UI selection, so the injection gets an attacker nothing they couldn't already get by selecting root. Real robustness bug, not a security boundary break. Worth validating the Drive-id format anyway.
- **Log injection via unsanitized Drive file names**: confirmed on both the console and file log sinks. Low severity, confuses a log reader rather than enabling code execution or data access.
- **Unbounded export of native Google Docs/Sheets**: confirmed. The `size` preflight guard silently no-ops for native mime types since Drive never populates `size` for them, so `extract_file_content` has no byte cap on that path. The byte-capped REST client path exists in the codebase but has zero production callers, it's dead code. Moderate severity (bounded by Docs' own ~1M-character limit per file, but adds up across a folder).
- **CI never installs the `gmail`/`google-drive` extras**: confirmed. The real `google-api-python-client`/`google-auth` code path never runs against actual libraries in CI. Separately, and pre-existing (not introduced by this PR): `dlt` isn't installed on the CI leg that collects `cognee/tests/integration/tasks/`, so this PR's own new connector integration tests are silently skipped there.
- **`.env.template` missing Gmail variables**: confirmed as a documentation asymmetry, not a blocker — they are documented in `cognee-frontend/README.md`.

### Checked and dismissed

- **Test coverage for orphan cleanup "silently narrowed"**: refuted. An untouched, unrelated test (`test_dlt_orphan_graph_vector_purge.py`) still exercises the real manifest-mode path end to end.
- **`record_sync_result` UnboundLocalError**: the code shape is real and reproduced standalone, but the trigger is refuted — `expire_on_commit=False` plus the fact that the same attributes are already read successfully earlier in the same call stack make the detached-instance scenario unreachable in practice. One-line hygiene fix, not a live bug.
- **`file_id` path interpolation into a URL**: refuted as reachable. The functions using it (`export_file`, `download_file`) have zero production callers, dead code.
- **Cross-provider state-secret confusion**: design gap exists, but the install-nonce cookie fully closes it in every ordering tested. Not exploitable today.
- **Token encryption at rest (AES-256-GCM)**: clean. No key fallback, no hardcoded key, correct nonce handling.
- **`dataset_summary()` filtering by `user_id` instead of `workspace_id`**: not a bug. The Google connectors never set `workspace_id` at all, so filtering by `user_id` is correct.

### Cloud-backend fit (design input, not a blocker on this PR)

`OAuthIntegration.sync_now()` collapses OAuth handling and the `remember()`-calling ingest into one in-process method, with no `SyncScope`-style hand-off — this is exactly why the separate SaaS backend built its own connector framework. That said, `source_factory`/`build_service` already provide a clean, token-injectable seam for that hand-off. Only `sync.py`/`run_sync` (roughly 200 of the ~1800 new lines) would need rewriting for a control-plane/pod split; the cursor, tombstone, and content-extraction logic in the connectors themselves is reusable as is.

### Verdict

NEEDS CHANGES. Three things should be fixed before merge: the cross-account cursor collision (Gmail resource naming), the credential resurrection race (add a status precondition to `upsert_credential`'s refresh path, plus a per-iteration status check in the sync loop), and the deletion-never-propagates cluster (three separate paths). Everything else is lower severity or a design note for the future cloud integration.

goran-radonic (dismissed):
LGTM

goran-radonic (approved):

PR #5168, opened 2026-09-22
COG-6624 fix(api): Scope provenance graph to caller's readable data

goran-radonic:
## Description

Ran a swarm review on some uncommitted schema/provenance scoping work on dev and found the fix wasn't complete. A plain tenant member could still see every role in the tenant (with full membership) and every user in the tenant, regardless of whether they had access to any dataset those roles/users touched. Added the same scoping that already existed for datasets/files/agents to roles and users too: a role stays in the view only if it has a grant on a dataset the caller can read, a user stays only if they own, hold a direct grant on, or belong to a role with a grant on one of those datasets.

Also fixed two smaller bugs from the same original patch: the scope check was running outside the route's try/except so a DB hiccup there returned a raw 500 instead of the normal error response, and a routine "not an admin" check was logging at ERROR level on every request instead of DEBUG.

Linear: COG-6624

## Acceptance Criteria

- Non-admin tenant member no longer sees roles/users unrelated to their own dataset access
- Admin view is unchanged, still sees everything
- Scope failures return the normal error response, not a 500
- Non-admin check no longer spams ERROR logs

## Type of Change
- [ ] Code refactoring
- [ ] Other (please specify):

## Pre-submission Checklist

## DCO Affirmation


NMZivkovic (commented):
Read this one closely because it is an access-control change, and it holds up. The leak is real, the narrowing is fail-closed on every path I could find, and the tests pin it rather than describe it. No blockers.

## The bug is corroborated outside this repo

Worth adding to the description, because it turns "a swarm review found this" into "Cloud is already paying for it". The Cloud frontend hit this and routed around it — `cognee-saas-frontend/src/app/(app)/business/BusinessView.tsx:107`:

> Correctness: `/v1/schema/provenance/json` scopes datasets by tenant (Dataset.tenant_id), so its first entry can be one this caller holds no read grant on, and the graph fetch for it then fails; `/v1/datasets/` answers with exactly the readable set.

That is this exact bug, seen in production. After the bump that comment becomes half-true — the two endpoints will agree for a plain member and still differ for an administrator — so it wants a follow-up edit in the SaaS repo rather than being left to rot.

## What I verified rather than took on trust

**The empty-list pitfall.** The classic way this patch shape fails is `if scope_dataset_ids:` — a caller who may read nothing gets a falsy `[]`, the gate is skipped, and they see everything. All eight gates use `is not None`, and `test_empty_dataset_scope_shows_no_datasets_rather_than_all_of_them` pins it.

**Two residual leaks I went hunting for, both closed.** The agent filter keeps an agent when *any* of its datasets is in scope and never prunes the ref list, so the edge loop builds `agent -> dataset` for out-of-scope datasets too; a kept role likewise carries its whole membership. Neither escapes, because `add_edge` only emits when both endpoints are already nodes. Comment inline — that guard is now load-bearing for the fix and nothing says so.

**Role membership is consistent, not over-wide.** A kept role's `user_ids` can only name users the new user query also admits, since its third clause admits every member of a role holding a grant on an in-scope dataset. Same set both sides: no dangling edges, no extra exposure.

**Every error path narrows.** `_administers_tenant` catches only the two expected exceptions and answers `False`, the narrower answer; anything else propagates into the route's `try`, which you moved it inside, and comes back 409 with no data. Both 409 tests fail against dev.

**The signature change is safe.** `PermissionDeniedError` really does take `log_level` (`exceptions.py:49`), and all ten production callers of `has_user_management_permission` keep the `ERROR` default — only the new read-side caller passes `DEBUG`, so no mutation gate changes behaviour.

**The subtle one you got right:** `scoped` now includes `scope_dataset_ids is not None`, so the relational memory fallback narrows too. `test_dataset_scope_alone_still_scopes_the_relational_memory_fallback` is the test for it, and it fails on dev.

## Tests

I reverted all three source files to dev's versions and kept the tests: **15 of 30 fail**, all 30 pass restored. They cover the leak itself, the empty scope, role and user narrowing, agents and sessions on out-of-scope datasets, the 409, and the log level. Another 44 pass across the users/permissions and provenance suites. Ruff clean, 169/169 green.

## One thing for the body

`examples/guides/memory_provenance.py` is now the only unscoped call site left. That is correct — the docstring documents unscoped as the OSS single-user default — but since this PR makes scoping the rule, it is worth one line there saying the omission is deliberate, so nobody reads it as a missed call site.

cognee/api/v1/visualize/memory_provenance.py
```diff
@@ -707,6 +772,20 @@ async def get_memory_provenance_graph(
         allowed_user_ids = set(user_ids)
         agents = [a for a in agents if a.get("user_id") in allowed_user_ids]
     sessions = await _read_sessions(user_ids, agents)
+    if scope_dataset_ids is not None:
+        # A dataset scope is a caller who may read only some of the workspace,
```
NMZivkovic, line 776: This keeps an agent when *any* of its datasets is in scope, and the ref list is never pruned — so the edge loop below still runs `add_edge(agent -> dataset:<out-of-scope id>)` for the rest of them, `writes` included. Same shape for a kept role's full `user_ids`.

Neither actually leaks, because `add_edge` emits only `if source in nodes and target in nodes`, and the out-of-scope dataset was never added as a node. I chased both paths before I convinced myself.

That guard is now load-bearing for this fix and nothing says so. Could you note it here ("out-of-scope refs survive on the record and are dropped at edge construction, which only connects admitted nodes") and in the module docstring? A future refactor that hoists or "optimises" that check reopens the leak with no test failing — none of the new tests assert on a dangling edge.

cognee/api/v1/visualize/memory_provenance.py
```diff
@@ -656,6 +688,31 @@ async def get_memory_provenance_graph(
             user_stmt = user_stmt.where(User.tenant_id.in_(scope_tenant_ids))
         elif scope_user_ids is not None:
             user_stmt = user_stmt.where(User.id.in_(scope_user_ids))
+        if scope_dataset_ids is not None:
+            # Same rule as roles above, applied to the user roster: a tenant
```
NMZivkovic, line 692: This three-clause rule is the right one, and I checked it composes with the role query rather than fighting it: the third clause admits exactly the membership of the roles the role query keeps, so every `user_ids` entry on a kept role resolves to an admitted node. Consistent both ways.

One behaviour worth knowing, not a change request: this AND-s with the tenant filter above, so a user holding a read grant on a dataset in *another* tenant has it dropped from their own provenance view. That is the safe direction and probably unreachable today, but it is a real asymmetry with `/v1/datasets/`, which would list it.

cognee/api/v1/visualize/routers/get_schema_router.py
```diff
@@ -55,6 +55,65 @@ class ErrorResponse(BaseModel):
     error: str
 
 
+async def _provenance_scope(user: User) -> tuple[list | None, list | None, list | None]:
```
NMZivkovic, line 58: Good shape: one function owns the decision, both routes call it, and it is inside the `try` so a DB failure while computing scope returns the normal error instead of a raw 500 — which is also the fail-closed answer, since no data goes out. Deferring the admin test to `has_user_management_permission` rather than re-deriving it keeps this endpoint honest about what the rest of the read-side API already enforces.

The parenthetical about tenant-mutation call sites checking `tenant.owner_id` directly is the kind of note I would rather find here than rediscover later.

NMZivkovic (approved):

PR #5172, opened 2026-09-22
fix(config): Resolve .env from the working directory first, once, for every reader (SDK-781)

dexters1:
Fixes [SDK-781](https://linear.app/cognee/issue/SDK-781/env-is-resolved-from-the-installed-package-location-so-a-project-env). From a user report: "Cognee's docs say setting `ENABLE_BACKEND_ACCESS_CONTROL=false` in your `.env` disables access control. On a standard pip install, it doesn't." Verified on `dev`.

## What was wrong

`cognee/__init__.py` and `cognee/shared/__init__.py` each called `dotenv.load_dotenv(override=True)` with no path. python-dotenv then searches upward from the **calling frame's file**, which is the installed package, not the user's project. Whether the project `.env` was found depended on where the environment lived:

| layout | project `.env` found |
| -- | -- |
| `python -m venv .venv` at the project root | yes, walking up from site-packages reaches the project |
| conda, pipx, Homebrew python, a venv kept elsewhere | **no** |
| REPL, notebook, `python -c`, under a debugger | yes, dotenv switches to the working directory |

Two things made it worse than a missed file:

1. **The file was half-read.** The settings classes (`BaseSettings` with `env_file=".env"`) resolve their file from the working directory, so `LLM_MODEL`, `GRAPH_DATABASE_PROVIDER`, `CACHE_BACKEND` and every other settings field **did** take effect, while everything read with `os.getenv` at the point of use did not: `ENABLE_BACKEND_ACCESS_CONTROL`, `AUTO_FEEDBACK`, `ALLOW_CYPHER_QUERY`, `ALLOW_HTTP_REQUESTS`, `COGNEE_PROVENANCE_MODE`, `DATASET_QUEUE_ENABLED`, `DEFAULT_USER_PASSWORD`, the log settings. The user saw their file plainly working and concluded one setting was broken.
2. **A stranger's `.env` could win.** Any `.env` above the installed package was loaded with `override=True` and then outranked the project's own file for every settings field too.

Nothing logged which file was loaded. The debugger switch meant the bug vanished under VS Code's debugger or under coverage.

## The change

One resolver, one load, one log line. New `cognee/shared/env_file.py`:

- `resolve_env_file()` searches `.env` in the **working directory and its parents** first, then in the **cognee package directory and its parents** (today's behaviour, kept so an environment nested inside the project keeps working when a script runs from elsewhere), else none. The search is explicit, so python-dotenv's REPL/notebook/debugger heuristics no longer apply.
- `load_env_file()` loads the resolved file once per process with `override=True` and returns the same answer on later calls. Both former call sites now use it.
- `cognee/__init__.py` logs the resolved path, or "No .env file found", once `setup_logging()` has run.

The settings classes are untouched. With the file already in `os.environ` before any of them is instantiated, pydantic reads the file's values too (measured below), so every reader agrees.

**Precedence, deliberately:** the `.env` file beats variables already set in the shell, for every reader, with no opt-out. Preset shell variables have repeatedly broken users who did not know they had them; the project file is the thing they can see and edit. This is what `override=True` already did on `dev` for the readers that got the file at all, so it is a consistency fix rather than a new policy, but it now applies everywhere. **Deployments must not ship a `.env` inside an image** and expect the container's environment to win.

## Measured, worktree of this branch, cognee installed outside the project

```
A) project .env in cwd, no .env near the package
   Loaded settings from .../projA/.env; its values take precedence over preset environment variables
   os.environ ENABLE_BACKEND_ACCESS_CONTROL: false      (dev: unset)
   backend_access_control_enabled():        False      (dev: True)

B) shell has LLM_MODEL=openai/PRESET-IN-SHELL, .env says openai/MODEL-FROM-ENV
   after import, os.environ LLM_MODEL  = openai/MODEL-FROM-ENV
   pydantic get_llm_config().llm_model = openai/MODEL-FROM-ENV   (file wins for both readers)

C) run from projA/deep/er, .env two levels up      -> projA/.env loaded
E) no .env near cwd, .env beside the package       -> package-side .env loaded (old layout keeps working)
F) .env in cwd AND beside the package              -> cwd file loaded, package-side ignored
D) no .env anywhere
   No .env file found; using process environment variables only.
```

## Tests

`cognee/tests/unit/shared/test_env_file.py`, 9 tests: the four layouts (working directory, ancestor, package-side fallback, none); a preset variable losing to the file for a raw `os.getenv` read **and** for a `BaseSettings` field; the load happening once even after a `chdir`; the log message; a deleted working directory not breaking import. The existing test that patches `dotenv.load_dotenv` at import still passes. Lint, format, `ty` and pre-commit clean.

## Docs

CLAUDE.md "Environment Setup" now says where the file is looked for and that it wins over preset variables. Release note for the changelog:

> `.env` is now found from the working directory (or any parent) first, then from the cognee package location, once per process, and its values take precedence over environment variables already set in the shell for every setting. Previously the file was found only when the virtualenv lived inside the project, and flags read with `os.getenv` (such as `ENABLE_BACKEND_ACCESS_CONTROL`) ignored it even then. The loaded path is logged at import. Do not ship a `.env` inside a container image if the container's environment is meant to win.

## Type of Change


cognee/shared/env_file.py
```diff
@@ -0,0 +1,90 @@
...
+Cognee reads settings two ways: settings classes (``BaseSettings`` with
+``env_file=".env"``) and plain ``os.getenv`` at the point of use. Both must see
+the same file with the same precedence, so the file is resolved here once and
+loaded into the process environment with ``override=True`` — the file wins over
+variables already present in the shell, which is what users expect from a
+project ``.env`` and what preset shell variables kept breaking.
+
+Where the file comes from, in order:
+
+1. ``.env`` in the working directory or any of its parents — the project.
```
dexters1, line 12: This ordering was chosen because users were having issues with their process environments overwritting their .env files for Cognee and causing issues

cognee/shared/env_file.py
```diff
@@ -0,0 +1,90 @@
...
+_resolved: str | None = None
+
+
+def _find_upward(start: Path, filename: str = ".env") -> str | None:
+    """The first ``filename`` found in ``start`` or any parent, or None."""
+    try:
+        start = start.resolve()
+    except OSError:
+        return None
+    for directory in (start, *start.parents):
```
NMZivkovic, line 42: This walks from the working directory all the way to `/`, and the result is loaded with `override=True` and no way to turn it off. On `dev` that was unreachable for a plain script — resolution was package-relative, and only the REPL/notebook heuristic ever looked at the cwd — so this combination is new.

The case I would worry about is a scratch or CI directory under a writable parent: run from `/tmp/whatever`, and a `/tmp/.env` someone else left there is loaded ahead of the real environment. The blast radius is not cosmetic — a `.env` reaching this path can set `ALLOW_HTTP_REQUESTS` (which gates SSRF protection in `tasks/web_scraper/ssrf_protection.py:55` and `tasks/code_graph/resolve_repo.py:251`), `ENABLE_BACKEND_ACCESS_CONTROL`, `ALLOW_CYPHER_QUERY`, `DEFAULT_USER_PASSWORD` (`base_config.py:138`), and `LLM_ENDPOINT`/`LLM_API_KEY`, i.e. point every prompt and retrieved passage somewhere else.

Walking up is conventional on its own (python-dotenv's `find_dotenv(usecwd=True)` does it) — it is the unconditional override plus these particular flags that earns a bound. Cheapest first: stop at a project marker (`.git` / `pyproject.toml`); or skip a candidate that is world-writable or not owned by the caller; or only look at the working directory itself and not its parents. Any of the three keeps the reported bug fixed.

cognee/shared/env_file.py
```diff
@@ -0,0 +1,90 @@
...
+    """Load the resolved ``.env`` once, the file winning over preset variables.
+
+    Returns the path that was loaded, or None when there was nothing to load.
+    Later calls return the same answer without touching the environment again.
+    """
+    global _loaded, _resolved
+    if not _loaded:
+        _resolved = resolve_env_file()
+        if _resolved is not None:
+            dotenv.load_dotenv(_resolved, override=True)
```
NMZivkovic, line 81: `override=True` with no opt-out is a defensible default and I agree with the reasoning — preset shell variables users forgot about are a real support cost, and this was already the behaviour for readers that found the file.

But the description tells operators "do not ship a `.env` inside an image if the container's environment is meant to win", and right now that is the only remedy on offer. A `COGNEE_ENV_FILE` to pin an explicit path (skipping the search entirely) and a flag to skip loading altogether would turn that warning into something an operator can act on, and both are a few lines here. Fine as a follow-up if you would rather keep this PR tight.

cognee/__init__.py
```diff
@@ -21,15 +21,18 @@
...
+from cognee.shared.env_file import load_env_file, describe_resolution
 
-dotenv.load_dotenv(override=True)
+_env_file = load_env_file()
 
 # NOTE: Log level can be set with the LOG_LEVEL env variable
 from cognee.shared.logging_utils import setup_logging
 
 logger = setup_logging()
+logger.info(describe_resolution(_env_file))
```
NMZivkovic, line 35: Nice detail that is easy to get wrong: the load happens before `setup_logging()` so `LOG_LEVEL` from the file still applies, but the *log line* is emitted after, so it actually lands. Logging the resolved path and not its contents is also the right call — this line will end up in user bug reports.

The missing half is the negative case being equally visible, and `describe_resolution` already covers it with "No .env file found; using process environment variables only." That is the line that would have short-circuited the original report.

NMZivkovic (commented):
Real bug, well diagnosed, and the fix is the right shape — one resolver, one load, one log line, with the settings classes left alone. The `os.getenv`-versus-`BaseSettings` split in the description is the part that explains why this reached us as "one setting is broken" instead of "my file is ignored", and it is worth keeping in the commit message. One design concern below, nothing blocking.

## The red CI is not yours

`Run Add with Docling Test` fails on `HfHubHTTPError: 429 Client Error: Too Many Requests` pulling `docling-project/docling-models`. Conclusive that it is environmental: the **first commit on this branch passed the whole suite**, and the commit that went red only shortens two log strings (-12/+2, all f-string text). A HuggingFace rate limit cannot come from that. `Test Completion Status` is just the aggregate following it down. Docling is green on #5168, #5111 and #5002, so it is intermittent rather than broken — worth a re-run before merge rather than an investigation.

## I reproduced the resolution difference

Built the reported layout (project `.env`, package installed outside it) and ran both resolvers:

| layout | old (package-relative) | new (cwd-first) |
| -- | -- | -- |
| venv outside the project | **None** | `…/proj/.env` |
| run from `proj/deep/er` | None | `…/proj/.env` |

Matches the user report exactly.

## The deployment scenario your framing invites — I checked, and it is a non-issue

Because the body says *"deployments must not ship a `.env` inside an image"*, the first thing a reviewer will reach for is `docker-compose.yml`, which mounts `.env` at `/app/.env` and sets `SYSTEM_ROOT_DIRECTORY`, `DATA_ROOT_DIRECTORY`, `DB_HOST` and `DEFAULT_USER_PASSWORD` in `environment:`, with `WORKDIR /app`.

It is **unchanged**. The package sits at `/app/cognee`, so the old package-relative walk already found `/app/.env` and already applied `override=True` — I verified both resolvers return the same path for that shape. The mounted file already beat `environment:` on `dev`. Please put that in the description; without it every reader has to re-derive it, and the warning reads scarier than the change actually is.

Cloud pods are likewise unaffected: the v2 chart injects everything through k8s `env:`/`valueFrom` with no `.env` mount, so resolution finds nothing and the process environment is used as-is.

## The one thing that really changes, measured

For a settings-class field where a shell variable and the `.env` disagree, the winner flips:

```
pydantic WITHOUT dotenv preload (old, venv outside project): openai/PRESET-IN-SHELL
pydantic AFTER override preload  (this PR):                 openai/FROM-ENV-FILE
```

pydantic ranks `env_settings` above `dotenv_settings`, so the shell used to win there and now the file does. That is the intended consistency fix — I just want it stated this crisply somewhere, because it is the case that will surprise someone relying on a shell override.

## Otherwise

The memo makes the load genuinely once per process (verified, including after a `chdir`), `OSError` on a deleted working directory is handled, the log line reports the path and never values, and the import order works out — `cognee/__init__.py` importing `env_file` triggers `cognee/shared/__init__.py`'s load first and the top-level call returns the cached answer. 9 targeted tests pass; ruff and format clean.

Two notes inline, one on the upward walk and one on the missing escape hatch.

cognee/shared/env_file.py
dexters1: Done. The search now stops at the project root: the nearest directory holding a `.git` or `pyproject.toml`. A `.env` above that is never loaded, and the same bound applies to the package-side search.

I went with the marker because it keeps the "run from a subdirectory of the project" case working and needs no ownership or permission checks.

One deliberate gap: when no marker exists anywhere above the starting directory, the search still continues upward. That's the plain folder of scripts, with a `.env` and a `main.py` and no `pyproject.toml`, which is a very common way people try cognee, and I didn't want to break it. It means your `/tmp/scratch` example is still reachable when nothing above it is a project. For anyone who needs a guarantee there, `COGNEE_ENV_FILE` (below) pins the exact file and skips the search.

Tests pin all three cases: stops at the marker, continues upward without one, and the subdirectory case with a marker.

cognee/shared/env_file.py
dexters1: Added `COGNEE_ENV_FILE` in this PR.
- **Unset or empty:** the default search runs, unchanged.
- **Set:** that exact file is loaded with the same precedence, and no search happens.
- **Set to a path that isn't a file:** import raises `FileNotFoundError` naming the variable. I made that an error instead of falling back to the search, so a typo can't silently load a different file.

I didn't add a separate "skip loading" flag. Pointing `COGNEE_ENV_FILE` at an empty file gives the same result, and it keeps one knob instead of two. That also gives operators an actionable answer for the image case: pin the file, or pin an empty one.

PR #5174, opened 2026-09-22
chore(docs): use slim Docling for MCP document processing

dexters1:
## Summary

- Use the slim `cognee[docling]` extra for MCP instead of `cognee[docs]`; remove MCP's now-unneeded direct Torch dependency and CPU-index override.
- Remove Pandoc-based converters from the optional Unstructured extra and stop routing ODT, RTF and EPUB to Unstructured.
- Include Docling's OpenDocument support, including when MCP resolves an older supported Cognee release.
- Regenerate both lockfiles and document the format requirements.

## Format impact

ODT and EPUB use slim Docling. RTF and legacy Office conversions require a host LibreOffice installation; it is not added to the MCP image. Default PDF text extraction stays on PyPDF. Explicit model-based Docling PDF/image processing requires `cognee[docling-full]`. The slim MCP install does not include Unstructured's advanced PDF processing. Org/reStructuredText conversion is not provided by these Cognee loaders.

## Validation

Tested locally on macOS arm64 / Python 3.12, in separate clean environments:

- `uv sync --locked --dev --python 3.12` in MCP, plus `uv pip check`.
- MCP suite against the local Cognee source: `python -m pytest -c pyproject.toml tests/ -q` — **182 passed**.
- `python -m pytest cognee/tests/unit/test_document_dependency_profiles.py -q` — **6 passed**.
- `python cognee/tests/test_docling_slim_loader.py` against both local Cognee and the published version in MCP's lock: DOCX, PPTX, XLSX, ODT, EPUB and default PyPDF extraction pass; explicit Docling PDF gives the expected `docling-full` install hint. No Torch, NVIDIA, Unstructured or Pandoc Python distributions in the slim environments.
- `python cognee/tests/test_unstructured_document_profile.py` in a separate locked `docs` environment: DOCX, PPTX and XLSX pass without either Pandoc Python distribution. Standard NLTK tokenizer data is required by the locked Unstructured release.
- Both `uv lock --check` commands, Ruff and `git diff --check` pass.

No LLM calls were needed for these checks. LibreOffice-based conversions, model-based PDF/OCR and a Docker image build were not exercised locally.

## DCO affirmation


dexters1:
Closing: none of the licenses this PR works around would change cognee's Apache-2.0 license. pandoc (GPL) runs as a separate program through pypandoc, which is aggregation rather than a combined work, and the unstructured[pdf] dependencies are weak copyleft (LGPL, MPL) or proprietary (NVIDIA CUDA). #5184 now checks for the case that would matter: strong copyleft (GPL, AGPL) in code cognee imports or native libraries it loads.

PR #5175, opened 2026-09-22
ci: review dependency license changes in pull requests

dexters1:
## Summary

Add a prospective dependency-license review check for pull requests.

- Compare added/updated packages in all tracked Python `uv.lock` and npm `package-lock.json` files, including locked transitive, optional, development and platform-specific dependencies.
- Evaluate SPDX expressions against an explicit permissive-license policy. Unknown/unapproved terms, unsupported package sources and metadata lookup errors require review instead of silently passing.
- Support exact-version, source-specific approvals with a reason and review link.
- Require the root and MCP Python manifests to match their lockfiles.
- Run without repository secrets or write permissions, with pinned Actions and no installation of the dependencies being scanned.
- Document review procedures, coverage and limitations.

This is a conservative metadata review gate, not a legal determination that every unlisted license is incompatible. Bundled binaries, vendored code, model weights, runtime downloads and OS/container packages still require distribution-level review.

## Validation

- `uv run --no-project --python 3.12 --with license-expression==30.4.4 python -m unittest discover -s scripts/tests -p 'test_dependency_licenses.py' -v` — **11 tests passed**, covering permissive/restrictive and compound expressions, unknown metadata, registry errors, version/source-scoped approvals and lockfile diffs.
- Exact-version public PyPI metadata smoke test (`attrs==25.3.0`) passes the MIT policy.
- CLI comparison against dev correctly reports no dependency changes for this workflow-only branch.
- Root/MCP `uv lock --check`, `actionlint`, Ruff and `git diff --check` pass.

## Activation

After merge, make **`dependency-license-check`** a required status check on protected branches. This PR does not change branch protection. Maintainers must review policy/exception changes.

## DCO affirmation


dexters1:
Closing in favour of #5184, a lighter check that installs every extra on Linux Python 3.13 and scans bundled native programs for copyleft notices. It catches the pandoc case in seconds without auditing every remote artifact.

https://claude.ai/code/session_01EcTtT5LxYj8C43hRrhvfxi

PR #5103, opened 2026-09-17
fix(neo4j): report get_edges endpoints in true edge direction

NitinKumar004:
## What

`Neo4jAdapter.get_edges` matches relationships **undirected** and returns tuples in the order of the *traversal* sides (`n`, `m`) rather than the edge's true stored direction:

```cypher
MATCH (n: `{BASE_LABEL}`{id: $node_id})-[r]-(m)
RETURN n, r, m
```
```python
(result["n"]["id"], result["m"]["id"], {"relationship_name": result["r"][1]})
```

Because `n` is always the *queried* node, an edge actually stored as `(other)-[:REL]->(queried)` comes back with its endpoints **inverted** — the queried node is reported as the source:

```
stored:   (node-b) -[:reports_to]-> (node-a)
get_edges("node-a") → [('node-a', 'node-b', {'relationship_name': 'reports_to'})]   # wrong
expected:            [('node-b', 'node-a', {'relationship_name': 'reports_to'})]
```

## Flow

How the same stored edge `(node-b) -[:reports_to]-> (node-a)` travels through `get_edges("node-a")`, before and after this fix:

```mermaid
flowchart TD
    subgraph store["🗄️  Stored in Neo4j"]
        E["(node-b) —[:reports_to]→ (node-a)"]
    end

    E --> M["get_edges('node-a')<br/>MATCH (n {id:'node-a'})-[r]-(m)<br/><i>undirected match → n = queried node</i>"]

    M --> B["❌ BEFORE<br/>RETURN n, r, m<br/>tuple = (n.id, m.id)<br/><b>('node-a', 'node-b', reports_to)</b><br/>direction inverted"]
    M --> A["✅ AFTER<br/>RETURN startNode(r).id, endNode(r).id, r<br/>tuple = (source_id, target_id)<br/><b>('node-b', 'node-a', reports_to)</b><br/>true edge direction"]

    B --> BF["format_edges<br/>'node-a —reports_to→ node-b'<br/>⚠️ swapped endpoints in LLM prompt"]
    A --> AF["format_edges<br/>'node-b —reports_to→ node-a'<br/>✔️ correct neighbourhood"]

    BF --> LLM["consolidate_entity_descriptions<br/>(LLM entity-description consolidation)"]
    AF --> LLM

    style B fill:#5b1a1a,stroke:#e5534b,color:#fff
    style BF fill:#5b1a1a,stroke:#e5534b,color:#fff
    style A fill:#14432a,stroke:#3fb950,color:#fff
[...]

PR #5179, opened 2026-09-22
fix: return real edge direction from ladybug and neptune get_edges

iuiu-py:
### Summary
- `get_edges` on the ladybug (the default graph backend) and neptune adapters matched edges **undirected** and reported the matched node as source. Stored `(other) -> (queried)` incoming edges were returned flipped, breaking the direction contract established for `get_edges` in #4967 — but that fix's PR (#5103) only covered the neo4j adapter.

### Changes
- **neptune**: `RETURN id(startNode(r)) AS source_id, id(endNode(r)) AS target_id` — the same pattern the file already uses for relationship streaming.
- **ladybug**: two directed `MATCH` passes (outgoing + incoming) returning the edge's own start/end nodes in order, mirroring the directed `<-[r:EDGE]-` patterns already used elsewhere in the adapter.
- Zero file overlap with #5103; the two fixes compose.


ArmagedonFlamer:
Thanks @iuiu-py for extending the direction fix to the Ladybug and Neptune adapters. We are closing this together with #4967 and the Neo4j PRs: since #4613 moved entity-description consolidation onto get_connections(), get_edges has no callers left in library code, so the flipped incoming edges no longer reach any pipeline. If we do tighten the get_edges contract later, it will be a single interface-level change across all adapters, with tests, rather than adapter-by-adapter patches. Two small things for next time: PRs in this repo target dev rather than main, and a regression test alongside an adapter fix makes review much faster. Thanks again for the careful work across two backends, and we look forward to your next contribution!

